PR #267, opened 2019-06-11
Safer buffering of Istanbul future messages

asaj:
### Description

This PR is a resurrection of https://github.com/celo-org/geth/pull/182, which had some bugs preventing it from working.

Istanbul is vulnerable to DDoS attack from other validators in two ways:
* retention of general Istanbul messages for future sequences
* retention of round change messages for future rounds in the same sequence

This PR implements a new backlog:
* always discards messages too many sequences in the future. `acceptMaxFutureViews` (as implemented by PBFT)
* keeps track of messages buffered on a per validator basis, and caps that at `acceptMaxFutureMsgsFromOneValidator`
* maintains an overall bound on future msgs, `acceptMaxFutureMessages`, and prunes the whole buffer by the most distant sequence
* fixes an issue where `round>1000` would overflow and deliver buffered messages in an unexpected order

And also changes round change buffering:
* We now only buffer a single future round change message from each validator; subsequent messages from the same validator with an equal or higher round replace the existing one.
* Correctness argument:
  * A round change message for `r'>r` is received  from a node `N`, and the round change message for 'r' from `N` is discarded.  
    * If `N` is Byzantine, then the discard does not prevent the node receiving `2f+1` round change messages for `r` from honest nodes (plus itself).
    * If `N` is honest, then N sent the message because it itself had seen a timeout at `r'-1` or `f+1` other round change messages (and hence at least one from an honest validator) for round `r'`. There are two cases:
      * Can no longer reach `2f+1` round change messages at `r` because this message was discarded, and hence a new round is not started at this node. Other honest nodes timeout and start a new round change. The new round change messages have `r''>r` and replace any existing ones sent from the same (honest) validator and stored at other nodes (since honest validators only ever increase rounds).
      * Can still reach `2f+1` round change messages at `r` from other nodes, and hence round change goes ahead as previously. 

### Tested

* Unit tests updated and added 
* Testnet running with 15/50 nodes not sending messages: http://34.83.56.171/
* Testnet running with 10/50 nodes sending future round change messages: **TODO**


asaj:
Closing as this is stale

PR #315, opened 2019-07-01
Encrypt enode announce messages

nicholasguo:
### Description

The enodeURL within the gossiped ValidatorAnnounce message is now encrypted by all the potential validators' public key.

### Tested

Geth unit tests

### Related issues

- Fixes https://github.com/celo-org/geth/issues/237

### Backwards compatibility


consensus/istanbul/backend/announce.go
```diff
@@ -155,9 +213,80 @@ func (sb *Backend) sendIstAnnounce() error {
...
+			pubKeyBytes = g.Alloc[addr].PublicKey
+		}
+		// ECDSAKey, err := crypto.UnmarshalPubkey(pubKeyBytes)
+		// if err != nil {
+		// 	log.Info("bigsad")
+		// 	continue;
+		// }
+		// pubKey := ecies.ImportECDSAPublic(ECDSAKey)
+		pubKey, err := p2p.ImportPublicKey(pubKeyBytes)
+		encryptedEnodeUrl, err := ecies.Encrypt(rand.Reader, pubKey, []byte(enodeUrl), nil, nil)
```
kevjue: I think it would be great if @kobigurk reviewed this piece.  I know that there could be some important ways randomness needs to be used with signing (which I assume encryption would have the same requirement), so having a crypto expert take a look at this would be great.

consensus/istanbul/backend/announce.go
```diff
@@ -155,9 +213,80 @@ func (sb *Backend) sendIstAnnounce() error {
...
+		if err != nil {
+			log.Error("Unable to unmarshal public key", "err", err)
+		} else {
+			encryptedEnodeUrls[addr] = encryptedEnodeUrl
+		}
+	}
+
+	encryptedEnodeData, err := json.Marshal(encryptedEnodeUrls)
```
kevjue: Do we need to serialize the `encryptedEnodeUrls`?  The whole announceMessage is going to be serialized.

consensus/istanbul/backend/announce.go
```diff
@@ -226,12 +355,24 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
 		return errUnauthorizedAnnounceMessage
 	}
 
+	// Decrypt the EnodeURL
+	encryptedEnodeData := msg.EncryptedEnodeData
+	var encryptedEnodeUrls map[common.Address][]byte
+	json.Unmarshal(encryptedEnodeData, &encryptedEnodeUrls)
+	encryptedEnodeUrl := encryptedEnodeUrls[sb.Address()]
```
kevjue: What happens if `encryptedEnodeUrl` is nil?

consensus/istanbul/backend/announce.go
```diff
@@ -226,12 +355,24 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
+	enodeUrlBytes, err := sb.Decrypt(encryptedEnodeUrl)
+	enodeUrl := string(enodeUrlBytes)
+
 	// Save in the valEnodeTable if mining
-	if sb.coreStarted {
+	if sb.coreStarted && enodeUrl != "" {
```
kevjue: What would cause the enodeUrl to be "empty string"?

consensus/istanbul/backend/backend.go
```diff
@@ -371,6 +377,16 @@ func (sb *Backend) verifyValSetDiff(proposal istanbul.Proposal, block *types.Blo
 	return nil
 }
 
+// Sign implements istanbul.Backend.Decrypt
+func (sb *Backend) Decrypt(data []byte) ([]byte, error) {
+	if sb.decryptFn == nil {
```
kevjue: I think that we should require that the decryptFn be non nil.

core/genesis.go
```diff
@@ -299,6 +302,13 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
 		return nil, err
 	}
 
+	genesisJSON, err := g.MarshalJSON()
+	err = db.Put(DBGenesisKey, genesisJSON)
```
kevjue: If there are multiple accounts in the genesis json file, wouldn't this be overwritten by all but one of those accounts?

consensus/istanbul/backend/announce.go
nicholasguo: I serialized it using `json` because the serialization we do for the object does not serialize go maps

kevjue: Seems like the ethereum wiki does have a suggestion on how to encode maps:  https://github.com/ethereum/wiki/wiki/Design-Rationale#rlp.

So EncodeRLP could transform the go map as described in the wiki, and DecodeRLP would then do the opposite transformation.

In general, all of the networking messages used within Ethereum uses RLP encoding, so for consistency, it would be nice if this field does as well.

core/genesis.go
nicholasguo: Why? This just stores the entire genesis object

kevjue: Yep.  Your right.  I misread the code.

consensus/istanbul/backend/announce.go
```diff
@@ -17,27 +17,78 @@
...
 	"io"
-	"math/rand"
+	mrand "math/rand"
+	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
+	"github.com/ethereum/go-ethereum/core"
```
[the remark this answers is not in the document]
kobigurk: As a side note, I think `core` is imported in many istanbul packages, mainly for the `ErrSmartContractIsNotDeployed` error.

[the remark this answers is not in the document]
kevjue: Thanks for the heads up!  @jarmg 's PR is actually going to move that error to another package, and we are going to change our imports accordingly.

consensus/istanbul/backend/announce.go
```diff
@@ -155,9 +197,76 @@ func (sb *Backend) sendIstAnnounce() error {
...
+		return err
+	}
+
+	validatorsAddress, err := sb.regAdd.GetRegisteredAddressAtCurrentHeader(params.ValidatorsRegistryId)
+	if err != nil {
+		log.Warn("Registry address lookup failed", "err", err)
+	}
+
+	encryptedEnodeUrls := make(map[common.Address][]byte)
+	for addr := range regVals {
```
kevjue: I had a separate conversation with @kobigurk on if it's possible to create one encrypted enodeURL cipher text such that only the validators can decrypt.  The purpose is to drastically reduce the payload size of the `Announce` message.  Here is his response:

_I don't know of a direct and easy way.
Before getting into heavy crypto, one easy optimization you can make is encrypt an ephemeral decryption key for each validator rather than the whole message, and then encrypt the message once.

There's a method called broadcast encryption that I haven't used, which is almost exactly what you need. It does require a new kind of public keys though, so there's this overhead... I'll be happy to take a look at it if the trade off of adding a new kind of public key makes sense.

I'll try to think if there are non direct ways this could be done._

_To remain within ECIES encryption, for simplicity, how about that in each epoch, the proposer publishes a group key to the validators? This makes the epoch message larger, and that might be fine since it's daily. The proposer will distribute this group key encrypted under each public key, doesn't have to be in the actual block (though then we'll have to make sure it gets to all the validators).

The validators would have to verify this key. This could be done by either:
1. A ZKP that all the encrypted messages contain the same key (e.g., a Schnorr signature-like proof, called DLEQ).
2. In the message that the proposer distributes, they encrypt a fixed string, e.g., the block hash, the validators attempt to decrypt it with their group key. (edited)_ 

cc'ing @nicholasguo

consensus/istanbul/backend/backend.go
kevjue: Just read this again, and you are doing exactly that :).

consensus/istanbul/backend/announce.go
kevjue: On a related note, Hotstuff BFT uses what is calls `authenticator complexity` to measure the cost of the algorithm.  

> The complexity measure we care about is authenticator complexity, which specifically is
> the sum, over all replicas i ∈ [n], of the number of authenticators received by replica i in the protocol to reach
> a consensus decision after GST. (Again, before GST, a consensus decision might not be reached at all in the worst case [27].) Here, an authenticator is either a partial signature or a signature. Authenticator complexity is a useful measure of communication complexity for several reasons. First, like bit complexity and unlike message complexity, it hides unnecessary details about the transmission topology. For example, n messages carrying one authenticator count the same as one message carrying n authenticators. Second, authenticator complexity is better suited than bit complexity for capturing costs in protocols like ours that reach consensus repeatedly, where each consensus decision (or each view proposed on the way to that consensus decision) is identified by a monotonically increasing counter. That is, because such a counter increases indefinitely, the bit complexity of a protocol that sends such a counter cannot be bounded. Third, since in practice, cryptographic operations to produce or verify digital signatures and to produce or combine partial signatures are typically the most computationally intensive operations in protocols that use them, the authenticator complexity provides insight into the computational burden of the protocol, as well.

If we are to use the same complexity measurement, then seems like the announce message will be n squared cost to the system.  If we eventually want to implement hotstuff BFT (to get to linear cost), then seems like it would be nice if we can somehow also get this message to be constant size.

consensus/istanbul/backend/announce.go
kobigurk: The choice of crypto.rand as a randomness source is good - it uses OS randomness, it's thread-safe and is the go-to choice for many libraries.

I have some general comments about the design:
* I think the same key is used for signing and encryption (correct me if I'm wrong). This is not good key hygiene in general. I think this was something that was added when istanbul was implemented, and if I recall correctly, signing keys and node p2p keys are unrelated in vanilla Geth. Worth thinking more about.
* EIP8 introduced the ciphertext length as the S2 parameter. I want to recall why and if it's important for our use.
* Something @nicholasguo and I discussed once - if we're using the signing key as the p2p key (which I think is the current design), this makes using a hardware wallet harder.

consensus/istanbul/backend/announce.go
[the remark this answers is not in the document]
nicholasguo: Is there an example of this being done?

consensus/istanbul/backend/announce.go
kobigurk: Maybe my information is out of date - do we eventually have distinct keys?
Currently the code, I believe, verifies that the enode public key derives a valid validator address.

consensus/istanbul/backend/announce.go
```diff
@@ -155,9 +213,80 @@ func (sb *Backend) sendIstAnnounce() error {
...
+		if err != nil {
+			log.Error("Unable to retrieve genesis from db", "err", err)
+		}
+	}
+
+	encryptedEnodeUrls := make(map[common.Address][]byte)
+	for addr := range regVals {
+		var pubKeyBytes []byte
+		if useValidatorContract {
+			state, err := sb.stateAt(block.Header().ParentHash)
```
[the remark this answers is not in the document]
nicholasguo: what about the header? Can that be ommitted as well?

consensus/istanbul/backend/announce.go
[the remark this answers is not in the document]
kevjue: https://github.com/celo-org/celo-blockchain/compare/rebuild_ievmh#diff-e968de38f75ea48cdf62c94527dbec22R427

consensus/istanbul/backend/announce.go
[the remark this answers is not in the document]
kevjue: Yes, that can be omitted.

consensus/istanbul/backend/announce.go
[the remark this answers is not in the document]
kevjue: Here's an example of an interface created for the blockchain object:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/consensus.go#L34

consensus/istanbul/backend/announce.go
kevjue: Thanks for the feedback @kobigurk !

To clarify, @nicholasguo 's current implementation has the following properties:

1) The istanbul consensus message signing key and enodeURL encryption key are the same.
2) The key in 1) would be able to be stored in a hardware wallet.
3) All p2p communication will be using the node key to encrypt (via the rlpx protocol).  This key def. should be different than the key in 1), but if a user configures their geth to have their node key be the same as key in 1), then we wouldn't stop them.

I hope that clarifies everything!

kobigurk: Hey @kevjue,

Thanks! How can decryption be done without having access to the key in 1 outside of a hardware wallet? I think I'm missing something obvious.

kobigurk: I looked again at the area of code I thought was deriving an address from the enodeURL, and it seems I was wrong, so I think there's no problem having the enodeURL different than the signing key, which is great.

Maybe a good compromise regarding 1 would be to have different keys for actual block signatures and "just" istanbul messages. If it makes sense, I'll also make the BLS code treat it like that.

consensus/istanbul/backend/announce.go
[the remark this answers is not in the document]
nicholasguo: closing because not using genesis obj anymore

consensus/istanbul/backend/backend.go
nicholasguo: closing because no longer using decryptfn

consensus/istanbul/backend/announce.go
nicholasguo: the decrypted result should be set to an empty slice

consensus/istanbul/backend/announce.go
```diff
@@ -190,6 +253,8 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
 		return err
 	}
 
+	log.Error("Handling an IstanbulAnnounce message", "actual", msg.IncompleteEnodeURL)
```
kevjue: nit:  Please change this severity to Debug.

consensus/istanbul/backend/announce.go
```diff
@@ -145,35 +157,86 @@ func (sb *Backend) sendAnnounceMsgs() {
...
+		return nil, nil
 	}
 
-	enodeUrl := enode.String()
+	enodeUrl := selfEnode.String()
 	view := sb.core.CurrentView()
+	incompleteEnodeUrl := enodeUrl[:strings.Index(enodeUrl, "@")]
+	ipData := enodeUrl[strings.Index(enodeUrl, "@"):]
+
+	regVals, err := sb.retrieveRegisteredValidators()
```
kevjue: nit: Since `regVals` is now containing both registered and active validators, perhaps rename this to regAndActiveVals.

consensus/istanbul/backend/announce.go
```diff
@@ -209,29 +274,47 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
 	} else if err != nil {
 		sb.logger.Error("Error in retrieving the registered validators", "err", err)
 		return err
 	}
 
+	// Add active validators regardless
+	block := sb.currentBlock()
+	valSet := sb.getValidators(block.Number().Uint64(), block.Hash())
+	for _, val := range valSet.List() {
+		regVals[val.Address()] = true
```
kevjue: nit: same comment as above regarding renaming the `regVals` variable.

consensus/istanbul/backend/announce.go
```diff
@@ -209,29 +274,47 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
+			encryptedIP = entry[1]
+		}
+	}
+	IPBytes, err := nodeKey.Decrypt(encryptedIP, nil, nil)
+	if err != nil {
+		sb.logger.Warn("Error in decrypting ip", "err", err)
+	}
+	enodeUrl := msg.IncompleteEnodeURL + string(IPBytes)
+
+	log.Error("Handling an IstanbulAnnounce message", "actual", msg.IncompleteEnodeURL, "lol", enodeUrl)
```
kevjue: nit:  LOL!  Also, please change severity to trace.

consensus/istanbul/backend/engine_test.go
```diff
@@ -47,6 +47,7 @@ func newBlockChain(n int, isFullChain bool) (*core.BlockChain, *Backend) {
 	config := istanbul.DefaultConfig
 	// Use the first key as private key
 	address := crypto.PubkeyToAddress(nodeKeys[0].PublicKey)
+
```
kevjue: Please remove the extra line.

p2p/rlpx.go
```diff
@@ -507,8 +507,8 @@ func readHandshakeMsg(msg plainDecoder, plainSize int, prv *ecdsa.PrivateKey, r
 	return buf, s.Decode(msg)
 }
 
-// importPublicKey unmarshals 512 bit public keys.
-func importPublicKey(pubKey []byte) (*ecies.PublicKey, error) {
+// ImportPublicKey unmarshals 512 bit public keys.
+func ImportPublicKey(pubKey []byte) (*ecies.PublicKey, error) {
```
kevjue: Do you still need to rename this function so it's "exportable"?  When I do a search for this function in this PR diff, I don't see it being used.

consensus/istanbul/backend/announce.go
```diff
@@ -145,35 +157,86 @@ func (sb *Backend) sendAnnounceMsgs() {
...
+	if selfEnode == nil {
 		sb.logger.Error("Enode is nil in sendIstAnnounce")
-		return nil
+		return nil, nil
 	}
 
-	enodeUrl := enode.String()
+	enodeUrl := selfEnode.String()
 	view := sb.core.CurrentView()
+	incompleteEnodeUrl := enodeUrl[:strings.Index(enodeUrl, "@")]
```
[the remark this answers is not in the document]
nicholasguo: I just need the front half of the message. The parsing function gives me the node object, which isn't what i need here.

consensus/istanbul/backend/announce.go
nicholasguo: I think this is a debug message i forgot to remove, will remove

consensus/istanbul/backend/announce.go
nicholasguo: also forgot to remove!

consensus/istanbul/backend/announce.go
```diff
@@ -34,14 +39,15 @@ import (
 // define the istanbul announce message
 
 type announceMessage struct {
-	Address   common.Address
-	EnodeURL  string
-	View      *istanbul.View
-	Signature []byte
+	Address            common.Address
+	IncompleteEnodeURL string //TODO(nguo) remove this field
+	EncryptedIPData    [][][]byte
```
kevjue: nit:  Perhaps rename this to encrypted endpoints?

consensus/istanbul/backend/announce.go
```diff
@@ -34,14 +39,15 @@ import (
 // define the istanbul announce message
 
 type announceMessage struct {
-	Address   common.Address
-	EnodeURL  string
-	View      *istanbul.View
-	Signature []byte
+	Address            common.Address
+	IncompleteEnodeURL string //TODO(nguo) remove this field
```
kevjue: IIUC, the incompleteEnodeURL contains the node ID (e.g. something like this:  `enode://352352235235235...@`.

One suggestion is to make the field more structured in that you replace it with the nodeID.

PR #325, opened 2019-07-09
Recovering Randomness

nicholasguo:
### Description

Currently randomness is done through a commit and reveal scheme in which a validator generates some randomness, adds a commitment to the block, stores the randomness in their db, then reveals the randomness at a later time. However, if the db is corrupted or lost before they can reveal, then the randomness is lost.

This PR changes the way randomness is determined to make it deterministic based on the validator's private key and current block, so that it can be recovered. More specifically, the randomness at a particular block will be hash(sign(some string) + block.ParentHash).

Design Doc: https://docs.google.com/document/d/1ko0OKbkHC0RXD_gXCekl-aEmQEqvrHW85kLQP6-tIww/edit?usp=sharing

### Tested

Migrations ran successfully, produced random values appeared random on inspection
Temporarily added logic that corrupts randomness data from being stored in the db, randomness values recovered still match the commitments

e2e tests: https://github.com/celo-org/celo-monorepo/pull/86

### Other changes

Added field to randomness contract 

### Related issues

- Fixes https://github.com/celo-org/geth/issues/282


core/random.go
```diff
@@ -146,33 +145,32 @@ func (r *Random) GetLastRandomness(coinbase common.Address, db *ethdb.Database,
 		return common.Hash{}, nil
 	}
 
-	randomness := common.Hash{}
-	randomnessSlice, err := (*db).Get(commitmentDbLocation(lastCommitment))
+	parentBlockHashBytes, err := (*db).Get(commitmentDbLocation(lastCommitment))
```
m-chrzan: How are we dealing with setting first randomness to 0 right now?

nicholasguo: If the smart contract does not contain the a previous commitment, the randomness of 0 is returned. In this case its value is common.Hash{}. https://github.com/celo-org/celo-blockchain/pull/325/files#diff-55aa515bdb3e537871a9f3140b9a65cdL146

core/random.go
```diff
@@ -146,33 +145,32 @@ func (r *Random) GetLastRandomness(coinbase common.Address, db *ethdb.Database,
...
-	randomness := common.BytesToHash(randomBytes[:])
+	randomness := crypto.Keccak256Hash(append(seed, header.ParentHash.Bytes()...))
 	// TODO(asa): Make an issue to not have to do this via StaticCall
-	_, err = r.iEvmH.MakeStaticCall(*r.address(), computeCommitmentFuncABI, "computeCommitment", []interface{}{randomness}, &commitment, gasAmount, header, state)
-	err = (*db).Put(commitmentDbLocation(commitment), randomness[:])
+	_, err := r.iEvmH.MakeStaticCall(*r.address(), computeCommitmentFuncABI, "computeCommitment", []interface{}{randomness}, &commitment, gasAmount, header, state)
+	err = (*db).Put(commitmentDbLocation(commitment), header.ParentHash.Bytes())
 	if err != nil {
-		log.Error("Failed to save randomness to the database", "err", err)
+		log.Error("Failed to save last block proposed to the database", "err", err)
```
asaj: nit: We're not saving the last block proposed here, are we? We're saving the parenthash of that block

PR #365, opened 2019-07-29
Separate phone number decryption to a separate key

### LINKED ISSUE #321: Validators SBAT specify a different attestations encryption key
### Expected Behavior

If we don't move the reveal off-chain, we should allow validators to use a separate key for decrypting phone numbers

### Current Behavior

Validators currently use their consensus private key to decrypt phone numbers sent to them in reveal txs

nambrot:
### Description

This change allows validators to specify the account with which phone numbers in reveal transactions should be decrypted.

### Tested

- [ ] On a testnet
- [ ] TOFIX: I didn't actually change the decryption, i changed the account to identify if the attestation request goes to our miner. I should actually pass through the decryption

### Related issues

- Fixes #321 

### Backwards compatibility

- We will no longer decrypt phone numbers with the coinbase account

### DIFF
--- a/abe/abe.go
+++ b/abe/abe.go
@@ -73,8 +73,8 @@ func sendSms(phoneNumber string, message string, account common.Address, issuer 
 	return err
 }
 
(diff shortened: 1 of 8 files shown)


nambrot:
No longer relevant

PR #366, opened 2019-07-30
Adds Prepared Certificates to ensure Istanbul liveness

### LINKED ISSUE #116: Devs SBAT be sure of Istanbul liveness
Istanbul has a liveness issue described in:
https://arxiv.org/pdf/1901.07160.pdf
https://drive.google.com/open?id=1v5kOvu5eppGHXdHqHoCIUMVmV7eAsDwpLljDtvVBR_U

and also described here:
https://github.com/ethereum/EIPs/issues/650#issuecomment-311836076

Istanbul learns of new blocks via miner/worker.go line 428, which results in a new FinalCommittedEvent being posted to the engine, handled at backend/handler.go line 123. Currently this just starts a new round.

trianglesphere:
### Description

This PR removes the locking mechanism in IBFT and adds:
- Optional PREPARED certificates to ROUND CHANGE messages, which are included when the node has seen 2F+1 PREPARE/COMMIT messages or a PREPARED certificate for this sequence.
- ROUND CHANGE certificates to PREPREPARE messages, which are required for rounds > 0. If any of the messages in this certificate contain a PREPARED certificate, the PREPREPARE must propose the PREPARED block with the highest round number.

In addition, this overhauls the round change logic:
- On `timeout` or `f+1` round change messages, transition to `StateWaitingForNewRound` and if the new desired round (one past the timeout round or the RC message view), it will set the new desired round, send a round change message for the desired round, set a timeout for the desired round. (All these actions done in `core.waitForDesiredRound`)
- On `quorum RC messages` or a `round change certificate` where the round associated with the messages is >= desired round, start a new round at the desired round and proceed with the normal consensus protocol.

This addresses a bug in which F nodes could lock onto block A, while 2F nodes could lock onto block B, causing a permanent loss of liveness. 

This PR also changes how the randomness source is verified to fix an issues with validators correctly re-proposing blocks.

### Tested

Unit tests in consensus pass.
Testnets for old versions were working, going to spin up a testnet with latest.

Testing Changes:
* Changing the round change message payload is needed to match what validators
actually do.
* errIgnored -> nil exists b/c I no longer return errIgnored from handleRoundChange
* Check for consistent commits in `TestCommitsBlockAfterRoundChange` and `TestPreparedCertificatePersistsThroughRoundChange` because prepared certificates are generated, but no single node has committed, so it is fine to commit a different block than the original block from validator 0 as long as there is not a fork.

### Other changes

- Verify the CommittedSeal included in COMMIT messages 
- Adds create empty block with empty randomness. Used for a null value when creating a prepared certificate.
- Lots of testing changes, logging + match other changes
- Removes `src instanbul.Validator` from being passed around endlessly
- Namespaces istanbul types
- BLS changes merged in (some slight changes in verifying commit seal)
- More logging. Testing only shows `core.current` is `nil` in `startNewRound` and not in the message handlers, but I still included `nil` checks there.

### Related issues

- Fixes #116 

### Backwards compatibility
No.


asaj:
Can you include more information in the PR description? A good place to start would be the description of the stale liveness PR

consensus/istanbul/core/types.go
```diff
@@ -69,94 +66,6 @@ func (s State) Cmp(y State) int {
 	return 0
 }
 
-const (
```
trianglesphere: Moved to `consensus/istanbul/types.go`. TBH not sure why.

consensus/istanbul/core/roundchange.go
```diff
@@ -48,45 +48,125 @@ func (c *core) sendRoundChange(round *big.Int) {
...
 		Msg:  payload,
 	})
 }
 
-func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
-	logger := c.logger.New("state", c.state, "from", src.Address().Hex())
+func (c *core) handleRoundChangeCertificate(roundChangeCertificate istanbul.RoundChangeCertificate) error {
+	logger := c.logger.New("state", c.state, "cur_round", c.current.Round(), "cur_seq", c.current.Sequence(), "func", "handleRoundChangeCertificate")
+
+	if len(roundChangeCertificate.RoundChangeMessages) > c.valSet.Size() || len(roundChangeCertificate.RoundChangeMessages) < 2*c.valSet.F()+1 {
```
trianglesphere: @kevjue this might be a manual quorum size that sneaks in

consensus/istanbul/core/roundchange.go
```diff
@@ -170,3 +251,21 @@ func (rcs *roundChangeSet) MaxRound(num int) *big.Int {
 	}
 	return maxRound
 }
+
+func (rcs *roundChangeSet) getCertificate(r *big.Int, f int) (istanbul.RoundChangeCertificate, error) {
+	rcs.mu.Lock()
+	defer rcs.mu.Unlock()
+
+	round := r.Uint64()
+	if rcs.roundChanges[round].Size() > 2*f {
```
trianglesphere: I think that this `2F` comes from PBFT directly. The IBFT correctness paper does not touch on this, but I'd like to double check if this should be a quorum instead.

consensus/istanbul/core/prepare.go
```diff
@@ -19,49 +19,124 @@ package core
...
+	logger.Trace("Accepted prepare", "Number of prepares or commits", c.current.GetPrepareOrCommitSize())
 
-	// Change to Prepared state if we've received enough PREPARE messages or it is locked
-	// and we are in earlier state before Prepared state.
-	if ((c.current.IsHashLocked() && prepare.Digest == c.current.GetLockedHash()) || c.current.GetPrepareOrCommitSize() > 2*c.valSet.F()) &&
-		c.state.Cmp(StatePrepared) < 0 {
-		c.current.LockHash()
+	// Change to Prepared state if we've received enough PREPARE messages and we are in earlier state
+	// before Prepared state.
+	if (c.current.GetPrepareOrCommitSize() > 2*c.valSet.F()) && c.state.Cmp(StatePrepared) < 0 {
```
trianglesphere: Another >= quorum?

trianglesphere:
Quorum sizing should be fixed (only rely on `minQuorumSize()` for quorums). `.F()` is used for `F+1` for round change conditions, but that's expected/fine.

consensus/istanbul/core/prepare.go
```diff
@@ -19,49 +19,124 @@ package core
...
-		Code: msgPrepare,
+	logger.Trace("Sending prepare")
+	c.broadcast(&istanbul.Message{
+		Code: istanbul.MsgPrepare,
 		Msg:  encodedSubject,
 	})
 }
 
-func (c *core) handlePrepare(msg *message, src istanbul.Validator) error {
+func (c *core) handlePreparedCertificate(preparedCertificate istanbul.PreparedCertificate) error {
```
asaj: Do we process the messages in the `preparedCertificate`? e.g. if somehow the preparedCertificate contained 2F+1 commit messages, would we then commit the block?

consensus/istanbul/core/prepare.go
```diff
@@ -19,49 +19,124 @@ package core
...
+	logger.Trace("Accepted prepare", "Number of prepares or commits", c.current.GetPrepareOrCommitSize())
 
-	// Change to Prepared state if we've received enough PREPARE messages or it is locked
-	// and we are in earlier state before Prepared state.
-	if ((c.current.IsHashLocked() && prepare.Digest == c.current.GetLockedHash()) || c.current.GetPrepareOrCommitSize() >= c.valSet.MinQuorumSize()) &&
-		c.state.Cmp(StatePrepared) < 0 {
-		c.current.LockHash()
+	// Change to Prepared state if we've received enough PREPARE messages and we are in earlier state
+	// before Prepared state.
+	if (c.current.GetPrepareOrCommitSize() >= c.valSet.MinQuorumSize()) && c.state.Cmp(StatePrepared) < 0 {
```
asaj: Consider calling `c.current.GetPrepareOrCommitSize()` and `c.valSet.MinQuorumSize()` only once

consensus/istanbul/core/prepare_test.go
```diff
@@ -28,6 +28,62 @@ import (
...
+	N := uint64(4) // replica 0 is the proposer, it will send messages to others
+	F := uint64(1)
+	sys := NewTestSystemWithBackend(N, F)
+	view := istanbul.View{
+		Round:    big.NewInt(0),
+		Sequence: big.NewInt(1),
+	}
+	proposal := makeBlock(0)
+
+	testCases := []struct {
```
asaj: Are there any other error cases we're missing?

consensus/istanbul/core/prepare_test.go
```diff
@@ -28,6 +28,62 @@ import (
...
+			c := backend.engine.(*core)
+			err := c.handlePreparedCertificate(test.certificate)
+			if err != test.expectedErr {
+				t.Errorf("error mismatch: have %v, want %v", err, test.expectedErr)
+			}
+		}
+	}
+}
+
+// TODO(asa): Test with PREPARED certificate
```
asaj: How hard would it be to add a test case for this?

consensus/istanbul/core/preprepare.go
```diff
@@ -55,9 +57,20 @@ func (c *core) handlePreprepare(msg *message, src istanbul.Validator) error {
 		return errFailedDecodePreprepare
 	}
 
+	// If round > 0, handle the ROUND CHANGE certificate.
```
asaj: Should we require that the ROUND CHANGE certificate is empty if round == 0?

consensus/istanbul/core/preprepare.go
```diff
@@ -75,20 +89,30 @@ func (c *core) handlePreprepare(msg *message, src istanbul.Validator) error {
...
 	// Check if the message comes from current proposer
-	if !c.valSet.IsProposer(src.Address()) {
+	if !c.valSet.IsProposer(msg.Address) {
 		logger.Warn("Ignore preprepare messages from non-proposer")
 		return errNotFromProposer
 	}
 
+	if preprepare.View.Round.Cmp(common.Big0) > 0 {
+		// If we have a PREPARED certificate after handling the ROUND CHANGE, the proposal must match.
+		// TODO(asa): Does it make a difference if this PREPARED certificate came from the ROUND CHANGE certificate vs
```
asaj: Wdyt @trianglesphere?

consensus/istanbul/core/preprepare.go
```diff
@@ -98,27 +122,12 @@ func (c *core) handlePreprepare(msg *message, src istanbul.Validator) error {
 		return err
 	}
 
-	// Here is about to accept the PRE-PREPARE
+	// TODO(asa): Can we skip to COMMIT if we have a PREPARED certificate already?
```
asaj: Wdyt @trianglesphere?

consensus/istanbul/core/preprepare_test.go
```diff
@@ -55,85 +56,139 @@ func TestHandlePreprepare(t *testing.T) {
...
+				roundChangeCertificate := sys.getRoundChangeCertificate(t, *(sys.backends[0].engine.(*core).currentView()), istanbul.EmptyPreparedCertificate())
+				roundChangeCertificate.RoundChangeMessages[1] = roundChangeCertificate.RoundChangeMessages[0]
+				return roundChangeCertificate
+			},
+			makeBlock(1),
+			errInvalidRoundChangeCertificateDuplicate,
+			false,
+		},
+		{
+			// ROUND CHANGE certificate contains PREPARED certificate for a different block.
```
asaj: Can we add a test case with a valid ROUND CHANGE certificate? And one with a valid ROUND CHANGE certificate that contains a valid PREPARED certificate?

consensus/istanbul/core/roundchange.go
```diff
@@ -48,45 +48,125 @@ func (c *core) sendRoundChange(round *big.Int) {
...
+	logger.Trace("Sending round change message", "rcs", c.roundChangeSet)
+	c.broadcast(&istanbul.Message{
+		Code: istanbul.MsgRoundChange,
 		Msg:  payload,
 	})
 }
 
-func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
-	logger := c.logger.New("state", c.state, "from", src.Address().Hex())
+func (c *core) handleRoundChangeCertificate(roundChangeCertificate istanbul.RoundChangeCertificate) error {
```
asaj: Ditto, does this make more sense in `prepare.go`? I wasn't sure when I initially wrote this which file it should live in

consensus/istanbul/core/roundchange.go
```diff
@@ -48,45 +48,125 @@ func (c *core) sendRoundChange(round *big.Int) {
...
+
+		decodedMessages[i] = *roundChange
+	}
+
+	for i, message := range roundChangeCertificate.RoundChangeMessages {
+		// No need to process ROUND CHANGE messages we've seen before.
+		if !c.backend.IsKnownMessage(message) {
+			err := c.handleDecodedCheckedRoundChange(&message, &decodedMessages[i])
+			// We want to continue to process ROUND CHANGE messages if they're for future rounds.
+			// TODO(asa): Should we just process every message regardless of if there's an error?
```
asaj: Wdyt @trianglesphere?

consensus/istanbul/core/roundchange.go
```diff
@@ -48,45 +48,125 @@ func (c *core) sendRoundChange(round *big.Int) {
...
+	logger := c.logger.New("state", c.state, "from", msg.Address, "cur_round", c.current.Round(), "cur_seq", c.current.Sequence(), "func", "handleDecodedCheckedRoundChange", "tag", "handleMsg")
+
 	cv := c.currentView()
 	roundView := rc.View
 
+	// Handle the PREPARED certificate if present.
+	var num int
+	if rc.HasPreparedCertificate() {
+		if err := c.handlePreparedCertificate(rc.PreparedCertificate); err != nil {
+			// TODO(asa): Should we still accept the round change message without the certificate if this fails?
```
asaj: Wdyt @trianglesphere? My hunch is no?

consensus/istanbul/core/roundchange.go
```diff
@@ -95,6 +175,7 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
 	// try to catch up the round number.
 	if c.waitingForRoundChange && num == c.valSet.F()+1 {
 		if cv.Round.Cmp(roundView.Round) < 0 {
+			// TODO(asa): If we saw a PREPARED certificate, do we include one in our round change message?
```
asaj: Wdyt @trianglesphere? Looks like I left a lot of open questions 🤦‍♂

consensus/istanbul/core/roundchange_test.go
```diff
@@ -91,3 +92,371 @@ func TestRoundChangeSet(t *testing.T) {
...
+		if !ok {
+			t.Error("Error reading block")
+		}
+		// Wait for all backends to finalize the block.
+		<-time.After(2 * time.Second)
+		for i, b := range sys.backends {
+			// TODO(asa): Do something with proposer here
+			committed, _ := b.LastProposal()
+			// We expect to commit the block proposed by the first proposer.
+			// TODO(joshua): This can be flaky in my local tests.
```
asaj: Any insights into this?

consensus/istanbul/core/testbackend_test.go
```diff
@@ -323,6 +419,54 @@ func (t *testSystem) NewBackend(id uint64) *testSystemBackend {
 	return backend
 }
 
+func (t *testSystem) F() uint64 {
+	return t.f
+}
+
+func (t *testSystem) MinQuorumSize() uint64 {
+	// TODO(joshua) Fix this to be based off of total number of validators.
```
asaj: +1, this seems worth doing

consensus/istanbul/core/types.go
asaj: I think it needed to be used in `backend` and `core` and there was some issue with circular dependencies

consensus/istanbul/backend/backend.go
```diff
@@ -298,11 +298,17 @@ func (sb *Backend) Verify(proposal istanbul.Proposal, src istanbul.Validator) (t
...
 	}
 
-	// verify the header of proposed block
-	if block.Header().Coinbase != src.Address() {
+	// The author should be the first person to propose the block to ensure that randomness matches up.
+	addr, err := sb.Author(block.Header())
+	if err != nil {
+		sb.logger.Error("Could not recover orignal author of the block to verify the randomness", "err", err, "func", "Verify")
+		return 0, errInvalidProposal
+	} else if addr != block.Header().Coinbase {
```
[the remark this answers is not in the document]
trianglesphere: The coinbase was co-opted by randomness to the source of the randomness. We're checking that coinbase matches the proposer (only can propose randomness for yourself). We need to use `Author` or `msg.Address` (on the proposal stored in the prepared certificate) to deal with safely re-proposing blocks through round changes.

consensus/istanbul/core/backlog.go
```diff
@@ -75,10 +75,10 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 	return nil
 }
 
-func (c *core) storeBacklog(msg *message, src istanbul.Validator) {
-	logger := c.logger.New("from", src, "state", c.state)
+func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
+	logger := c.logger.New("from", msg.Address, "state", c.state)
 
-	if src.Address() == c.Address() {
+	if msg.Address == c.Address() {
```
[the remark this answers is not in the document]
trianglesphere: I'd like to get rid of `src instanbul.Validator` everywhere, but didn't get around to it here b/c the backlog uses `src` as the key rather than `src.Address`

consensus/istanbul/core/core.go
```diff
@@ -208,9 +207,9 @@ func (c *core) commit() {
 func (c *core) startNewRound(round *big.Int) {
 	var logger log.Logger
 	if c.current == nil {
-		logger = c.logger.New("old_round", -1, "old_seq", 0)
+		logger = c.logger.New("old_round", -1, "old_seq", 0, "func", "startNewRound", "tag", "stateTransition")
```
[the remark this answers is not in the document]
trianglesphere: Yea, `startNewRound` really is the main outlier

consensus/istanbul/core/commit.go
```diff
@@ -72,30 +75,32 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
...
-	err = blscrypto.VerifySignature(validator.BLSPublicKey(), seal, []byte{}, msg.CommittedSeal, false)
-	if err != nil {
-		return err
+	if err := c.verifyCommittedSeal(commit.Digest, msg.CommittedSeal, validator); err != nil {
+		return errInvalidCommittedSeal
 	}
 
-	c.acceptCommit(msg, src)
+	c.acceptCommit(msg)
+	logger.Trace("Accepted commit", "Number of commits", c.current.Commits.Size())
```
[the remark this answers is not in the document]
trianglesphere: It could be. I think it's fine here. Particularly that it's recording how many commits it has each time it gets a new commit.

consensus/istanbul/core/roundchange_test.go
trianglesphere: Unfortunately not. I just saw it be flaky.

consensus/istanbul/backend/backend.go
```diff
@@ -274,7 +274,7 @@ func (sb *Backend) EventMux() *event.TypeMux {
 }
 
 // Verify implements istanbul.Backend.Verify
-func (sb *Backend) Verify(proposal istanbul.Proposal, src istanbul.Validator) (time.Duration, error) {
+func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
```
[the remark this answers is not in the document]
trianglesphere: It's done as part of the general consensus message checking. `preprepares` are checked that the validator sending it is the proposer, and then from there only messages that extend that `preprepare` are acted upon thus ensuring that the original validator was the proposer. It is true that this info seems to be lost once the block is produced.

consensus/istanbul/core/preprepare.go
trianglesphere: Probably should to keep it cleaner, but I don't think it hurts either way.

consensus/istanbul/core/prepare.go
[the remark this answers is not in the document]
trianglesphere: Could be, not a really a priority for me in this PR.

trianglesphere: No we don't. I don't think there is a case where the certificate should have 2F+1 commits, but even in that case, I don't think we lose liveness by not committing in that case.

consensus/istanbul/core/preprepare.go
trianglesphere: I don't think that it makes a difference.

When a node handles round change messages, it resets its `prepared certificate`, thus the proposals should match. I think we may need to double check to see that we make sure we start the new round at the correct time.

consensus/istanbul/core/preprepare.go
trianglesphere: Not sure. My gut feeling is no, but I'd need some time to know either way. Feels out of scope for now.

consensus/istanbul/core/roundchange.go
trianglesphere: I'm mixed. I think that it could be fine **given** the validator sending the new round is honest, that validator can ensure that it won't fail.

I'm starting to think that we might need some specific logic around `Round Change Certificates` (`view-change` from PBFT) that treats them differently than `round change` or `new view` messages. (In particular pick out the `prepared certificate` with the highest `view` or `round` as the `prepared certificate` that the validator should lock on.

consensus/istanbul/core/roundchange.go
trianglesphere: I think it doesn't really matter given again **honest** nodes **should not** send invalid prepared certificates. I think that this could open up a minor risk of byzantine nodes trying to split honest node's rounds. I also think cutting the `prepared certificate` when invalid shouldn't hurt much.

consensus/istanbul/core/roundchange.go
trianglesphere: For this we actually do include that `prepared certificate`. Again I feel like it doesn't make a huge difference either way if it's included or not.

consensus/istanbul/core/prepare_test.go
trianglesphere: Missing:
* `errInvalidPreparedCertificateProposal`
* `errInvalidPreparedCertificateMsgSignature`
* Plus errors directly return from `message.PayloadNoSig`, `c.validateFn`, and `c.verifyCommittedSeal`

consensus/istanbul/core/prepare_test.go
trianglesphere: Apparently not too hard.

trianglesphere:
The latest batch of commits is pretty big, but hopefully it simplifies stuff (and still works).
Some of the biggest changes are around the implicit state machine wrt round changes and how messages are dealt with.
* New Round Messages: the view of the message is the view that the validator wants to transition to. Therefore it is rejected as old if it matches the current view.
* Pre-prepares: must have a valid round change certificate if round != 0. Future pre-prepares are considered and the round change certificate can cause the validator to round change to it's round.
* On the first of `timeout` or `f+1 round change` messages for a certain round a node will send a round change message, but not change rounds. As it is waiting for a new round it rejects all messages that are not round change messages as future messages (including pre-prepares).
* On either `quorum round change messages` or a `round change certificate` it will complete the round change (even if not waiting for it).
* It is okay for the node to try to start a new round with the same round number as it's current round, it just won't do anything. This is done to simplify `handleRoundChangeCertificate`.

consensus/istanbul/core/request.go
```diff
@@ -33,8 +41,9 @@ func (c *core) handleRequest(request *istanbul.Request) error {
 	logger.Trace("handleRequest", "number", request.Proposal.Number(), "hash", request.Proposal.Hash())
 
 	c.current.pendingRequest = request
-	if c.state == StateAcceptRequest {
-		c.sendPreprepare(request)
+	// Must go through startNewRound to send a proposal that matches the round change certificate if not on the first round.
```
asaj: The round change certificate does not necessarily specify a proposal, right?

consensus/istanbul/core/roundchange.go
```diff
@@ -32,79 +32,151 @@ func (c *core) sendNextRoundChange() {
...
 
 	cv := c.currentView()
 	if cv.Round.Cmp(round) >= 0 {
-		logger.Error("Cannot send out the round change", "current round", cv.Round, "target round", round)
+		logger.Error("Cannot send out the round change")
 		return
 	}
 
-	c.catchUpRound(&istanbul.View{
+	nextView := &istanbul.View{
```
asaj: Why don't we need to `catchUpRound` here anymore?

consensus/istanbul/core/roundchange.go
```diff
@@ -32,79 +32,151 @@ func (c *core) sendNextRoundChange() {
...
-	if c.waitingForRoundChange && num == c.valSet.F()+1 {
-		if cv.Round.Cmp(roundView.Round) < 0 {
-			c.sendRoundChange(roundView.Round)
-		}
-		return nil
-	} else if num == c.valSet.MinQuorumSize() && (c.waitingForRoundChange || cv.Round.Cmp(roundView.Round) < 0) {
-		// We've received the minimum quorum size ROUND CHANGE messages, start a new round immediately.
+	// On f+1 round changes we send a round change and wait for the next round if we haven't done so already
+	// On quorum round change messages we go to the next round immediately.
+	if num == c.valSet.F()+1 && !c.waitingForNewRound {
```
asaj: Why go from `c.waitingForRoundChange` to `!c.waitingForNewRound` here?

consensus/istanbul/core/core.go
```diff
@@ -263,52 +272,44 @@ func (c *core) startNewRound(round *big.Int) {
...
+		if request != nil {
+			c.sendPreprepare(request, roundChangeCertificate)
 		}
 	}
 	c.newRoundChangeTimer()
 
 	logger.Debug("New round", "new_round", newView.Round, "new_seq", newView.Sequence, "new_proposer", c.valSet.GetProposer(), "valSet", c.valSet.List(), "size", c.valSet.Size(), "isProposer", c.isProposer())
 }
 
-func (c *core) catchUpRound(view *istanbul.View) {
```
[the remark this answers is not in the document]
trianglesphere: I rolled this functionality into `startNewRound` because this would basically cause a round change except it would not update the next proposer which was causing a couple of issues.

consensus/istanbul/core/core.go
```diff
@@ -263,52 +274,54 @@ func (c *core) startNewRound(round *big.Int) {
...
-			c.sendPreprepare(r)
-		} else if c.current.pendingRequest != nil {
-			c.sendPreprepare(c.current.pendingRequest)
+		}
+		if request != nil {
+			c.sendPreprepare(request, roundChangeCertificate)
 		}
 	}
 	c.newRoundChangeTimer()
+	// If going into an immediate waiting state, need round change timer for the next round & need to send the round change message.
```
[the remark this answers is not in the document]
trianglesphere: The round change timer on the previous line carries the current round view, but if we've flagged ourselves as in a waiting for round change state we'll ignore round change timeouts from our current round (there is a race condition if we go into a waiting state on f+1 round change messages and get a timeout for our current round which causes us to jump ahead an extra round), so we need to register a timeout for the next round.

It's part of the round change implicit state machine.

consensus/istanbul/core/roundchange.go
```diff
@@ -32,79 +32,151 @@ func (c *core) sendNextRoundChange() {
...
+		}
+
+		var roundChange *istanbul.RoundChange
+		if err := message.Decode(&roundChange); err != nil {
+			logger.Error("Failed to decode ROUND CHANGE in certificate", "err", err)
+			return err
+		}
+
+		// Verify ROUND CHANGE message is for a proper view
+		// TODO(joshua): May be able to relax to the proposal is >= than the round change message.
```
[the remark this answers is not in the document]
trianglesphere: This is actually a little bit of me thinking ahead to buffering round change messages and not super important for this. I think it'd work if we move into the minimum view.

consensus/istanbul/core/roundchange.go
```diff
@@ -32,79 +32,151 @@ func (c *core) sendNextRoundChange() {
...
+				return errInvalidPreparedCertificateDigestMismatch
+			}
+
+			if err := c.verifyPreparedCertificate(roundChange.PreparedCertificate); err != nil {
+				return err
+			}
+		}
+
+		decodedMessages[i] = *roundChange
+		// TODO(joshua): startNewRound needs these round change messages to generate a
```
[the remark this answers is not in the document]
trianglesphere: https://github.com/celo-org/celo-blockchain/blob/trianglesphere/istanbul-liveness/consensus/istanbul/core/core.go#L256 checks if we can generate a prepared certificate (which requires 2f+1 round change messages). If it can't it assumes that it's waiting for a round change (even if it is trying to start a new round because it got a prepared certificate) which causes it to block.

Also part of the implicit round change state machine causing problems.

consensus/istanbul/core/request.go
trianglesphere: No it doesn't, but we check every message in it to see if there is a proposal we must use, and then if there isn't we can use the pendingRequest.

It's less start new round and more the logic that selects what the next pre-prepare should be (which is in start new round, but could be refactored out).

consensus/istanbul/core/roundchange.go
trianglesphere: I've placed the responsibility of doing what `catchUpRound` did (bumping waiting round) with whoever calls `sendRoundChange`

consensus/istanbul/core/handler.go
```diff
@@ -109,11 +109,14 @@ func (c *core) handleEvents() {
 					c.logger.Error("Error in handling istanbul message that was sent from a backlog event", "err", err)
 				}
 			}
-		case _, ok := <-c.timeoutSub.Chan():
+		case event, ok := <-c.timeoutSub.Chan():
 			if !ok {
 				return
 			}
-			c.handleTimeoutMsg()
+			switch ev := event.Data.(type) {
```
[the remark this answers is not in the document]
trianglesphere: No clue really, I just copied what other cases did to get the data from the event.

consensus/istanbul/core/backlog.go
```diff
@@ -57,14 +58,15 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 		return errOldMessage
 	}
 
-	if c.waitingForRoundChange {
+	// Round change messages are already let through.
+	if c.waitingForNewRound {
```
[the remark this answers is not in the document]
trianglesphere: It was clearer to me that it's waiting for a new round round-change rather than a round-change message.

consensus/istanbul/core/backlog.go
```diff
@@ -40,10 +40,11 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 		return errInvalidMessage
 	}
 
-	if msgCode == msgRoundChange {
+	// Round change messages should be in the same sequence but have a larger round.
+	if msgCode == istanbul.MsgRoundChange {
 		if view.Sequence.Cmp(c.currentView().Sequence) > 0 {
 			return errFutureMessage
-		} else if view.Cmp(c.currentView()) < 0 {
+		} else if view.Cmp(c.currentView()) <= 0 {
```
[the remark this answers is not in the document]
trianglesphere: Round change messages carry the view that they'd like the nodes to move to (and thus are not in yet). Also part of the round change state machine.

consensus/istanbul/core/core.go
```diff
@@ -263,52 +304,47 @@ func (c *core) startNewRound(round *big.Int) {
...
-		// If we have pending request, propose pending request
-		if c.current.IsHashLocked() {
-			r := &istanbul.Request{
-				Proposal: c.current.Proposal(), //c.current.Proposal would be the locked proposal by previous proposer, see updateRoundState
-			}
-			c.sendPreprepare(r)
-		} else if c.current.pendingRequest != nil {
-			c.sendPreprepare(c.current.pendingRequest)
-		}
+	if roundChange && c.isProposer() && c.current != nil && request != nil {
```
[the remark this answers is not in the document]
trianglesphere: In `handleRequest`

consensus/istanbul/core/core.go
```diff
@@ -263,52 +304,47 @@ func (c *core) startNewRound(round *big.Int) {
...
-	c.newRoundChangeTimer()
+	logger.Debug("Waiting for desired round")
 
-	logger.Trace("Catch up round", "new_round", view.Round, "new_seq", view.Sequence, "new_proposer", c.valSet)
+	desiredView := &istanbul.View{
+		Sequence: new(big.Int).Set(c.current.Sequence()),
+		Round:    new(big.Int).Set(r),
+	}
+	// Perform all of the updates
+	c.setState(StateWaitingForNewRound)
```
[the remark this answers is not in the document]
trianglesphere: There's two components to this state: `desiredRound` and `WaitingForNewRound`. Because we can timeout in `StateWaitingForNewRound`, we should not bail out if we're already in the waiting state, but **should** bail if the given desired round is not greater than new desired round.

`StateWaitingForNewRound` is pretty much only used to reject incoming messages.

consensus/istanbul/core/core.go
```diff
@@ -263,52 +304,47 @@ func (c *core) startNewRound(round *big.Int) {
...
+	c.current.SetDesiredRound(r)
+	// TODO(joshua): Double check that this works with skipped proposers
```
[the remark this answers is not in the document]
trianglesphere: This was the same general concern about the statefulness of who the proposer is. Double checking this, I think it's fine. Last proposer is just the proposer of the last block, and I think `CalcProposer` just uses that + round number.

consensus/istanbul/core/roundchange.go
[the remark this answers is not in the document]
trianglesphere: My bad, `prepared certificate` should be `round change certificate`. Thanks for catching this.

consensus/istanbul/core/handler.go
```diff
@@ -165,37 +180,28 @@ func (c *core) handleCheckedMsg(msg *istanbul.Message, src istanbul.Validator) e
...
-	// if the max round is larger than current round.
-	if !c.waitingForRoundChange {
-		maxRound := c.roundChangeSet.MaxRound(c.valSet.F() + 1)
-		if maxRound != nil && maxRound.Cmp(c.current.Round()) > 0 {
-			c.sendRoundChange(maxRound)
-			return
-		}
-	}
+func (c *core) handleTimeoutMsg(timeoutView *istanbul.View) {
+	// logger := c.logger.New("func", "handleTimeoutMsg", "round", timeoutView.Round)
```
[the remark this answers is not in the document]
trianglesphere: I'm not really sure. Perhaps that receiving a committed block for a large sequence does not end the instance of the IBFT protocol for that height and does not start the next. It does seem a little odd.

consensus/istanbul/core/backlog.go
```diff
@@ -40,10 +40,11 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 		return errInvalidMessage
 	}
 
+	// Round change messages should be in the same sequence but be after the desired round
```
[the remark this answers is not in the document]
trianglesphere: They actually can be. This comment wasn't updated when I switched to the state machine like design.

consensus/istanbul/core/preprepare.go
```diff
@@ -66,59 +88,43 @@ func (c *core) handlePreprepare(msg *istanbul.Message, src istanbul.Validator) e
...
 			// 1. The proposer needs to be a proposer matches the given (Sequence + Round)
 			// 2. The given block must exist
-			if valSet.IsProposer(src.Address()) && c.backend.HasProposal(preprepare.Proposal.Hash(), preprepare.Proposal.Number()) {
+			if valSet.IsProposer(msg.Address) && c.backend.HasProposal(preprepare.Proposal.Hash(), preprepare.Proposal.Number()) {
+				logger.Trace("Sending a commit message for an old block", "view", preprepare.View, "block hash", preprepare.Proposal.Hash())
 				c.sendCommitForOldBlock(preprepare.View, preprepare.Proposal.Hash())
 				return nil
 			}
 		}
+		// Already in the next round b/c handleRoundChangeCertificate moves us to the next round.
```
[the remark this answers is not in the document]
trianglesphere: We basically shouldn't be able to get there.
If the round change certificate was valid, we move to the view associated with it.
If that view is old, we don't transition and we can get an old message error.
If that view is in the future, we move to it and thus the pre-prepare should have the same view and not `errFutureMessage`
If the round change certificate was invalid, we would return an error earlier and not get to this part of the code.

consensus/istanbul/core/roundchange.go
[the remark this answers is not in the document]
trianglesphere: Let's talk offline about this. This comment isn't about this PR, but adapting `istanbul-future-rc` to liveness

consensus/istanbul/core/roundchange.go
```diff
@@ -32,79 +32,158 @@ func (c *core) sendNextRoundChange() {
...
+			return errInvalidRoundChangeCertificateMsgView
+		}
+
+		if roundChange.HasPreparedCertificate() {
+			if err := c.verifyPreparedCertificate(roundChange.PreparedCertificate); err != nil {
+				return err
+			}
+			// The prepared certificate with the highest round number carries the proposal we must use
+			preparedView := roundChange.PreparedCertificate.View()
+			if preparedView != nil && preparedView.Round.Cmp(maxRound) > 0 {
```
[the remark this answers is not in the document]
trianglesphere: Good catch. My feeling is that we should reject the round change certificate for having an invalid round change message. I'd probably also need to change `handleRoundChange` to verify the prepared certificate view against the round change message view.

consensus/istanbul/backend/backend.go
```diff
@@ -481,6 +487,19 @@ func (sb *Backend) HasBadProposal(hash common.Hash) bool {
 	return sb.hasBadBlock(hash)
 }
 
+func (sb *Backend) IsKnownMessage(msg istanbul.Message) bool {
```
trianglesphere: This is no longer used. Should it be removed?

consensus/istanbul/core/commit.go
```diff
@@ -72,30 +75,37 @@ func (c *core) handleCommit(msg *istanbul.Message, src istanbul.Validator) error
...
 
 	// Commit the proposal once we have enough COMMIT messages and we are not in the Committed state.
 	//
 	// If we already have a proposal, we may have chance to speed up the consensus process
 	// by committing the proposal without PREPARE messages.
-	if c.current.Commits.Size() >= c.valSet.MinQuorumSize() && c.state.Cmp(StateCommitted) < 0 {
-		// Still need to call LockHash here since state can skip Prepared state and jump directly to the Committed state.
-		c.current.LockHash()
+	// TODO(joshua): Remove state comparisons (or change the cmp function)
+	if numberOfCommits >= minQuorumSize && c.state.Cmp(StateCommitted) < 0 {
```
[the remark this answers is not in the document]
trianglesphere: I don't think so. I've added `StateWaitingForNewRound` at the end, and in this state, we don't want to do anything with committing or preparing. In theory, `checkMessage` should reject all messages when in `StateWaitingForNewRound` so this check should just be that we have not double committed or (for the one below) have not already got a prepared certificate.

consensus/istanbul/backend/backend.go
asaj: Why is it no longer used?

trianglesphere: I removed the check against known messages (perhaps incorrectly) when adding round change messages to the round change message set in `handleRoundChangeCertificate`

consensus/istanbul/core/preprepare.go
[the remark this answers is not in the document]
asaj: Hmm okay I think I was just confused by the comment, consider making it more verbose/explicit

trianglesphere:
@timmoreton @asaj Are you ready yet?

Issue #362, opened 2019-07-24
Developers SBAT interact with the validator contract from the blockchain via a contract_comm interface


jarmg:
### Expected Behavior
Interaction with the validator contract should use the function collection pattern established in `contract_comm` for other contract interactions. This pattern was established in the iEvmH build in #317 

### Current Behavior
Calls to the validator contract are executed using an evm throughout the codebase.

https://github.com/celo-org/celo-blockchain/pull/317#discussion_r305975072

PR #372, opened 2019-08-02
Move validator contract calls to contract_comm interface

jarmg:
### Description
This is a follow-on PR to #317 which performs a similar refactor for Validator contract calls

### Tested
e2e and geth tests were run. No tests were added.

### Other changes
None 

### Related issues
#362 

### Backwards compatibility
Yes, this is backwards compatible.


contract_comm/evm.go
```diff
@@ -227,6 +227,7 @@ func makeCallWithContractId(scRegistryId string, abi abi.ABI, funcName string, a
 			log.Warn("Contract Address Registry not yet deployed")
 			return 0, err
 		} else {
+			log.Error("Error in contract communication", "contract id", scRegistryId, "error", err)
```
jarmg: Required to deprecate the functionality of the `errValidatorsContractNotRegistered` error object.

consensus/istanbul/backend/announce.go
```diff
@@ -203,11 +205,11 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
 	}
 
 	// If the message is not within the registered validator set, then ignore it
-	regVals, err := sb.retrieveRegisteredValidators()
+	regVals, err := validators.RetrieveRegisteredValidators(nil, nil)
```
jarmg: Is it okay to not use pending state in announce?

kevjue: Ya, it's fine to use the latest finalized block on the blockchain.

Issue #304, opened 2019-06-26
Expose GasPriceMinimum functions in RPC


jarmg:
### Expected Behavior
We have added `api_backend` functions 
- `SuggestPriceInCurrency`
- `GasPriceMinimum`
- `InfrastructureFraction`
And we should should expose these through the rpc for the light client and full node

### Current Behavior
They are not exposed

mcortesi:
We've already have `SuggestGastPriceMinimun` with support for multiple currencies. The rest are deprioritized

PR #373, opened 2019-08-06
Add gas price minimum calls to rpc

jarmg:
### Description
This PR adds calls to interact with gas price minimum functionality to the rpc interface.

### Tested

### Other changes

### Related issues
#304 

### Backwards compatibility
These changes only add commands to the rpc and are fully backwards compatible.


jarmg:
Status: This first call works on full nodes but not light nodes

PR #385, opened 2019-08-09
Makes Makefile friendlier to use for Android builds

kobigurk:
Currently, `make android` assumes you have relevant Rust targets installed and requires an optional environment variable NDK_VERSION to be set.
This PR adds the targets automatically assuming you have rustup installed and removes the dependency on the environment variable.

Ran `make android -j` locally.


Makefile
```diff
@@ -26,16 +26,20 @@ bls-zexe: vendor/github.com/celo-org/bls-zexe/bls/target/release/libbls_zexe.a
...
 ndk_bundle: check_android_env
+	@test $${NDK_VERSION?Please set environment variable NDK_VERSION}
 	curl --silent --show-error --location --fail --retry 3 --output /tmp/$(NDK_VERSION).zip \
 		https://dl.google.com/android/repository/$(NDK_VERSION)-linux-x86_64.zip && \
 	mkdir $(ANDROID_NDK_HOME) && \
 	unzip -q /tmp/$(NDK_VERSION).zip -d $(ANDROID_NDK_HOME) && \
 	rm /tmp/$(NDK_VERSION).zip
 
 bls-zexe-android: check_android_env
+	rustup target add aarch64-linux-android
```
ashishb: why not add a check to ensure that rustup is defined command?

kobigurk: Right. Let me do that.

kobigurk: Done, take a look?

Makefile
```diff
@@ -25,21 +34,31 @@ geth: bls-zexe
...
-	PATH="$$PATH:$(ANDROID_NDK_HOME)/toolchains/llvm/prebuilt/linux-x86_64/bin:$(ANDROID_NDK_HOME)/toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin"; ln -s $(ANDROID_NDK_HOME)/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android21-clang $(ANDROID_NDK_HOME)/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android-clang; cd vendor/github.com/celo-org/bls-zexe/bls && cargo build --release --target=x86_64-linux-android --lib
+ifeq ("$(RUSTUP_exists)","")
+	$(error "No rustup in PATH, consult https://github.com/celo-org/celo-monorepo/blob/master/SETUP.md")
+else
+	rustup target add aarch64-linux-android
+	rustup target add armv7-linux-androideabi
+	rustup target add i686-linux-android
+	rustup target add x86_64-linux-android
+	PATH="$$PATH:$(ANDROID_NDK)/toolchains/llvm/prebuilt/$(OS)-x86_64/bin:$(ANDROID_NDK)/toolchains/aarch64-linux-android-4.9/prebuilt/$(OS)-x86_64/bin"; ln -s $(ANDROID_NDK)/toolchains/llvm/prebuilt/$(OS)-x86_64/bin/aarch64-linux-android21-clang $(ANDROID_NDK)/toolchains/llvm/prebuilt/$(OS)-x86_64/bin/aarch64-linux-android-clang; cd vendor/github.com/celo-org/bls-zexe/bls && cargo build --release --target=aarch64-linux-android --lib
```
ashishb: Please don't write such gigantic lines, it is hard to edit them.

ashishb: Also use `&&` and not `;` so that `make` stops if a command fails.

PR #404, opened 2019-08-16
[cleanup]Remove references to Celo latest sync mode

ashishb:
### Description

Remove references to Celo latest sync mode

### Tested

`make -j geth && make -j lint`

### Related issues

- Addresses https://github.com/celo-org/celo-monorepo/issues/596

### Backwards compatibility

Removes `CeloLatestsync` mode.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -171,7 +171,7 @@ var (
 	defaultSyncMode = eth.DefaultConfig.SyncMode
 	SyncModeFlag    = TextMarshalerFlag{
 		Name:  "syncmode",
-		Usage: `Blockchain sync mode ("fast", "full", "light", or "celolatest")`,
+		Usage: `Blockchain sync mode ("fast", "full", "light", or "ultralight")`,
 		Value: &defaultSyncMode,
 	}
 	GCModeFlag = cli.StringFlag{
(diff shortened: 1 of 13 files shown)


cmd/utils/flags.go
```diff
@@ -1490,7 +1488,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node) (chain *core.BlockChain, chai
 	var err error
 	chainDb = MakeChainDatabase(ctx, stack)
 	config, _, err := core.SetupGenesisBlock(chainDb, MakeGenesis(ctx))
-	if ctx.GlobalString(SyncModeFlag.Name) == "celolatest" {
+	if ctx.GlobalString(SyncModeFlag.Name) == "ultralight" {
```
asaj: This is weird, should this have originally been checking if the syncmode was "ultralight" *or* "celolatest"?

If we weren't setting this config variable before for `ultralight`, perhaps it's not necessary now that `celolatest` is deprecated?

consensus/clique/clique.go
```diff
@@ -378,7 +378,7 @@ func (c *Clique) verifyCascadingFields(chain consensus.ChainReader, header *type
 		parent = chain.GetHeader(header.ParentHash, number-1)
 	}
 
-	// Added to bypass the situation when the parent is missing in celolatest mode
+	// Added to bypass the situation when the parent is missing in ultralight mode
```
asaj: To that point, we would have previously been entering this when in ultralight mode, right?

ashishb: Yes, the comment was incomplete earlier.

cmd/utils/flags.go
ashishb: Yeah. It is a bug. It does not impact the normal execution *but* execution of specific functions - `importChain`, `exportChain`, `importPreimages`, `exportPreimages`, and `dump`. And in all those cases, when we start Geth, `fullHeaderChainAvailable` will get the correct value since it is not being filled from this code path.

timmoreton:
@ashishb  Can we hold off merging this? Geth 1.9 introduces an `ultralight` sync mode and we need to rename ours. Merging this now will make merging upstream harder.

PR #405, opened 2019-08-16
[CircleCI]Build walletkit explicitly

ashishb:
### Description

Earlier, `yarn` would build walletkit in the post install step.
That has been removed in celo-monorepo, therefore, build walletkit and some other packages
explicitly to avoid https://circleci.com/gh/celo-org/celo-blockchain/7304

### Tested

Circle CI execution will test this

Fixes https://github.com/celo-org/celo-monorepo/issues/616


trianglesphere:
Fixed in https://github.com/celo-org/celo-blockchain/pull/411

PR #407, opened 2019-08-17
Don't wait for bls-zexe to checkout celo-monorepo or geth linting

ashishb:
### Description

Don't wait for bls-zexe to checkout celo-monorepo or geth linting

Partially addresses  https://github.com/celo-org/celo-monorepo/issues/616

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,46 +1,45 @@
 version: 2
 
-defaults: &defaults
+geth-defaults: &geth-defaults
   docker:
     - image: circleci/golang:1.12
-  working_directory: /go/src/github.com/celo-org/geth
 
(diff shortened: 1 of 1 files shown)


kobigurk:
The governance test has a weird worrying error: `Error applying workspace layer for job 0e6e9243-3c9c-455a-afb3-01f3cddccf07: Concurrent upstream jobs persisted the same file(s)`

ashishb:
> The governance test has a weird worrying error: Error applying workspace layer for job 0e6e9243-3c9c-455a-afb3-01f3cddccf07: Concurrent upstream jobs persisted the same file(s)

I will dig more into it once #405 is merged since many other tests should pass after that.

PR #409, opened 2019-08-17
Istanbul: adds minimum quorum violation logging

kobigurk:
### Description

Currently, the Istanbul engine only fails with a generic error. Added logging for the specific case where the quorum size is not big enough.

### Tested

Tested against integration which had an older quorum size logic and currently the local node fails to sync with integration.


kobigurk:
@timmoreton maybe you can merge until tests get fixed?

PR #416, opened 2019-08-21
Fixes LightPeers equal to MaxPeers

trianglesphere:
### Description

`eth/backend.go` rejects configs where `LightPeers >= MaxPeers`.
https://github.com/celo-org/celo-blockchain/pull/395 broke our tests/maybe geth and this fixes it.

### Tested

Tests should be passing now.

### Other changes

N/A

### Related issues

N/A

### Backwards compatibility

N/A


timmoreton:
oh, good catch. the value of having passing tests..

PR #418, opened 2019-08-22
Hash registry id

bricewang:
### Description

Hashes registry ID before making EVM call to `Registry.getAddressFor`.

### Tested

No tests added.

### Related issues

- Sister PR of https://github.com/celo-org/celo-monorepo/pull/657

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
           command: |
             set -euo pipefail
             export CELO_MONOREPO_DIR="$PWD"
-            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b master
+            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b bricewang/registry-lookup
             cd ${CELO_MONOREPO_DIR}/packages
             # TODO(ashishb): Delete unnecessary packages to speed up build time.
             # It would be better whitelist certain packages and delete the rest.
--- a/contract_comm/evm.go
+++ b/contract_comm/evm.go
@@ -125,12 +125,12 @@ type InternalEVMHandler struct {
 	chain ChainContext
 }
 
-func MakeStaticCall(scRegistryId [32]byte, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, header *types.Header, state vm.StateDB) (uint64, error) {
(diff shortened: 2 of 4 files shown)


m-chrzan:
The registry hashes the input string, which is work that could be easily done off-chain.

I don't feel strongly about whether we should leave the old interface available as well. Accepting only hashes, as this PR does, leaves less clutter in the contract.

asaj:
What do you mean by "on-chain"? We're either hashing here in the EVM or outside the EVM, but since the former is a precompile I would imagine it just calls the same code.

If we actually wanted efficiency gains we would only compute the hash of the identifier once

m-chrzan:
> If we actually wanted efficiency gains we would only compute the hash of the identifier once

Oh yes, of course. We can totally do that, I'll add that change.

But the change in monorepo does already give us efficiency gains in any contract that calls out to other contracts, since in each spot like that we call `getAddressFor`, and there we do compute them once before deployment in `UsingRegistry`.

asaj:
> But the change in monorepo does already give us efficiency gains in any contract that calls out to other contracts, since in each spot like that we call getAddressFor, and there we do compute them once before deployment in UsingRegistry.

👍

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           command: |
             set -euo pipefail
             export CELO_MONOREPO_DIR="$PWD"
-            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b master
+            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b bricewang/registry-lookup
```
asaj: Same comment as on fixidity, you will only need to repoint the end-to-end tests for the first PR, and then the second PR can point to master

contract_comm/evm.go
```diff
@@ -141,7 +141,7 @@ func MakeCallWithAddress(scAddress common.Address, abi abi.ABI, funcName string,
 	return executeEVMFunction(scAddress, abi, funcName, args, returnObj, gas, value, header, state, true)
 }
 
-func GetRegisteredAddress(registryId string, header *types.Header, state vm.StateDB) (*common.Address, error) {
+func GetRegisteredAddress(registryId [32]byte, header *types.Header, state vm.StateDB) (*common.Address, error) {
```
asaj: nit: `registryId` vs `scRegistryId`

PR #423, opened 2019-08-29
Enable recommitting of blocks in IBFT

### LINKED ISSUE #420: Users SBAT send sequence of txs in sequential blocks
### Expected Behavior

A web3 client waiting for a tx to get included in a block should be notified early enough that a subsequent transaction can land in the very next block.

### Current Behavior

We see gaps between transactions during migrations and regular use

e.g https://alfajores-blockscout.celo-testnet.org/address/0xcea3ef8e187490a9d85a1849d98412e5d27d1bb3/transactions?block_number=174320&index=0

asaj:
### Description
This PR reverts the disabling of recommits in Istanbul, allowing for higher throughput of transactions for clients awaiting the confirmation of transaction `N` before submitting `N+1`.

Previously, after achieving consensus on block `N` at time `T`, the proposer for `N+1` would immediately create their proposal, and wait until `T+BLOCK_TIME` to send the preprepare, thus ignoring any transactions that were broadcast between `T` and `T+BLOCK_TIME`.

Now, between `T` and `T+BLOCK_TIME` the proposer will create a new proposal every `RECOMMIT_INTERVAL` seconds, allowing it to incorporate any new transactions that it has heard about.
 
### Tested

Ran the end to end tests and confirmed that each block contained a transaction. Previously, transactions would only be included every other block.

### Other changes

- Adds a `StateSentPreprepare` to ensure that we don't hit a race condition which would let a proposer send two preprepares.

### Related issues

- Fixes #420 

### Backwards compatibility

This should be backwards compatible

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
           command: |
             set -euo pipefail
             export CELO_MONOREPO_DIR="$PWD"
-            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b asaj/https
+            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b asaj/governance-e2e
             cd ${CELO_MONOREPO_DIR}/packages
             # TODO(ashishb): Delete unnecessary packages to speed up build time.
             # It would be better whitelist certain packages and delete the rest.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ jobs:
             set -euo pipefail
             export PATH=$PATH:$HOME/.cargo/bin
             make bls-zexe
-            
+
       - persist_to_workspace:
           root: ~/repos
           paths:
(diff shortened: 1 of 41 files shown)


kevjue:
> I'm still reviewing these changes, but wanted to ask a more general question first.
> 
> Have you thought of moving the original delay logic from consensus/istanbul/backend/engine.go:Seal to consensus/istanbul/backend/engine.go:Prepare?
> 
> Specifically, place the delay logic here:
> 
> https://github.com/celo-org/celo-blockchain/blob/f28a2119af377f9851f4346993a4f1ccca7a09c2/consensus/istanbul/backend/engine.go#L406
> 
> Basically have the proposer wait till it's allowed to construct a new block before retrieving pending tx to construct a block proposal (as opposed to it waiting after it constructs a proposal).
> 
> My sense is that will reduce the added complexity within the Istanbul engine (removing the need of a new state and multiple outstanding block proposals for the same block number).

Ah, just realized that the new state is to prevent multiple outstanding proposals.  However, I still think that doing the delay earlier will reduce complexity (specifically reduce the need to recommit).

asaj:
> Ah, just realized that the new state is to prevent multiple outstanding proposals. However, I still think that doing the delay earlier will reduce complexity (specifically remove the need to recommit).

Yeah that makes sense, done.

PR #441, opened 2019-09-06
Add precompiles to access validator set

### LINKED ISSUE #360: Attestations SBAT access current validator set
### Expected Behavior

Provide a lightweight way of retrieving the current validator set, preferably via a precompiled contract that returns the current list compiled from the extradata of block headers going back over the previous epochs. 

### Current Behavior

We don't have a current view of the entire validator set from smart contracts, just the set that would be elected if an election was held now.

m-chrzan:
### Description

Sister monorepo PR: https://github.com/celo-org/celo-monorepo/pull/874

This PR adds two precompiled contracts that give the EVM access to the current validator set's information.

* `getValidator` at `0xfa` - takes as input an index, returns the address of the validator at that index. Reverts if out of bounds.
* `numberValidators` at `0xf9` - returns the size of the current validator set.

### Tested

Geth e2e tests in monorepo.

### Related issues

- Fixes #360 

### Backwards compatibility

No breaking changes, only adds new precompiles.


kevjue:
The code LGTM.  However, seems like there are some unit tests that are failing.  I would of expected that, since the evmcontext object (which now has a new field) is well used within the unit tests.

I'll approve once the unit tests pass.

asaj:
Why not have a single precompile that returns the entire set? Seems like that would be more flexible going forward

m-chrzan:
>However, seems like there are some unit tests that are failing. I would of expected that, since the evmcontext object (which now has a new field) is well used within the unit tests.

Working through those right now.

> Why not have a single precompile that returns the entire set? Seems like that would be more flexible going forward

I'd imagine we would set a higher gas cost if we went with that approach. I wanted to allow contract authors to be able to optimize gas costs if they only need to read a few validator addresses (like our use case with requesting attestations.

asaj:
> I'd imagine we would set a higher gas cost (proportional to size of validator set) if we went with that approach. I wanted to allow contract authors to be able to optimize gas costs if they only need to read a few validator addresses (like our use case with requesting attestations.

Should we include an `isValidator` precompile? I'm worried these precompiles are a bit too specific to our attestations use case, and won't generalize well (though we can always add more later on)

m-chrzan:
> Should we include an `isValidator` precompile?

I've actually been considering adding this as well recently. Feels like it could easily be a follow up PR though.

PR #442, opened 2019-09-06
Limited size buffer of future round change messages

trianglesphere:
### Description

Note: Not yet ready.

This now only stores one round change message from each other validator (the one with the highest round number).

When there are `f+1` round change messages, it chooses the minimum round of the set and waits for that round.

When the are quorum round change messages (in the same round) it starts a new round for that round.

The theory is that there are quorum honest nodes that are able to stabilize on the same round thus while the `f+1` round change messages can be split of different rounds, progress cannot be made until quorum is in the same round (which will happen via timeouts & round changes).

Potential Todos:
- Test "clearing" the round change set
- What happens if there is a quorum round for a lower round than the desired round.

### Tested

Not, :(.

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

I think so.


trianglesphere:
Cleaned up git history. Will reopen when ready.

PR #446, opened 2019-09-09
fixed geth's reading of exchange rates

kevjue:
### Description

There is an existing bug in the logic around converting and comparing two denominations of different currencies within geth.  Basically, when geth was reading the exchange rates from the smart contracts, it erroneously switched the numerator and denominator around.

Specifically, it should be using this format:

<asset we’re pegging to> / Celo Gold 

instead of 

Celo Gold / <asset we’re pegging to> 

### Tested

Ran celo dollar gas currency e2e tests.
Ran geth's unit tests.

### Backwards compatibility

Is backward compatible.

### DIFF
--- a/contract_comm/currency/currency.go
+++ b/contract_comm/currency/currency.go
@@ -140,9 +140,9 @@ func Convert(val *big.Int, currencyFrom *common.Address, currencyTo *common.Addr
 	}
 
 	// Given value of val and rates n1/d1 and n2/d2 the function below does
-	// (val * n1 * d2) / (d1 * n2)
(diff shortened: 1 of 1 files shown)


asaj:
> Ran celo dollar gas currency e2e tests.
> Ran geth's unit tests.

How hard would it be to add a test that would have caught this bug?

kevjue:
> > Ran celo dollar gas currency e2e tests.
> > Ran geth's unit tests.
> 
> How hard would it be to add a test that would have caught this bug?

@asaj  - The new e2e celo dollar gas currency tests will catch this bug.  In fact, that's how I found it.

Issue #453, opened 2019-09-13
Eth nodes SNBAT to find celo nodes via a v5 bootnode


tkporter:
### Expected Behavior

Network ID & celo salt is used, similar to https://github.com/celo-org/celo-blockchain/pull/436

### Current Behavior

There are no checks that filter out non-celo nodes using v5 discovery

PR #454, opened 2019-09-13
Allow v4/v5 on a bootnode simultaneously, allow mobile to use discv5

### LINKED ISSUE #443: Bootnodes SBAT run discovery v4 and v5 at the same time
### Expected Behavior

When started with a specific flag (or both --v4 --v5, doesn't matter), a bootnode can use discovery v4 and v5 at the same time

### Current Behavior

Entirely different v4/v5 bootnodes are required. LES requires discv5, and currently our tx-node/validators use v4 when discovering each other. Tx-nodes & validators are already able to use v4/v5 at the same time.

Required for https://github.com/celo-org/celo-monorepo/issues/67

tkporter:
### Description

Mobile/LES requires discovery v5-- to avoid changing the code to use v4 or to require us to deploy a v5 bootnode, I made it so a bootnode can use v4, v5, or both at the same time (using a similar approach to using v4/v5 at the same time that full nodes use).

I previously added a flag called `NoDiscovery` to disable discovery, but I will include that in a later pr

I also added a salt to the discv5 p2p messages, similar to the approach in #436. I didn't add the network ID, however-- I've created an issue here https://github.com/celo-org/celo-blockchain/issues/453, it just felt slightly outside of the scope of this PR

### Tested

Ran my own VM testnet & was able to get an app to peer!

### Other changes

Changed `Dockerfile.alltools` to only recompile rust dependencies if vendored dependencies are updated (which is where we have our only rust deps).

Also just got rid of the hardcoded default bootnode enodes that were just annoying when testing.

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/67
- Fixes #443 

### Backwards compatibility

A newer version of `react-native-geth` and `monorepo` will be needed in order to take advantage of the additions


kevjue:
> ### Other changes
> Changed `Dockerfile.alltools` to only recompile rust dependencies if vendored dependencies are updated (which is where we have our only rust deps).
> 

Thank you! 🙏🙏🙏🙏🙏

It was getting annoying to wait for my google cloud builds.

cmd/bootnode/main.go
```diff
@@ -124,22 +128,37 @@ func main() {
 		}
 	}
 
-	if *runv5 {
-		if _, err := discv5.ListenUDP(nodeKey, conn, "", restrictList); err != nil {
-			utils.Fatalf("%v", err)
+	var unhandled chan discover.ReadPacket
```
kevjue: Please add a comment with a link to server.go that contains the same logic.

mobile/geth.go
```diff
@@ -86,6 +86,9 @@ type NodeConfig struct {
 	// WhisperEnabled specifies whether the node should run the Whisper protocol.
 	WhisperEnabled bool
 
+	// PeerDiscovery indicates whether the node should participate in p2p discovery
+	PeerDiscovery bool
```
kevjue: How does this get set?

For consistency sake, what are your thoughts of keeping `nodiscovery`?  I think at this point, we all know what flag means.

tkporter: Gets set here in `react-native-geth`: https://github.com/celo-org/react-native-geth/pull/17/files#diff-67cba49e3e2aa654dc9f8955ad86296bR95

Personally I'm fine with going back to `nodiscovery`

PR #472, opened 2019-09-18
set max gas to double of the charged gas for the 'intrinsic' smart contract calls

kevjue:
# Description

Set the max gas for intrinsic smart contracts calls to double the charged gas.

Note that we plan to modify geth to read these values from a smart contract.

### Tested

Ran the unit tests.

### Backwards compatibility

Is backward compatible.  Specifically, the charged intrinsic gas is not changed.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -212,7 +212,7 @@ func (st *StateTransition) canBuyGas(accountOwner common.Address, gasNeeded *big
 		return st.state.GetBalance(accountOwner).Cmp(gasNeeded) > 0
 	}
 	balanceOf, gasUsed, err := currency.GetBalanceOf(accountOwner, *gasCurrency, params.MaxGasToReadErc20Balance, st.evm.GetHeader(), st.evm.GetStateDB())
-	log.Debug("canBuyGas called", "gasCurrency", *gasCurrency, "gasUsed", gasUsed)
+	log.Debug("balanceOf called", "gasCurrency", *gasCurrency, "gasUsed", gasUsed)
 
 	if err != nil {
 		return false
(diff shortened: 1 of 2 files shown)


params/protocol_params.go
```diff
@@ -139,11 +139,16 @@ const (
 	// TODO(asa): Make these operations less expensive by charging only what is used.
 	// The problem is we don't know how much to refund until the refund is complete.
 	// If these values are changed, "setDefaults" will need updating.
-	MaxGasForDebitFromTransactions      uint64 = 38 * 1000
 	ExpectedGasForDebitFromTransactions uint64 = 23 * 1000
-	MaxGasForCreditToTransactions       uint64 = 32 * 1000
-	MaxGasToReadErc20Balance            uint64 = 15 * 1000
-	MaxGasToReadTobinTax                uint64 = 50 * 1000
+	MaxGasForDebitFromTransactions      uint64 = 46 * 1000
```
m-chrzan: Can you add a comment explaining that this is currently a workaround where we're overestimating the gas?

kevjue: Done.

kevjue:
> Thanks. Did you also want to add logs for actual gas used with intrinsic operations in this PR?

Good point.  I added debug messages for those operations.

core/state_transition.go
```diff
@@ -211,7 +211,9 @@ func (st *StateTransition) canBuyGas(accountOwner common.Address, gasNeeded *big
 	if gasCurrency == nil {
 		return st.state.GetBalance(accountOwner).Cmp(gasNeeded) > 0
 	}
-	balanceOf, _, err := currency.GetBalanceOf(accountOwner, *gasCurrency, params.MaxGasToReadErc20Balance, st.evm.GetHeader(), st.evm.GetStateDB())
+	balanceOf, gasUsed, err := currency.GetBalanceOf(accountOwner, *gasCurrency, params.MaxGasToReadErc20Balance, st.evm.GetHeader(), st.evm.GetStateDB())
+	log.Debug("canBuyGas called", "gasCurrency", *gasCurrency, "gasUsed", gasUsed)
```
m-chrzan: Question: did you consider logging the contract function name (`balanceOf`) instead of the go one (`canBuyGas`)?

kevjue: Done

PR #475, opened 2019-09-19
Reduce istanbul default timeout, cap exp backoff

timmoreton:
### Description

* reduce default istanbul round change timeout from 10s to 3s 
* limit exponential backoff to 5 rounds, so backoff at round=5 is 3+2**4 = 35 secs
* add the blockperiod on to the timeout for the first round of every sequence 

### Tested

By running a testnet and e2e tests.

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -342,11 +342,14 @@ func (c *core) stopTimer() {
 func (c *core) newRoundChangeTimer() {
 	c.stopTimer()
 
-	// timeout for subsequent rounds adds an exponential backup, capped at 2**4 = 16s
 	timeout := time.Duration(c.config.RequestTimeout) * time.Millisecond
 	round := c.current.Round().Uint64()
-	if round > 0 {
-		timeout += time.Duration(math.Pow(2, math.Min(float64(round), 4.))) * time.Second
+	if round == 0 {
+		// timeout for first round takes into account expected block period
+		timeout += time.Duration(c.config.BlockPeriod) * time.Second
+	} else {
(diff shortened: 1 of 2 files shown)


consensus/istanbul/core/core.go
```diff
@@ -342,11 +342,14 @@ func (c *core) stopTimer() {
...
 	timeout := time.Duration(c.config.RequestTimeout) * time.Millisecond
 	round := c.current.Round().Uint64()
-	if round > 0 {
-		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
+	if round == 0 {
+		// timeout for first round takes into account expected block period
+		timeout += time.Duration(c.config.BlockPeriod) * time.Second
+	} else {
+		// timeout for subsequent rounds adds an exponential backup, capped at 2**5 = 32s
+		timeout += time.Duration(math.Pow(2, math.Min(float64(round), 5.))) * time.Second
```
kevjue: Not sure how important this is, but should the timeout for rounds > 0 also include the blockperiod?  Seems like if it's not, then the timeout may not be strictly increasing as rounds increase.

timmoreton: Good question. The timeout for round 1 can only over occur after we've had a timeout for round 0 -- i.e a Byzantine quorum of nodes have timed out and sent RC messages. The idea is that subsequent timeouts don't need to include the time we spend waiting to reach the time of last block + block period. Does that make sense?

kevjue: Yes makes sense.

PR #476, opened 2019-09-19
Add input length checks for precompiled contracts

### LINKED ISSUE #448: Users SNBAT call precompiles with invalid inputs
### Expected Behavior

`fractionMulExp` and `transfer` both check the length of `input` before executing

### Current Behavior

No check, invalid input length will crash node

yerdua:
### Description

This adds checks for the input length of some precompiled contracts, and throws a comprehensible error when the input is too short or too long. Without these checks it's possible to spam the contracts with empty input that crashes the node and doesn't take gas fees from the caller.

### Tested

For the modified contracts that don't require an instance of `evm`, unit tests for the input length are included in this PR. For the ones that do need an instance of `evm`, like `transfer`, the overhead in adding a unit test is high, so they'll be tested in end-to-end instead.

### Related issues

- Fixes #448


core/vm/contracts.go
```diff
@@ -475,6 +475,10 @@ func (c *transfer) Run(input []byte, caller common.Address, evm *EVM, gas uint64
 		return nil, gas, err
 	}
 
+	if len(input) != 96 {
```
kevjue: Please add comments that state the argument types expected and their lengths.

PR #477, opened 2019-09-19
Check message address against signature

timmoreton:
### Description

Ensure we check the address of the validator signing a message with the value at `msg.Address` 

### Tested

New unit tests.

### Related issues

- Fixes celo-labs#107

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/core/testbackend_test.go
+++ b/consensus/istanbul/core/testbackend_test.go
@@ -129,6 +129,7 @@ func (self *testSystemBackend) Verify(proposal istanbul.Proposal, src istanbul.V
 }
 
 func (self *testSystemBackend) Sign(data []byte) ([]byte, error) {
+	// We overload the signature with the signer address so we can return the latter from CheckValidatorSignature.
 	return self.Address().Bytes(), nil
 }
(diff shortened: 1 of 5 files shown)


consensus/istanbul/core/testbackend_test.go
```diff
@@ -130,16 +129,15 @@ func (self *testSystemBackend) Verify(proposal istanbul.Proposal, src istanbul.V
 }
 
 func (self *testSystemBackend) Sign(data []byte) ([]byte, error) {
-	testLogger.Warn("not sign any data")
-	return data, nil
+	return self.Address().Bytes(), nil
```
asaj: Why return the address here?

timmoreton: because that's the simplest way to be able to mock out returning the address from `CheckValidatorSignature`

asaj: Please add a comment explaining why this is done

timmoreton: We now use real signing in this test.

PR #481, opened 2019-09-23
Checking minimum client version

mrsmkl:
### Description

Checks the minimum client version from a smart contract, exits with error message if required version is larger than current version.

### Tested

Still need to figure out how to test.

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/823

PR for the smart contract: https://github.com/celo-org/celo-monorepo/pull/1081

### DIFF
--- a/contract_comm/blockchain_parameters/blockchain_parameters.go
+++ b/contract_comm/blockchain_parameters/blockchain_parameters.go
@@ -65,7 +65,7 @@ func init() {
 	}
 }
 
-func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
+func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionInfo, error) {
 	version := [3]*big.Int{big.NewInt(0), big.NewInt(0), big.NewInt(0)}
 	var err error
 	_, err = contract_comm.MakeStaticCall(
@@ -78,15 +78,21 @@ func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
 		header,
 		state,
 	)
+	if err != nil {
+		return nil, err
+	}
+	return &params.VersionInfo{version[0].Uint64(), version[1].Uint64(), version[2].Uint64()}, nil
+}
+
+func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
+	version, err := GetMinimumVersion(header, state)
 
 	if err != nil {
 		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
 		return
 	}
 
-	if params.VersionMajor < version[0].Uint64() ||
-		params.VersionMajor == version[0].Uint64() && params.VersionMinor < version[1].Uint64() ||
-		params.VersionMajor == version[0].Uint64() && params.VersionMinor == version[1].Uint64() && params.VersionPatch < version[2].Uint64() {
+	if params.CurrentVersionInfo.Cmp(version) == -1 {
 		log.Crit("Client version older than required", "current", params.Version, "required", version)
 	}
 
--- a/contract_comm/blockchain_parameters/blockchain_parameters.go
+++ b/contract_comm/blockchain_parameters/blockchain_parameters.go
@@ -75,6 +75,7 @@ func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
 
 	if err != nil {
 		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
+		return
 	}
 
 	if params.VersionMajor < version[0].Uint64() ||
--- a/contract_comm/blockchain_parameters/blockchain_parameters.go
+++ b/contract_comm/blockchain_parameters/blockchain_parameters.go
@@ -59,7 +59,7 @@ var (
 	blockchainParametersABI, _ = abi.JSON(strings.NewReader(blockchainParametersABIString))
 )
 
-func CheckMinimumVersion(header *types.Header, state vm.StateDB) error {
+func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
 	version := [3]*big.Int{big.NewInt(0), big.NewInt(0), big.NewInt(0)}
 	var err error
 	_, err = contract_comm.MakeStaticCall(
@@ -75,7 +75,6 @@ func CheckMinimumVersion(header *types.Header, state vm.StateDB) error {
 
 	if err != nil {
 		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
-		return nil
 	}
 
 	if params.VersionMajor < version[0].Uint64() ||
@@ -84,5 +83,4 @@ func CheckMinimumVersion(header *types.Header, state vm.StateDB) error {
 		log.Crit("Client version older than required", "current", params.Version, "required", version)
 	}
 
-	return nil
 }
--- a/contract_comm/blockchain_parameters/blockchain_parameters.go
+++ b/contract_comm/blockchain_parameters/blockchain_parameters.go
@@ -19,6 +19,7 @@ package blockchain_parameters
 import (
 	"math/big"
 	"strings"
+	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/contract_comm"
@@ -93,7 +94,17 @@ func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
 	}
 
 	if params.CurrentVersionInfo.Cmp(version) == -1 {
+		time.Sleep(10 * time.Second)
 		log.Crit("Client version older than required", "current", params.Version, "required", version)
 	}
 
+}
+
+func SpawnCheck() {
+	go func() {
+		for {
+			time.Sleep(60 * time.Second)
+			CheckMinimumVersion(nil, nil)
+		}
+	}()
 }
--- a/contract_comm/blockchain_parameters/blockchain_parameters.go
+++ b/contract_comm/blockchain_parameters/blockchain_parameters.go
@@ -55,9 +55,15 @@ const (
 
 const defaultGasAmount = 2000000
 
-var (
-	blockchainParametersABI, _ = abi.JSON(strings.NewReader(blockchainParametersABIString))
-)
+var blockchainParametersABI abi.ABI
+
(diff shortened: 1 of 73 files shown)


contract_comm/blockchain_params/blockchain_params.go
```diff
@@ -0,0 +1,82 @@
...
+		header,
+		state,
+	)
+
+	if err != nil {
+		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParamsRegistryId)
+		return nil
+	}
+
+	log.Warn("Version", "version", minVersion)
```
m-chrzan: Why is this a warning?

contract_comm/blockchain_params/blockchain_params.go
```diff
@@ -0,0 +1,82 @@
...
+		[]interface{}{},
+		&minVersion,
+		defaultGasAmount,
+		header,
+		state,
+	)
+
+	if err != nil {
+		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParamsRegistryId)
+		return nil
```
m-chrzan: What sorts of errors are we expecting here? Is it fine to just ignore them all or should we propagate `err` to the caller?

contract_comm/blockchain_params/blockchain_params.go
mrsmkl: Just forgot to remove debug message.

contract_comm/blockchain_params/blockchain_params.go
mrsmkl: Mostly the error would be that the contract has not yet been deployed. So should be ok to ignore.

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,88 @@
...
+			"payable": false,
+			"stateMutability": "view",
+			"type": "function"
+	}]`
+)
+
+const defaultGasAmount = 2000000
+
+var (
+	blockchainParametersABI, _ = abi.JSON(strings.NewReader(blockchainParametersABIString))
```
mcortesi: Shouldn't we use `init()` and check the error here?  (golang newbie here)
@kevjue ?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,88 @@
...
+
+	if err != nil {
+		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
+		return nil
+	}
+
+	if params.VersionMajor < version[0].Uint64() ||
+		params.VersionMajor == version[0].Uint64() && params.VersionMinor < version[1].Uint64() ||
+		params.VersionMajor == version[0].Uint64() && params.VersionMinor == version[1].Uint64() && params.VersionPatch < version[2].Uint64() {
+		log.Crit("Client version older than required", "current", params.Version, "required", version)
```
mcortesi: does log.Crit() exits? shouldn't we exit after we fail?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,88 @@
...
+	}]`
+)
+
+const defaultGasAmount = 2000000
+
+var (
+	blockchainParametersABI, _ = abi.JSON(strings.NewReader(blockchainParametersABIString))
+)
+
+func CheckMinimumVersion(header *types.Header, state vm.StateDB) error {
```
mcortesi: this function always returns `nil`

mrsmkl: oops

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,86 @@
...
+		defaultGasAmount,
+		header,
+		state,
+	)
+
+	if err != nil {
+		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
+	}
+
+	if params.VersionMajor < version[0].Uint64() ||
```
mcortesi: we are entering here if `err != nil`, which seems incorrect.

contract_comm/blockchain_parameters/blockchain_parameters.go
mcortesi: so, do we want to kill the node if we find this?
@kevjue

mrsmkl:
About killing the node, perhaps we should check for a confirmed block?

asaj:
> About killing the node, perhaps we should check for a confirmed block?

What do you mean by checking for a confirmed block? Do you mean waiting for an additional block?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,87 @@
...
+		state,
+	)
+
+	if err != nil {
+		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
+		return
+	}
+
+	if params.VersionMajor < version[0].Uint64() ||
+		params.VersionMajor == version[0].Uint64() && params.VersionMinor < version[1].Uint64() ||
```
asaj: Consider defining a `Version` struct and a `Cmp` function to make this more readable.

contract_comm/blockchain_parameters/blockchain_parameters.go
asaj: I believe we do want to kill the node when we find this

mrsmkl:
Yeah, I was thinking that it would read the value from some parent block.

core/state_processor.go
```diff
@@ -73,6 +74,8 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 			return nil, nil, 0, err
 		}
 	}
+	blockchain_parameters.CheckMinimumVersion(header, statedb)
```
asaj: Is this the right place to check this? I think we process all proposed blocks before they're finalized.

Since it's not critical that we shut down out-of-date node immediately, consider instead a separate go routine that just checks the version against the latest finalized block every few minutes.

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2017 The Celo Authors
```
asaj: Per a suggestion from @timmoreton, consider adding a flag that allows users to disable this check. This will allow the network to recover more quickly in the event that a malicious actor is able to manipulate the minimum client version

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,110 @@
...
+				"type": "uint256"
+			  }
+			],
+			"payable": false,
+			"stateMutability": "view",
+			"type": "function"
+	}]`
+)
+
+const defaultGasAmount = 2000000
```
asaj: Consider adding this to `protocol_params.go`

core/state_processor.go
```diff
@@ -73,6 +73,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 			return nil, nil, 0, err
 		}
 	}
+
```
asaj: Revert changes to this file?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,110 @@
...
+func CheckMinimumVersion(header *types.Header, state vm.StateDB) {
+	version, err := GetMinimumVersion(header, state)
+
+	if err != nil {
+		log.Warn("Error checking client version", "err", err, "contract id", params.BlockchainParametersRegistryId)
+		return
+	}
+
+	if params.CurrentVersionInfo.Cmp(version) == -1 {
+		time.Sleep(10 * time.Second)
```
asaj: Why sleep here?

mrsmkl: When testing, it's useful to know that the node cannot exit immediately when making the transaction.

contract_comm/blockchain_parameters/blockchain_parameters.go
mcortesi: @mrsmkl can we move this to an init and check the error?

params/protocol_params.go
```diff
@@ -156,3 +157,5 @@ const (
 	// We charge for reading the balance, 1 debit, and 3 credits (refunding gas, paying the gas fee recipient, sending to the infrastructure fund)
 	AdditionalGasForNonGoldCurrencies uint64 = 3*ExpectedGasForCreditToTransactions + ExpectedGasForDebitFromTransactions + ExpectedGasToReadErc20Balance
 )
+
+const ContractCommGas uint64 = 2000000 // gas amount used for querying system contracts
```
asaj: I don't think this is entirely accurate, as if you look above we specify different amounts for different calls to system contracts. Instead, consider specifying one specific to this call, e.g. `MaxGasForGetMinimumClientVersion`

contract_comm/gasprice_minimum/gasprice_minimum.go
```diff
@@ -97,7 +97,7 @@ const (
   ]`
 )
 
-const defaultGasAmount = 2000000
+const defaultGasAmount = params.ContractCommGas
```
asaj: Ditto

contract_comm/random/random.go
```diff
@@ -84,7 +84,7 @@ const (
       "type": "function"
     }
 ]`
-	gasAmount = 1000000
+	gasAmount = params.ContractCommGas
```
asaj: Ditto

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,110 @@
...
+
+const defaultGasAmount = params.ContractCommGas
+
+var blockchainParametersABI abi.ABI
+
+func init() {
+	var err error
+	blockchainParametersABI, err = abi.JSON(strings.NewReader(blockchainParametersABIString))
+	if err != nil {
+		log.Error("Error reading ABI for BlockchainParameters", "err", err)
```
mcortesi: probably fatal here, since this happens on startup and the erros means a developer bug

PR #483, opened 2019-09-24
Log validator result elections

### LINKED ISSUE #469: Validators SBAT see log messages when elected/unelected
### Expected Behavior

The operator of a validator node can easily audit logs to see during which time period/block range they were elected to validate or not.

### Current Behavior

No clear logs?

mcortesi:
### Description

After a lengthly research; in order to detect the correct place where to  log, my findings were

1) Needs to be done **After** the block has been finalized. So, we react upon looking at the parent block, not the block that is being mined. That way we know it's finalized.

2) Needs to happen at the end/beginning of each epoch, that when the election is run

3) 
worker listens to newBlocks, and call the mainLoop (writes to newWorkCh),

mainLoop on `newWorkCh` call consensus `NewChainHead`

```go
		case req := <-w.newWorkCh:
			if h, ok := w.engine.(consensus.Handler); ok {
				h.NewChainHead()
			}
			w.commitNewWork(req.interrupt, req.noempty, req.timestamp)
```

`NewChainHead` checks if the last block (already mined) was the last epoch block; and that's where we can log

```go
func (sb *Backend) NewChainHead() error {
	sb.coreMu.RLock()
	defer sb.coreMu.RUnlock()
	if !sb.coreStarted {
		return istanbul.ErrStoppedEngine
	}

	// If the last block of the epoch has just been added to the blockchain, then
	// establish 'validator' type connections to all validators in the upcoming epoch
	// and disconnect from the ones that are no longer in the val set.
	currentBlock := sb.currentBlock()
	if istanbul.IsLastBlockOfEpoch(currentBlock.Number().Uint64(), sb.config.Epoch) {
		sb.logger.Trace("At end of epoch and going to refresh validator peers", "current block number", currentBlock.Number().Uint64())
		go sb.RefreshValPeers(sb.getValidators(currentBlock.Number().Uint64(), currentBlock.Hash()))
	}

	go sb.istanbulEventMux.Post(istanbul.FinalCommittedEvent{})
	return nil
}
```

### Tested

Run e2e tests

### Related issues

- Fixes #469

### DIFF
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -111,7 +111,13 @@ func (sb *Backend) NewChainHead() error {
 	currentBlock := sb.currentBlock()
 	if istanbul.IsLastBlockOfEpoch(currentBlock.Number().Uint64(), sb.config.Epoch) {
 		sb.logger.Trace("At end of epoch and going to refresh validator peers", "current block number", currentBlock.Number().Uint64())
-		go sb.RefreshValPeers(sb.getValidators(currentBlock.Number().Uint64(), currentBlock.Hash()))
+		valset := sb.getValidators(currentBlock.Number().Uint64(), currentBlock.Hash())
+		if _, val := valset.GetByAddress(sb.Address()); val == nil {
+			sb.logger.Info("Validators Election Results: Node OUT ValidatorSet")
+		} else {
+			sb.logger.Info("Validators Election Results: Node IN ValidatorSet")
+		}
+		go sb.RefreshValPeers(valset)
 	}
 
 	go sb.istanbulEventMux.Post(istanbul.FinalCommittedEvent{})
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,18 +70,13 @@ jobs:
             set -euo pipefail
             export CELO_MONOREPO_DIR="$PWD"
             git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b master
-            cd ${CELO_MONOREPO_DIR}/packages
-            # TODO(ashishb): Delete unnecessary packages to speed up build time.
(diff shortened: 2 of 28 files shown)


consensus/istanbul/backend/backend.go
```diff
@@ -524,10 +524,12 @@ func (sb *Backend) RefreshValPeers(valset istanbul.ValidatorSet) {
 
 	// Disconnect all validator peers if this node is not in the valset
 	if _, val := valset.GetByAddress(sb.Address()); val == nil {
+		sb.logger.Info("Validators Election Results: Node OUT ValidatorSet")
```
kevjue: I am a bit concerned of the logging being within `RefreshValPeers`.  It seems like the logging is a bit misaligned with the purpose of the function.

Have you considered moving it to `NewChainHead()` and added it's own `if address in valset else` logic there?

mcortesi: sure!
I agree that is not at all related.

consensus/istanbul/backend/handler.go
```diff
@@ -111,7 +111,13 @@ func (sb *Backend) NewChainHead() error {
 	currentBlock := sb.currentBlock()
 	if istanbul.IsLastBlockOfEpoch(currentBlock.Number().Uint64(), sb.config.Epoch) {
 		sb.logger.Trace("At end of epoch and going to refresh validator peers", "current block number", currentBlock.Number().Uint64())
-		go sb.RefreshValPeers(sb.getValidators(currentBlock.Number().Uint64(), currentBlock.Hash()))
+		valset := sb.getValidators(currentBlock.Number().Uint64(), currentBlock.Hash())
+		if _, val := valset.GetByAddress(sb.Address()); val == nil {
+			sb.logger.Info("Validators Election Results: Node OUT ValidatorSet")
```
m-chrzan: Nit: "OUT *of* ValidatorSet"?

PR #498, opened 2019-09-26
Allow validators to add/remove/inspect sentries

### LINKED ISSUE #484: Validator SBAT specify it's being proxied
This is a child issue of https://github.com/celo-org/celo-blockchain/issues/192

### Expected Behavior

Sentry proxied validator's UX will need to be modified.

- [x] New command line to specify a validator is proxied.  If enabled, it should automatically turn off discovery
- [x] New RPC `admin.addSentry`
- [x] New RPC `admin.removeSentry`
- [x] New RPC `admin.sentryInfo`

### Current Behavior

Currently there is no support for validator to be proxied by a sentry.

tkporter:
### Description

Added the RPCs: `admin_addSentry`, `admin_removeSentry`, `admin_sentryInfo`, and the geth flag `--proxied` to indicate that the node will use sentries as a proxy. Sentry peers do not contribute to the max peer limit. `admin_sentryInfo` outputs the peer info that `admin_peers` prints out but only for sentry nodes

Example REPL output:

```
> admin.addSentry('enode://c5987f5df7d63aba328ecdf441437ec520e4dc42a17303aae17afc1b5f522d0655503f65f7179f62250fba3d6b94ed982f57cf9e79653c8620dbfd91587588f5@127.0.0.1:30304')
true
> admin.sentryInfo

[{
    caps: ["eth/63", "les/1", "les/2"],
    enode: "enode://c5987f5df7d63aba328ecdf441437ec520e4dc42a17303aae17afc1b5f522d0655503f65f7179f62250fba3d6b94ed982f57cf9e79653c8620dbfd91587588f5@127.0.0.1:30304",
    id: "141ffb9266bd56b0aa31bb97d1688822f4044531c4648a34ce142e00f8543864",
    name: "Geth/v1.8.23-stable-45ffb873/darwin-amd64/go1.11.13",
    network: {
      inbound: false,
      localAddress: "127.0.0.1:64114",
      remoteAddress: "127.0.0.1:30304",
      sentry: true,
      static: true,
      trusted: false,
      validator: false
    },
    protocols: {
      eth: {
        difficulty: 17179869184,
        head: "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3",
        version: 63
      }
    }
}]
> admin.removeSentry('enode://c5987f5df7d63aba328ecdf441437ec520e4dc42a17303aae17afc1b5f522d0655503f65f7179f62250fba3d6b94ed982f57cf9e79653c8620dbfd91587588f5@127.0.0.1:30304')
true
```

One thing to note is I've made the assumption that a peered sentry node will not be also a validator peer ( https://github.com/celo-org/celo-blockchain/pull/498/files#diff-40b2cd29997bbd3f578f9d537789b9bbR933 )-- I think this is a safe assumption but I wanted to highlight it in case that's not the case

Here's the help text for `--proxied`:
```
NETWORKING OPTIONS:
  --bootnodes value                Comma separated enode URLs for P2P discovery bootstrap (set v4+v5 instead for light servers)
  --bootnodesv4 value              Comma separated enode URLs for P2P v4 discovery bootstrap (light server, full nodes)
  --bootnodesv5 value              Comma separated enode URLs for P2P v5 discovery bootstrap (light server, light nodes)
  --port value                     Network listening port (default: 30303)
  --maxpeers value                 Maximum number of network peers (network disabled if set to 0) (default: 25)
  --maxpendpeers value             Maximum number of pending connection attempts (defaults used if set to 0) (default: 0)
  --nat value                      NAT port mapping mechanism (any|none|upnp|pmp|extip:<IP>) (default: "any")
  --nodiscover                     Disables the peer discovery mechanism (manual peer addition)
  --v5disc                         Enables the experimental RLPx V5 (Topic Discovery) mechanism
  --netrestrict value              Restricts network communication to the given IP networks (CIDR masks)
  --nodekey value                  P2P node key file
  --nodekeyhex value               P2P node key as hex (for testing)
  --proxied                        Indicates the node will be proxied by sentry nodes. Disables discovery.
  --ping-ip-from-packet            Has the discovery protocol use the IP address given by a ping packet
  --use-in-memory-discovery-table  Specifies whether to use an in memory discovery table
```

### Tested

Ran a few geth instances on my local computer. I was able to add other nodes as sentries, print info about sentries, and tested that they don't contribute to the max peer count.

### Other changes

Found a superfluous if statements ( https://github.com/celo-org/celo-blockchain/pull/498/files#diff-40b2cd29997bbd3f578f9d537789b9bbL790 ), so I removed those

I also moved the order of some groupings of flags on the help output-- I put `DEPRECATED OPTIONS` and `MISC OPTIONS` at the bottom, which felt a little more natural

I also noticed we use `GlobalIsSet` to check values of boolean flags, which should instead be `GlobalBool`

### Related 
- Fixes #484 

### Backwards compatibility

This is backwards compatible

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -189,16 +189,18 @@ type Server struct {
 	quit            chan struct{}
 	addvalidator    chan *enode.Node
 	removevalidator chan *enode.Node
-	addstatic       chan *enode.Node
-	removestatic    chan *enode.Node
-	addtrusted      chan *enode.Node
-	removetrusted   chan *enode.Node
-	posthandshake   chan *conn
-	addpeer         chan *conn
-	delpeer         chan peerDrop
-	loopWG          sync.WaitGroup // loop, listenLoop
-	peerFeed        event.Feed
-	log             log.Logger
+	addsentry       chan *enode.Node
+	// removesentry    chan *enode.Node
+	addstatic     chan *enode.Node
+	removestatic  chan *enode.Node
+	addtrusted    chan *enode.Node
+	removetrusted chan *enode.Node
+	posthandshake chan *conn
+	addpeer       chan *conn
+	delpeer       chan peerDrop
+	loopWG        sync.WaitGroup // loop, listenLoop
+	peerFeed      event.Feed
+	log           log.Logger
 }
 
 type peerOpFunc func(peers map[enode.ID]*Peer, valNodes map[enode.ID]*valNodeInfo)
@@ -217,6 +219,7 @@ const (
 	inboundConn
 	trustedConn
 	validatorConn
+	sentryConn
 )
 
 // conn wraps a network connection with information gathered
@@ -270,6 +273,9 @@ func (f connFlag) String() string {
 	}
 	if f&validatorConn != 0 {
 		s += "-validator"
+	}
+	if f&sentryConn != 0 {
+		s += "-sentry"
 	}
 	if s != "" {
 		s = s[1:]
@@ -358,6 +364,14 @@ func (srv *Server) RemoveValidatorPeer(node *enode.Node) {
 	}
 }
 
+// AddSentryPeer assigns the given node as a sentry node.  It will set it as a static and trusted node.
+func (srv *Server) AddSentryPeer(node *enode.Node) {
+	select {
+	case srv.addsentry <- node:
+	case <-srv.quit:
+	}
+}
+
 // AddPeer connects to the given node and maintains the connection until the
 // server is shut down. If the connection fails for any reason, the server will
 // attempt to reconnect the peer.
@@ -393,7 +407,7 @@ func (srv *Server) RemoveTrustedPeer(node *enode.Node) {
 	}
 }
 
-// SubscribePeers subscribes the given channel to peer events
+// SubscribeEvents subscribes the given channel to peer events
 func (srv *Server) SubscribeEvents(ch chan *PeerEvent) event.Subscription {
 	return srv.peerFeed.Subscribe(ch)
 }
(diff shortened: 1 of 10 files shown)


p2p/server.go
```diff
@@ -910,7 +922,11 @@ running:
 				}
 			}
 		case n := <-srv.removesentry:
-			if isSentryNode(n.ID()) && srv.Proxied {
+			if isSentryNode(n.ID()) {
+				if !srv.Proxied {
```
kevjue: Same comment here.  srv.Proxied should always be True at this point.

tkporter:
@timmoreton, I agree! I added a warning if `--mine` isn't enabled, and opened a ticket for adding sentries when geth is started: https://github.com/celo-org/celo-blockchain/issues/503

p2p/server.go
```diff
@@ -57,6 +57,10 @@ const (
 
 	// Maximum amount of time allowed for writing a complete message.
 	frameWriteTimeout = 20 * time.Second
+
+	// The maximum number of sentry nodes a proxied node can have.
+	// This will increase in the future.
+	maxSentryNodes = 1
```
kevjue: Move this to `istanbul`

p2p/server.go
```diff
@@ -114,6 +118,9 @@ type Config struct {
 	// IP networks contained in the list are considered.
 	NetRestrict *netutil.Netlist `toml:",omitempty"`
 
+	// Proxied indicates the node is proxied by sentry nodes.
+	Proxied bool
```
kevjue: This should be in the `node` config.

p2p/server.go
```diff
@@ -189,6 +196,8 @@ type Server struct {
 	quit            chan struct{}
 	addvalidator    chan *enode.Node
 	removevalidator chan *enode.Node
+	addsentry       chan *enode.Node
```
kevjue: move the sentries and validators management into the `istanbul` package.

kevjue:
Closing this PR as it's now folded into this PR:  https://github.com/celo-org/celo-blockchain/pull/515

PR #504, opened 2019-09-30
txpool invalidate txns for txs with gasPrice less than gpm

### LINKED ISSUE #502: Validators SNBAT stop mining if it has zero gas-priced txns
### Expected Behavior

A validator should not stop mining it's txns if the highest priority txn has a 0 gas price.

### Current Behavior

A validator will stop mining it's txns if the highest priority txn has a 0 gas price.

The worker will not fill up a block after it first encounters a txn that is below the gas price minimum:  https://github.com/celo-org/celo-blockchain/blob/master/miner/worker.go#L818

I believe that we should modify the tx pool's validateTx to check if the txn's gas price is below the gpm, but that needs to be thought out more, since the gpm can change per block.

kevjue:
### Description

This PR modifies the tx pool so that it will verify that submitted txns' price is greater than or equal to the current gpm.

### Tested

* Unit tests ran
* Added an e2e test to verify that a txn w/ too low of a gasprice is invalidated by the txpool.

### Other changes

None

### Related issues

- Fixes #502 

### Backwards compatibility

Will not add txns that have a txn gas price < current gas price minimum

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -32,9 +32,10 @@ import (
 )
 
 var (
-	errInsufficientBalanceForGas    = errors.New("insufficient balance to pay for gas")
-	errNonWhitelistedGasCurrency    = errors.New("non-whitelisted gas currency address")
-	errGasPriceDoesNotExceedMinimum = errors.New("gasprice does not exceed gas price minimum")
+	ErrGasPriceDoesNotExceedMinimum = errors.New("gasprice does not exceed gas price minimum")
+
+	errInsufficientBalanceForGas = errors.New("insufficient balance to pay for gas")
+	errNonWhitelistedGasCurrency = errors.New("non-whitelisted gas currency address")
 )
 
 /*
@@ -297,7 +298,7 @@ func (st *StateTransition) preCheck() error {
 	// Make sure this transaction's gas price is valid.
 	if st.gasPrice.Cmp(st.gasPriceMinimum) < 0 {
 		log.Error("Tx gas price does not exceed minimum", "minimum", st.gasPriceMinimum, "price", st.gasPrice)
-		return errGasPriceDoesNotExceedMinimum
+		return ErrGasPriceDoesNotExceedMinimum
 	}
 
 	return nil
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/contract_comm/currency"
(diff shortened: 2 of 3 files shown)


core/tx_pool.go
```diff
@@ -667,6 +668,13 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		log.Debug("validateTx gas less than intrinsic gas", "tx.Gas", tx.Gas(), "intrinsic Gas", intrGas)
 		return ErrIntrinsicGas
 	}
+
+	gasPriceMinimum, _ := gpm.GetGasPriceMinimum(tx.GasCurrency(), nil, nil)
```
tkporter: any reason to ignore this error?

timmoreton:
@tkporter 's point above, plus check `tx.gasPrice()` can't be `nil`

core/state_transition.go
```diff
@@ -32,9 +32,10 @@ import (
 )
 
 var (
-	errInsufficientBalanceForGas    = errors.New("insufficient balance to pay for gas")
-	errNonWhitelistedGasCurrency    = errors.New("non-whitelisted gas currency address")
-	errGasPriceDoesNotExceedMinimum = errors.New("gasprice does not exceed gas price minimum")
+	ErrGasPriceDoesNotExceedMinimum = errors.New("gasprice does not exceed gas price minimum")
```
ashishb: Nit: *Capitalize* the first letters of each sentence.

core/tx_pool.go
ashishb: I agree, we should never silently ignore errors. At the least, we should log them.

kevjue: Great point.  I also agree that we shouldn't ignore errors from `GetGasPriceMinimum`.  However, if you look at the `GetGasPriceMinimum` implementation, it will never return a non nil err.

I think that there is a broader error handling issue in `GetGasPriceMinimum`, and I'm going to create a separate issue to address that.  I'm going to merge this PR and we can address the GetGasPriceMinimum ignoring error separately.

@tkporter @ashishb - Let me know if you guys have an issue with that.

core/state_transition.go
kevjue: FWIW, all errors in geth start with a lower case, so for the sake of consistency, I am going to keep it lowercase.

kevjue:
> @tkporter 's point above, plus check `tx.gasPrice()` can't be `nil`

If a new txn is created with a nil gasprice, the gasprice will be set to 0.

core/tx_pool.go
kevjue: I actually went ahead and added error handling in txpool for get price minimum.  For now, that code path will never get executed, but one the error handling for `GetGasPriceMinimum` is fixed, then that code will become relevant.

tkporter: Cool, sounds good to me

PR #506, opened 2019-10-01
Proxied validators send a ValEnodeShare message to sentry nodes

### LINKED ISSUE #487: Proxied validators SBAT sync it's validator enode table with it's sentry
This is a child issue of #192

### Expected Behavior

A proxied validator SBAT sync it's validator enode table with it's sentry.

### Current Behavior

There currently is no support for a proxied validator.

tkporter:
Don't merge-- going to merge all sentry changes together

### Description

Heads up, I started this branch off of #498 which can't be merged right now due to failing tests, so the diff includes whats contained in that PR as well. This diff against the branch in #498 might be easier for review: https://github.com/celo-org/celo-blockchain/compare/trevor/sentry-rpc...trevor/val-enode-share-msg

This PR has a proxied validator send a `ValEnodeShare` message to all its sentry nodes every minute. Right now, the entire ValEnodeTable of the validator is sent as the payload and the sentry just upserts all the entries

A couple things that would be fixed in the short term:
1. Right now sentries aren't aware of who the validator is they're serving as a proxy for, so there is no check by the sentry if the address that sent the ValEnodeShare message is the correct proxied validator
2. Because there's no way for a sentry to know if the validator it's a proxy for is in the validator set, I added a `forcePeer` parameter to the validator enode table `upsert` function that has the proxy peer with the other validators even if it's not in the validator set

### Tested

I created a 3-validator, 1-sentry setup. 1 validator was using the sentry as a proxy. I made sure messages arrived at the sentry, the validator enode table changed appropriately, and the sentry peered with the validators.

### Other changes

n/a

### Related issues

- Fixes #487

### Backwards compatibility

This is backwards compatible

### DIFF
--- a/consensus/istanbul/backend/val_enode_share.go
+++ b/consensus/istanbul/backend/val_enode_share.go
@@ -39,9 +39,9 @@ type sharedValidatorEnode struct {
 }
 
 type valEnodeShareMessage struct {
-	Address           common.Address
-	ValEnodes         []sharedValidatorEnode
-	Signature         []byte
+	Address   common.Address
+	ValEnodes []sharedValidatorEnode
+	Signature []byte
 }
 
 func (sm *valEnodeShareMessage) String() string {
@@ -60,9 +60,9 @@ func (sm *valEnodeShareMessage) EncodeRLP(w io.Writer) error {
 // DecodeRLP implements rlp.Decoder, and load the am fields from a RLP stream.
 func (sm *valEnodeShareMessage) DecodeRLP(s *rlp.Stream) error {
 	var msg struct {
-		Address     common.Address
-        ValEnodes   []sharedValidatorEnode
-		Signature   []byte
+		Address   common.Address
+		ValEnodes []sharedValidatorEnode
+		Signature []byte
 	}
 
 	if err := s.Decode(&msg); err != nil {
@@ -89,9 +89,9 @@ func (sm *valEnodeShareMessage) Sign(signingFn func(data []byte) ([]byte, error)
 	// Construct and encode a message with no signature
 	var payloadNoSig []byte
 	payloadNoSig, err := rlp.EncodeToBytes(&valEnodeShareMessage{
-		Address:     sm.Address,
-		ValEnodes:   sm.ValEnodes,
-		Signature:   []byte{},
+		Address:   sm.Address,
+		ValEnodes: sm.ValEnodes,
+		Signature: []byte{},
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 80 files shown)


consensus/istanbul/backend/val_enode_share.go
```diff
@@ -0,0 +1,249 @@
...
+//
+// define the validator enode share message
+
+type sharedValidatorEnode struct {
+	Address  common.Address
+	EnodeURL string
+	View     *istanbul.View
+}
+
+type valEnodeShareMessage struct {
```
kevjue: Could you use the istanbul.Message type to store your new message.  Could you also please refactor the `Announce` message to also use it. 🙏

That type used to be a private type within istanbul.core package, but Asa recently made it a public type in the istanbul package, specifically so that Announce and related messages can use it.

consensus/istanbul/backend/val_enode_share.go
```diff
@@ -0,0 +1,249 @@
...
+			ticker.Stop()
+			return
+		}
+	}
+}
+
+func (sb *Backend) generateValEnodeShareMsg() ([]byte, error) {
+	sharedValidatorEnodes := make([]sharedValidatorEnode, len(sb.valEnodeTable.valEnodeTable))
+	i := 0
+	for address, validatorEnode := range sb.valEnodeTable.valEnodeTable {
```
kevjue: as discussed, it's probably good to get a read lock on the valenodetable mutex.

consensus/istanbul/backend/val_enode_share.go
```diff
@@ -0,0 +1,249 @@
...
+	sb.logger.Trace("Received an Istanbul Validator Enode Share message", "msg", msg.String())
+
+	block := sb.currentBlock()
+	valSet := sb.getValidators(block.Number().Uint64(), block.Hash())
+	for _, sharedValidatorEnode := range msg.ValEnodes {
+		valEnode := &validatorEnode{
+			enodeURL: sharedValidatorEnode.EnodeURL,
+			view:     sharedValidatorEnode.View,
+		}
+		if err := sb.valEnodeTable.upsert(sharedValidatorEnode.Address, valEnode, valSet, sb.Address(), true); err != nil {
```
kevjue: Everytime `upsert` is called, it will lock on the validator enode table mutex.

I think that it should lock once per `handleValEnodeShareMsg` invocation.   Also, this will handle the case in case two separate `handleValEnocdeShareMsg` invocations run concurrently (which would be rare, but not impossible).

consensus/istanbul/backend/val_enode_table.go
```diff
@@ -81,7 +81,7 @@ func (vet *validatorEnodeTable) getUsingEnodeURL(enodeURL string) (common.Addres
 }
 
 // This function will update or insert a validator enode entry.  It will also do the associated set/remove to the validator connection.
-func (vet *validatorEnodeTable) upsert(remoteAddress common.Address, newValEnode *validatorEnode, valSet istanbul.ValidatorSet, localAddress common.Address) error {
+func (vet *validatorEnodeTable) upsert(remoteAddress common.Address, newValEnode *validatorEnode, valSet istanbul.ValidatorSet, localAddress common.Address, forcePeer bool) error {
```
kevjue: Proxied validators shouldn't connect/disconnect to/from validator peers.

kevjue:
Closing this PR as it is now a part of https://github.com/celo-org/celo-blockchain/pull/515

PR #508, opened 2019-10-01
Include signature from previous block in current block

### LINKED ISSUE #456: Block proposers SBAT include signatures for the previous block
### Expected Behavior

Block proposers should include signatures seen for the previous block in the block body. This issue can be broken down to the following tasks:

- [x] Previous signatures (and signer bit vector) should be added to the Block struct.
- [x] Clients should validate signatures when processing a block by checking that the aggregated signature matches the aggregated public key of the previous block. This can be tricky at the epoch boundary as the previous block's validator set may not be easily known (e.g. after a fast sync). It may have to be inferred from the validator diff in the header.
- [ ] This (and past) signatures and bit vectors should be stored in a smart contract so that they can be accessed by other contracts (API for other smart contracts are TBD).

### Current Behavior

Not included

gakonst:
### Description

First part of #456. Adds a new struct including an aggregate signature (expressed as a bytes slice) and a bitmap for the validators that signed. This just augments the block structure and the functions that use it, it does not include any functionality about including the signatures in blocks or verifying them

### Tested

Added a test that ensures the new field is properly RLP Encoded. All the other tests pass.

### Other changes

All functions that relate to downloading/broadcasting blocks were modified to take the new information into account.

### Related issues

- #456


eth/downloader/queue.go
```diff
@@ -397,6 +398,7 @@ func (q *queue) Results(block bool) []*fetchResult {
 				size += tx.Size()
 			}
 			size += result.Randomness.Size()
+			size += result.ParentSeal.Size() // TODO: Ensure that the Size() method is properly implemented for BlockSeal
```
asaj: This feels important to do, no?

miner/worker.go
```diff
@@ -102,6 +102,7 @@ type environment struct {
 	txs        []*types.Transaction
 	receipts   []*types.Receipt
 	randomness *types.Randomness // The types.Randomness of the last block by mined by this worker.
+	parentSeal *types.BlockSeal  // The types.BlockSeal of the last block mined by this worker
```
asaj: Comment here is slightly off, perhaps "The types.BlockSeal containing the signatures for the parent block seen by this worker."?

eth/downloader/queue.go
gakonst: Added a test for it.

gakonst:
@asaj I've addressed your comments, `end-to-end-monorepo-checkout` seems to be failing for an unrelated reason

asaj:
> @asaj I've addressed your comments, end-to-end-monorepo-checkout seems to be failing for an unrelated reason.

Yeah this was a bug that should now be fixed if you merge in the latest master.

> Also, do you want to merge this as-is in master, and then create the follow up PR for the second bullet of #456, or should we have a separate feature branch for this feature, merge each part there, and finally merge the whole feature on master? Either works for me.

Either works for me as well, we typically do the former so I would suggest that.

PR #511, opened 2019-10-03
Update to geth 1.9.8 + Announce v2

timmoreton:
# Description

The pr includes an update with upstream [geth version 1.9.8](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.8); and some other PRs; which are:

 * #815 Android & iOS build fix on CI
 * #816 Announce Protocol v2
 * #817 Get gas price suggestion from the correct block
 * #830 Add proxy information to celostats
 * #840  Prevent ethstats from blocking chainhead publishing

## 1.9 Notable Changes

Check [blog post](https://blog.ethereum.org/2019/07/10/geth-v1-9-0/)

* Use go 1.13
* Use go modules (no more vendor pkg)
* Use golint-ci for linting (much better, can be integrated with vscode)

* consensu.Engine interface has `Finalize()` and `FinalizeAndAssemble()` (before it was only finalize)
* Istanbul Fork support (there's some istanbul related flags)
* Geth separated its database in 2: recent blocks, freezer (old blocks)
* Geth has a new graphql service
* Better hardware wallets support
* Signing/keystore now belongs to "Clef". which is an interface and eventually an independent service to sign things.
* Geth has its own ultraligh client mode. (ours has been renamed to lightest)
* Light clients used hardcoded checkpoints to avoid downloading all headers. Now there are onchain checkpoints. For that
geth queries smart contracts in a similar way we do.
* Metrics improved: Geth v1.9.0 features 3 independent mechanisms: ExpVars, InfluxDB and Prometheus.
* Geth moved to use ENR (Ethreum Node Records) instead of enodeUrls
* Security: By default you can't unlock account if you have http, websocket or graphql interface open
* The --lightserv and --lightpeers flags were renamed to --light.serve and --light.maxpeers respectively
* The JSON-RPC server has been rewritten and now supports bi-directional communication. 
* The Less client/server has been splitted into client & server. Before it was a single type/class.
* Less client/server rewrite. Ligth Client tracking, many other stuff.
* Deprecated Protocol eth/62 & les/1
* Added protocol eth/64 (extending the protocol handshake with a forkid field)
* LES servers now advertise the les capability via ENR 
* Add eth.fillTransaction to create-but-not-sign a transaction  (this looks interesting!!!)
* IstanbulFork: Implement EIP-152: Blake2B F compression function precompile (#19972).
* IstanbulFork: Implement EIP-2028: Transaction data gas cost reduction (#19931).
* IstanbulFork: Implement EIP-2200: Net SSTORE optimizations (#19964).

## Fixes / Merge decisions

* vendor/bls is now crypto/bls is now a submodule (not actually a submodule, something similar)
* Move CI docker images to go 1.13

### Gateway Fee nil on RPC.EstimateGas

RPC CallArgs defines arguments for Call/Estimate Gas. Fields changed from hexutil.Int to a pointer of that. Which meant
that if we don't pass GatewayFee, before it defaulted to zero, and now it defaults to Nil. This breaks state transition logic.

The fix was to use a concrete type, and have then json decoder to default to Zero. But this is a quick fix. IMO it's better to
use pointer, and fail if the user didn't send GatewayFee as he should. Also EstimateGast should behave exactly the same as 
SendTransaction but that's not the case, since the latter call setDefaults() over the arguments.

### Missing Etherbase

Config moved Etherbase to cfg.Miner.Etherbase, but kept the old location `cfg.Etherbase` so we failed to obtain in on les clients.

We set it 7bfc95191cce079db6a30930af35682f37ba4971

After we split Etherbase and Validator we should keep only one Etherbase on the config.

### On init, light nodes tried to access eth.Ethereum

When running a node.. init runs:

```go
if ctx.GlobalInt(utils.LightLegacyServFlag.Name) > 0 || ctx.GlobalInt(utils.LightServeFlag.Name) > 0 {
  ...
```

which surprisingly, it will enter the if, since by default those flags are > 0. We should do this when syncmode
is full or fast.

Fixed on a7b48772f2bb049b68d8725f66677fcdff869e97

### Problem with Gas Calculation

Istanbul introduces gas cost changes. When merging we end up using those gas cost, but we shouldn't since we are not activating it now.

Fixed on 5008499b4300ae36cfe402478f421f659d552ac8

### Engine Finalize & FinalizeAndAssemble

1.9 split Engine interface Finalize in 2. `Finalize` should finalize the block header, but does not assemble the block.

On the merge, we didn't do anything on finalize; which resulted into a bug. `Finalize()` its used during block validation, and `FinalizeAndAssemble` during block creation. Both should modify the state in the same manner. And they didn't.

Fix ba7bfff1a8feccb1e099aefe50146c94b75640e8

### evm.DontMeterGas

On master we have a `evm.DontMeterGas` flag to avoid metering gas with some calls. 

The problem is that EVM execution code changed, and we were actually metering gas for some opcodes. Thus we ended with outOfGas errors on some of this
supposedly unmetered calls.

Fix 9e23bac9f76eb515a791af4841393e4fb75dba94

### Didn't get any istanbul messages

The way peer connection works is that for each connection an ordered list of protocols is assigned.

Each protocol has a length (# of messages). And peer has a logic to assign message code uniquely by stacking prococol lengths.

Istanbul IBFT, expands the Ethereum Protocol that is returned on `eth.Ethereum.Protocols()`. That is, it adds new messages code. So we
need to update the protocol length for that protocol.

Fix eb1cf12176cf57e537ce3683ffe65ea6c58c8d73

### ENR vs EnodeUrls

To generate announce Messages, we need our enodeURL, that we obtain by `node.String()`.

On 1.9, `String()` returns the ENR, not the EnodeURL. So announce failed.

In the future we should use ENR, but for now we keep using EnodeURL

Fix 8d1bfd3e6032040d8c87bb1d1a469ee9dd4ca3aa

### Handshake with more than one protocol Problem

There's a problem when during the handshake we have more than one protocol (see https://github.com/celo-org/celo-blockchain/pull/333)

Now, on 1.9 There are 2 Ethereum protocols eth/63 and eth/64. And none of them was marked as primary.

So, the fix was to disable eth/63 and make eth/64 the primary.

**Probably we'll have a problem here since eth/64 was introduced on 1.9. So, it's not backward compatible**

Fix c2b4a45fd9c4729c9f299da2de9a911f4b61dd3b

### Double Hashing on msg signing

Msg Signing logic on geth changed due to the introduction of Clef.

Before we hashed and then called the sign function. Now the sign function hashes, so we had to removed the hashing we did.

Fix 47004932134b16e414e3c36f799ce2446db6e621

### LES Rewrite

Light client features had to be completely rewritten since they changed massively on 1.9

One change here, is that before we had a map of peer -> etherebase on the peerSet. And now each
peer has a etherbase field.

This code should be reviewed.

# Tested

* Reviewed
* Unit Tests
* e2e Tests
* Tested with faulty network
* Tested against baklava
* Deployed to baklava-integration and others
* A Good amount of 🙏 

# Backwards compatibility

* A previous version was used to keep backward compatibility with baklava
* Current version is not due to changes in `master`


nategraf:
This PR will also fix https://github.com/celo-org/celo-blockchain/issues/768

eth/config.go
```diff
@@ -45,7 +45,7 @@ var DefaultConfig = Config{
 	},
 	NetworkId:          1,
 	LightPeers:         99,
-	LightServ:          0,
+	LightServ:          50,
```
H34D: If `LightServ` is set to something higher than 0 it will trigger this code in eth/backend.go: 

```
	if s.config.LightServ > 0 {
		if s.config.LightPeers != 0 && s.config.LightPeers >= srvr.MaxPeers {
			return fmt.Errorf("invalid peer config: light peer count (%d) >= total peer count (%d)", s.config.LightPeers, srvr.MaxPeers)
		}
		maxPeers -= s.config.LightPeers
	}

which will ultimately result in a fail with the default run command like this:

```
./bin/geth --datadir /tmp/test/0-node/datadir --syncmode full --debug --port 30303 --rpcvhosts=* --networkid 1101 --verbosity=1 --consoleoutput=stdout --consoleformat=term --nat extip:127.0.0.1 --rpc --rpcport 8545 --rpccorsdomain=* --rpcapi=eth,net,web3,debug,admin,personal,txpool,istanbul --wsorigins=* --ws --wsport 8546 --wsapi=eth,net,web3,debug,admin,personal,txpool,istanbul --mine --minerthreads=10 --nodekeyhex=ad..6c --istanbul.blockperiod 1 --nodiscover --password=/dev/null --unlock=0 --allow-insecure-unlock --ethstats=0-node@localhost:3000 --etherbase=0
Fatal: Error starting protocol stack: invalid peer config: light peer count (99) >= total peer count (50)
```

Comment (nategraf) [reply eth/config.go]: Rsolved in https://github.com/celo-org/celo-blockchain/commit/8521e133dcb5e5631c1243cc35e91c830970e2f1

File: .circleci/config.yml [hunk 2]
```diff
@@ -1,66 +1,107 @@
...
-    <<: *rust-defaults
+  e2e:
+    docker:
+      - image: celohq/node10-gcloud
+    working_directory: ~/repos/celo-monorepo/packages/celotool
+    environment:
+      GO_VERSION: "1.13.7"
+      CELO_MONOREPO_BRANCH_TO_TEST: mc/geth1.9-final
+jobs:
+  build-rust:
```
jeanregisser: Maybe `build-bls-zexe` is a better name? Otherwise it makes you think this job builds rust 😉

PR #517, opened 2019-10-08
Buffer future messages

### LINKED ISSUE #519: Validators SNBAT DDOS each other with future sequence messages
### Expected Behavior

There should be a limit on messages in future sequences that can be stored.

### Current Behavior

Validators store future messages without limit.

trianglesphere:
### Description

This limits the number of messages in the future that are stored. It is limited on a per validator basis to prevent byzantine validators from stopping honest messages from going through.

### Tested
Not yet.

### Other changes

### Related issues

- Fixes #519 

### Backwards compatibility

Should be backwards compatible with other nodes.


consensus/istanbul/core/backlog.go
```diff
@@ -103,114 +116,181 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
-			if err != nil {
-				if err == errFutureMessage {
-					logger.Trace("Stop processing backlog", "msg", msg)
-					backlog.Push(msg, prio)
-					isFuture = true
-					break
+				err := c.checkMessage(msg.Code, view)
+				if err != nil {
+					if err == errFutureMessage {
+						// TODO(asa): Why is this unexpected? It could be for a future round...
```
nambrot: How would this happen? I thought `seq` is the current sequence, and all the messages in this callback should be from this sequence? I would go as far to say that this should be an error level message?

consensus/istanbul/core/core.go
```diff
@@ -75,8 +76,10 @@ type core struct {
 	valSet     istanbul.ValidatorSet
 	validateFn func([]byte, []byte) (common.Address, error)
 
-	backlogs   map[istanbul.Validator]*prque.Prque
-	backlogsMu *sync.Mutex
+	backlogBySeq      map[uint64]*prque.Prque
+	backlogCountByVal map[common.Address]int
```
nambrot: Should this be an lru cache? Not sure if this is a practical concern, but otherwise we'll keep keys of old addresses around forever even if they have been rotated

nambrot: Should be this be `uint` instead of `int`?

consensus/istanbul/core/backlog.go
```diff
@@ -103,114 +116,181 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
-				logger.Trace("Skip the backlog event", "msg", msg, "err", err)
-				continue
-			}
-			logger.Trace("Post backlog event", "msg", msg)
+				logger.Trace("Post backlog event", "msg", msg)
 
-			go c.sendEvent(backlogEvent{
-				src: src,
-				msg: msg,
+				go c.sendEvent(backlogEvent{
```
nambrot: So this may be wrong, but I'm wondering if we have a bug here in that when we put messages on the backlog, we check that the address signing the messages is in the current validator set. However when we pull from the backlog, it seems to me that we no longer check that the address is still in the validator set (as it might be from a different epoch). If I see correctly, this means that a validator could possibly double their voting power by sending a future messages during the old epoch, change their key, and then be able to send another PREPARE/COMMIT with their new key. Admittedly, they would have to know the block of the future, but that still sounds bad?

consensus/istanbul/core/backlog.go
```diff
@@ -103,114 +116,181 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
+						view = rc.View
+					}
 				}
-			case istanbul.MsgRoundChange:
-				var rc *istanbul.RoundChange
-				err := msg.Decode(&rc)
-				if err == nil {
-					view = rc.View
+				if view == nil {
+					logger.Debug("Nil view", "msg", msg)
```
nambrot: Should this not be a more worrying level given that a message in this sequence backlog somehow we can extract the view of?

consensus/istanbul/core/backlog.go
asaj: What if they're from a future round?

consensus/istanbul/core/core.go
asaj: > Should this be an lru cache? Not sure if this is a practical concern, but otherwise we'll keep keys of old addresses around forever even if they have been rotated

Won't we clear them out when we drain the backlog?

consensus/istanbul/core/backlog.go
asaj: Nice catch. Changed the handler for `backlogEvents` to call `handleMsg` instead of `handleCheckedMessage`

consensus/istanbul/core/backlog.go
asaj: Changed to `Error`

consensus/istanbul/core/backlog.go
```diff
@@ -95,115 +108,178 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
-					view = sub.View
+				if view == nil {
+					logger.Error("Nil view", "msg", msg)
+					// continue
+					return
 				}
-			case istanbul.MsgRoundChange:
-				var rc *istanbul.RoundChange
-				err := msg.Decode(&rc)
+				err := c.checkMessage(msg.Code, view)
```
asaj: This feels buggy

consensus/istanbul/core/backlog.go
```diff
@@ -95,115 +108,178 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
+
+// Drain a backlog for a given sequence, passing each to optional callback.
+// Call with backlogsMu held.
+func (c *core) drainBacklogForSeq(seq uint64, cb func(*istanbul.Message, istanbul.Validator)) {
+	backlogForSeq := c.backlogBySeq[seq]
+	if backlogForSeq == nil {
+		return
+	}
+
+	for !backlogForSeq.Empty() {
```
timmoreton: check the size here, only iterate that many times, and afterwards only delete if it's empty.

consensus/istanbul/core/backlog.go
timmoreton: what specifically?

consensus/istanbul/core/backlog.go
```diff
@@ -95,115 +108,178 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
+		m := backlogForSeq.PopItem()
+		msg := m.(*istanbul.Message)
+		if cb != nil {
+			_, src := c.valSet.GetByAddress(msg.Address)
+			if src != nil {
+				cb(msg, src)
+			}
+		}
+		c.backlogCountByVal[msg.Address]--
```
timmoreton: if == 0, we should remove the key

consensus/istanbul/core/core.go
timmoreton: yeah, I think if we just remove entries when the value falls to zero should be ok.  Then the map is at most 10k entries worst case. (and in practice, way smaller)

consensus/istanbul/core/handler.go
```diff
@@ -113,9 +113,12 @@ func (c *core) handleEvents() {
 					c.logger.Debug("Error in handling istanbul message", "err", err)
 				}
 			case backlogEvent:
-				// No need to check signature for internal messages
-				if err := c.handleCheckedMsg(ev.msg, ev.src); err != nil {
-					c.logger.Warn("Error in handling istanbul message that was sent from a backlog event", "err", err)
+				if payload, err := ev.msg.Payload(); err != nil {
+					c.logger.Error("Error in retrieving payload from istanbul message that was sent from a backlog event", "err", err)
+				} else {
+					if err := c.handleMsg(payload); err != nil {
```
timmoreton: This shouldn't have changed and should be `handleCheckedMsg`

consensus/istanbul/core/backlog.go
asaj: After double checking, we actually were okay here, but no problem in calling `handleMsg` so for safety let's do that

consensus/istanbul/core/backlog.go
asaj: I don't remember TBH

consensus/istanbul/core/backlog.go
```diff
@@ -95,115 +108,178 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
+
+	backlogForSeq.Push(msg, toPriority(msg.Code, v))
+
+	// Keep backlog below total max size by pruning future-most sequence first
+	// (we always leave one sequence's entire messages and rely on per-validator limits)
+	if c.backlogTotal > acceptMaxFutureMessages {
+		backlogSeqs := c.getSortedBacklogSeqs()
+		for i := len(backlogSeqs) - 1; i > 0; i-- {
+			seq := backlogSeqs[i]
+			if seq <= c.currentView().Sequence.Uint64() ||
```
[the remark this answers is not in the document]
asaj: Not sure I agree, don't we want to stop draining if we hit the current sequence?

consensus/istanbul/core/backlog.go
```diff
@@ -95,115 +108,178 @@ func (c *core) storeBacklog(msg *istanbul.Message, src istanbul.Validator) {
...
+
+	c.backlogsMu.Lock()
+	defer c.backlogsMu.Unlock()
+
+	// Check and inc per-validator future message limit
+	if c.backlogCountByVal[msg.Address] > acceptMaxFutureMsgsFromOneValidator {
+		logger.Trace("Dropping: backlog exceeds per-src cap", "src", src)
+		return
+	}
+	c.backlogCountByVal[src.Address()]++
```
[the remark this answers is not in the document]
asaj: Done

consensus/istanbul/core/handler.go
asaj: Are you sure?

consensus/istanbul/core/core.go
nambrot: Basically was looking for https://github.com/celo-org/celo-blockchain/pull/517/files/6f0aed30022a89bb06009932c12969fc7a521cd8#r348840207

consensus/istanbul/core/backlog.go
nambrot: Out of curiousity, how were we ok?

consensus/istanbul/core/backlog.go
nambrot: We are draining for the current sequence? https://github.com/celo-org/celo-blockchain/pull/517/files/6f0aed30022a89bb06009932c12969fc7a521cd8#diff-3bcb34249a8e59b2eca3deb619ec7238R233

consensus/istanbul/core/backlog.go
timmoreton: Looking at this, I think we can evade signature checking here still. When a new message comes in, `handleMsg` checks signature + validator in current valset. Then it calls `handleCheckedMsg` which is where for future messages `storeBacklog` is called. By the time this callback handles the message again, the epoch could have changed, so the validator that sent it may no longer be in the active set. WDYT?

consensus/istanbul/core/backlog.go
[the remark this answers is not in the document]
timmoreton: not sure it's a big deal. If everyone has moved on to a new sequence, we could be throwing away those messages here (but only in the event that we are being heavily spammed by other msgs in that next sequence?).

consensus/istanbul/core/backlog.go
timmoreton: yes let's be super sure of this!

mcortesi:
I'm doing a review + cleanup of the backlog messages... 

Basically, there are 2 flows:

A) When a message is received for the fist time it goes:
   1. `handleMessage()` (do some checkings)
   2. `handleCheckedMessage()`
   3. `handleXXX` where XXX is the message type (Preprepare, prepare, etc)
   4. `handleXXX` will call `checkMessage()` which will determine is the msg is old, future or current
   5. when it returns to `handleCheckedMessage()` we will check if it's a futureMessage and add it to the backlog.

B) message comes from the backlog, same as (A), but we skip (1) (at least before this PR)

Now, a few findings:

1. `handleMessage` will verify if the sender is in the current valset; thus "future messages" from the next epoch won't ever be stored in the backlog.

2. Backlog messages where "verified" the first time... where verification means:
      * Check that the msg signature is from a validator in the current validator set
      * Check that `msg.address` is from a validator in the current validator set (which is redundant since we already check the signature, and address from signature and msg.address must match)

Now, if we **asume** that the valSet remains constant between the time the msg was received for the first time, and the time the msg is reprocessed from the backlog. Then, we don't need to check the message again.

Again, from finding (1) we can assume that.

This should answer https://github.com/celo-org/celo-blockchain/pull/517#discussion_r348858936

Also, finding nro 1, seems like something that could be problematic when you are a little behind... but i suppose this means you need to first sync to start participating.

mcortesi:
Second comment...

I think that `checkMessage()` could use some ❤️.

In fact; when we check RoundChange message i've found that If the `msgView.Sequence is < current.Sequence` it should always be and oldMessage…
still in that case.. we check the round number… (which we should only check if sequences are the same, then it could be either oldMessage or nil (which signals we will process the msg)

```go
// Round change messages should be in the same sequence but be >= the desired round
	if msgCode == istanbul.MsgRoundChange {
		if msgView.Sequence.Cmp(c.current.Sequence()) > 0 {
			return errFutureMessage
		} else if msgView.Round.Cmp(c.current.DesiredRound()) < 0 {
			return errOldMessage
		}
		return nil
	}
```

The function pseudocode imo should be something like:

```python
if msgView < currentView:
   # check if commit from last block, if not oldMessage
   return oldMessage
else if msgView.Sequence > currentView.Sequence:
   return futureMessage #(or tooFarIntTheFuture message)
else:
   # same sequence, msgView.round >= currentView.round
   # handle RoundChange message, check round, desiredRound and msg.Code
   # against current state
```

consensus/istanbul/core/handler.go
mcortesi: check my comments below

mcortesi:
> I'm doing a review + cleanup of the backlog messages...
> 
> Basically, there are 2 flows:
> 
> A) When a message is received for the fist time it goes:
> 
> 1. `handleMessage()` (do some checkings)
> 2. `handleCheckedMessage()`
> 3. `handleXXX` where XXX is the message type (Preprepare, prepare, etc)
> 4. `handleXXX` will call `checkMessage()` which will determine is the msg is old, future or current
> 5. when it returns to `handleCheckedMessage()` we will check if it's a futureMessage and add it to the backlog.
> 
> B) message comes from the backlog, same as (A), but we skip (1) (at least before this PR)
> 
> Now, a few findings:
> 
> 1. `handleMessage` will verify if the sender is in the current valset; thus "future messages" from the next epoch won't ever be stored in the backlog.
> 2. Backlog messages where "verified" the first time... where verification means:
>    
>    * Check that the msg signature is from a validator in the current validator set
>    * Check that `msg.address` is from a validator in the current validator set (which is redundant since we already check the signature, and address from signature and msg.address must match)
> 
> Now, if we **asume** that the valSet remains constant between the time the msg was received for the first time, and the time the msg is reprocessed from the backlog. Then, we don't need to check the message again.
> 
> Again, from finding (1) we can assume that.
> 
> This should answer [#517 (comment)](https://github.com/celo-org/celo-blockchain/pull/517#discussion_r348858936)
> 
> Also, finding nro 1, seems like something that could be problematic when you are a little behind... but i suppose this means you need to first sync to start participating.

Actually, there's a case where `handleMEssage` is needed for backlog messages.

A validator from the current set sends a future message for the next epoch... It would be accepted and stored, but later when epoch changes it won't be valid (if the validator is not on the new epoch valset)

consensus/istanbul/core/backlog.go
mcortesi: See below.. 

 * only messages from the current valSet are stored on the backlog
 * if a future messages was stored, but belongs to the previous epoch we should check it again

mcortesi:
Guys, please check https://github.com/celo-org/celo-blockchain/pull/672

Which continues the development of this PR, and updates it to the last changes

asaj:
Closing in favor of https://github.com/celo-org/celo-blockchain/pull/672

Issue #396, opened 2019-08-14
Validators SBAT recover consensus state on restart


timmoreton:
### Expected Behavior

Make Istanbul crash consistent. PBFT relies on a node persisting its progress through a round (i.e has it sent a prepare).  

### Current Behavior

Behavior if a validator node restarts and rejoins a consensus round that it already participated in is undefined.

ashishb:
#520 merged. This issue can be closed as soon as #531 is merged in.

mcortesi:
Solved on #632

PR #520, opened 2019-10-10
[IBFT]Persist Proposer state across restarts

ashishb:
### Description

Persist Proposer state across restarts. So, if a proposer crashes immediately after proposing a new block then on restart, it will propose the same block again. This prevents double-signing as well as potential deadlock.

### Tested

Change the code in `preprepare.go`. Create a module-level boolean `crashOnReceivingPrepareAsProposer`, initialize it false and in `getPreprepareMessage , 
set it to `crashOnReceivingPrepareAsProposer = existingPreparedMessage == nil`

In `handlePreprepare`, `panic` if `crashOnReceivingPrepareAsProposer` is true.

Build the docker images, deploy a new testnet and verify that the testnet progresses slowly with crashes being visible in validators-[0-3] logs.

### Related issues

- Partially addresses #396


consensus/istanbul/core/persistent_state.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return err
+	}
+	defer fp.Close()
+
+	_, err2 := fp.Write(data)
+	if err2 != nil {
+		return err2
+	}
+	err3 := fp.Sync()
```
timmoreton: I don't think calling `Sync` on the file descriptor is enough -- the file is new and so its directory entry also needs syncing. see https://linux.die.net/man/2/open

consensus/istanbul/core/persistent_state.go
```diff
@@ -0,0 +1,96 @@
...
+	fp, err := os.OpenFile(fileName, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	if err != nil {
+		return err
+	}
+	defer fp.Close()
+
+	_, err2 := fp.Write(data)
+	if err2 != nil {
+		return err2
+	}
```
timmoreton: it would be good to wrap whether we call sync in a flag, so we can test its performance impact

consensus/istanbul/core/persistent_state.go
```diff
@@ -0,0 +1,96 @@
...
+	return data, err
+}
+
+// deletePrepareMessageFromDisk deletes all files from this round
+func (c *core) deletePrepareMessageFromDisk(roundNumber *big.Int) error {
+	dir := c.backend.GetDataDir()
+	// This pattern must be  similar to the filenames generated by
+	// generateFileName function.
+	filePattern := filepath.Join(dir,
+		fmt.Sprintf("geth_istanbul_state_type_.*_round_%s_.*", roundNumber.String()))
```
timmoreton: we shouldn't delete files associated with (possibly subsequent) sequences all at once here.

consensus/istanbul/core/persistent_state.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return nil
+}
+
+func (c *core) generateFileName(
+	messageType uint64,
+	roundNumber *big.Int,
+	sequenceNumber *big.Int) (string, error) {
+	dir := c.backend.GetDataDir()
+	fileName := fmt.Sprintf("geth_istanbul_state_type_%d_round_%s_sequence_%s",
```
timmoreton: makes more sense to have this structure be `(sequence, round, type)`

consensus/istanbul/core/persistent_state.go
```diff
@@ -0,0 +1,96 @@
...
+	if os.IsNotExist(err) {
+		log.Debug("getPreprepareMessageFromDisk/file does not exist", "file", fileName)
+		return nil, nil
+	}
+	log.Debug("getPreprepareMessageFromDisk/file found on the disk", "file", fileName)
+	return data, err
+}
+
+// deletePrepareMessageFromDisk deletes all files from this round
+func (c *core) deletePrepareMessageFromDisk(roundNumber *big.Int) error {
```
timmoreton: consider renaming - since it deletes files of any type of message

consensus/istanbul/core/persistent_state.go
```diff
@@ -0,0 +1,96 @@
+package core
+
+import (
+	"fmt"
+	"github.com/ethereum/go-ethereum/log"
```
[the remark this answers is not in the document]
ashishb: Done

consensus/istanbul/backend/snapshot_test.go
```diff
@@ -256,7 +258,8 @@ func TestValSetChange(t *testing.T) {
 			headers: make(map[uint64]*types.Header),
 		}
 
-		engine := New(config, db).(*Backend)
+		dataDir := filepath.Join("/tmp", string(rand.Int()))
```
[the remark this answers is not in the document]
ashishb: 1. I couldn't find one in the codebase.
2. I tried importing but it didn't work, so, I decided to duplicate. Since this duplication is only in the test code, I think it is not a major issue.

consensus/istanbul/core/preprepare.go
```diff
@@ -29,27 +30,61 @@ func (c *core) sendPreprepare(request *istanbul.Request, roundChangeCertificate
...
+		RoundChangeCertificate: roundChangeCertificate,
+	})
+	if err != nil {
+		logger.Error("Failed to encode", "view", curView)
+		return nil, err
+	}
+	err2 := c.savePrepareMessageToDisk(messageType, roundNumber, sequenceNumber, preprepare)
+	if err2 != nil {
+		logger.Error("Failed to write prepare message to the disk", "msg", preprepare)
+		return nil, err2
```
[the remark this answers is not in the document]
ashishb: And risk double-signing?

eth/backend.go
```diff
@@ -273,6 +275,30 @@ func CreateConsensusEngine(ctx *node.ServiceContext, chainConfig *params.ChainCo
 	}
 }
 
+func getDataDirOrFail(ctx *node.ServiceContext) string {
+	dataDir := ctx.ResolvePath("istanbul")
```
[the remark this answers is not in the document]
ashishb: Done and moved the constant to the top of the file.

miner/worker_test.go
```diff
@@ -244,7 +246,8 @@ func getAuthorizedIstanbulEngine() consensus.Istanbul {
 		return signatureBytes, nil
 	}
 
-	engine := istanbulBackend.New(istanbul.DefaultConfig, ethdb.NewMemDatabase())
+	dataDir := filepath.Join("/tmp", string(rand.Int()))
```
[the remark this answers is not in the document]
ashishb: Extracted into a variable.

consensus/istanbul/core/persistent_state.go
ashishb: Interesting, I didn't know that. I added dir syncing to `writeToDisk`. Please confirm that it looks correct to you.

consensus/istanbul/core/persistent_state.go
ashishb: How about doing that in a follow-on PR? Adding flags is easy, it modifies multiple files though making PR look unusually big.

consensus/istanbul/core/persistent_state.go
ashishb: Changed the code to be conservative and delete only the files from current round and sequence.

consensus/istanbul/core/persistent_state.go
ashishb: Changed to `geth_istanbul_sequence_%s_round_%s_type_%d`

consensus/istanbul/core/persistent_state.go
ashishb: Renamed to `deleteMessageFromDisk`

consensus/istanbul/core/persistent_state.go
trianglesphere: My one concern with how we delete old messages is that we need to be careful in when we call this function, likely for the range of [oldest prepare found, curr seq) on each new seq if there is a file saved here or just for the previous seq when starting a new seq.

It's the challenge of trying to match numeric equality to a regex.

consensus/istanbul/core/core.go
```diff
@@ -289,6 +289,7 @@ func (c *core) startNewRound(round *big.Int) {
 	var roundChangeCertificate istanbul.RoundChangeCertificate
 	var request *istanbul.Request
 	if roundChange {
+		c.deletePrepareMessageFromDisk(c.current.Round())
```
trianglesphere: Might also have to delete old saved prepare messages when starting a new sequence.

consensus/istanbul/core/persistent_state.go
ashishb: Fair point. @trianglesphere , for now, I have reduced the code to only delete files for a particular sequence number and round number. Missing deletion does not affect correctness of the code and more can be added as a follow-on.

consensus/istanbul/core/core.go
ashishb: Changed to delete messages for a particular (sequence, round) pair.

PR #522, opened 2019-10-11
Governable block gas limit

### LINKED ISSUE #474: Protocol SBAT set block gas limit via governance
### Expected Behavior

- [ ] Governable parameter for block gas limit. Default value should probably be >=10M.

- [ ] Read this limit from smart contract at the start of every block.

### Current Behavior

Determined by miners/validators as per Ethereum.

mrsmkl:
### Description

Reads the block gas limit from a governable contract.

### Tested

I'll add test to the corresponding PR https://github.com/celo-org/celo-monorepo/pull/1253

### Related issues

- Fixes #474


asaj:
@mrsmkl when this is ready to review would you mind assigning me (or someone else?)

consensus/istanbul/backend/engine_test.go
```diff
@@ -249,7 +249,7 @@ func makeHeader(parent *types.Block, config *istanbul.Config) *types.Header {
 	header := &types.Header{
 		ParentHash: parent.Hash(),
 		Number:     parent.Number().Add(parent.Number(), common.Big1),
-		GasLimit:   core.CalcGasLimit(parent, parent.GasLimit(), parent.GasLimit()),
+		GasLimit:   core.CalcGasLimit(parent, nil, parent.GasLimit(), parent.GasLimit()),
```
asaj: Do we need to store this in the block header at all anymore, since it can be read from the state trie?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -106,3 +120,21 @@ func SpawnCheck() {
...
+
+func GetBlockGasLimit(header *types.Header, state vm.StateDB) (uint64, error) {
+	var gasLimit *big.Int
+	_, err := contract_comm.MakeStaticCall(
+		params.BlockchainParametersRegistryId,
+		blockchainParametersABI,
+		"blockGasLimit",
+		[]interface{}{},
+		&gasLimit,
+		params.MaxGasForBlockGasLimit,
```
asaj: Should we have a shared `MaxGasForReadingBlockchainParameter` parameter? Since presumably the cost will be the same for all?

consensus/istanbul/backend/engine_test.go
mrsmkl: Well there is still the time before the contracts are deployed, how should that be handled?

asaj: Good question. Should we have a fallback value that we define in protocol_params.go?

consensus/istanbul/backend/engine_test.go
```diff
@@ -270,7 +270,7 @@ func makeHeader(parent *types.Block, config *istanbul.Config) *types.Header {
 	header := &types.Header{
 		ParentHash: parent.Hash(),
 		Number:     parent.Number().Add(parent.Number(), common.Big1),
-		GasLimit:   core.CalcGasLimit(parent, parent.GasLimit(), parent.GasLimit()),
+		GasLimit:   core.CalcGasLimit(parent, nil),
```
asaj: I think we should remove this field from the block header

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -146,3 +160,21 @@ func SpawnCheck() {
...
+		[]interface{}{},
+		&gasLimit,
+		params.MaxGasForReadBlockchainParameter,
+		header,
+		state,
+	)
+	if err != nil {
+		return params.DefaultGasLimit, err
+	}
+	return gasLimit.Uint64(), err
```
asaj: nit: s/err/nil would be mildly more readable IMO

params/protocol_params.go
```diff
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	GasLimitBoundDivisor uint64 = 1024    // The bound divisor of the gas limit, used in update calculations.
-	MinGasLimit          uint64 = 5000    // Minimum the gas limit may ever be.
-	GenesisGasLimit      uint64 = 4712388 // Gas limit of the Genesis block.
+	GasLimitBoundDivisor uint64 = 1024     // The bound divisor of the gas limit, used in update calculations.
```
asaj: Is this still used?

params/protocol_params.go
```diff
@@ -24,9 +24,10 @@ import (
...
+	MinGasLimit          uint64 = 5000     // Minimum the gas limit may ever be.
```
asaj: Is this still used?

params/protocol_params.go
```diff
@@ -24,9 +24,10 @@ import (
...
+	GenesisGasLimit      uint64 = 4712388  // Gas limit of the Genesis block.
```
asaj: Is this still used? How is it different from `DefaultGasLimit`?

params/protocol_params.go
```diff
@@ -24,9 +24,10 @@ import (
...
+	DefaultGasLimit      uint64 = 10000000 // Gas limit of the Genesis block.
```
asaj: Make this 20M

asaj:
It looks like there's an end-to-end test failure

params/protocol_params.go
mrsmkl: only used in puppeth, seems safe to remove

params/protocol_params.go
mrsmkl: only used in puppeth, seems safe to remove

params/protocol_params.go
mrsmkl: Changing this broke some tests, will investigate further

consensus/istanbul/backend/engine_test.go
mrsmkl: Perhaps this should be in another PR, changing the header seems to break many things in the tests.

asaj: Works for me!

PR #524, opened 2019-10-11
Add incentives to proof-of-stake

asaj:
### Description

This PR adds incentives to proof-of-stake, specifically epoch payments for validators and groups, and epoch rewards for voters.

### Tested

end-to-end tests

### Other changes

- Precompile for epoch size
### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Not backwards compatible


contract_comm/evm.go
```diff
@@ -67,6 +67,11 @@ func NewEVMContext(msg types.Message, header *types.Header, chain ChainContext,
 		beneficiary = *author
 	}
 
+	var engine consensus.Engine
+	if chain != nil {
```
kevjue: I have two comments regarding this nil check:

1) Be aware that you are checking if the interface value is nil, NOT if the underlying value of the interface is nil.  So if this function is ever invoked with a `chain` parameter that is set to nil, then the next line will panic (as opposed to if the function is invoked with a "constant" `nil` parameter for that parameter).
2) When would chain be nil?  Seems like if it is, then your new `getEpochSize` precompile will panic, as it assumes that engine will be non nil.

consensus/istanbul/backend/engine.go
```diff
@@ -415,10 +363,8 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
 	// If this is the last block of the epoch, then get the validator set diff, to save into the header
 	log.Trace("Called UpdateValSetDiff", "number", header.Number.Uint64(), "epoch", sb.config.Epoch)
 	if istanbul.IsLastBlockOfEpoch(header.Number.Uint64(), sb.config.Epoch) {
-		newValSet, err := validators.GetValidatorSet(header, state)
-		if err != nil {
-			log.Error("Istanbul.Finalize - Error in retrieving the validator set. Using the previous epoch's validator set", "err", err)
-		} else {
+		newValSet, err := sb.getNewValidatorSet(header, state)
+		if err == nil {
```
kevjue: Why remove the error message is err is non nil?  Also, seems like if there ever was an error from here: 
 https://github.com/celo-org/celo-blockchain/pull/524/files#diff-6e58dd6edf43343da874fb58a00f6727R368, it would be silently ignored.

Also, should this return the error if it's non nil?

consensus/istanbul/backend/engine.go
```diff
@@ -443,95 +389,166 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
+		header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	}
 
-		if governanceAddress != nil {
-			state.AddBalance(*governanceAddress, infrastructureBlockReward)
-			totalBlockRewards.Add(totalBlockRewards, infrastructureBlockReward)
+	if istanbul.IsLastBlockOfEpoch(header.Number.Uint64(), sb.config.Epoch) {
+		err = sb.updateValidatorScoresAndDistributeEpochPaymentsAndRewards(header, state)
+		if err == nil {
+			header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
```
kevjue: Can you just call `state.IntermediateRoot` once at the end of the method?

consensus/istanbul/backend/engine.go
```diff
@@ -443,95 +389,166 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
+	return sb.config.Epoch
+}
+
 // Finalize runs any post-transaction state modifications (e.g. block rewards)
 // and assembles the final block.
 //
 // Note, the block header and state database might be updated to reflect any
 // consensus rules that happen at finalization (e.g. block rewards).
 func (sb *Backend) Finalize(chain consensus.ChainReader, header *types.Header, state *state.StateDB, txs []*types.Transaction, uncles []*types.Header, receipts []*types.Receipt, randomness *types.Randomness) (*types.Block, error) {
+	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
```
kevjue: Looks like all of the errors will be ignored in this method.  Is that intended?

consensus/istanbul/backend/engine.go
```diff
@@ -443,95 +389,166 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
-			state.AddBalance(*lockedGoldAddress, stakerBlockReward)
-			totalBlockRewards.Add(totalBlockRewards, stakerBlockReward)
-			_, err := contract_comm.MakeCallWithAddress(*lockedGoldAddress, setCumulativeRewardWeightFuncABI, "setCumulativeRewardWeight", []interface{}{stakerBlockReward}, nil, 1000000, common.Big0, header, state)
-			if err != nil {
-				log.Error("Unable to send epoch rewards to LockedGold", "err", err)
-				return nil, err
-			}
+func (sb *Backend) distributeEpochPayments(header *types.Header, state *state.StateDB, valSet []istanbul.Validator) error {
+	for _, val := range valSet {
+		log.Info("Distributing epoch payment for address", "address", val.Address())
```
kevjue: sb.logger?

consensus/istanbul/backend/engine.go
```diff
@@ -443,95 +389,166 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
+	}
+	return sb.increaseGoldTokenTotalSupply(header, state, totalEpochRewards)
+}
+
+func (sb *Backend) updateValidatorScores(header *types.Header, state *state.StateDB, valSet []istanbul.Validator) error {
+	for _, val := range valSet {
+		// TODO: Use actual uptime metric.
+		// 1.0 in fixidity
+		uptime := math.BigPow(10, 24)
+		log.Info("Updating validator score for address", "address", val.Address(), "uptime", uptime.String())
```
kevjue: sb.logger?

params/protocol_params.go
```diff
@@ -97,8 +97,9 @@ const (
 	// TODO: make this cost variable- https://github.com/celo-org/geth/issues/250
 	FractionMulExpGas uint64 = 1050 // Cost of performing multiplication and exponentiation of fractions to an exponent of up to 10^3.
 	// TODO(kobigurk):  Figure out what the actual gas cost of this contract should be.
-	ProofOfPossessionGas uint64 = 1050 // Cost of verifying a BLS proof of possession.
-	GetValidatorGas      uint64 = 5000 // Cost of reading a validator's address
+	ProofOfPossessionGas uint64 = 50000 // Cost of verifying a BLS proof of possession.
+	GetValidatorGas      uint64 = 5000  // Cost of reading a validator's address
+	GetEpochSizeGas      uint64 = 1000
```
kevjue: Seems like there be a comment here (at least for consistency sake).

consensus/istanbul/backend/engine.go
kevjue: As we discussed, it does appear that the contract_call library functions will output an evm error if it's encountered.

contract_comm/election/election.go
```diff
@@ -0,0 +1,208 @@
...
+					lesser = voteTotals[i-1].Group
+				}
+				if i+1 < len(voteTotals) {
+					greater = voteTotals[i+1].Group
+				}
+				break
+			}
+		}
+		_, err := contract_comm.MakeCall(params.ElectionRegistryId, electionABI, "distributeEpochRewards", []interface{}{group, reward, lesser, greater}, nil, params.MaxGasForDistributeEpochRewards, common.Big0, header, state)
+		if err != nil {
```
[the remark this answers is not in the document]
kevjue: As we discussed, your plan is to rollback the state if there is an error in the loop.

consensus/istanbul/backend/engine.go
```diff
@@ -443,95 +389,166 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
+		err := validators.UpdateValidatorScore(header, state, val.Address(), uptime)
+		if err != nil {
```
kevjue: Here's a case where a "partial update" would be done if there ever is a failure mid-loop.

contract_comm/election/election.go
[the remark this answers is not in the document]
kevjue: Also, note that you'll need to "rollback" the totalRewards variable as well.

contract_comm/election/election.go
```diff
@@ -0,0 +1,208 @@
...
+	_, err := contract_comm.MakeStaticCall(params.ElectionRegistryId, electionABI, "getGroupEpochRewards", []interface{}{group, totalEpochRewards}, &groupEpochRewards, params.MaxGasForGetGroupEpochRewards, header, state)
+	if err != nil {
+		return nil, err
+	}
+	return groupEpochRewards, nil
+}
+
+func DistributeEpochRewards(header *types.Header, state vm.StateDB, groups []common.Address, totalEpochRewards *big.Int) (*big.Int, error) {
+	totalRewards := big.NewInt(0)
+	voteTotals, err := getTotalVotesForEligibleValidatorGroups(header, state)
```
mcortesi: So we have two moments:
 - beginning of epoch (in reality end of the previous one)
 - end of the epoch

It seems we are querying state about eligible groups at the end, when we should do so at the beginning? is this correct?

contract_comm/evm.go
kevjue: section 7.5.1 of the golang book has a good description about this.

In summary, the way to solve it is that to make sure that the arguments passed to the function is an interface type, as opposed to a "real" type.

consensus/istanbul/backend/engine.go
asaj: Great catch. Using `Snapshot` and `RevertToSnapshot` should fix this

consensus/istanbul/backend/engine.go
```diff
@@ -443,95 +389,166 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
+	totalEpochRewards := big.NewInt(0)
+
+	// Fixed epoch reward to the infrastructure fund.
+	infrastructureEpochReward := big.NewInt(params.Ether)
+	governanceAddress, err := contract_comm.GetRegisteredAddress(params.GovernanceRegistryId, header, state)
+	if err != nil {
+		return totalEpochRewards, err
+	}
+
+	if governanceAddress != nil {
```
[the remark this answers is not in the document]
asaj: Yup! Very much a WIP, added a comment

contract_comm/election/election.go
asaj: The way we distribute rewards to voters is by increasing the number of active votes for various groups. When we increase the number of votes for an eligible group, we need to sort that group into the list of all eligible groups. In order to know where to sort that group, we need to know the list of all eligible groups and their vote totals, which is what we're fetching here.

contract_comm/election/election.go
```diff
@@ -0,0 +1,208 @@
...
+			return totalRewards, err
+		}
+		rewards[i] = reward
+	}
+
+	for i, group := range groups {
+		reward := rewards[i]
+		for _, voteTotal := range voteTotals {
+			if voteTotal.Group == group {
+				voteTotal.Value.Add(voteTotal.Value, reward)
```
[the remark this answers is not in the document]
asaj: `voteTotal.Value` is a `*big.Int`, so I believe so, but please double check

contract_comm/election/election.go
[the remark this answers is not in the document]
asaj: > Also, note that you'll need to "rollback" the totalRewards variable as well.

Since we roll this all into `updateValidatorScoresAndDistributeEpochPaymentsAndRewards`, and revert if that fails, we should be doing this, but please double check

contract_comm/evm.go
asaj: I don't have the book handy, how do I do this?

contract_comm/election/election.go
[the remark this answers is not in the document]
mcortesi: mutable!

contract_comm/evm.go
```diff
@@ -232,5 +238,9 @@ func makeCallWithContractId(registryId [32]byte, abi abi.ABI, funcName string, a
 		}
 	}
 
-	return executeEVMFunction(*scAddress, abi, funcName, args, returnObj, gas, value, header, state, shouldMutate)
+	gasLeft, err := executeEVMFunction(*scAddress, abi, funcName, args, returnObj, gas, value, header, state, shouldMutate)
+	if err != nil {
```
asaj: Should log the error here as well

core/vm/contracts.go
```diff
@@ -662,15 +665,33 @@ func (c *numberValidators) RequiredGas(input []byte) uint64 {
 }
 
 func (c *numberValidators) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
-	blockNumber := evm.Context.BlockNumber
-	validators := evm.Context.Engine.GetValidators(blockNumber, evm.Context.GetHash(blockNumber.Uint64()))
 	gas, err := debitRequiredGas(c, input, gas)
 	if err != nil {
```
timmoreton: we could check `len(input)!=0` here too

core/vm/contracts.go
```diff
@@ -630,13 +632,14 @@ func (c *getValidator) RequiredGas(input []byte) uint64 {
 }
 
 func (c *getValidator) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
-	blockNumber := evm.Context.BlockNumber
-	validators := evm.Context.Engine.GetValidators(blockNumber, evm.Context.GetHash(blockNumber.Uint64()))
 	gas, err := debitRequiredGas(c, input, gas)
 	if err != nil {
```
timmoreton: we could check `len(input)!=0` here too

contract_comm/election/election.go
```diff
@@ -54,3 +124,85 @@ func GetElectedValidators(header *types.Header, state vm.StateDB) ([]common.Addr
 	}
 	return newValSet, nil
 }
+
+type voteTotal struct {
+	Group common.Address
+	Value *big.Int
+}
+
+func getTotalVotesForEligibleValidatorGroups(header *types.Header, state vm.StateDB) ([]voteTotal, error) {
```
timmoreton: I think this is the first time we've introduced the notion of a group in celo-blockchain -- previously just limited to the contracts. Not saying it's a problem per se, just a layering observation.

contract_comm/election/election.go
```diff
@@ -54,3 +124,85 @@ func GetElectedValidators(header *types.Header, state vm.StateDB) ([]common.Addr
...
+	var groups []common.Address
+	var values []*big.Int
+	_, err := contract_comm.MakeStaticCall(params.ElectionRegistryId, electionABI, "getTotalVotesForEligibleValidatorGroups", []interface{}{}, &[]interface{}{&groups, &values}, params.MaxGasForGetEligibleValidatorGroupsVoteTotals, header, state)
+	if err != nil {
+		log.Error("DistributeEpochRewards, error calling getTotalVotesForEligibleValidatorGroups", "err", err)
+	}
+
+	voteTotals := make([]voteTotal, len(groups))
+	for i, group := range groups {
+		log.Info("Got group vote total", "group", group, "value", values[i])
```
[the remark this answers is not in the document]
asaj: Trace

core/vm/contracts.go
asaj: I think we want len(input) to equal 0 here

contract_comm/evm.go
kevjue: IMO, I think it's best to make it so that the concrete value of `chain` is always non-nil and to not do the `if` check.

If that's not doable, then it will get messy.  E.g. you could use introspective via the `reflect` package, and do something like this:

`if c == nil || (reflect.ValueOf(c).Kind() == reflect.Ptr && reflect.ValueOf(c).IsNil())`

where `c` is the interface variable.

kevjue: IMO, I think it would be best to always assume that chain is not `nil` and it's contract value is not nil, so that the `if` condition is removed.   Seems like we have precompiles that now depend on it not being nil, such as your `epochsize` and the validator set related ones.

kevjue:
Please see this comment:  https://github.com/celo-org/celo-blockchain/pull/524#discussion_r337844746

consensus/istanbul/backend/engine.go
```diff
@@ -499,6 +433,130 @@ func (sb *Backend) Finalize(chain consensus.ChainReader, header *types.Header, s
...
+	return nil
+}
+
+func (sb *Backend) distributeEpochRewards(header *types.Header, state *state.StateDB, valSet []istanbul.Validator) (*big.Int, error) {
+	totalEpochRewards := big.NewInt(0)
+
+	// Fixed epoch reward to the infrastructure fund.
+	// TODO(asa): This should be a fraction of the overal reward to stakers.
+	infrastructureEpochReward := big.NewInt(params.Ether)
+	governanceAddress, err := contract_comm.GetRegisteredAddress(params.GovernanceRegistryId, header, state)
```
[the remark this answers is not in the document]
asaj: Yeah that's intended behavior.

contract_comm/evm.go
asaj: Still not totally clear on this, let's chat in person

asaj: Removed this check

PR #525, opened 2019-10-13
Add & check seals from parent block in current block

### LINKED ISSUE #456: Block proposers SBAT include signatures for the previous block

gakonst:
### Description

Solves:
>  Clients should validate signatures when processing a block by checking that the aggregated signature matches the aggregated public key of the previous block. This can be tricky at the epoch boundary as the previous block's validator set may not be easily known (e.g. after a fast sync). It may have to be inferred from the validator diff in the header.

1. Modify the worker to gather the seals it knows from the parent block and insert them in the current block's `parentSeal` struct.
1. Refactor the BLS signature check on a validator set to be callable on the parent block's validator set as well
1. Make the Istanbul backend check that the parent seal on the current block is signed correctly by the parent block's validator set

### Related issues

- part of #456

###

Questions for @asaj:

1. The parentSeal is not part of the header, but we need to verify that it matches the parent's seal. Should we make the `parentSeal` part of the header's extra and move it away from being next to the `Randomness` field? This was approach implemented in #526


gakonst:
Test currently failing on E2E-Governance test due to the validator set changing

```
  1) governance tests Validators.validatorAddressFromCurrentSet() after the validator set changes should revert when asked for an out of bounds validator:
     AssertionError: Expected "revert", got Error: Transaction was not mined within750 seconds, please make sure your transaction was properly sent. Be aware that it might still be mined! instead
      at Object.<anonymous> (src/e2e-tests/utils.ts:457:24)
      at Generator.throw (<anonymous>)
      at rejected (src/e2e-tests/utils.ts:5:65)

  2) governance tests when the validator set is changing should have produced blocks with the correct validator set:

      AssertionError: expected 1 to equal 3
      + expected - actual

      -1
      +3
      
      at Context.<anonymous> (src/e2e-tests/governance_tests.ts:498:31)
      at Generator.next (<anonymous>)
      at src/e2e-tests/governance_tests.ts:7:71
      at new Promise (<anonymous>)
      at __awaiter (src/e2e-tests/governance_tests.ts:3:12)
      at Context.<anonymous> (src/e2e-tests/governance_tests.ts:496:20)
```

consensus/istanbul/backend/backend.go
```diff
@@ -251,6 +255,27 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, bitmap *big.Int, seals []b
 	if err != nil {
 		return err
 	}
+
+	parent := sb.chain.GetHeader(h.ParentHash, h.Number.Uint64()-1)
+	// If the parent was not the genesis block then try to use the parent's
+	// extras as the seal for this block
+	if parent != nil {
+		parentExtras, err := types.ExtractIstanbulExtra(parent)
```
asaj: If we do things this way, I believe we will only ever write 2F+1 signatures to the `parentSeal`, even if we received more signatures than that

consensus/istanbul/backend/backend.go
```diff
@@ -251,6 +255,27 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, bitmap *big.Int, seals []b
...
+		if err != nil {
+			return err
+		}
+
+		block = block.WithParentSeal(
```
asaj: Is `Commit` the right place to do this? I would have thought we would do this in `Prepare`, after we sleep, to give us the most time possible to collect signatures

consensus/istanbul/backend/backend.go
```diff
@@ -251,6 +255,27 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, bitmap *big.Int, seals []b
...
+			&types.BlockSeal{
+				Bitmap: parentExtras.Bitmap,
+				Seal:   parentExtras.CommittedSeal,
+			},
+		)
+	} else {
+		// otherwise initialize with default parent seal
+		block = block.WithParentSeal(&types.EmptyBlockSeal)
```
asaj: I'm assuming this is illegal for all non-genesis blocks?

consensus/istanbul/backend/backend.go
```diff
@@ -313,7 +338,35 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
+		}
+
+		// Get the validator set from the previous block
+		validators := snap.ValSet.Copy()
+
+		// Get the parent seals for this block (ie signature & bitmap)
+		parentSeal := block.ParentSeal()
+
+		// Get the parent block's proposal seal
+		proposalSeal := istanbulCore.PrepareCommittedSeal(block.ParentHash())
```
asaj: I find the naming here a bit confusing.  IIUC, `proposalSeal` is the message that we expect validators to have signed, whereas `parentSeal` is the actual signatures and bitmap.

The latter feels more like a seal (to me) than the former, wdyt?

(Though I understand you just copied this variable name from elsewhere in the code to be consistent)

consensus/istanbul/backend/engine.go
```diff
@@ -334,34 +335,36 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
 	}
 
+	// 1. Get committed seals from current header
 	validators := snap.ValSet.Copy()
-	// Check whether the committed seals are generated by parent's validators
-
 	proposalSeal := istanbulCore.PrepareCommittedSeal(header.Hash())
-	// 1. Get committed seals from current header
-	myValidatorIndex, myValidator := validators.GetByAddress(sb.Address())
+	fmt.Println("Current parent seal: ", extra.CommittedSeal)
```
asaj: Is this the parent?

consensus/istanbul/backend/engine.go
```diff
@@ -454,6 +457,7 @@ func (sb *Backend) IsLastBlockOfEpoch(header *types.Header) bool {
 // Note, the block header and state database might be updated to reflect any
 // consensus rules that happen at finalization (e.g. block rewards).
 func (sb *Backend) Finalize(chain consensus.ChainReader, header *types.Header, state *state.StateDB, txs []*types.Transaction, uncles []*types.Header, receipts []*types.Receipt, randomness *types.Randomness, parentSeal *types.BlockSeal) (*types.Block, error) {
+	// Ensure that the parentSeal
```
asaj: ?

consensus/istanbul/backend/backend.go
```diff
@@ -313,7 +338,35 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
+
+		fmt.Println("Proposal Seal:", common.ToHex(proposalSeal))
```
asaj: s/fmt.Println/sb.logger.X everywhere for any log messages you want to keep around

consensus/istanbul/backend/engine.go
```diff
@@ -559,6 +564,20 @@ func (sb *Backend) Seal(chain consensus.ChainReader, block *types.Block, results
 		return err
 	}
 
+	parentExtras, err := types.ExtractIstanbulExtra(parent)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Got seal from extras: ", parentExtras.Bitmap, parentExtras.CommittedSeal)
+	block = block.WithParentSeal(
```
asaj: I think the block passed to `Seal` should already have the `parentSeal` populated

consensus/istanbul/backend/engine.go
```diff
@@ -618,6 +637,18 @@ func (sb *Backend) updateBlock(parent *types.Header, block *types.Block) (*types
 		return nil, err
 	}
 
+	parentExtras, err := types.ExtractIstanbulExtra(parent)
+	if err != nil {
+		return nil, err
+	}
+
+	block = block.WithParentSeal(
```
asaj: Ditto

asaj:
> The main reason we need the parentSeal in the first place is that different nodes will not necessarily have a consistent view of which validators actually signed each block.

Also note that the seal for a block must necessarily be excluded from the block hash

consensus/istanbul/backend/backend.go
gakonst: Correct

consensus/istanbul/backend/backend.go
gakonst: PR still needs some extra work done, will be cleaning it up from all debugging related logs/println's I have added.

consensus/istanbul/backend/engine.go
gakonst: Typo when debugging

PR #526, opened 2019-10-13
Parent block seals prepare, verify and add to headers

gakonst:
### Description

Adds the seals & bitmap from the parent block to the block header, as an `extra` field.

### Tested

Tests have been modified to pass. We must add a negative-case test for when the parent signature fails.


consensus/istanbul/core/core.go
```diff
@@ -185,36 +185,49 @@ func (c *core) commit() {
...
+		return
+	}
+}
 
-			bitmap.SetBit(bitmap, int(j), 1)
+func (c *core) calculateBitmapAndSeal(proposal istanbul.Proposal) (*big.Int, []byte) {
+	bitmap := big.NewInt(0)
+	publicKeys := [][]byte{}
+	committedSeals := make([][]byte, c.current.Commits.Size())
+	for i, v := range c.current.Commits.Values() {
```
gakonst: This should be replaced with the pubkeys which were present in the last proposal.

consensus/istanbul/backend/engine.go
```diff
@@ -404,6 +405,15 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 		header.Time = big.NewInt(time.Now().Unix())
 	}
 
+	parentExtra, err := types.ExtractIstanbulExtra(parent)
```
gakonst: @asaj Is this what you had in mind? In the `Prepare` we are modifying the header to also include the parent's info.

PR #527, opened 2019-10-14
Governable gas price for transactions paid in non-gold currencies

### LINKED ISSUE #473: Geth SBAT to read the gas values for intrinsic system smart contract calls from a smart contract
### Expected Behavior

Geth should read the gas values for intrinsic system smart contract calls from a smart contract.

### Current Behavior

Those values are currently hard coded in the geth code.  See here:  https://github.com/celo-org/celo-blockchain/blob/e2ed93b9c5647b12ef085104ef7bcf983cf608b4/params/protocol_params.go#L146

mrsmkl:
### Description

Adds a governable parameter to set the extra gas used if a transaction is paid using a non-gold currency.

### Tested

Added tests in https://github.com/celo-org/celo-monorepo/pull/1253

### Related issues

- Fixes #473


asaj:
If I understand correctly, #473 is about storing `ExpectedGasForCreditToTransactions`,  + `ExpectedGasForDebitFromTransactions`, and `ExpectedGasToReadErc20Balance` rather than  `AdditionalGasForNonGoldCurrencies`

@kevjue is that correct?

mrsmkl:
Currently the `ExpectedGas` constants are just used to calculate the `AdditionalGas`, so that's why I use only that here...

kevjue:
That's correct.  It's about storing the individual constants.

core/state_transition.go
```diff
@@ -315,11 +313,13 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 	contractCreation := msg.To() == nil
 
 	// Calculate intrinsic gas.
-	gas, err := IntrinsicGas(st.data, contractCreation, homestead, msg.GasCurrency())
+	gas, err := IntrinsicGas(st.data, contractCreation, homestead)
 	if err != nil {
 		return nil, 0, false, err
 	}
 
+	gas += blockchain_parameters.GetCurrencyGasCost(st.evm.GetHeader(), st.state, msg.GasCurrency())
```
asaj: Why did we pull this cost out of the intrinsic gas calculation?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -83,6 +127,49 @@ func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionI
...
+
+func GetGasForCreditToTransactions(header *types.Header, state vm.StateDB) uint64 {
+	return GetGasCost(header, state, params.ExpectedGasForCreditToTransactions, "gasForCreditToTransactions")
+}
+
+func GetGasToReadErc20Balance(header *types.Header, state vm.StateDB) uint64 {
+	return GetGasCost(header, state, params.ExpectedGasToReadErc20Balance, "gasToReadErc20Balance")
+}
+
+func GetCurrencyGasCost(header *types.Header, state vm.StateDB, gasCurrency *common.Address) uint64 {
```
asaj: nit: Consider a slightly more verbose name here, i.e. `GetIntrinsicGasForAlternateGasCurrency`

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -83,6 +127,49 @@ func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionI
...
+		&gas,
+		params.MaxGasForReadBlockchainParameter,
+		header,
+		state,
+	)
+	if err != nil {
+		log.Info("Default gas", "gas", defaultGas, "method", method)
+		return defaultGas
+	}
+	log.Info("Reading gas", "gas", gas)
```
asaj: `Info` feels pretty verbose here, perhaps `Trace`?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -83,6 +127,49 @@ func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionI
...
+		blockchainParametersABI,
+		method,
+		[]interface{}{},
+		&gas,
+		params.MaxGasForReadBlockchainParameter,
+		header,
+		state,
+	)
+	if err != nil {
+		log.Info("Default gas", "gas", defaultGas, "method", method)
```
asaj: Ditto

params/protocol_params.go
```diff
@@ -159,7 +159,6 @@ const (
 	AdditionalGasForNonGoldCurrencies uint64 = 3*ExpectedGasForCreditToTransactions + ExpectedGasForDebitFromTransactions + ExpectedGasToReadErc20Balance
```
asaj: Are these parameters used anymore?
`AdditionalGasForNonGoldCurrencies`, `ExpectedGasForDebitFromTransactions`, `ExpectedGasForCreditToTransactions`, `ExpectedGasToReadErc20Balance`?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -83,6 +127,49 @@ func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionI
...
+
+func GetGasForCreditToTransactions(header *types.Header, state vm.StateDB) uint64 {
+	return GetGasCost(header, state, params.ExpectedGasForCreditToTransactions, "gasForCreditToTransactions")
+}
+
+func GetGasToReadErc20Balance(header *types.Header, state vm.StateDB) uint64 {
+	return GetGasCost(header, state, params.ExpectedGasToReadErc20Balance, "gasToReadErc20Balance")
+}
+
+func GetIntrinsicGasForAlternateGasCurrency(header *types.Header, state vm.StateDB, gasCurrency *common.Address) uint64 {
```
asaj: Sorry to flip-flop, but I'm becoming more convinced that your original approach of storing just one `IntrinsicGasForAlternativeGasCurrency` approach is better, can we move back to that?

params/protocol_params.go
asaj: Can we increase the `MaxGasForDebitFromTransactions`, `MaxGasForCreditToTransactions, and `MaxGasToReadErc20Balance` all to 100k?

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -83,6 +127,49 @@ func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionI
...
+	if gasCurrency == nil {
```
asaj: Can we do this check inside `IntrinsicGas` instead? Otherwise, would recommend renaming this function to `GetIntrinsicGasForGasCurrency`

contract_comm/blockchain_parameters/blockchain_parameters.go
asaj: @kevjue what do you think?

params/protocol_params.go
mrsmkl: They are only used as defaults before the contract is deployed.

mrsmkl: They are only used as defaults before the contract is deployed.

asaj: In that case, would you mind deleting the above variables and adding this:

```
	// Default intrinsic gas cost of transactions paying for gas in alternative currencies.
        IntrinsicGasForAlternativeGasCurrency uint64 = 150000
```

And moving the `MaxGas*` variables down with the other `MaxGas*` variables?

PR #528, opened 2019-10-14
Add iOS build with bls

jeanregisser:
### Description

This PR adds iOS build support with bls.

`make ios` generates the static framework and bls lib that can be used on iOS (device and simulator).
The `CeloBlockchain` podspec is to be used by final iOS clients. It's directly compatible with Podfile using `use_frameworks!`. In that case it creates a dynamic framework linking against both geth and bls.

I had to disable bitcode for now since the rust generated lib isn't yet compatible with apple's bitcode. See https://github.com/rust-lang/rust/issues/35968 (I tried the latest rust nigthly but I couldn't make it work).

It also exposes the `IPCPath` config for the node so that iOS can use a relative path to workaround the 104 chars path limit for unix domain sockets.

Big thank to @cmcewen who did the initial working prototype!

### Tested

Tested that it builds and runs correctly on both iOS and Android.

### Other changes

- Ignore `ndk_bundle` dir used by the Makefile

### Related issues

- celo-org/celo-monorepo#1165

### Backwards compatibility

Yes.


Makefile
```diff
@@ -86,6 +87,16 @@ else
...
 
+bls-zexe-ios:
+ifeq ("$(RUSTUP_exists)","")
+	$(error "No rustup in PATH, consult https://github.com/celo-org/celo-monorepo/blob/master/SETUP.md")
+else ifeq ("$(CARGO_LIPO_exists)","")
+	$(error "No cargo lipo in PATH, run cargo install cargo-lipo")
+else
+	rustup target add aarch64-apple-ios armv7-apple-ios x86_64-apple-ios i386-apple-ios
+	cd vendor/github.com/celo-org/bls-zexe/bls && cargo lipo --release --targets=aarch64-apple-ios,armv7-apple-ios,x86_64-apple-ios,i386-apple-ios
+endif
```
kobigurk: Isn't there an endif missing here?

build/ci.go
```diff
@@ -942,6 +942,8 @@ func doXCodeFramework(cmdline []string) {
 
 	// Build the iOS XCode framework
 	build.MustRun(goTool("get", "golang.org/x/mobile/cmd/gomobile", "golang.org/x/mobile/cmd/gobind"))
+	// Patch gomobile to disable bitcode for now (rust generated bls lib output is not compatible)
+	build.MustRunCommand("sed", "-i", "", `/^[[:space:]]*cflags += \" -fembed-bitcode\"$/s/^/\/\//`, filepath.Join(build.GOPATH(), "src/golang.org/x/mobile/cmd/gomobile/env.go"))
```
kobigurk: Too bad it doesn't work :(
Can you try with Rust 1.37? I'm upgrading the build toolchain to that anyway. Seems like people succeeded with recent nightlies, so it might work there: https://github.com/rust-lang/rust/issues/35968#issuecomment-529618704

Makefile
```diff
@@ -86,6 +87,16 @@ else
 			 cd vendor/github.com/celo-org/bls-zexe/bls && cargo build --release --target=x86_64-linux-android --lib
 endif
 
+bls-zexe-ios:
+ifeq ("$(RUSTUP_exists)","")
+	$(error "No rustup in PATH, consult https://github.com/celo-org/celo-monorepo/blob/master/SETUP.md")
+else ifeq ("$(CARGO_LIPO_exists)","")
+	$(error "No cargo lipo in PATH, run cargo install cargo-lipo")
```
kobigurk: How about installing cargo-lipo for the user here?

Makefile
jeanregisser: Good question 👍  this syntax works but someone mentions it's failing with GNU Make 3.80:
https://stackoverflow.com/a/12297193/158525

My machine has GNU Make 3.81 which was first released in 2006.
I think we're safe, but let me know if you still wanna change it.

Makefile
jeanregisser: Good idea! I'll do that.

build/ci.go
jeanregisser: I tried with Rust 1.38 stable and 1.40.0-nightly (2748a9fd9 2019-10-08).
Both failed to link with gomobile when bitcode was enabled (Xcode 10.3). Though the generated rust lib do contain bitcode markers.

Makefile
jeanregisser: Ah I just changed the structure anyway so no questions about this anymore :D

asaj:
Do you need reviews from all five of us?

PR #531, opened 2019-10-15
[IBFT]Prevent validator from double-signing

### LINKED ISSUE #396: Validators SBAT recover consensus state on restart

ashishb:
### Description

 Validator before signing a message would check if it has already signed a message with same round and sequence number before. This prevents a malicious proposer from making a validator sign a different message after it has signed a message and rebooted.

### Tested

Verified that all the current tests and a deployment testnet work. So, this change does not break anything. I don't have a test to verify this new behavior.

### Related issues

- Fixes #396

### DIFF
--- a/consensus/istanbul/core/persistent_state.go
+++ b/consensus/istanbul/core/persistent_state.go
@@ -10,12 +10,26 @@ import (
 	"path/filepath"
 )
 
+func (c *core) savePreprepareMessageToDisk(
+	messageType uint64,
+	roundNumber *big.Int,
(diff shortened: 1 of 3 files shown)


mcortesi:
Closed, since we are using a different strategy now.

Issue #193, opened 2019-05-14
Validators SBAT persistently cache others validators' nodes


timmoreton:
### Expected Behavior

Depends on #191 

Store cache of other validators' enodes on disk so we can pre-populate cache on startup and quickly rejoin consensus protocol.

### Current Behavior

With #191, we would forget all details of other validators we'd seen if we restart.

PR #534, opened 2019-10-17
Cleanup val_enode_table

mcortesi:
### Description

Changes:
 - Move val_enode_table to an internal module to provide encapsulation
 - Test val_enode_table
 - Remove duplicated logic to obtain registered & active validators
 - Move logic to add/removeValidatorPeer outside val_enode_table,
 leaving that module only with responsability to persist enodes
 - Fix String() bad usage of fmt.Sprintf
 - Change validator contract comm to reflect actual contract return type
 (an array, not a map)

### Tested

unit tests

### Related issues

- Preparation work for #193


consensus/istanbul/backend/backend.go
```diff
@@ -220,20 +222,20 @@ func (sb *Backend) Gossip(valSet istanbul.ValidatorSet, payload []byte, msgCode
...
 func (sb *Backend) Enode() *enode.Node {
 	if sb.broadcaster != nil {
 		return sb.broadcaster.GetLocalNode()
-	} else {
-		return nil
 	}
+	return nil
 }
 
+// GetNodeKey gets the Node PrivateKey
```
asaj: nit: I would be careful here to make it clear that this returns the nodekey, i.e. the key used to encrypt messages in the p2p layer, and not, for example, the private key used for signing blcoks

consensus/istanbul/backend/internal/enodes/val_enode_table.go
```diff
@@ -0,0 +1,145 @@
...
+		return enode.enodeURL, ok
+	}
+	return "", false
+}
+
+// GetAddressFromEnodeURL will return the address for a enodeURL if it's kwown
+func (vet *ValidatorEnodeTable) GetAddressFromEnodeURL(enodeURL string) (common.Address, bool) {
+	vet.valEnodeTableMu.RLock()
+	defer vet.valEnodeTableMu.RUnlock()
+	address, ok := vet.reverseValEnodeTable[enodeURL]
```
asaj: Why the inconsistency between the way this and `GetEnodeURLFromAddress` handle not finding an entry?

consensus/istanbul/backend/internal/enodes/val_enode_table.go
```diff
@@ -0,0 +1,145 @@
...
+	return address, ok
+}
+
+// Upsert will update or insert a validator enode entry.  It will also do the associated set/remove to the validator connection.
```
asaj: This seems no longer to be true

consensus/istanbul/backend/internal/enodes/val_enode_table.go
```diff
@@ -0,0 +1,145 @@
...
+func (vet *ValidatorEnodeTable) Upsert(remoteAddress common.Address, enodeURL string, view *istanbul.View) (string, error) {
+	vet.valEnodeTableMu.Lock()
+	defer vet.valEnodeTableMu.Unlock()
+
+	var oldEnodeURL string
```
asaj: I think `old` here is a bit misleading, since this entry may actually be more recent than the one we're trying to upsert. Wdyt of `current` or `existing`?

consensus/istanbul/backend/internal/enodes/val_enode_table.go
```diff
@@ -0,0 +1,145 @@
...
+	if oldValEnode, ok := vet.valEnodeTable[remoteAddress]; ok {
+		// If it is an old message, ignore it.
+		if view.Cmp(oldValEnode.view) <= 0 {
+			return oldEnodeURL, errOldAnnounceMessage
```
asaj: Should this return `"", errOldAnnounceMessage`? IIUC the first return value of this function is supposed to be the value we replaced, but we don't replace anything if the announce message is old.

edit: I see now that this is essentially what we do. Would it be more readable to explicitly return the empty string?

PR #538, opened 2019-10-18
Include CommittedSeals from parent in current block header

gakonst:
- Adds ParentSeal to extras
- Listen for **previous** round Commits instead of rejecting them and add them to `core.current.ParentSeals`
- Verify seals from the previous sequence against the validator set from the previous sequence, to account for the epoch block edge cases
- On round change pass over the `ParentSeals` to the next round's `ParentSeals`
- On sequence change, pass over the `Commits` to the next sequence's `ParentSeals`
- Before making a Prepare, gather all ParentSeals and aggregate them

Currently uses a [branch of celo-monorepo](https://github.com/celo-org/celo-monorepo/tree/modify-extra-header) with the edited genesis block to account for the new `extraData` fields.

By leveraging this feature, we can:
1. Get a notion of network's actual uptime and avoid rewarding "lazy" validators, since we pipeline, and don't do tendermint style Precommit-Commit to know who participated in a roudn
2. (In the future) Reward validators for putting as many signatures as possible in their `ParentSeals` field


consensus/istanbul/core/core.go
```diff
@@ -345,10 +354,41 @@ func (c *core) waitForDesiredRound(r *big.Int) {
 
 func (c *core) updateRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, roundChange bool) {
 	// TODO(Joshua): Include desired round here.
-	if roundChange && c.current != nil {
-		c.current = newRoundState(view, validatorSet, nil, c.current.pendingRequest, c.current.preparedCertificate, c.backend.HasBadProposal)
+	if roundChange { // Can a roundChange occur with c.current being nil?
+		var pendingRequest *istanbul.Request
```
asaj: If we update `newRoundState()` to take `parentSeals` as an argument, could we just do:
```
if roundChange && c.current != nil {
		c.current = newRoundState(view, validatorSet, nil, c.current.pendingRequest, c.current.preparedCertificate, c.current.ParentSeals, c.backend.HasBadProposal)
```
?

consensus/istanbul/core/roundstate.go
```diff
@@ -29,19 +29,26 @@ import (
...
-		Prepares:            newMessageSet(validatorSet),
-		Commits:             newMessageSet(validatorSet),
+		round:        view.Round,
+		desiredRound: view.Round,
+		sequence:     view.Sequence,
+		Preprepare:   preprepare,
+		Prepares:     newMessageSet(validatorSet),
+		Commits:      newMessageSet(validatorSet),
+		// todo not necessarily this validator set, might be the parent ones
+		ParentSeals:         newMessageSet(validatorSet),
```
asaj: Should we call this `ParentCommits` for consistency? Technically the `seal` is included in the `Commit` message, and it looks like we're storing the whole message here

consensus/istanbul/core/handler.go
```diff
@@ -23,6 +23,14 @@ import (
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
 )
 
+func (c *core) ParentSeals() *messageSet {
```
asaj: Imagine that a node is the proposer one block after becoming a validator. They won't have seen any `Commit` messages for the previous block, since they were not a validator during that block. However, they *will* have seen a seal for the previous block representing 2F+1 validator signatures, and we should be sure to include those signatures in the `parentSeal` that we include in our proposal, otherwise the block will be invalid.

Essentially, we should always set the parent seal to be the union of "committed seal on previous block", "commits seen for previous block". Note that these can be different even in the case where we've been a validator for a while, since if someone else sees 2F+1 Commits before we do, they will broadcast the block with that committed seal, we will accept it, and move on to a new sequence.

consensus/istanbul/backend/engine.go
```diff
@@ -390,6 +390,17 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 	// wait for the timestamp of header, use this to adjust the block period
 	delay := time.Unix(header.Time.Int64(), 0).Sub(now())
 	time.Sleep(delay)
+
+	// modify the block header to include all the ParentSeals
+	parentCommitedSeals := sb.core.ParentSeals()
+	if parentCommitedSeals != nil {
```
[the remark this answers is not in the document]
gakonst: I do not think it should return an error, since for whatever reason maybe your connection dropped and you were not gossiped any ParentCommits. Maybe on error we could fetch the seals from our previous block and just use that (which is 2f+1, but better than 0)

consensus/istanbul/backend/engine.go
```diff
@@ -897,6 +922,35 @@ func writeSeal(h *types.Header, seal []byte) error {
 	return nil
 }
 
+// writeParentSeals writes the extra-data field of a block header with given committed seals.
+func writeParentSeals(h *types.Header, bitmap *big.Int, parentSeals []byte) error {
+	if len(parentSeals) == 0 {
+		return errInvalidCommittedSeals
+	}
+
+	if len(parentSeals) != types.IstanbulExtraCommittedSeal {
```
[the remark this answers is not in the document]
gakonst: Correct, it's the BLS aggregated signature.

consensus/istanbul/backend/engine.go
```diff
@@ -897,6 +922,35 @@ func writeSeal(h *types.Header, seal []byte) error {
...
+
+	istanbulExtra, err := types.ExtractIstanbulExtra(h)
+	if err != nil {
+		return err
+	}
+
+	istanbulExtra.ParentSeal = make([]byte, len(parentSeals))
+	copy(istanbulExtra.ParentSeal, parentSeals)
+	istanbulExtra.ParentBitmap = big.NewInt(0)
+	istanbulExtra.Bitmap.Set(bitmap)
```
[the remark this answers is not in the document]
gakonst: Nice catch, thanks

consensus/istanbul/backend/engine.go
```diff
@@ -860,12 +871,26 @@ func assembleExtra(header *types.Header, oldValSet []istanbul.ValidatorData, new
 			"addedValidators", common.ConvertToStringSlice(addedValidatorsAddresses), "removedValidators", removedValidators.Text(16))
 	}
 
+	var bitmap *big.Int
+	var parentSeal []byte
+	extras, err := types.ExtractIstanbulExtra(header)
+	if err != nil {
```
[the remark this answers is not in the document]
gakonst: Same response: https://github.com/celo-org/celo-blockchain/pull/538#discussion_r338158747

consensus/istanbul/core/backlog.go
```diff
@@ -54,7 +57,17 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
...
 	}
 
+	// as soon as we move on to the next sequence, we discard any messages from
+	// a previous sequence as invalid old messages if they are not Commits
 	if view.Cmp(c.currentView()) < 0 {
+		// let commits from previous views through so that the handler adds them
+		// to the parent view's commits.
+		// TODO: Should we let old rounds through as well? The view.Cmp function
+		// returns <0 if the sequences are equal and uses the round as a tiebreaker.
+		parentSequencePlusOne := big.NewInt(0).Add(view.Sequence, common.Big1)
```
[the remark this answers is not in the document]
gakonst: This is `view.Sequence + 1` (not `c.currentView().Sequence + 1`

If `view.Sequence + 1 == c.currentView().Sequence` then it means that this particular`view` corresponds to the parent block

consensus/istanbul/core/commit.go
```diff
@@ -84,38 +85,67 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
...
+		3. check that the block hash that i signed last round, is the same as the
+		one i received
+	*/
+
+	parentSequencePlusOne := big.NewInt(0).Add(commit.View.Sequence, common.Big1)
+	if msg.Code == istanbul.MsgCommit && c.currentView().Sequence.Cmp(parentSequencePlusOne) == 0 {
+		// Retrieve the validator set for the previous proposal (which should
+		// match the one broadcast
+		lastProposal, _ := c.backend.LastProposal()
+		parentValset := c.backend.Validators(lastProposal)
```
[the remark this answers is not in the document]
gakonst: Using [`ParentValidators`](https://github.com/celo-org/celo-blockchain/blob/2ba24548d31c25a81f736b920ae3cc4b7275b3cf/consensus/istanbul/core/commit.go#L104) which calls snapshot under the hood

consensus/istanbul/core/commit.go
```diff
@@ -84,38 +85,67 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
...
+		_, validator := parentValset.GetByAddress(msg.Address)
+		if validator == nil {
+			return errInvalidValidatorAddress
+		}
+		if err := c.verifyCommittedSeal(commit.Digest, msg.CommittedSeal, validator); err != nil {
+			return errInvalidCommittedSeal
+		}
+		// TODO: What further verification checks are needed? Maybe ensure that
```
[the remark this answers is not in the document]
gakonst: How would you get the previous block that's stored? Getting the proposal as follows fails currently:

```go
// Ensure that the commit's digest (ie the received proposal's hash)
		// matches the saved last proposal's hash
		lastProposal, _ := c.backend.LastProposal()
		if commit.Digest != lastProposal.Hash() {
			return errors.New(fmt.Sprintf("parent block does not match. Expected %v. Got %v", commit.Digest.String(), lastProposal.Hash().String()))
		}
```

consensus/istanbul/core/handler.go
gakonst: I agree with this approach:
> we should always set the parent seal to be the union of "committed seal on previous block", "commits seen for previous block"

The problem is the aggregation. "committed seal on previous block" is a BLS aggregated signature. The commits seen for previous block are signatures which we have not aggregated yet. In order to get the union of the seals in this block and the previous block's we have to get the seals themselves (and not their aggregate version), which I believe the node does not store currently.

gakonst:
```
ok  	github.com/ethereum/go-ethereum/eth	11.619s
--- FAIL: TestFakedSyncProgress62 (0.81s)
    downloader_test.go:1349: final progress mismatch:
        have {StartingBlock:0 CurrentBlock:1004 HighestBlock:1003 PulledStates:0 KnownStates:0}
        want {StartingBlock:0 CurrentBlock:1003 HighestBlock:1003 PulledStates:0 KnownStates:0}
FAIL
```

This is a flaky test for some reason, doesn't get reproduced locally. Either way, this is good for review again @asaj

consensus/istanbul/core/roundstate.go
```diff
@@ -27,21 +27,31 @@ import (
 )
 
 // newRoundState creates a new roundState instance with the given view and validatorSet
-func newRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, preprepare *istanbul.Preprepare, pendingRequest *istanbul.Request, preparedCertificate istanbul.PreparedCertificate, hasBadProposal func(hash common.Hash) bool) *roundState {
+func newRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, preprepare *istanbul.Preprepare, pendingRequest *istanbul.Request, preparedCertificate istanbul.PreparedCertificate, parentCommits *messageSet, hasBadProposal func(hash common.Hash) bool) *roundState {
+	// in case we passed empty parent commits, initialize them
+	if parentCommits == nil {
+		parentCommits = newMessageSet(validatorSet)
```
kevjue: Should the `validatorSet` parameter be the one from the last block?  This one seems to be the current block's one.

consensus/istanbul/core/types.go
```diff
@@ -27,6 +27,8 @@ type Engine interface {
 	Stop() error
 	CurrentView() *istanbul.View
 	SetAddress(common.Address)
+	// Validator -> CommittedSeal from Parent Block
+	ParentCommits() *messageSet
```
kevjue: Can this function be removed from the interface?   The functions listed here are normally for packages that are outside of the top level `consensus` package.

consensus/istanbul/core/backlog.go
```diff
@@ -54,7 +57,19 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 		return errFutureMessage
 	}
 
+	// as soon as we move on to the next sequence, we discard any messages from
+	// a previous sequence as invalid old messages if they are not Commits
 	if view.Cmp(c.currentView()) < 0 {
+		// let commits from previous views through so that the handler adds them
+		// to the view's parent commits.
+		// TODO: Should we let old rounds through as well? The view.Cmp function
+		// returns <0 if the sequences are equal and uses the round as a tiebreaker.
```
kevjue: If this node will want to maximize rewards from parent seals, then I think it should keep old rounds as well.

Once a block has been "locked" by the consensus algorithm, then all of the subsequent proposals from later rounds of the same sequence should be the same (which means all those signatures should be valid and hence "reward-able").

consensus/istanbul/core/roundstate.go
```diff
@@ -27,21 +27,31 @@ import (
...
 		Commits:             newMessageSet(validatorSet),
+		ParentCommits:       parentCommits,
 		mu:                  new(sync.RWMutex),
 		pendingRequest:      pendingRequest,
 		preparedCertificate: preparedCertificate,
 		hasBadProposal:      hasBadProposal,
 	}
 }
 
+// sequence -> sequence: commit -> parentcommit
```
kevjue: nit:  Please make this comment more clear.

consensus/istanbul/core/handler.go
kevjue: > The problem is the aggregation. "committed seal on previous block" is a BLS aggregated signature. The commits seen for previous block are signatures which we have not aggregated yet. In order to get the union of the seals in this block and the previous block's we have to get the seals themselves (and not their aggregate version), which I believe the node does not store currently (since they are stored in [`c.current.Commits`](https://github.com/celo-org/celo-blockchain/blob/0c903dd7f906bf5f4d649c4737e9b517012c8d55/consensus/istanbul/core/core.go#L178) which is not persisted)

I had a quick chat with @mstraka100 about this, and he says that it is possible to "incrementally" aggregate an aggregated BLS signature.  The operation to do so should be the same as the one that was used to generate the original aggregation.

consensus/istanbul/backend/engine.go
```diff
@@ -860,12 +880,26 @@ func assembleExtra(header *types.Header, oldValSet []istanbul.ValidatorData, new
 			"addedValidators", common.ConvertToStringSlice(addedValidatorsAddresses), "removedValidators", removedValidators.Text(16))
 	}
 
+	var bitmap *big.Int
```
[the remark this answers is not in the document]
gakonst: Thanks, this should be fixed in https://github.com/celo-org/celo-blockchain/pull/538/commits/11f86bf41e634ffdee39dd68f330eb2663da1444, which also addresses @asaj's comment here: https://github.com/celo-org/celo-blockchain/pull/538#discussion_r337208903

consensus/istanbul/core/types.go
gakonst: The goal here is to access the data inside `<core>.current.ParentCommits` of `core` from `backend.go`. [`core` is a field of `Backend`](https://github.com/celo-org/celo-blockchain/blob/eb9ad80d99ddc43e48a3d133e359e4c98385ce3c/consensus/istanbul/backend/backend.go#L99) which implements the Engine interface. 

I did not find any mentions of `CurrentView` and `SetAddress` outside of `consensus/istanbul/core` (ie outside of the top level package), so it made sense to expose the desired data over the interface and add an extra function in the implementation of the interface under the `core/handler.go` file (where the rest of the interface is also implemented).

In order to access the core as a `core` type, we'd have to export the core (as `Core`) in order to do something like `sb.core.(*Core)`, and then get the data, but it doesn't make much sense to me to do that.

Does that convince you to go with the interface approach as I already have, or do you prefer the exporting option? Do you have any alternative suggestions?

consensus/istanbul/core/roundstate.go
gakonst: Yeah so there should be no need to initialize this with the current valset, removed this step

consensus/istanbul/core/backlog.go
gakonst: Actually this TODO is leftover from a previous iteration, the current functionality lets the previous sequence through, independently of its round (which indeed as you said: once a block has been locked, all later rounds are the same)

consensus/istanbul/core/backlog.go
```diff
@@ -54,7 +57,19 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 		return errFutureMessage
 	}
 
+	// as soon as we move on to the next sequence, we discard any messages from
+	// a past sequence as invalid old messages unless they are Commits of the
+	// previous sequence
 	if view.Cmp(c.currentView()) < 0 {
+		// let commits from previous sequencse through so that the handler adds them
```
asaj: nit: "sequence"

consensus/istanbul/core/backlog.go
```diff
@@ -54,7 +57,19 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
...
+		// to the view's parent commits (the round does not matter in this case,
+		// since all rounds after a block is "locked" should be the same)
+		viewSequencePlusOne := big.NewInt(0).Add(view.Sequence, common.Big1)
+		// if the received view's sequence +1 equals the current view, then the
+		// received view corresponsd to the parent block
```
asaj: nit: "corresponds"

consensus/istanbul/core/commit.go
```diff
@@ -84,38 +86,62 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
...
-		return err
-	}
-
-	_, validator := c.valSet.GetByAddress(msg.Address)
-	if validator == nil {
-		return errInvalidValidatorAddress
-	}
+	msgSequencePlusOne := big.NewInt(0).Add(commit.View.Sequence, common.Big1)
+	// if the received view's sequence +1 equals the current view, then the
+	// received view corresponsd to the parent block
```
asaj: corresponds

consensus/istanbul/core/commit.go
```diff
@@ -84,38 +86,62 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
...
+	if c.currentView().Sequence.Cmp(msgSequencePlusOne) == 0 {
+		// Retrieve the validator set for the previous proposal (which should
+		// match the one broadcast
```
asaj: Close parentheses

consensus/istanbul/core/commit_test.go
```diff
@@ -190,6 +231,15 @@ OUTER:
 			}
 		}
 
+		// core should have received a parent seal from each of its neighbours
+		// how can we add our signature to the ParentCommit? Broadcast to ourselve
```
asaj: Good question, this is a tricky one. I *think* we don't need to do anything here, since if we were online and participating in consensus, we would have included our own commit in whatever block we wind up getting. So if, when proposing, we use the union of the commits we saw from others, and the commits we have in the previous block, we should get included by way of the latter.


@kevjue, is that your understanding as well?

core/types/istanbul.go
```diff
@@ -50,10 +51,16 @@ type IstanbulExtra struct {
...
+	// CommittedSeal is an aggregated BLS signature resulting from signatures by
+	// each validator that signed this block
 	CommittedSeal []byte
 	// EpochData is a SNARK-friendly encoding of the validator set diff (WIP)
 	EpochData []byte
+	// ParentBitmap is a bitmap having an active bit for each validator that
+	// signed on the previous block
+	ParentBitmap *big.Int
+	// ParentCommit is the aggregated BLS signature resulting from commits of the validators which signed on the previous block
+	ParentCommit []byte
```
asaj: Wdyt of `ParentSeal` here? Would be more consistent with `CommittedSeal`

consensus/istanbul/backend/engine.go
```diff
@@ -390,6 +390,26 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
+	// modify the block header to include all the ParentCommits
+	parentCommittedSeals := sb.core.ParentCommits()
+	if parentCommittedSeals != nil {
+		if parentCommittedSeals.Size() != 0 {
+			sb.logger.Info("Storing", "parent seals from", parentCommittedSeals.String())
+			bitmap, asig := istanbulCore.AggregateSeals(parentCommittedSeals)
+			writeCommittedSeals(header, bitmap, asig, true)
+		}
+	} else {
+		// if we were not gossipped any commits, then just copy over the seals
```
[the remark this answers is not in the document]
asaj: We should also be wary of the case when the parent commit messages are for a different block than what winds up in our chain. In *theory* I don't think this should happen (right @kevjue?) but I'd like to check the digests just in case

consensus/istanbul/backend/engine.go
```diff
@@ -860,20 +866,30 @@ func assembleExtra(header *types.Header, oldValSet []istanbul.ValidatorData, new
...
 	}
 
-	ist := &types.IstanbulExtra{
-		AddedValidators:           addedValidatorsAddresses,
-		AddedValidatorsPublicKeys: addedValidatorsPublicKeys,
-		RemovedValidators:         removedValidators,
-		Seal:                      []byte{},
-		CommittedSeal:             []byte{},
+	extras, err := types.ExtractIstanbulExtra(header)
+	if err != nil {
```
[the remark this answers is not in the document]
gakonst: I have actually realized that this is the first time the header's `Extra` field is populated (I've confirmed the fact that the `else` branch never gets reached). 

As a result, it does not make sense to even try to extract any existing extras. 
This means that the name `updateValidatorSetDiff` does not make much sense, and it's better to rename it back to `assembleExtra` and maintain the previous functionality.

consensus/istanbul/core/commit_test.go
gakonst: I agree that the union approach should fix this.

consensus/istanbul/backend/engine.go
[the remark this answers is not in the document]
kevjue: Ya, that shouldn't happen.  All blocks that are inserted in the chain have their header verified: https://github.com/celo-org/celo-blockchain/blob/master/core/blockchain.go#L1137

consensus/istanbul/core/commit_test.go
kevjue: I also agree with the approach.

consensus/istanbul/core/types.go
kevjue: You approach makes sense to me.

consensus/istanbul/backend/engine.go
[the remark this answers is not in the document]
gakonst: `writeCommittedSeals` has been amended to also initialize the header's extra if it could not extract any extras from the existing header. 

This is done because it the first time the header's extras are accessed is in the `Prepare` function when we're writing the parent commits. 

Both functions are now consistent with each other

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           command: |
             set -euo pipefail
             export CELO_MONOREPO_DIR="$PWD"
-            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b master
+            git clone --depth 1 https://github.com/celo-org/celo-monorepo.git ${CELO_MONOREPO_DIR} -b modify-extra-header
```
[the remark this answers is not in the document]
gakonst: Yeah once we get this approved, I'm opening the PR for the above change

core/types/istanbul.go
gakonst: Had originally called it "ParentCommits" because of: https://github.com/celo-org/celo-blockchain/pull/538#discussion_r337616746. Either name works for me.

gakonst:
There seems to be some edge case with ultralight mode in the e2e sync & transfer tests.

consensus/istanbul/backend/engine.go
```diff
@@ -317,29 +315,66 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
-	// Check whether the committed seals are generated by parent's validators
+	err = sb.checkValidatorSignatures(header.Hash(), validators, extra.Bitmap, extra.CommittedSeal)
+	if err != nil {
+		return err
+	}
 
-	proposalSeal := istanbulCore.PrepareCommittedSeal(header.Hash())
-	// 1. Get committed seals from current header
-	myValidatorIndex, myValidator := validators.GetByAddress(sb.Address())
+	if len(parents) > 1 {
```
asaj: Why do we only check the parent signatures if `len(parents) > 1`? Will there be cases where `parents` is empty? IIRC, there's no guarantee that parents will actually be populated

consensus/istanbul/backend/engine.go
```diff
@@ -390,6 +425,27 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
+			return err
+		}
+		bitmap := parentExtra.Bitmap
+		asig := parentExtra.CommittedSeal
+
+		parentGossipedSeals := sb.core.ParentCommits()
+		// if we had any seals gossiped to us, proceed to add them to the
+		// already aggregated signature
+		if parentGossipedSeals != nil && parentGossipedSeals.Size() != 0 {
+			bitmap, asig = istanbulCore.UnionOfSeals(bitmap, asig, parentGossipedSeals)
```
asaj: Just to be clear, @kevjue we believe that there is no risk that `parentCommits` will have messages for a different block than what we wound up finalizing?

Is it worth checking the hash that was signed before we compute the union or do we not have to worry about this

consensus/istanbul/backend/engine.go
gakonst: if the length of parents == 0, then we're at the genesis block, and there's no parent
if the length is 1, then the parent is going to be the genesis block which validators never signed on
if the length is 2, then the parent is going to be the block #1, which was a block generated normally via the consensus process, which will have validator sigs, so only then does it start to make sense to check the signatures.

I could make this check `if block.Number.Uint64() > 1` if you'd like

kevjue: I strongly agree with @asaj 's concern as well.  The `VerifiyHeader` function in `engine.go` will ultimately pass a `nil` value for parents.

consensus/istanbul/backend/engine.go
asaj: After talking this over with @kevjue, it does seem like there is an issue here.

Imagine I'm the block proposer, and I send you block A and everyone else block B. I then send you a COMMIT for block A. Everyone else agrees on B, and it eventually gets broadcast to you with 2F+1 signatures. You accept block B, and move on to a new sequence, copying the COMMIT I sent you for block A into your `ParentCommits`.

If you are the next block proposer, you will wind up proposing a block with parent signatures from both block A and block B, which will be invalid. 

To fix this, it seems that we have two options:
- Before copying `Commits` over to `ParentCommits`, check that the most recent block's hash matches those in the `Commit` messages.
- Before doing the union here, check that the block hashes match in the same manner.

I think we have a slight preference for option 1, all things being equal.

gakonst: This fix should be address this @asaj: https://github.com/celo-org/celo-blockchain/pull/538/commits/32ed4438e77603500a1b04c4c5654000c265fcc8

gakonst:
https://circleci.com/gh/celo-org/celo-blockchain/13810 fails on CI but passed locally. Unclear why it's flaky.

consensus/istanbul/backend/engine.go
gakonst: This should be addressed in https://github.com/celo-org/celo-blockchain/pull/538/commits/b8627f90bc7114d21088b95729215c122b368849

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+	// The genesis block is skipped since it has no parents
+	// The first block is also skipped since its parent
+	// is the genesis block which contains no parent signatures
+	if number > 1 {
+		sb.logger.Debug("verifyCommittedSeals: verifying parent block for block", "num", number)
+		var parentValidators istanbul.ValidatorSet
+		// parents of epoch blocks have different validator sets since they
+		// belong to the previous epoch.
+		if number%sb.config.Epoch == 0 {
+			sb.logger.Debug("verifyCommittedSeals: it's an epoch block!")
```
asaj: Remove

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+			var parent *types.Header
+			var cache []*types.Header
+			if len(parents) > 1 {
+				cache = parents[:len(parents)-1]
+				parent = parents[len(parents)-1]
+				sb.logger.Debug("verifyCommittedSeals: was given a cache", "cache-len", len(cache), "parent-num-from-cache", parent.Number.Uint64())
+			} else {
+				cache = parents
+				parent = chain.GetHeaderByNumber(number - sb.config.Epoch)
```
asaj: I think we can probably get away with not fetching the parent anywhere in this function and just using the block number for `sb.snapshot`. Would clean a lot of this code up

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+	}
+
+	// Check parent signatures only after block 1.
+	// The genesis block is skipped since it has no parents
+	// The first block is also skipped since its parent
+	// is the genesis block which contains no parent signatures
+	if number > 1 {
+		sb.logger.Debug("verifyCommittedSeals: verifying parent block for block", "num", number)
+		var parentValidators istanbul.ValidatorSet
+		// parents of epoch blocks have different validator sets since they
```
asaj: ```The first block in an epoch will have a different validator set than the block before it. If the current block is the first block in an epoch, we need to fetch the previous validator set to validate the parent signatures.```

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+		// belong to the previous epoch.
+		if number%sb.config.Epoch == 0 {
```
asaj: We actually want to check if this is 1. `number%sb.config.Epoch == 0` means `number` is the last block of the epoch.

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
 		return errEmptyCommittedSeals
 	}
 
+	// Check the signatures on the current header
+	snap, err := sb.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, parents)
```
asaj: You probably noticed this, but just want to highlight here that we call snapshot on the previous block here, because snapshot(blockN) gives us the validator set that needs to sign blockN+1

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+					return nil
+				}
+				sb.logger.Debug("verifyCommittedSeals: no cache provided", "cache-len", len(cache), "parent-num", parent.Number.Uint64())
+			}
 
-	proposalSeal := istanbulCore.PrepareCommittedSeal(header.Hash())
-	// 1. Get committed seals from current header
-	myValidatorIndex, myValidator := validators.GetByAddress(sb.Address())
+			// Get the validator set from the parent block's snapshot.
+			// If in ultralight mode, the parent block is going to be the
```
asaj: Actually, the way `snapshot` is written, it can handle arbitrary block numbers, even in ultralight mode. It does the logic of fetching only epoch blocks under the hood. So we should always just call snapshot here on the block before the parent block

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+			// previous epoch block, while in other modes it will be the n-1'th
+			// block. Both cases are handled in the same way.
+			snap, err := sb.snapshot(chain, parent.Number.Uint64(), parent.Hash(), cache)
```
asaj: Just highlighting that snapshot returns the validator set that is needed to sign off on the block *following* the block that is passed to it. I know it's a confusing API. So I think here we would want `parent.Number.Uint64()`, or better yet, `number - 2`

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
+			if err != nil {
+				return err
+			}
+			parentValidators = snap.ValSet.Copy()
+		} else {
+			sb.logger.Debug("verifyCommittedSeals: not an epoch block!")
+			// if the current block is not an epoch block, then we can get the previous validator set from the previous block
+			// note(gakonst): I thought that just using the same validator set as the current block would be sufficient, but for some reason it did not work.
```
asaj: Your intuition is correct, we should be able to use the same validator set.

That this didn't work suggests an off by one error to me. I think (hope) this should be solved by checking `if number%sb.config.Epoch == 1` instead.

consensus/istanbul/backend/engine.go
```diff
@@ -272,29 +270,103 @@ func (sb *Backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
 			pubKey := validators.GetByIndex(uint64(i)).BLSPublicKey()
-			if myValidatorIndex >= 0 && bytes.Equal(pubKey, myValidator.BLSPublicKey()) {
-				sb.logger.Debug("Our backend participated in consensus", "number", number)
-			}
 			publicKeys = append(publicKeys, pubKey)
 		}
 	}
+	sb.logger.Debug("checkValidatorSignatures: args", "bitmap", bitmap.Bits(), "seal", common.ToHex(seal), "proposal", common.ToHex(proposalSeal), "pubkeys", publicKeys, "validators", validators.List())
 
 	// The length of validSeal should be larger than number of faulty node + 1
```
asaj: This comment is incorrect
```The length of a valid seal should be greater than the minimum quorum size```

File: consensus/istanbul/backend/engine.go [hunk 36]
```diff
@@ -786,20 +877,34 @@ func assembleExtra(header *types.Header, oldValSet []istanbul.ValidatorData, new
 			"addedValidators", common.ConvertToStringSlice(addedValidatorsAddresses), "removedValidators", removedValidators.Text(16))
 	}
 
-	ist := &types.IstanbulExtra{
+	// if the extras could not be calculated, initialize a new struct with
```
asaj: Should we be trying to `ExtractIstanbulExtra` before this we create this struct, much like we do for `writeCommittedSeals`?

consensus/istanbul/core/backlog_test.go
```diff
@@ -51,9 +51,46 @@ func TestCheckMessage(t *testing.T) {
 	testStates := []State{StateAcceptRequest, StatePreprepared, StatePrepared, StateCommitted, StateWaitingForNewRound}
 	testCode := []uint64{istanbul.MsgPreprepare, istanbul.MsgPrepare, istanbul.MsgCommit, istanbul.MsgRoundChange}
 
-	// future sequence
+	// accept Commits from previous sequence
```
asaj: How hard would it be to test that we reject commits from the previous sequence if their digest doesn't match the latest block in our chain?

consensus/istanbul/core/core.go
```diff
@@ -344,10 +381,46 @@ func (c *core) waitForDesiredRound(r *big.Int) {
 
 func (c *core) updateRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, roundChange bool) {
 	// TODO(Joshua): Include desired round here.
-	if roundChange && c.current != nil {
-		c.current = newRoundState(view, validatorSet, nil, c.current.pendingRequest, c.current.preparedCertificate, c.backend.HasBadProposal)
+	if c.current != nil {
+		var commits *messageSet
```
asaj: Can we call this `parentCommits`?

consensus/istanbul/core/core.go
```diff
@@ -344,10 +381,46 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
+				if lastProposal.Hash() == preprepareProposal.Hash() {
+					c.logger.Debug("hashes-matched", "num", preprepareProposal.Number(), "hash", preprepareProposal.Hash().String())
+					// if it was not a round change (ie. a sequence change) and the
+					// preprepare proposal hash matches the chain head,
+					// we use this sequence's commits as the ParentCommits field in the next round
+					commits = c.current.Commits
+				} else {
+					c.logger.Debug("preprepare", "num", preprepareProposal.Number(), "hash", preprepareProposal.Hash().String())
+					c.logger.Debug("lastproposal", "num", lastProposal.Number(), "hash", lastProposal.Hash().String())
+					// if the hashes did not match, this means there was a partition and
```
asaj: 👍 👍 👍

consensus/istanbul/core/core.go
```diff
@@ -344,10 +381,46 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
+					// the lastProposal is different from the one we were listening to.
+					// as a result, we discard ths current round's commits, and initialize an empty
+					// ParentCommits message set, which we use to listen for messages which match the
+					// lastProposal as a parent.
+					commits = newMessageSet(c.backend.Validators(lastProposal))
+				}
+			} else {
+				c.logger.Debug("nil-preprepare", "num", lastProposal.Number(), "hash", lastProposal.Hash().String())
+				// if the preprepare was nil, then just initialize empty parent commits with the current valset
+				commits = newMessageSet(validatorSet)
```
asaj: Should we use the previous validator set here instead?

consensus/istanbul/core/core.go
```diff
@@ -344,10 +381,46 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
+					commits = c.current.Commits
+				} else {
+					c.logger.Debug("preprepare", "num", preprepareProposal.Number(), "hash", preprepareProposal.Hash().String())
+					c.logger.Debug("lastproposal", "num", lastProposal.Number(), "hash", lastProposal.Hash().String())
+					// if the hashes did not match, this means there was a partition and
+					// the lastProposal is different from the one we were listening to.
+					// as a result, we discard ths current round's commits, and initialize an empty
+					// ParentCommits message set, which we use to listen for messages which match the
+					// lastProposal as a parent.
+					commits = newMessageSet(c.backend.Validators(lastProposal))
```
asaj: Is there an off-by-one error here? i.e. does `Validators` return the validator set that signs off on `lastProposal` or `lastProposal+1`. This is one of the most unnecessarily confusing parts of this code base.

consensus/istanbul/backend/engine.go
gakonst: Yeah that's why I write the `-1` and `ParentHash`, IMO the whole codebase needs to be refactored to not require that, it's very confusing

consensus/istanbul/core/backlog_test.go
gakonst: We don't exactly reject the "commits", we're just initializing new roundstates with an empty parent commits field. 

There are no tests for `updateRoundState` available for this functionality currently. I can write some unit tests for it, but would require mocking everything (ie a mock `core`/`core.current`/`core.backend`) such that all execution paths in the `updateRoundState` function are explored.

What's hard is figuring out when this will actually happen with a real (ie non-mocked) chain, which I do not have an estimate for.

consensus/istanbul/backend/engine.go
gakonst: I think I like it as is, do you mind me not changing it? Don't think the attempt to create consistency between the two is that valuable here

consensus/istanbul/core/backlog.go
```diff
@@ -54,7 +57,19 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 		return errFutureMessage
 	}
 
+	// as soon as we move on to the next sequence, we discard any messages from
+	// a past sequence as invalid old messages unless they are Commits of the
+	// previous sequence
 	if view.Cmp(c.currentView()) < 0 {
+		// let commits from previous sequence through so that the handler adds them
+		// to the view's parent commits (the round does not matter in this case,
```
asaj: nit: I would change this comment slightly to:

`Let commits from the previous sequence through so that the handler adds them to the view's parent commits. The round does not matter in this case, as we only care that the message's committed seal matches the previous block.`

consensus/istanbul/core/commit.go
```diff
@@ -84,38 +86,62 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
...
-	if err := c.verifyCommit(commit); err != nil {
-		return err
-	}
-
-	_, validator := c.valSet.GetByAddress(msg.Address)
-	if validator == nil {
-		return errInvalidValidatorAddress
-	}
+	msgSequencePlusOne := big.NewInt(0).Add(commit.View.Sequence, common.Big1)
+	// if the received view's sequence +1 equals the current view, then the
```
asaj: nit: `equals the current sequence`

consensus/istanbul/core/commit.go
```diff
@@ -84,38 +86,62 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
...
+	// received view corresponds to the parent block
+	if c.currentView().Sequence.Cmp(msgSequencePlusOne) == 0 {
+		// Retrieve the validator set for the previous proposal (which should
+		// match the one broadcast)
+		parentValset := c.backend.ParentValidators(c.current.Proposal())
```
asaj: Is this an off-by-one error? IIUC, `ParentValidators` will give us the validator set that needs to sign off on the block we pass in, in this case, `c.current.Proposal()` which is the block we're currently trying to finalize. 

I think instead we want to pass `c.backend.LastProposal()`, but please double check me here

consensus/istanbul/backend/engine.go
```diff
@@ -345,6 +381,40 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
+		if parentGossipedSeals != nil && parentGossipedSeals.Size() != 0 {
+			sb.logger.Trace("backend.Prepare: taking union of parent seals", "valset", parentGossipedSeals.String(), "num", number)
+			// if we had any seals gossiped to us, proceed to add them to the
+			// already aggregated signature
+			sb.logger.Trace("backend.Prepare: parent seal before", "bitmap", bitmap.Bits(), "sig", common.ToHex(asig))
+			newBitmap, newAsig := istanbulCore.UnionOfSeals(bitmap, asig, parentGossipedSeals)
+			parentValidators := sb.getValidators(parent.Number.Uint64(), parent.Hash())
+			// only update to use the union if we indeed provided a valid aggregate signature for this block
+			if err := sb.checkValidatorSignatures(parent.Hash(), parentValidators, newBitmap, newAsig); err != nil {
+				sb.logger.Trace("backend.Prepare: tried to create invalid aggregate signature. not updating to union")
```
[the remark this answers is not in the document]
gakonst: So this was occurring _some_ times. Fixing the 2 off-by-1 cases you mentioned has eliminated any occurrences of this in the logs of all E2E tests, so I'll make it an Error, and use it as a signal that your suggestions on off-by-1's were correct

consensus/istanbul/backend/engine.go
```diff
@@ -345,6 +381,40 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
+		asig := parentExtra.CommittedSeal
+
+		parentGossipedSeals := sb.core.ParentCommits()
+		if parentGossipedSeals != nil && parentGossipedSeals.Size() != 0 {
+			sb.logger.Trace("backend.Prepare: taking union of parent seals", "valset", parentGossipedSeals.String(), "num", number)
+			// if we had any seals gossiped to us, proceed to add them to the
+			// already aggregated signature
+			sb.logger.Trace("backend.Prepare: parent seal before", "bitmap", bitmap.Bits(), "sig", common.ToHex(asig))
+			newBitmap, newAsig := istanbulCore.UnionOfSeals(bitmap, asig, parentGossipedSeals)
+			parentValidators := sb.getValidators(parent.Number.Uint64(), parent.Hash())
```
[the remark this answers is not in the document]
gakonst: Nice catches again.

consensus/istanbul/core/commit.go
gakonst: Yeah, this and the other off-by-one eliminated the errors in the backend which were appearing some times

asaj: For posterity, we agreed that it was

consensus/istanbul/backend/engine.go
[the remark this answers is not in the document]
asaj: For posterity, we agreed that it should be

Issue #465, opened 2019-09-14
Users SBAT see validator info in ethstats


timmoreton:
- [x] Add "Elected" as extra column to right of "Is mining"/"Validator" and populate with whether that validator is elected
- [x] Rename "Last Blocks Miner" to "Last Proposer" (field at top) -- and fix it -- it doesn't seem to work right now
- [ ] Read 'name' from Validator and group address (or even 'name' from Group)

diminator:
"Last Blocks Miner" has been changed to "Recent Validators"

Issue #462, opened 2019-09-14
Users SBAT see num elected validators on ethstats


timmoreton:
In one of the info panels at the top (ideally with 'node count' beneath it).

diminator:
what do you mean with "node count"
- all full nodes (validator + tx nodes)
- only all registered validators
?

diminator:
closed by #543

Issue #460, opened 2019-09-14
Ethstats SBAT list elected validators that are not sending data


timmoreton:
Any elected validator should be shown in the list, and grayed out.

diminator:
closed by #543

Issue #459, opened 2019-09-14
Ethstats operator SBAT block stats submission on a per-node basis


timmoreton:
* TLS proxy
* Token based auth
* Figure out best approach to integrate this with 'elected validators' in #458 - maybe they get a single global token.

diminator:
With the changes made for #458, operators have a "trusted" list of nodes that they accept upon authorization. Stats are also signed for non-repudiation. However, there's no handshake protocol, so nodes could sniff and re-submit packages of trusted nodes. This would allow them to occupy a connection, but they won't be able to change the signed content

There's a "blocked" list in the ethstats operator, that can hold IP addresses.

PR #543, opened 2019-10-22
Ethstats (client) sends signed stats to operators along with validator information

### LINKED ISSUE #458: Ethstats operators SBAT receive stats from all elected validators
* Right now, there is a single global token allowing anyone to publish. (So only CeloLabs validators appear on the list).
* We should use the list of validators in groups which have one or more validator elected as the basis for whose signature to authenticate for stats -- this would allow any validator to publish in a permissionless way.  
* Also support blacklisting of addresses to prevent malicious validators spamming ethstats instances

diminator:
### Description

(see https://github.com/celo-org/ethstats-server/pull/6 for the operator part of this PR)

Stats-operators share a secret with nodes that want to publish their stats on a websocket provided by the operator. This makes it hard to publish stats dynamically.

Nodes now sign their stats before publishing to an operator, proving that _a_ node with `publicKey` signed the hash of the stats object.

```json
{
    "stats": "<som stats object (like before)>",
    "proof": {
        "address": "<etherbase address>",
        "publicKey": "<etherbase public key in hex encoding>",
        "msgHash": "<keccak hash of stats and the message to sign in hex encoding>",
        "signature": "<ecdsa signature in hex encoding>"
    }
}
```

Operators verify the signed stats published by the nodes and ensure `publicKey` is a node they trust. The operator has a list of trusted (and banned) nodes initialized with a [config file](https://github.com/diminator/ethstats-server/blob/diminator/feature-458-ethstats-ops-should-receive-from-validators/lib/utils/config.js).

The list of active validators is also sent (`address`, `blsPublicKey`), and can be used for dynamically updating the "trusted" nodes in the ethstats-operator side. Only validators that have the `--mine` option enabled can send this list.

### Tested

The changes were tested by connecting `geth` to `integration` and monitoring to ethstats while modifying the "trusted" node list. Automated tests to be added (WIP).

### Related issues

- Fixes #458, #462, #459, #460, #465

### Backwards compatibility

- no more `WS_SECRET` required 
- stats operators have to initialize the config file with trusted addresses

celostats-local


ethstats/ethstats.go
```diff
@@ -92,21 +99,28 @@ func New(url string, ethServ *eth.Ethereum, lesServ *les.LightEthereum) (*Servic
 		return nil, fmt.Errorf("invalid netstats url: \"%s\", should be nodename:secret@host:port", url)
 	}
 	// Assemble and return the stats service
-	var engine consensus.Engine
+	var (
+		engine consensus.Engine
+		id     string
+	)
 	if ethServ != nil {
+		etherBase, _ := ethServ.Etherbase()
```
timmoreton: worth doing `id =` outside the `if`, checking err here and only setting it to the address if `err == nil` ?

ethstats/ethstats.go
```diff
@@ -551,16 +637,49 @@ func (s *Service) assembleBlockStats(block *types.Block) *blockStats {
...
 
+	// Add validator enode list
+	valsElectedWithoutEnode := s.backend.GetValidators(block.Number(), block.Hash())
+
+	valsElected := make([]validatorInfo, 0, len(valsElectedWithoutEnode))
+	for i := range valsElectedWithoutEnode {
+		valsElected = append(valsElected, validatorInfo{
+			Address:      valsElectedWithoutEnode[i].Address(),
+			BLSPublicKey: valsElectedWithoutEnode[i].BLSPublicKey(),
+			// NodePublicKey: s.backend.valEnodeTable.GetEnodeURLFromAddress(valsElectedWithoutEnode[i].Address()),
```
timmoreton: is this not present yet?

diminator: It is - but I think we're doing our best not to expose this (sentries, encrypted enode table, ...). So I figured I leave it out for now. WDYT?

diminator:
> Could you make it so we keep the structure around and only update it every N (=10?) stats msgs

That makes sense. I was thinking we might find a way to listen to changes on the validator set using channels. maybe that's food for a future issue and now we proceed with every 10 blocks?

ethstats/ethstats.go
```diff
@@ -86,27 +95,36 @@ type Service struct {
...
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("invalid netstats url: \"%s\", should be nodename@host:port", url)
 	}
 	// Assemble and return the stats service
-	var engine consensus.Engine
+	var (
+		engine consensus.Engine
+		id     string
+	)
+	fmt.Println("parts", parts)
```
tkporter: Is this intentional? if so I think it's best practice to use `log` when logging within a goroutine

ethstats/ethstats.go
```diff
@@ -284,65 +302,76 @@ func (s *Service) readLoop(conn *websocket.Conn) {
...
+				}
+				select {
+				case s.histCh <- numbers:
+					continue
+				default:
+				}
 			}
+		default:
+			// Report anything else and continue
+			log.Info("Unknown stats message", "msg", msg)
```
tkporter: If this is an expected case where the server wants a pong maybe we should change this log to indicate that

ethstats/ethstats.go
```diff
@@ -382,8 +415,12 @@ func (s *Service) login(conn *websocket.Conn) error {
 		network = fmt.Sprintf("%d", infos.Protocols["les"].(*les.NodeInfo).Network)
 		protocol = fmt.Sprintf("les/%d", les.ClientProtocolVersions[0])
 	}
+	if s.eth != nil {
+		etherBase, _ = s.eth.Etherbase()
+	}
 	auth := &authMsg{
-		ID: s.node,
+		ID:      s.node,
+		Address: etherBase,
```
tkporter: Does it matter that this could be nil if `s.eth` is nil up above? If so should we fail?

ethstats/ethstats.go
```diff
@@ -515,10 +581,8 @@ func (s *Service) reportBlock(conn *websocket.Conn, block *types.Block) error {
 		"id":    s.node,
 		"block": details,
 	}
-	report := map[string][]interface{}{
-		"emit": {"block", stats},
-	}
-	return websocket.JSON.Send(conn, report)
+
+	return s.sendStats(conn, "block", stats)
```
tkporter: Nothing major but wondering if we should put these string literals (block/node-ping/etc) as consts up here https://github.com/celo-org/celo-blockchain/blob/aa3d9d5f5d175f78ef9f80bd32661fc25ad25eb0/ethstats/ethstats.go#L54

ethstats/ethstats.go
```diff
@@ -577,9 +650,68 @@ func (s *Service) assembleBlockStats(block *types.Block) *blockStats {
...
+		valData, _ = validators.GetValidator(
+			s.eth.BlockChain().CurrentHeader(),
+			state,
+			address)
+
+		valsRegistered = append(valsRegistered, validatorInfo{
+			Address:      address,
+			Score:        fmt.Sprintf("%d", valData.Score),
+			Name:         address.String(),
+			Url:          "no url",
```
tkporter: is this the correct value?

ethstats/ethstats.go
```diff
@@ -92,21 +102,30 @@ func New(url string, ethServ *eth.Ethereum, lesServ *les.LightEthereum) (*Servic
...
-		pass:   parts[3],
-		host:   parts[4],
-		pongCh: make(chan struct{}),
-		histCh: make(chan []uint64, 1),
+		eth:     ethServ,
+		les:     lesServ,
+		engine:  engine,
+		backend: engine.(*istanbulBackend.Backend),
+		node:    id,
+		pass:    parts[3],
```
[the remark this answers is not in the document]
tkporter: Whoops this was from before

ethstats/ethstats.go
```diff
@@ -284,65 +302,76 @@ func (s *Service) readLoop(conn *websocket.Conn) {
...
-			case s.pongCh <- struct{}{}:
-				// Pong delivered, continue listening
-				continue
-			default:
-				// Ping routine dead, abort
-				log.Warn("Stats server pinger seems to have died")
+
+		switch packet := msg.(type) {
+		case map[string]interface{}:
+			msgEmit, _ := packet["emit"].([]interface{})
```
tkporter: Also we should check err here and everywhere else that they're ignored with an underscore

diminator: yeah, that seems like a bigger thing. For now I think it's ok (many of those come from upstream).

ethstats/ethstats.go
```diff
@@ -535,33 +536,56 @@ type txStats struct {
...
+	wallet, errWallet := s.eth.AccountManager().Find(account)
+	if errWallet != nil {
+		return errWallet
+	}
+	pubkey, errPubkey := wallet.GetPublicKey(account)
+	pubkeyBytes := crypto.FromECDSAPub(pubkey)
+	if errPubkey != nil {
+		return errPubkey
+	}
+	fmt.Println("pubkey to address", crypto.PubkeyToAddress(*pubkey).Hex())
```
tkporter: Change to log

accounts/accounts.go
```diff
@@ -105,6 +106,7 @@ type Wallet interface {
 	SignHashBLS(account Account, hash []byte) ([]byte, error)
 	SignMessageBLS(account Account, msg []byte, extraData []byte) ([]byte, error)
 	GenerateProofOfPossession(account Account) ([]byte, error)
+    GetPublicKey(account Account) (*ecdsa.PublicKey, error)
```
tkporter: Might want to run gofmt to fix this indentation issue

PR #545, opened 2019-10-23
Fix validator pre-compiles

timmoreton:
### Description

We were incorrectly attempting to obtain a hash of the current block using `evm.Context.GetHash` which has a nasty failure mode in that if it cannot find a given block number (usually protected only by input validation in the `BLOCKHASH` op call), it builds a cache of the entire chain, which at many hundreds of thousands of blocks was taking 10secs, and therefore causing consensus to timeout when verifying a received block.  

Note: added this comment about semantics of these precompiles:

```
Return the number of validators that are required to sign this current, not yet sealed, block. 
If this block is the last in an epoch, note that that may mean one or more of those validators 
may no longer be elected for subsequent blocks.
```
### Tested

By fixing `integration`

### Other changes

* Avoid building a cache forever in `evm.go` `GetHashFn` 
* Validation check on `numberValidators` pre-compile
* For Tobin Transfers, avoid a lookup for the Registry if the value is zero.

### Backwards compatibility

Yes


core/evm.go
```diff
@@ -92,8 +92,8 @@ func GetHashFn(ref *types.Header, chain ChainContext) func(n uint64) common.Hash
 		if hash, ok := cache[n]; ok {
 			return hash
 		}
-		// Not cached, iterate the blocks and cache the hashes
-		for header := chain.GetHeader(ref.ParentHash, ref.Number.Uint64()-1); header != nil; header = chain.GetHeader(header.ParentHash, header.Number.Uint64()-1) {
+		// Not cached, iterate the blocks and cache the hashes (up to a limit of 256)
```
asaj: Please add a comment here and in `contrac_comm/evm.go` that when one is changed so should the other

core/vm/contracts.go
```diff
@@ -83,8 +83,10 @@ var PrecompiledContractsByzantium = map[common.Address]PrecompiledContract{
 
 // RunPrecompiledContract runs and evaluates the output of a precompiled contract.
 func RunPrecompiledContract(p PrecompiledContract, input []byte, contract *Contract, evm *EVM) (ret []byte, err error) {
+	log.Trace("Running precompiled contract", "input", input, "contract caller address", contract.CallerAddress, "contract gas", contract.Gas)
```
asaj: It would be helpful to include the address of the precompiled contract

core/vm/contracts.go
```diff
@@ -645,6 +645,8 @@ func (c *getValidator) Run(input []byte, caller common.Address, evm *EVM, gas ui
 
 	index := (&big.Int{}).SetBytes(input[0:32])
 
+	validators := evm.Context.Engine.GetValidators(evm.Context.BlockNumber, common.Hash{})
```
asaj: Include a comment explaining why the empty hash is okay to pass

core/vm/contracts.go
```diff
@@ -662,15 +664,14 @@ func (c *numberValidators) RequiredGas(input []byte) uint64 {
...
 func (c *numberValidators) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
-	blockNumber := evm.Context.BlockNumber
-	validators := evm.Context.Engine.GetValidators(blockNumber, evm.Context.GetHash(blockNumber.Uint64()))
 	gas, err := debitRequiredGas(c, input, gas)
-	if err != nil {
+	if err != nil || len(input) != 0 {
 		return nil, gas, err
 	}
 
+	validators := evm.Context.Engine.GetValidators(evm.Context.BlockNumber, common.Hash{})
```
asaj: Ditto

core/vm/contracts.go
```diff
@@ -662,15 +664,14 @@ func (c *numberValidators) RequiredGas(input []byte) uint64 {
 }
 
 func (c *numberValidators) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
-	blockNumber := evm.Context.BlockNumber
-	validators := evm.Context.Engine.GetValidators(blockNumber, evm.Context.GetHash(blockNumber.Uint64()))
 	gas, err := debitRequiredGas(c, input, gas)
-	if err != nil {
+	if err != nil || len(input) != 0 {
```
asaj: The length check should be separate, and return an error specific to invalid input if it fails

core/vm/evm.go
```diff
@@ -515,29 +515,30 @@ func getOrComputeTobinTaxFunctionSelector() []byte {
...
 		}
 
-		// Expected size of ret is 64 bytes because getOrComputeTobinTax() returns two uint256 values,
-		// each of which is equivalent to 32 bytes
-		if binary.Size(ret) == 64 {
-			numerator := new(big.Int).SetBytes(ret[0:32])
-			denominator := new(big.Int).SetBytes(ret[32:64])
-			tobinTax := new(big.Int).Div(new(big.Int).Mul(numerator, amount), denominator)
+		if err == nil {
+			ret, gas, err := evm.Call(AccountRef(sender), *reserveAddress, getOrComputeTobinTaxFunctionSelector(), gas, big.NewInt(0))
```
asaj: This should be moved into a `contract_comm` function that returns the tobin tax

core/vm/contracts.go
asaj: What will this return when called during the last block of an epoch? Note that elections are run at the *end* of the block. This may be like `BLOCKHASH` where we can't call it on the current block. Please think carefully about this

core/vm/evm.go
timmoreton: unfortunately an import cycle makes this very difficult.

core/vm/contracts.go
timmoreton: Semantics are now to retrieve the validator set that needs to sign this block (i.e. we call snapshot with the parent block number and hash).

contract_comm/evm.go
```diff
@@ -97,8 +100,8 @@ func GetHashFn(ref *types.Header, chain ChainContext) func(n uint64) common.Hash
 		if hash, ok := cache[n]; ok {
 			return hash
 		}
-		// Not cached, iterate the blocks and cache the hashes
-		for header := chain.GetHeader(ref.ParentHash, ref.Number.Uint64()-1); header != nil; header = chain.GetHeader(header.ParentHash, header.Number.Uint64()-1) {
+		// Not cached, iterate the blocks and cache the hashes (up to a limit of 256)
+		for i, header := 0, chain.GetHeader(ref.ParentHash, ref.Number.Uint64()-1); header != nil && i < 257; i, header = i+1, chain.GetHeader(header.ParentHash, header.Number.Uint64()-1) {
```
nategraf: nit: `i <= 256` would be more readable

core/vm/contracts.go
```diff
@@ -645,6 +648,8 @@ func (c *getValidator) Run(input []byte, caller common.Address, evm *EVM, gas ui
 
 	index := (&big.Int{}).SetBytes(input[0:32])
 
+	validators := evm.Context.Engine.GetValidators(big.NewInt(int64(evm.Context.BlockNumber.Uint64()-1)), evm.Context.GetHash(evm.Context.BlockNumber.Uint64()-1))
```
nategraf: You might use `BlockNumber.Int64()` instead to avoid the  extra conversion to `int64`

contract_comm/currency/currency.go
```diff
@@ -221,17 +221,18 @@ func GetBalanceOf(accountOwner common.Address, contractAddress common.Address, g
 //-------------------------------
 func retrieveWhitelist(header *types.Header, state vm.StateDB) ([]common.Address, error) {
 	returnList := []common.Address{}
-	gasCurrencyWhiteListAddress, err := contract_comm.GetRegisteredAddress(params.GasCurrencyWhitelistRegistryId, nil, nil)
+
+	_, err := contract_comm.MakeStaticCall(params.GasCurrencyWhitelistRegistryId, getWhitelistFuncABI, "getWhitelist", []interface{}{}, &returnList, params.MaxGasForGetWhiteList, header, state)
 	if err != nil {
```
asaj: You can remove lines 226-235, as all of this error logging will happen at the contract_comm level (see https://github.com/celo-org/celo-monorepo/pull/1292/files for this feature being enhanced)

core/vm/contracts.go
```diff
@@ -629,9 +631,10 @@ func (c *getValidator) RequiredGas(input []byte) uint64 {
 	return params.GetValidatorGas
 }
 
+// Return the validators that are required to sign this current, not yet sealed, block. If this block is
```
asaj: Consider:
`Return the validators that are required to sign the current block.` or 
`Return the validators that are required to sign the current, possibly unsealed, block.`

Whether or not the block has been sealed yet depends on what role you're playing and when you receive the block. For example, for full nodes syncing the chain and validating blocks, the block will of course have a seal on it when doing the tx validation.

core/vm/contracts.go
```diff
@@ -661,16 +666,22 @@ func (c *numberValidators) RequiredGas(input []byte) uint64 {
 	return params.GetValidatorGas
 }
 
+// Return the number of validators that are required to sign this current, not yet sealed, block. If this block is
```
asaj: Ditto

contract_comm/currency/currency.go
timmoreton: I will leave in for now as this change may go out before the other PR

core/vm/contracts.go
timmoreton: that is very true

Issue #530, opened 2019-10-15
Validators SBAT use provided randomness for proposer ordering


nategraf:
### Expected Behavior

Validators should assigned proposers using round-robin ordering with respect to a pseudo-random order seeded with the block randomness.

### Current Behavior

Proposer ordering is determined by the particular way validators are added in the headers, roughly in oldest validator first ordering.

PR #549, opened 2019-10-24
Create shuffled round robin proposer selector [Followup to #536]

nategraf:
Followup to https://github.com/celo-org/celo-blockchain/pull/536

### Description

Prior to this PR the proposers are selected round robin with the order determined by the storage order, which is roughly equal to the order in which the validators are added.

This PR adds shuffled round robin proposer policy, which uses the randomness from the last block of each epoch to create a shuffle order to use for the round robin during that epoch.

### Tested

- Proposer selector tested in unit tests in `consensus/istanbul/validator`

### Other changes

Refactored a few pieces of `consensus/istanbul/validator` code including extracting the ProposerSelector methods to a new `selectors.go`

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/530

### Backwards compatibility

Validators running shuffled round robin will not be compatible with validators running round robin mode.


consensus/istanbul/backend/backend.go
```diff
@@ -516,6 +488,40 @@ func (sb *Backend) getValidators(number uint64, hash common.Hash) istanbul.Valid
 	return snap.ValSet
 }
 
+// validatorRandomnessAtBlockNumber calls into the EVM to get the randomness to use in proposer ordering at a given block.
+func (sb *Backend) validatorRandomnessAtBlockNumber(number uint64, hash common.Hash) (common.Hash, error) {
+	lastBlockInPreviousEpoch := number
+	if number > 0 {
+		lastBlockInPreviousEpoch = number - istanbul.GetNumberWithinEpoch(number, sb.config.Epoch)
+	}
+	header := sb.chain.GetHeader(hash, lastBlockInPreviousEpoch)
```
timmoreton: so looks like we will not be able to verify blocks without having already fetched the last header of that epoch? I feel it would be better to have all of the info we need to verify a block be self-contained in the latest block.

consensus/istanbul/backend/backend.go
```diff
@@ -516,6 +488,40 @@ func (sb *Backend) getValidators(number uint64, hash common.Hash) istanbul.Valid
...
+func (sb *Backend) getOrderedValidators(number uint64, hash common.Hash) istanbul.ValidatorSet {
+	valSet := sb.getValidators(number, hash)
+	if valSet.Size() == 0 {
+		return valSet
+	}
+
+	if valSet.Policy() == istanbul.ShuffledRoundRobin {
+		seed, err := sb.validatorRandomnessAtBlockNumber(number, hash)
+		if err != nil {
+			sb.logger.Error("Failed to set randomness for proposer selection", "block_number", number, "hash", hash, "error", err)
```
timmoreton: not sure we should fall through here

consensus/istanbul/backend/backend.go
nategraf: That would certainly be better, but we do need to pull those block headers to construct the snapshot, so in reality this is not a new requirement. What is new is pulling state at that point. If we were to put the randomness in the header that would work, but I am not sure whether the tradeoffs would be appropriate.

PR #551, opened 2019-10-24
Improve Tobin Tax handling

### LINKED ISSUE #550: Tobin Transfers should be free
### Expected Behavior

 We want to no longer charge for Tobin transfers.

### Current Behavior

Currently, we fail if we could not calculate the Tobin Tax, because the reserve contract was not deployed, or because insufficient gas was provided.  This differs from intended design where we want the transfer to proceed even if Tobin Tax cannot be calculated. However, there is an attack with this intended approach whereby a transfer can avoid the Tobin Tax by not providing enough gas to calculate it.

timmoreton:
### Description

Currently, we fail if we could not calculate the Tobin Tax, because the reserve contract was not deployed, or because insufficient gas was provided. This differs from intended design where we want the transfer to proceed even if Tobin Tax cannot be calculated. However, there is an attack with this intended approach whereby a transfer can avoid the Tobin Tax by not providing enough gas to calculate it.

To resolve this, we want to no longer charge gas for Tobin transfers.

Also, optimization to not calculate Tobin tax when value is zero. 

### Tested

Running unit tests and updating integration tests

### Other changes

* Fix out of date gas costs in tests
* New Circle CI config parameter to simplify re-pointing integration tests at branches: `CELO_MONOREPO_BRANCH_TO_TEST`

### Related issues

- Fixes #550 

### Backwards compatibility

No


params/protocol_params.go
```diff
@@ -150,6 +158,7 @@ const (
 	MaxGasForGetGasPriceMinimum                    uint64 = 2000000
 	MaxGasForGetGroupEpochRewards                  uint64 = 50 * 1000
 	MaxGasForGetMembershipInLastEpoch              uint64 = 1 * 1000000
+	MaxGasForGetOrComputeTobinTax                  uint64 = 100000
```
asaj: Please add an extra 0 here just to be extra cautious

PR #556, opened 2019-10-27
Upgrades proof-of-possession to use Ethereum address

### LINKED ISSUE #449: Validators SBAT submit their BLS signed ethereum address as proof-of-possession
### Expected Behavior

Validators prove possession of their private key by signing their ethereum address

### Current Behavior

Validators prove possession of their private key by signing their BLS public key

kobigurk:
### Description

Currently, the proof of possession is used with the BLS public key, opening the possibility for front-running attacks. This changes the behaviour such that the Ethereum address is used.

In conjuction with https://github.com/celo-org/celo-monorepo/pull/1494.

### Tested

Runs end-to-end tests in celo-monorepo.

### Other changes

* Upgrades to Rust 1.37.
* Upgrades to latest bls-zexe.

### Related issues

- Fixes #449.

### Backwards compatibility

Not backwards compatible.


accounts/keystore/keystore.go
```diff
@@ -374,7 +374,7 @@ func (ks *KeyStore) GenerateProofOfPossession(a accounts.Account) ([]byte, error
 	}
 	defer privateKey.Destroy()
 
-	signature, err := privateKey.SignPoP()
+	signature, err := privateKey.SignPoP(a.Address[:])
```
nategraf: Is this copy necessary? If so then could you add a comment for the benefit of anyone looking at this in the future (so they don't remove the copy).

kobigurk: > Is this copy necessary? If so then could you add a comment for the benefit of anyone looking at this in the future (so they don't remove the copy).

You mean the `a.Address[:]` part? It's not a copy then, the `Address` is an array, and SignPoP expects a slice.

kevjue: Perhaps use `a.Address.Bytes()`?  Seems to be a bit more readable.

PR #560, opened 2019-10-31
Adjust new interface from Accounts.sol

nambrot:
### Description

Adjust to new interface due to https://github.com/celo-org/celo-monorepo/pull/1392

### Tested

- Ran e2e test locally
- Will run e2e from the above PR

### Backwards compatibility

- Not compatible without the above PR

### DIFF
--- a/contract_comm/validators/validators.go
+++ b/contract_comm/validators/validators.go
@@ -144,7 +144,6 @@ func GetValidatorData(header *types.Header, state vm.StateDB, validatorAddresses
 	var validatorData []istanbul.ValidatorData
 	for _, addr := range validatorAddresses {
 		validator := struct {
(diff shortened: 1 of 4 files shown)


contract_comm/validators/validators.go
```diff
@@ -148,7 +144,6 @@ func GetValidatorData(header *types.Header, state vm.StateDB, validatorAddresses
 	var validatorData []istanbul.ValidatorData
 	for _, addr := range validatorAddresses {
 		validator := struct {
-			Name           string
```
asaj: Shouldn't URL be gone too?

nambrot: Indeed, it was removed from the ABI by someone else before already

PR #561, opened 2019-10-31
Use EpochRewards contract for proof-of-stake rewards

asaj:
### Description

Updates proof-of-stake incentives from hard coded values to dynamically calculated via `EpochRewards`, added in https://github.com/celo-org/celo-monorepo/pull/1558.

### Tested

End-to-end tests

### Other changes

None

### Related issues

- Fixes #TODO
### Backwards compatibility

Not backwards compatible


contract_comm/epoch_rewards/epoch_rewards.go
```diff
@@ -0,0 +1,77 @@
...
+      "name": "updateTargetVotingYield",
+      "outputs": [],
+      "payable": false,
+      "stateMutability": "nonpayable",
+      "type": "function"
+    }
+]
+`
+
+var epochRewardsABI, _ = abi.JSON(strings.NewReader(epochRewardsABIString))
```
mcortesi: shouldn't we do this in an `init()` function and fail if there's an error?

Issue #540, opened 2019-10-21
Refactor tx fee payments


timmoreton:
### Expected Behavior

In `state_transition.go` change how transaction fees are distributed: 
* Send the entire `base` transaction fee to the infrastructure fund
*  Send the `tip` to the validator.

And update the documentation.

Background: with our current scheme, there is an attack if any part of the base is allocated to an entity (i.e a validator or a full node) whereby they have an incentive to fill blocks with dummy transactions (with low or no cost, since they receive [a portion of] the `base`) and push the gas price minimum up.  

We will be adding a new fixed fee for payments to full nodes. See #541

nategraf:
The initial proposal here is to send the entire gas fee base to the infrastructure fund. We could alternatively burn part or even all of it (as decided through governance), which is closer to the original design discussed in EIP1559 [1].

One edge case worth noting is when the the Governance contract has not yet been deployed or otherwise cannot be found. In that case my intuition is to burn the gas fee, but that may be worth discussing.

[1] https://medium.com/@eric.conner/fixing-the-ethereum-fee-market-eip-1559-9109f1c1814b

PR #563, opened 2019-11-01
Split gas fee into base and tip with base going to infrastructure fund

nategraf:
### Description

As discussed in https://github.com/celo-org/celo-blockchain/issues/540, this PR splits the gas fee into a base and tip, sending the base to the infrastructure fund and the tip to the block proposer. Additionally the proposer fraction is no longer needed from the has price minimum contract and references to it are removed.

### Tested

Existing unit tests in this repo and e2e tests [WIP] in the monorepo

### Other changes

Changed p2p tests for invalid port format to use `strings.Contains` to work around differing output behavior from golang 11.5 and what is used in CI.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/540

### Backwards compatibility

This change is not backwards compatible


nategraf:
Marking `do not merge` util I've had a chance to ensure this is covered in e2e tests

nategraf:
Removing `do not merge` tag with the introduction of https://github.com/celo-org/celo-monorepo/pull/1585

contract_comm/gasprice_minimum/gasprice_minimum.go
```diff
@@ -98,17 +80,11 @@ const (
 )
 
 var (
-	gasPriceMinimumABI, _                      = abi.JSON(strings.NewReader(gasPriceMinimumABIString))
-	FallbackProposerFraction *ProposerFraction = &ProposerFraction{big.NewInt(0), big.NewInt(1)}
-	FallbackGasPriceMinimum  *big.Int          = big.NewInt(0) // gasprice min to return if contracts are not found
-	suggestionMultiplier     *big.Int          = big.NewInt(5) // The multiplier that we apply to the minimum when suggesting gas price
+	gasPriceMinimumABI, _            = abi.JSON(strings.NewReader(gasPriceMinimumABIString))
+	FallbackGasPriceMinimum *big.Int = big.NewInt(0) // gasprice min to return if contracts are not found
```
asaj: nit: Adjust comment to `gas price minimum to return if unable to fetch from contract`

core/state_transition.go
```diff
@@ -374,27 +373,29 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
...
-	if err = st.creditGas(st.evm.Coinbase, coinbaseTxFee, msg.GasCurrency()); err != nil {
+	if err = st.creditGas(st.evm.Coinbase, tipTxFee, msg.GasCurrency()); err != nil {
 		return nil, 0, false, err
 	}
-	if err = st.creditGas(*txFeeRecipient, recipientTxFee, msg.GasCurrency()); err != nil {
-		return nil, 0, false, err
+
+	governanceAddress, err := vm.GetRegisteredAddressWithEvm(params.GovernanceRegistryId, evm)
+	if err != nil {
+		log.Trace("Cannot credit gas fee to infrastructure fund: burning the fee", "error", err, "fee", baseTxFee)
```
asaj: My concern here is that burning the fee would break the invariant that `totalSupply` is the sum of balances over all accounts. Which I suppose nothing relies on strictly, but it makes me uneasy.

Is there somewhere else we could send it that would make sense?

p2p/discv5/node_test.go
```diff
@@ -73,7 +73,7 @@ var parseNodeTests = []struct {
 	{
 		rawurl: "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo",
 		// net/url.Parse(rawurl) returns an error with rawurl and why the parse failed.
-		wantError: `parse enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo: invalid port ":foo" after host`,
+		wantError: `invalid port`,
```
asaj: Why did this file change?

p2p/enode/urlv4_test.go
```diff
@@ -48,7 +48,7 @@ var parseNodeTests = []struct {
 	{
 		// net/url.Parse(rawurl) returns an error with rawurl and why the parse failed.
 		rawurl:    "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo",
-		wantError: `parse enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo: invalid port ":foo" after host`,
+		wantError: `invalid port`,
```
asaj: Ditto

p2p/discv5/node_test.go
nategraf: The output of this test depends on what version of go is being run. It was modified in https://github.com/celo-org/celo-blockchain/commit/f960cf71d8 to work for the version of Go used in Circle, but was failing for me locally with Go 1.11.5. I would like to make it work more universally so that I can run go test locally.

core/state_transition.go
nategraf: I see two options here:
1.  Send it to either the sender or the block proposer, both of which expose the vulnerability we are trying to eliminate, but only until the Governance contract is deployed.
2. Have a fixed address (like out `ce10` address for the registry) or one that is configured in the genesis block act as the account for the infrastructure fund so we know it is always available.

contract_comm/gasprice_minimum/gasprice_minimum.go
nategraf: 👍

core/state_transition.go
asaj: +1 to option 1, I think it's fine for this vulnerability to exist in that small window during network initialization. There even exists a world in which we encode the deployment of all the core contract into the genesis block.

I don't have a strong preference for sending it to the proposer vs the sender. For the former, we expose the attack where the block proposers will want to fill the block with fake txs to drive up fee revenue. For the latter, we expose the DOS risk of potentially free txs for users, which only occurs if tx senders specify the exact gas price minimum as their gas price. In the case where there's competition, users will need to specify higher gas prices, and I think this actually winds up reducing to be more or less equivalent to the market for txs in ethereum (though not exactly). So actually, would probably suggest sending it back to the sender.

nategraf: Sending the fees to the sender has the same gas price minimum (GPM) attack as if it was sent to proposer because the proposer can construct a number of empty transactions that are send from accounts they control, so it has both the GPM attack risk and the DOS attack risk.
One addition to this PR that could be made is to only enforce the GPM if the governance address is available in addition to the GPM parameter. That will encode the current Ethereum market system as the fallback and eliminate this window for a GPM attack to occur (as there will be no enforce GPM to push up)

nategraf: We discussed this offline: My assertion that the GPM attack is still feasible in the "refund to sender" scenario is not true because the proposers will have no incentive to push the GPM up (and a weak incentive to push it down as a high GPM in that scenario means the caller needs a certain amount of funds in their account)
We will go with the "refund to sender" approach

PR #572, opened 2019-11-06
Remove all identity-related code from celo-blockchain

### LINKED ISSUE #571: Developers should not be able to reference identity related code in celo-blockchain

nambrot:
### Description

With the move to make the reveal off-chain, we can remove all identity related code from celo-blockchain

### Tested

- [x] compiles

### Related issues

- Fixes #571 
### Backwards compatibility

- Not compatible with the contract version of Attestations that still reveals on-chain


asaj:
Is there a corresponding PR that removes this code from `Attestations.sol`?

nambrot:
Yes! https://github.com/celo-org/celo-monorepo/pull/1578

PR #515, opened 2019-10-07
Proxy v1

kevjue:
### Description

The first implementation of the Proxy node.

### Tested

* Made sure all the geth unit tests works
* Creates an e2e test for the sentry and verified that all the network connections are correctly created and that the network started mining and the contracts were successfully migrated.

### Other changes

- Moved all of the `validator` and `sentry` connection management to the `istanbul` package (originally that was in the `p2p` package).
- Announce message
    - Modified the `announce` message to use the istanbul.Message type
    - Removed the encryption of the enodeURL within the announce message
    - Now using the hash of the enodeURL for the lastAnnounceGossiped table

### Related issues

- Fixes #192 
- Fixes #510 

### Backwards compatibility

Is not backward compatible.

PR #583, opened 2019-11-13
refactor istanbul message

kevjue:
### Description

This is a follow up PR for this one:  https://github.com/celo-org/celo-blockchain/pull/515

It has the following changes:

1) Created a new subject type `CommittedSubject` that is used for commit messages.  This type contains a subject and a commitedSeal for it.  The `CommittedSeal` field that original was in the Istanbul `Message` type is moved to the new type.  @kobigurk - I would really appreciate it if you could take a glance at this.
2) Created a new istanbul submessage type `ForwardMessage` that is used by proxied validators to instruct the proxy forward it's consensus messages.

### Tested

* Ran geth unit tests
* Ran e2e proxy test

### Backwards compatibility

Is not backward compatible.

### DIFF
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -118,9 +118,16 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, peer consensus.Pe
 				return true, nil
 			}
 
-			fwdMsg := new(istanbul.Message)
-			if err := fwdMsg.FromPayload(data, nil); err != nil {
+			msg := new(istanbul.Message)
+			if err := msg.FromPayload(data, nil); err != nil {
 				sb.logger.Error("Failed to decode message from payload", "err", err)
+				return true, err
+			}
+
+			var fwdMsg *istanbul.ForwardMessage
+			err := msg.Decode(&fwdMsg)
+			if err != nil {
+				sb.logger.Error("Failed to decode a ForwardMessage", "err", err)
 				return true, err
 			}
 
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -118,14 +118,17 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, peer consensus.Pe
 				return true, nil
 			}
 
-			msg := new(istanbul.Message)
-			if err := msg.FromPayload(data, nil); err != nil {
+			istMsg := new(istanbul.Message)
+
+			// An Istanbul FwdMsg doesn't have a signature since it's coming from a trusted peer and
+			// the wrapped message is already signed by the proxied validator.
+			if err := istMsg.FromPayload(data, nil); err != nil {
 				sb.logger.Error("Failed to decode message from payload", "err", err)
 				return true, err
 			}
 
 			var fwdMsg *istanbul.ForwardMessage
-			err := msg.Decode(&fwdMsg)
+			err := istMsg.Decode(&fwdMsg)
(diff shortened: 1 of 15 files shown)


consensus/istanbul/backend/handler.go
```diff
@@ -118,12 +118,19 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, peer consensus.Pe
 				return true, nil
 			}
 
-			fwdMsg := new(istanbul.Message)
-			if err := fwdMsg.FromPayload(data, nil); err != nil {
+			msg := new(istanbul.Message)
```
asaj: Reused var name could be confusing

consensus/istanbul/backend/handler.go
```diff
@@ -118,12 +118,19 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, peer consensus.Pe
...
+			if err := msg.FromPayload(data, nil); err != nil {
 				sb.logger.Error("Failed to decode message from payload", "err", err)
 				return true, err
 			}
 
+			var fwdMsg *istanbul.ForwardMessage
+			err := msg.Decode(&fwdMsg)
```
asaj: IIUC, we're not checking the signature on the message to forward because we've already verified that the fwd message came from a proxied peer, correct?

consensus/istanbul/core/commit.go
```diff
@@ -51,36 +51,38 @@ func (c *core) generateCommittedSeal(digest common.Hash) ([]byte, error) {
...
-		return
+		logger.Error("Failed to encode committedSubject", committedSub)
 	}
 
 	istMsg := istanbul.Message{
-		Code:          istanbul.MsgCommit,
-		Msg:           encodedSubject,
-		CommittedSeal: committedSeal,
+		Code: istanbul.MsgCommit,
+		Msg:  encodedCommittedSubject,
```
asaj: 👍 👍 👍

consensus/istanbul/core/core.go
```diff
@@ -173,7 +173,14 @@ func (c *core) commit() {
 		committedSeals := make([][]byte, c.current.Commits.Size())
 		for i, v := range c.current.Commits.Values() {
 			committedSeals[i] = make([]byte, types.IstanbulExtraCommittedSeal)
-			copy(committedSeals[i][:], v.CommittedSeal[:])
+
+			var commit *istanbul.CommittedSubject
+			err := v.Decode(&commit)
+			if err != nil {
+				panic(fmt.Sprintf("commit: error in decoding committed subject for address %s", hex.EncodeToString(v.Address[:])))
```
asaj: As we discussed yesterday, would be better to return an error. Fine to save this for a follow-up PR as I believe we have an issue around this

consensus/istanbul/core/prepare.go
```diff
@@ -98,7 +111,8 @@ func (c *core) verifyPreparedCertificate(preparedCertificate istanbul.PreparedCe
 		// If COMMIT message, verify valid committed seal.
 		if message.Code == istanbul.MsgCommit {
 			_, src := c.valSet.GetByAddress(signer)
-			err := c.verifyCommittedSeal(subject.Digest, message.CommittedSeal, src)
+
```
asaj: Would it be cleaner to move this block into the `if message.Code == istanbul.MsgCommit` above?

consensus/istanbul/backend/handler.go
kevjue: Yes, that's correct.  I'll add a comment for that.

PR #584, opened 2019-11-13
Extra receipt for internal EVM call when finalizing blocks

### LINKED ISSUE #370: Internal EVM calls SBAT emit events
### Expected Behavior

If the inflation factor gets updated when calling `debitFrom` to pay for a transaction fee with Celo Dollars, an `InflationFactorUpdated` event should be emitted.

### Current Behavior

Haven't checked 100%, but from what I've seen, events from internal EVM calls are likely to be lost. So first step is to confirm this is actually a problem, but wanted to flag this.

mrsmkl:
### Description

When a block is finalized, the events emitted from internal EVM calls were not stored into the block, because there was no associated transaction and receipt. This PR adds an extra receipt in the case the block finalization emits events.

### Tested

Let's see if CI works.

### Related issues

- Fixes #370

### Backwards compatibility

Might cause some issues, for example the events are associated to tx with hash `0x00000000`.


asaj:
Looks like this may have broken fast sync?

mrsmkl:
Sorry, thought it would be working because transfer tests also fast sync. But validators aren't elected there...

core/blockchain.go
```diff
@@ -816,6 +816,16 @@ func SetReceiptsData(config *params.ChainConfig, block *types.Block, receipts ty
 			logIndex++
 		}
 	}
+	// Handle extra receipt
```
asaj: nit: wdyt of `Handle block finalization receipt`?

core/state_processor.go
```diff
@@ -84,8 +84,16 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		allLogs = append(allLogs, receipt.Logs...)
 	}
 	// Finalize the block, applying any consensus engine specific extras (e.g. block rewards)
+	statedb.Prepare(common.Hash{}, block.Hash(), len(block.Transactions()))
```
asaj: Why is this needed?

core/state_processor.go
```diff
@@ -84,8 +84,16 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
 	p.engine.Finalize(p.bc, header, statedb, block.Transactions(), block.Uncles(), receipts, block.Randomness())
 
+	if len(statedb.GetLogs(common.Hash{})) > 0 {
```
asaj: Not 100% clear on what's going on here. Would there be a single receipt for all blocks? If not, won't we have the null hash being mapped to many receipts?

asaj:
Thanks for working on this!

core/state_processor.go
mrsmkl: When logs are added, the associated transaction will be read from the instance variables that are set by this method.

core/state_processor.go
mrsmkl: Yeah, many receipts will be associated into null hash, but they can be separated because they are contained in different blocks.

Comparison between using null and block hash
* if block hash is used, there will be more hashes without corresponding transactions
* if null hash is used, (tx hash, log index) pairs will no longer be unique

I'll next check that tx hash in receipt is not consensus data.

core/blockchain.go
```diff
@@ -786,11 +786,11 @@ func SetReceiptsData(config *params.ChainConfig, block *types.Block, receipts ty
 	signer := types.MakeSigner(config, block.Number())
 
 	transactions, logIndex := block.Transactions(), uint(0)
-	if len(transactions) != len(receipts) {
+	if len(transactions) != len(receipts) && len(transactions)+1 != len(receipts) {
```
asaj: Please add a comment above this:
```The receipts may include an additional "block finalization" receipt```

File: core/blockchain.go (hunk 1 above)
Comment (asaj) [reply core/blockchain.go]: Wdyt?

File: miner/worker.go [hunk 5]
```diff
@@ -614,6 +614,9 @@ func (w *worker) resultLoop() {
 				// receipt/log of individual transactions were created.
 				for _, log := range receipt.Logs {
 					log.BlockHash = hash
+					if (log.TxHash == common.Hash{}) {
```
asaj: Please add a comment ```Handle block finalization receipt```

PR #592, opened 2019-11-14
Move duplicated code to core.ValidateTransactorBalanceCoversTx

### LINKED ISSUE #320: Devs SBAT share tx pool payment logic
### Expected Behavior

One function encapsulates the logic on whether a tx submitter has enough value to pay for a tx in the txpool

### Current Behavior

I copied and pasted some code from the regular transaction pool in this PR (https://github.com/celo-org/geth/pull/319), and it seems that it should be shared. Potentially as part of the iEVM work

jfoutts-celo:
### Description

Moves a block of code from core.TxPool.validateTx (that's duplicated in light.TxPool.validateTx) to core.ValidateTransactorBalanceCoversTx.

### Tested

Tested that core unit tests succeed and cover both cases (sufficient and insufficient transaction funds) for the base currency.  There are no core unit tests for transactions that are invalid due to insufficient ERC20 tx.GasCurrency().

### Related issues

- Fixes #320


core/tx_pool.go
```diff
@@ -1213,6 +1197,35 @@ func (pool *TxPool) demoteUnexecutables() {
 	}
 }
 
+// ValidateTransactorBalanceCoversTx valdiates transactor has enough funds to cover transaction cost: V + GP * GL.
```
asaj: nit: typo

PR #602, opened 2019-11-15
Replace GasFeeRecipient with GatewayFee

### LINKED ISSUE #541: Separate transaction payment to full nodes
### Expected Behavior

Related to #540 

* New field in transaction `gatewayFee` to be used a payment to full nodes denominated in the same currency as `gasCurrency`
* in `state_transition.go`, add this to the amount debited at the beginning of the transaction, and make the transfer of this amount to the address specified both for success and failure
* for betanet, we will enforce a minimum fee as a global constant (later we will want nodes to be able to specify their own fee).  When accepting transactions, add a check that the fee meets this minimum. 
* rename `gasCurrency` as `feeCurrency` to show this applies to both payments for gas and fee to gateway full node -- this may affect the wallet, contractKit and other tx generators!

nategraf:
### Description

As discussed in #541, paying light client serving full-nodes is being changed from being a distribution of the gas fees to a GasFeeRecipient to a separate GatewayFee with GatewayFeeRecipient.

This PR includes all changes needed to implement a basic form of the gateway fee scheme.
* Replace `GasFeeRecipient` with `GatewayFeeRecipient` (in most cases, but a rename)
* Rename `GasCurrency` to `FeeCurrency`
* Add `GatewayFee` to transaction structs
* Distribute the gateway fee in `core/state_transition.go`
* Add `GatewayFee` to config struct
* Add `gatewayfee` flag to allow setting the gateway fee on the full node
* Enforce the gateway fee in `les/handler.go`

This PR explicitly does not include support for clients to query the gateway fee of their peers or send transactions to nodes based on who offers a better rate. Nor does it support the full node providing priority service to transactions with higher gateway fees.

### Tested

Added unit tests to verify signing and serialization as well as enforcement in the LES handler.
Will be adding checks in the transfer test within in the monorepo in separate PR.

### Related issues

- Fixes #541 

### Backwards compatibility

Definitely not


core/state_transition.go
```diff
@@ -185,41 +187,47 @@ func (st *StateTransition) useGas(amount uint64) error {
 	return nil
 }
 
-func (st *StateTransition) buyGas() error {
-	mgval := new(big.Int).Mul(new(big.Int).SetUint64(st.msg.Gas()), st.gasPrice)
+// payFees deducts gas and gateway fees from sender balance and returns error if the sender cannot pay.
```
asaj: note that it also sets the initial gas amount

core/state_transition.go
```diff
@@ -387,7 +395,16 @@ func (st *StateTransition) distributeTxFees() error {
 	baseTxFee := new(big.Int).Mul(gasUsed, st.gasPriceMinimum)
 	tipTxFee := new(big.Int).Sub(totalTxFee, baseTxFee)
 
-	if err := st.creditGas(st.evm.Coinbase, tipTxFee, st.msg.GasCurrency()); err != nil {
+	// Pay gateway fee to the specified recipient.
+	if st.msg.GatewayFeeRecipient() != nil {
+		err := st.creditFee(*st.msg.GatewayFeeRecipient(), st.msg.GatewayFee(), st.msg.FeeCurrency())
```
asaj: nit: `if err := st.creditFee(*st.msg.GatewayFeeRecipient(), st.msg.GatewayFee(), st.msg.FeeCurrency()); err != nil {`

core/tx_pool.go
```diff
@@ -625,33 +625,33 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
 
 		if err != nil {
-			log.Debug("validateTx error in getting gas currency balance", "gasCurrency", tx.GasCurrency(), "error", err)
+			log.Debug("validateTx error in getting fee currency balance", "feeCurrency", tx.FeeCurrency(), "error", err)
 			return err
 		}
 
-		if gasCurrencyBalance.Cmp(new(big.Int).Mul(tx.GasPrice(), big.NewInt(int64(tx.Gas())))) < 0 {
-			log.Debug("validateTx insufficient gas currency", "gasCurrency", tx.GasCurrency(), "gasCurrencyBalance", gasCurrencyBalance)
+		if feeCurrencyBalance.Cmp(new(big.Int).Mul(tx.GasPrice(), big.NewInt(int64(tx.Gas())))) < 0 {
```
asaj: We should check that they have enough to cover the gateway fee here as well. Note that @jfoutts-celo is refactoring this code to be shared with the light tx pool

core/tx_pool_test.go
```diff
@@ -88,7 +88,7 @@ func transaction(nonce uint64, gaslimit uint64, key *ecdsa.PrivateKey) *types.Tr
 }
 
 func pricedTransaction(nonce uint64, gaslimit uint64, gasprice *big.Int, key *ecdsa.PrivateKey) *types.Transaction {
-	tx, _ := types.SignTx(types.NewTransaction(nonce, common.Address{}, big.NewInt(100), gaslimit, gasprice, nil, nil, nil), types.HomesteadSigner{}, key)
+	tx, _ := types.SignTx(types.NewTransaction(nonce, common.Address{}, big.NewInt(100), gaslimit, gasprice, nil, nil, nil, nil), types.HomesteadSigner{}, key)
```
asaj: Please add tests for the gateway fee

eth/backend.go
```diff
@@ -552,7 +558,15 @@ func (s *Ethereum) IsListening() bool                  { return true } // Always
 func (s *Ethereum) EthVersion() int                    { return int(s.protocolManager.SubProtocols[0].Version) }
 func (s *Ethereum) NetVersion() uint64                 { return s.networkID }
 func (s *Ethereum) Downloader() *downloader.Downloader { return s.protocolManager.downloader }
-func (s *Ethereum) GasFeeRecipient() common.Address    { return s.config.Etherbase }
+func (s *Ethereum) GatewayFee() *big.Int               { return s.gatewayFee }
+
+func (s *Ethereum) GatewayFeeRecipient() common.Address {
+	etherbase, err := s.Etherbase()
```
asaj: Is it more efficient for us to return `nil` here? That way we don't need to actually process the gateway fee

eth/config.go
```diff
@@ -89,7 +90,9 @@ type Config struct {
 	// Light client options
 	LightServ  int `toml:",omitempty"` // Maximum percentage of time allowed for serving LES requests
 	LightPeers int `toml:",omitempty"` // Maximum number of LES client peers
-	// The GasFeeRecipient light clients need to specify in order for their transactions to be accepted by this node.
+	// Minimum gateway fee value to serve a transaction from a light client
+	GatewayFee *big.Int `toml:",omitempty"`
+	// Etherebase is the GatewayFeeRecipient light clients need to specify in order for their transactions to be accepted by this node.
```
asaj: nit: Spelling

interfaces.go
```diff
@@ -113,14 +113,15 @@ type ChainSyncReader interface {
...
-	GasCurrency     *common.Address // 0 for the native currency
-	GasFeeRecipient *common.Address // 0 for returning gas fees to the sender
-	GasPrice        *big.Int        // wei <-> gas exchange ratio
-	Value           *big.Int        // amount of wei sent along with the call
-	Data            []byte          // input data, usually an ABI-encoded contract method invocation
+	From                common.Address  // the sender of the 'transaction'
+	To                  *common.Address // the destination contract (nil for contract creation)
+	Gas                 uint64          // if 0, the call executes with near-infinite gas
+	FeeCurrency         *common.Address // 0 for the native currency
+	GatewayFeeRecipient *common.Address // 0 for no gatway fee
```
asaj: spelling

les/api_backend.go
```diff
@@ -209,6 +210,11 @@ func (b *LesApiBackend) ServiceFilter(ctx context.Context, session *bloombits.Ma
 	}
 }
 
-func (b *LesApiBackend) GasFeeRecipient() common.Address {
+func (b *LesApiBackend) GatewayFeeRecipient() common.Address {
 	return b.eth.GetRandomPeerEtherbase()
 }
+
+func (b *LesApiBackend) GatewayFee() *big.Int {
+	// TODO(nategraf): Create a method to discover the gateway fee values of peers.
```
asaj: Definitely out of scope of this PR, but we can do this in the same message that we send to get the etherbase of our peers

miner/validatorenodes/LOG
```diff
@@ -0,0 +1,20 @@
+=============== Nov 15, 2019 (PST) ===============
```
asaj: Are all these new files intentionally checked in?

core/state_transition.go
nategraf: Added a note about that in the comments.

core/state_transition.go
nategraf: Done

core/tx_pool.go
nategraf: Added the gateway fee to this.

eth/backend.go
nategraf: I don't quite understand this comment? What do you mean by "we don't actually need to process the gateway fee"?

les/api_backend.go
nategraf: Adjusted to the TODO message to include that.

miner/validatorenodes/LOG
nategraf: They were not. Thanks for catching that.

core/tx_pool_test.go
nategraf: Thanks for noticing this. I've added gateway to the overall transaction cost and added a test in `tx_pool_test.go` to make sure transactions are rejected when their funds aren't enough.

eth/backend.go
nategraf: Ah I understand now. This is the full-node api code and a full-node has no reason to specify a gateway fee. I had misunderstood what this package was used for, but found the error just now while working on the `transfer_test.ts`

core/state_transition.go
```diff
@@ -187,7 +187,7 @@ func (st *StateTransition) useGas(amount uint64) error {
 	return nil
 }
 
-// payFees deducts gas and gateway fees from sender balance and returns error if the sender cannot pay.
+// payFees deducts gas and gateway fees from sender balance and adds the purchases amount of gas to the state.
```
asaj: nit: `purchased`

core/state_transition.go
```diff
@@ -397,13 +396,14 @@ func (st *StateTransition) distributeTxFees() error {
 
 	// Pay gateway fee to the specified recipient.
 	if st.msg.GatewayFeeRecipient() != nil {
-		err := st.creditFee(*st.msg.GatewayFeeRecipient(), st.msg.GatewayFee(), st.msg.FeeCurrency())
-		if err != nil {
+		log.Debug("Crediting gateway fee", "recipient", *st.msg.GatewayFeeRecipient(), "amount", st.msg.GatewayFee(), "feeCurrency", st.msg.FeeCurrency())
```
asaj: Please remove these log messages, or at least move them to `Trace`. From experience, these really clutter up the logs.

PR #608, opened 2019-11-19
Hide RoundState Implementation

mcortesi:
### Description

Hide `RoundState` and `MessageSet` behind interface to avoid bad usage. For example, access data to read/write without the proper lock setup

### Tested

unit tests

### Other changes

minor lint issues

### DIFF
--- a/consensus/istanbul/core/message_set.go
+++ b/consensus/istanbul/core/message_set.go
@@ -46,7 +46,6 @@ type MessageSet interface {
 	Values() (result []*istanbul.Message)
 	Size() int
 	Get(addr common.Address) *istanbul.Message
-	// String() string
 }
 
 type messageSetImpl struct {
--- a/consensus/istanbul/core/message_set.go
+++ b/consensus/istanbul/core/message_set.go
@@ -26,8 +26,8 @@ import (
 )
 
 // Construct a new message set to accumulate messages for given sequence/view number.
-func newMessageSet(valSet istanbul.ValidatorSet) *messageSet {
-	return &messageSet{
+func newMessageSet(valSet istanbul.ValidatorSet) MessageSet {
+	return &messageSetImpl{
 		messagesMu: new(sync.Mutex),
 		messages:   make(map[common.Address]*istanbul.Message),
 		valSet:     valSet,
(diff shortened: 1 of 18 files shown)


consensus/istanbul/core/message_set.go
```diff
@@ -36,13 +36,26 @@ func newMessageSet(valSet istanbul.ValidatorSet) *messageSet {
...
+	fmt.Stringer
+	Add(msg *istanbul.Message) error
+	GetAddressIndex(addr common.Address) (uint64, error)
+	GetAddressPublicKey(addr common.Address) ([]byte, error)
+	ValSetSize() uint64
+	Remove(address common.Address)
+	Values() (result []*istanbul.Message)
+	Size() int
+	Get(addr common.Address) *istanbul.Message
+	// String() string
```
asaj: Why is this commented out?

consensus/istanbul/core/roundstate.go
```diff
@@ -87,88 +125,107 @@ func (s *roundState) Subject() *istanbul.Subject {
...
 			Sequence: new(big.Int).Set(s.sequence),
 		},
-		Digest: s.Preprepare.Proposal.Hash(),
+		Digest: s.preprepare.Proposal.Hash(),
 	}
 }
 
-func (s *roundState) SetPreprepare(preprepare *istanbul.Preprepare) {
+func (s *roundStateImpl) Preprepare() *istanbul.Preprepare {
+	s.mu.RLock()
```
asaj: Why does this have locking but commits, prepares, and parentcommits don't?

consensus/istanbul/core/roundstate.go
```diff
@@ -177,22 +234,34 @@ func (s *roundState) CreateAndSetPreparedCertificate(quorumSize int) error {
...
 	return nil
 }
 
+func (s *roundStateImpl) PreparedCertificate() istanbul.PreparedCertificate {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.preparedCertificate
+}
+
+func (s *roundStateImpl) SetPreparedCertificate(preparedCertificate istanbul.PreparedCertificate) {
```
asaj: Is this only used for testing? If so, should we remove it?

consensus/istanbul/core/message_set.go
mcortesi: oh, that's the Stringer interface 
i'll remove the comment

consensus/istanbul/core/roundstate.go
mcortesi: yes

PR #612, opened 2019-11-19
Cleanup: newLogger & remove core.View()

mcortesi:
### Description

- Use the `newLogger` factory to create child loggers
- make newLogger private (was public)
- use c.current.View() instead of having a delegate method c.View()
- cleanup some for loops that used break/continue and nested if

### Tested

ci/unit

### DIFF
--- a/consensus/istanbul/core/backlog.go
+++ b/consensus/istanbul/core/backlog.go
@@ -42,7 +42,7 @@ func (c *core) checkMessage(msgCode uint64, view *istanbul.View) error {
 
 	// Round change messages should be in the same sequence but be >= the desired round
 	if msgCode == istanbul.MsgRoundChange {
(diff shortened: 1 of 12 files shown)


consensus/istanbul/core/backlog.go
```diff
@@ -178,28 +173,29 @@ func (c *core) processBacklog() {
...
+				logger.Trace("Post backlog event", "msg", msg)
+
+				go c.sendEvent(backlogEvent{
+					src: src,
+					msg: msg,
+				})
+			} else if err == errFutureMessage {
+				logger.Trace("Stop processing backlog", "msg", msg)
+				backlog.Push(msg, prio)
+				isFuture = true
```
asaj: break?

asaj: Nevermind, as discussed offline this was redundant

PR #617, opened 2019-11-19
Rename LastProposal/Proposer to Block/Author

mcortesi:
### Description

- We don't want a new name for Block.
- The author or coinbase is the name already used. We leave proposer for
the current round proposer (which not always will be the final
author/coinbase)
- Rename ParentValidators to ParentBlockValidators

### Tested

unit/ci

### DIFF
--- a/consensus/istanbul/backend.go
+++ b/consensus/istanbul/backend.go
@@ -70,8 +70,8 @@ type Backend interface {
 	// GetCurrentHeadBlock retrieves the last block
 	GetCurrentHeadBlock() Proposal
 
-	// GetCurrentHeadBlockAndAuthorAndAuthor retrieves the last block alongside the author for that block
-	GetCurrentHeadBlockAndAuthorAndAuthor() (Proposal, common.Address)
+	// GetCurrentHeadBlockAndAuthor retrieves the last block alongside the author for that block
+	GetCurrentHeadBlockAndAuthor() (Proposal, common.Address)
 
 	// LastSubject retrieves latest committed subject (view and digest)
 	LastSubject() (Subject, error)
--- a/consensus/istanbul/backend.go
+++ b/consensus/istanbul/backend.go
@@ -67,17 +67,17 @@ type Backend interface {
 	// the given validator
 	CheckSignature(data []byte, addr common.Address, sig []byte) error
 
-	// LastBlock retrieves the last block
-	LastBlock() Proposal
+	// GetCurrentHeadBlock retrieves the last block
+	GetCurrentHeadBlock() Proposal
 
-	// LastBlockAndAuthor retrieves the last block alongside the author for that block
-	LastBlockAndAuthor() (Proposal, common.Address)
+	// GetCurrentHeadBlockAndAuthorAndAuthor retrieves the last block alongside the author for that block
+	GetCurrentHeadBlockAndAuthorAndAuthor() (Proposal, common.Address)
 
 	// LastSubject retrieves latest committed subject (view and digest)
 	LastSubject() (Subject, error)
 
-	// HasBlockMatching checks if the combination of the given hash and height matches any existing blocks
-	HasBlockMatching(hash common.Hash, number *big.Int) bool
+	// HasBlock checks if the combination of the given hash and height matches any existing blocks
+	HasBlock(hash common.Hash, number *big.Int) bool
 
 	// AuthorForBlock returns the proposer of the given block height
 	AuthorForBlock(number uint64) common.Address
--- a/consensus/istanbul/backend.go
+++ b/consensus/istanbul/backend.go
@@ -67,20 +67,23 @@ type Backend interface {
 	// the given validator
 	CheckSignature(data []byte, addr common.Address, sig []byte) error
 
-	// LastProposal retrieves latest committed proposal and the address of proposer
-	LastProposal() (Proposal, common.Address)
+	// LastBlock retrieves the last block
+	LastBlock() Proposal
+
+	// LastBlockAndAuthor retrieves the last block alongside the author for that block
+	LastBlockAndAuthor() (Proposal, common.Address)
 
 	// LastSubject retrieves latest committed subject (view and digest)
 	LastSubject() (Subject, error)
 
-	// HasProposal checks if the combination of the given hash and height matches any existing blocks
-	HasProposal(hash common.Hash, number *big.Int) bool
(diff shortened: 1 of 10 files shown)


asaj:
Build failing?

mcortesi:
> Build failing?

fixed. Forgot to commit something

kevjue:
Are we sure that we want to rename `Proposal` to `Block`?  It seems like throughout Istanbul, `Proposal` is it's terminology for a `Block`.  E.g. it doesn't have a notion of what a block is.

Perhaps rename to `CommittedProposal` to where it's appropriate?

kevjue:
> Are we sure that we want to rename `Proposal` to `Block`? It seems like throughout Istanbul, `Proposal` is it's terminology for a `Block`. E.g. it doesn't have a notion of what a block is.
> 
> Perhaps rename to `CommittedProposal` where ever it's appropriate?

Actually, I take that back.  I think the original `Istanbul` had that pattern, but we've introduced the concept of a `block` from our changes.

consensus/istanbul/backend/backend.go
```diff
@@ -533,26 +538,18 @@ func (sb *Backend) CheckSignature(data []byte, address common.Address, sig []byt
 	return nil
 }
 
-// HasProposal implements istanbul.Backend.HasProposal
-func (sb *Backend) HasProposal(hash common.Hash, number *big.Int) bool {
+// HasBlockMatching implements istanbul.Backend.HasBlockMatching
```
kevjue: What about just `HasBlock`?

consensus/istanbul/core/core.go
```diff
@@ -343,14 +331,14 @@ func (c *core) startNewRound(round *big.Int) {
 	// New snapshot for new round
 	c.updateRoundState(newView, c.valSet, roundChange)
 	// Calculate new proposer
-	c.valSet.CalcProposer(lastProposer, newView.Round.Uint64())
+	c.valSet.CalcProposer(lastAuthor, newView.Round.Uint64())
 	c.setState(StateAcceptRequest)
-	if roundChange && c.isProposer() && c.current != nil && request != nil {
+	if roundChange && c.current != nil && c.valSet.IsProposer(c.address) && request != nil {
```
kevjue: Is it guaranteed that `c.valset` is not nil?  The original `isProposer` method checked for that.

consensus/istanbul/core/preprepare.go
```diff
@@ -28,7 +28,7 @@ func (c *core) sendPreprepare(request *istanbul.Request, roundChangeCertificate
 	logger := c.newLogger("func", "sendPreprepare")
 
 	// If I'm the proposer and I have the same sequence with the proposal
-	if c.current.Sequence().Cmp(request.Proposal.Number()) == 0 && c.isProposer() {
+	if c.current.Sequence().Cmp(request.Proposal.Number()) == 0 && c.valSet.IsProposer(c.address) {
```
kevjue: Ditto

consensus/istanbul/core/core.go
```diff
@@ -343,14 +331,14 @@ func (c *core) startNewRound(round *big.Int) {
...
 		c.sendPreprepare(request, roundChangeCertificate)
 	}
 	c.newRoundChangeTimer()
 
-	logger.Debug("New round", "new_round", newView.Round, "new_seq", newView.Sequence, "new_proposer", c.valSet.GetProposer(), "valSet", c.valSet.List(), "size", c.valSet.Size(), "isProposer", c.isProposer())
+	logger.Debug("New round", "new_round", newView.Round, "new_seq", newView.Sequence, "new_proposer", c.valSet.GetProposer(), "valSet", c.valSet.List(), "size", c.valSet.Size(), "isProposer", c.valSet.IsProposer(c.address))
```
kevjue: Ditto

consensus/istanbul/backend/backend.go
```diff
@@ -536,25 +533,32 @@ func (sb *Backend) getOrderedValidators(number uint64, hash common.Hash) istanbu
 	return valSet
 }
 
-func (sb *Backend) LastProposal() (istanbul.Proposal, common.Address) {
+// LastBlock retrieves the last block
+func (sb *Backend) LastBlock() istanbul.Proposal {
```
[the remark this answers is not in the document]
mcortesi: sure, i like those names better. I'll go fro `GetCurrentHeadBlock()`

Go has a convention of removing the `Get` part from getter, but i found that sometimes confusing, specially when the name of the variable can also indicate an action or mutation.

consensus/istanbul/core/core.go
```diff
@@ -270,40 +263,35 @@ func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbu
 
 // startNewRound starts a new round. if round equals to 0, it means to starts a new sequence
 func (c *core) startNewRound(round *big.Int) {
-	var logger log.Logger
-	if c.current == nil {
-		logger = c.logger.New("cur_round", -1, "cur_seq", 0, "next_round", 0, "next_seq", 0, "func", "startNewRound", "tag", "stateTransition")
-	} else {
-		logger = c.logger.New("cur_round", c.current.Round(), "cur_seq", c.current.Sequence(), "func", "startNewRound", "tag", "stateTransition")
-	}
+	logger := c.newLogger("func", "startNewRound", "tag", "stateTransition")
```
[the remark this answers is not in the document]
mcortesi: I haven't removed them. `c.newLogger()` creates a logger that already outputs: seq, round, state, address

consensus/istanbul/core/core.go
mcortesi: valSet is set at the beginning so it won't be nil.

Either way I'm planning to remove this all together in a future commit. This was prework, to expose that it actually comes from the valset, not core

consensus/istanbul/backend.go
```diff
@@ -67,20 +67,23 @@ type Backend interface {
 	// the given validator
 	CheckSignature(data []byte, addr common.Address, sig []byte) error
 
-	// LastProposal retrieves latest committed proposal and the address of proposer
-	LastProposal() (Proposal, common.Address)
+	// GetCurrentHeadBlock retrieves the last block
+	GetCurrentHeadBlock() Proposal
+
+	// GetCurrentHeadBlockAndAuthorAndAuthor retrieves the last block alongside the author for that block
+	GetCurrentHeadBlockAndAuthorAndAuthor() (Proposal, common.Address)
```
kevjue: Seems like a typo.

PR #618, opened 2019-11-20
Track uptime & distribute rewards based on it

### LINKED ISSUE #457: Full nodes SBAT track validator uptime within an epoch
### Expected Behavior

Whenever a block is processed, for each validator, a counter is incremented by one if that validator had at least one signature present in the last [10] blocks. This cannot be done during the first [9] blocks of the epoch. At each epoch, this counter is reset to 0.

### Current Behavior

What is the current behavior?

gakonst:
### Description

Will close #457. Idea is that we track a validator's uptime by the bitmap from my previous PR. 

- We can safely assume the bitmap represents 2/3rds of the validators, so we can multiply the number of 1s in the bitmap by 1.5 (3/2) to get an upper bound on the number of validators in that epoch.
- We store each epoch's uptimes, but we could probably just store the latest epoch only
- The "forgiving" factor means that a validator will get always a +1 point if they signed a block, and they will still get a +1, even if they did not sign it, as long as they had at least 1 signature in a sliding window before the current block. 
- We do not score for the first `window` blocks per epoch
 
### Tested

- Unit tests for the uptime tracking based on the [excel sheet draft](https://docs.google.com/spreadsheets/d/1HixYHoNOZamUn2R_IQqnCnEpg4b79WWtd5772qRy5HE/edit?usp=sharing)

### Other changes

- extra cli parameter for the window which defaults to 12


core/rawdb/accessors_chain.go
```diff
@@ -253,6 +256,39 @@ func ReadTd(db DatabaseReader, hash common.Hash, number uint64) *big.Int {
...
+	data, err := rlp.EncodeToBytes(uptime)
+	if err != nil {
+		log.Crit("Failed to RLP encode updated uptime", "err", err)
+	}
+	if err := db.Put(uptimeKey, data); err != nil {
+		log.Crit("Failed to store updated uptime", "err", err)
+	}
+}
+
+// DeleteAccumulatedUptime removes all block total difficulty data associated with a hash.
```
asaj: Associated with a hash?

consensus/istanbul/backend/pos.go
```diff
@@ -78,19 +79,33 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
...
 }
 
+func (sb *Backend) getUptime(validatorIndex int) (uint64, error) {
+	uptimes := rawdb.ReadAccumulatedUptime(sb.db)
+	if uptimes == nil {
+		return 0, errors.New("Invalid accumulated uptime")
+	}
+
+	// the uptimes array is aligned with the validator set
+	validatorUptime := uptimes[validatorIndex].Score / (sb.EpochSize() - sb.LookbackWindow() + 1)
```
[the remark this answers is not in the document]
gakonst: yeah logic in the pos file was still WIP, should be fixed now

core/blockchain.go
```diff
@@ -924,11 +930,104 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+	if bc.engine.Protocol().Name == "istanbul" {
+		// Get the bitmap from the previous block
+		extra, err := types.ExtractIstanbulExtra(block.Header())
+		if err != nil {
+			return NonStatTy, errors.New("could not extract block header extra")
+		}
+
+		signedValidatorsBitmap := extra.ParentAggregatedSeal.Bitmap
+		uptime := bc.GetAccumulatedUptime()
+		uptime = updateUptime(uptime, block.NumberU64(), signedValidatorsBitmap, bc.chainConfig.Istanbul.LookbackWindow)
```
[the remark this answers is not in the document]
gakonst: I changed the uptimes per epoch to be saved in different keys, so there is no direct need to save the data in order to calculate the uptime properly

core/blockchain.go
```diff
@@ -924,11 +931,106 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+		uptime = make([]istanbul.Uptime, validatorsSize)
+	} else {
+		validatorsSize = uint64(len(uptime))
+	}
+
+	for i := 0; i < len(uptime); i++ {
+		if bitmap.Bit(i) == 1 {
+			// update their latest signed block and reward them
+			uptime[i].LastSignedBlock = blockNumber
+			if blockNumber >= firstBlockInEpoch+window-1 {
```
kevjue: I think that this should be `firstBlockInEpoch + window + 1`.

cmd/utils/flags.go
```diff
@@ -661,6 +661,11 @@ var (
 		Usage: "Default minimum difference between two consecutive block's timestamps in seconds",
 		Value: uint64(eth.DefaultConfig.Istanbul.ProposerPolicy),
 	}
+	IstanbulLookbackWindowFlag = cli.Uint64Flag{
+		Name:  "istanbul.lookbackwindow",
+		Usage: "Default number of blocks to use for forgiving downtime",
```
asaj: nit: Consider something like "A validator's signature must be absent for this many consecutive blocks to be considered down."

consensus/istanbul/backend/engine.go
```diff
@@ -484,6 +489,7 @@ func (sb *Backend) Finalize(chain consensus.ChainReader, header *types.Header, s
 		state.RevertToSnapshot(snapshot)
 	}
 
+	sb.logger.Debug("Finalizing", "block", header.Number.Uint64(), "epochSize", sb.config.Epoch)
```
asaj: Trace?

core/blockchain.go
```diff
@@ -937,11 +948,114 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+	signedBlockWindowLastBlockNum := blockNumber - 1
+	signedBlockWindowFirstBlockNum := signedBlockWindowLastBlockNum - (window - 1)
+
+	for i := 0; i < len(uptime); i++ {
+		if bitmap.Bit(i) == 1 {
+			// update their latest signed block
+			uptime[i].LastSignedBlock = blockNumber - 1
+		}
+
+		// If we are within the validator uptime tally window, then update the validator's score if it's last signed block is within
```
gakonst:
```suggestion
		// If we are within the validator uptime tally window, then update the validator's score if its last signed block is within
```

consensus/istanbul/backend/pos.go
```diff
@@ -78,11 +81,27 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
 }
 
 func (sb *Backend) updateValidatorScores(header *types.Header, state *state.StateDB, valSet []istanbul.Validator) error {
-	for _, val := range valSet {
-		// TODO: Use actual uptime metric.
-		// 1.0 in fixidity
-		uptime := math.BigPow(10, 24)
-		sb.logger.Info("Updating validator score for address", "address", val.Address(), "uptime", uptime.String())
+	epoch := istanbul.GetEpochNumber(header.Number.Uint64(), sb.EpochSize())
+	sb.logger.Debug("uptime-trace: updateValidatorScores", "blocknum", header.Number.Uint64(), "epoch", epoch, "epochsize", sb.EpochSize(), "window", sb.LookbackWindow())
```
asaj: Trace?

consensus/istanbul/backend/pos.go
```diff
@@ -78,11 +81,27 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
...
+
+	// The denominator is the (last block - first block + 1) of the val score tally window
+	denominator := istanbul.GetValScoreTallyLastBlockNumber(epoch, sb.EpochSize()) - istanbul.GetValScoreTallyFirstBlockNumber(epoch, sb.EpochSize(), sb.LookbackWindow()) + 1
+
+	// get all the uptimes for this epoch
+	// note(@gakonst): `db` _might_ be possible to be replaced with `sb.db`,
+	// but I believe it's a different database handle
+	c := sb.chain.(*core.BlockChain)
```
asaj: meganit: consider `bc` since we see `c` so often as core

core/blockchain.go
```diff
@@ -937,11 +948,114 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+		// the lookback window
+		if valScoreTallyFirstBlockNum <= blockNumber && blockNumber <= valScoreTallyLastBlockNum {
```
gakonst: LGTM, nit: we could _maybe_ extract this and the `valScoreTally{First, Last}BlockNum` functions together to another function, `isInLookbackWindow`

consensus/istanbul/backend/pos.go
```diff
@@ -78,11 +81,27 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
...
+	db := c.GetDatabase()
+	uptimes := rawdb.ReadAccumulatedEpochUptime(db, epoch)
+	if uptimes == nil {
```
asaj: Please log an error here as well

consensus/istanbul/backend/pos.go
```diff
@@ -78,11 +81,27 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
...
+		return errors.New("no accumulated uptimes found, will not update validator scores")
+	}
+
+	for i, val := range valSet {
+		numerator := big.NewInt(0).Mul(big.NewInt(int64(uptimes[i].Score)), math.BigPow(10, 24))
```
asaj: Can we define `fixidity1 := math.BigPow(10, 24)` in protocol_params.go?

consensus/istanbul/backend/pos.go
```diff
@@ -78,11 +81,27 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
...
+		uptime := big.NewInt(0).Div(numerator, big.NewInt(int64(denominator)))
+
+		sb.logger.Debug("uptime-trace: Updating validator score for address", "index", i, "address", val.Address(), "uptime", uptime)
```
asaj: Trace?

consensus/istanbul/backend/pos.go
asaj: nit: Create a new logger here, with `"func", "updateValidatorScores"` (in addition to all the other things here)

consensus/istanbul/backend/pos.go
asaj: And use the logger you created above

consensus/istanbul/backend/pos.go
```diff
@@ -91,6 +110,33 @@ func (sb *Backend) updateValidatorScores(header *types.Header, state *state.Stat
...
+	if uptimes == nil {
+		panic(fmt.Sprintf("could not read uptimes, i: %v", validatorIndex))
+		return nil, errors.New("Invalid accumulated uptime")
+	}
+
+	// skip the math if we've counted more signatures than necessary for that epoch
+	// when does this happen?
+	if uptimes[validatorIndex].Score >= sb.EpochSize()-sb.LookbackWindow()+1 {
+		// 1.0 in fixidity
+		return math.BigPow(10, 24), nil
```
asaj: Use params.fixidity1

consensus/istanbul/backend/pos.go
```diff
@@ -91,6 +110,33 @@ func (sb *Backend) updateValidatorScores(header *types.Header, state *state.Stat
...
+	db := c.GetDatabase()
+	uptimes := rawdb.ReadAccumulatedEpochUptime(db, epoch)
+	if uptimes == nil {
+		panic(fmt.Sprintf("could not read uptimes, i: %v", validatorIndex))
+		return nil, errors.New("Invalid accumulated uptime")
+	}
+
+	// skip the math if we've counted more signatures than necessary for that epoch
+	// when does this happen?
+	if uptimes[validatorIndex].Score >= sb.EpochSize()-sb.LookbackWindow()+1 {
```
asaj: This should never happen... should we return an error instead?

We should at the very least log an error

core/blockchain.go
```diff
@@ -937,11 +948,114 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+			lastSignedBlock := uptime[i].LastSignedBlock
+			// TODO (kevjue) add comment about 2nd check
+			if signedBlockWindowFirstBlockNum <= lastSignedBlock && lastSignedBlock <= signedBlockWindowLastBlockNum {
```
gakonst:
```suggestion
                        // increase our score as long as we signed at least one block within the lookback window, 
                        // (even if we did not sign on the current block) 
			if signedBlockWindowFirstBlockNum <= lastSignedBlock && lastSignedBlock <= signedBlockWindowLastBlockNum {
```

consensus/istanbul/backend/pos.go
```diff
@@ -91,6 +110,33 @@ func (sb *Backend) updateValidatorScores(header *types.Header, state *state.Stat
 	return nil
 }
 
+func (sb *Backend) getUptime(validatorIndex int, epoch uint64) (*big.Int, error) {
```
asaj: Where is this used?

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
 // WriteBlockWithState writes the block and all associated state to the database.
 func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.Receipt, state *state.StateDB) (status WriteStatus, err error) {
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
+	// We are going to update the uptime tally.
+	if bc.engine.Protocol().Name == "istanbul" {
+		// The epoch's first block's aggregated parent signatures is for the previous epoch's valset.
+		// We can ignore updating the tally for that block.
+		if !istanbul.IsFirstBlockOfEpoch(block.NumberU64(), bc.chainConfig.Istanbul.Epoch) {
```
gakonst: Nice catch! There's probably merit in us documenting all the edge cases we've encountered so far in this and my previous PR about parent seals regarding epoch transitions

consensus/istanbul/types.go
```diff
@@ -26,6 +26,15 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
+type Uptime struct {
+	Score           uint64
```
asaj: nit: Consider something else (e.g. `Count`), since `Score` is an overloaded word. We use `Score` to refer to the exponential moving average

core/blockchain_test.go
```diff
@@ -42,6 +44,79 @@ var (
...
+		},
+		{
+			Score:           0,
+			LastSignedBlock: 211,
+		},
+		{
+			Score:           0,
+			LastSignedBlock: 211,
+		},
+		// plus 2 dummies due to the *1.5
```
gakonst: I do not have an immediate idea on how we can remove these dummies, although they're just going to be empty so don't think they're much harm.

core/rawdb/accessors_chain.go
```diff
@@ -253,6 +254,43 @@ func ReadTd(db DatabaseReader, hash common.Hash, number uint64) *big.Int {
 	return td
 }
 
+// ReadAccumulatedEpochUptime retrieves the so-far accumulated uptime array for the validators of the specified epoch
+func ReadAccumulatedEpochUptime(db DatabaseReader, epoch uint64) []istanbul.Uptime {
+	data, _ := db.Get(uptimeKey(epoch))
+	if len(data) == 0 {
+		log.Debug("uptime-trace: ReadAccumulatedEpochUptime = EMPTY", "epoch", epoch)
```
gakonst: nit: We could make these Traces or remove them altogether, had added them to get a sense of how the code executes during the E2E tests

```suggestion
		log.Trace("uptime-trace: ReadAccumulatedEpochUptime = EMPTY", "epoch", epoch)
```

core/rawdb/accessors_chain.go
```diff
@@ -253,6 +254,43 @@ func ReadTd(db DatabaseReader, hash common.Hash, number uint64) *big.Int {
...
+		return nil
+	}
+	uptime := new([]istanbul.Uptime)
+	if err := rlp.Decode(bytes.NewReader(data), uptime); err != nil {
+		log.Error("Invalid uptime RLP", "err", err)
+		return nil
+	}
+	log.Debug("uptime-trace: ReadAccumulatedEpochUptime(ret)", "epoch", epoch, "uptime", uptime)
```
gakonst: same comment as before on Debug->Trace

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
 	return nil
 }
 
+// GetAccumulatedEpochUptime retrieves the accumulated uptime for an epoch from the database,
+// caching it if found.
+func (bc *BlockChain) GetAccumulatedEpochUptime(epoch uint64) []istanbul.Uptime {
+	uptime := rawdb.ReadAccumulatedEpochUptime(bc.db, epoch)
```
asaj: Should we acquire a read lock? If not, should this be made non-public? Is this function even necessary?

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+	if uptime == nil {
+		return nil
+	}
+	return uptime
+}
+
+// WriteAccumulatedEpochUptime stores the accumulated uptime for an epoch into the database, also caching it
+// along the way.
+func (bc *BlockChain) WriteAccumulatedEpochUptime(epoch uint64, uptime []istanbul.Uptime) error {
```
asaj: Do we need this function at all?

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+	}
+	return uptime
+}
+
 // WriteBlockWithState writes the block and all associated state to the database.
 func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.Receipt, state *state.StateDB) (status WriteStatus, err error) {
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
+	// We are going to update the uptime tally.
```
asaj: Should we be locking during this?

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+	x = (x & m2) + ((x >> 2) & m2) //put count of each 4 bits into those 4 bits
+	x = (x + (x >> 4)) & m4        //put count of each 8 bits into those 8 bits
+	return int((x * h01) >> 56)    //returns left 8 bits of x + (x<<8) + (x<<16) + (x<<24) + ...
+}
+
+func updateUptime(uptime []istanbul.Uptime, blockNumber uint64, bitmap *big.Int, window uint64, epochNum uint64, epochSize uint64) []istanbul.Uptime {
+	var validatorsSize uint64
+	if len(uptime) == 0 {
+		// the number of validators is upper bound by 3/2 of the number of 1s in the bitmap
+		// TODO: due to this maybe we create an uptimeScore array which is 1 element bigger than required
```
asaj: Please remove this comment, since we can create an array that is much more than 1 element larger

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+		// is this a problem?
+		validatorsSize = uint64(math.Ceil(float64(bitCount(bitmap)) * 1.5))
+		// TODO: there is no need for more than the specific block if we know when the validator signed last itme
```
asaj: Remove this?

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
+	return int((x * h01) >> 56)    //returns left 8 bits of x + (x<<8) + (x<<16) + (x<<24) + ...
+}
+
+func updateUptime(uptime []istanbul.Uptime, blockNumber uint64, bitmap *big.Int, window uint64, epochNum uint64, epochSize uint64) []istanbul.Uptime {
+	var validatorsSize uint64
+	if len(uptime) == 0 {
+		// the number of validators is upper bound by 3/2 of the number of 1s in the bitmap
+		// TODO: due to this maybe we create an uptimeScore array which is 1 element bigger than required
+		// is this a problem?
+		validatorsSize = uint64(math.Ceil(float64(bitCount(bitmap)) * 1.5))
```
asaj: nit: `validatorsSizeUpperBound`

core/blockchain.go
```diff
@@ -937,11 +948,117 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
...
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
+	// We are going to update the uptime tally.
+	if bc.engine.Protocol().Name == "istanbul" {
+		// The epoch's first block's aggregated parent signatures is for the previous epoch's valset.
+		// We can ignore updating the tally for that block.
+		if !istanbul.IsFirstBlockOfEpoch(block.NumberU64(), bc.chainConfig.Istanbul.Epoch) {
+			epochNum := istanbul.GetEpochNumber(block.NumberU64(), bc.chainConfig.Istanbul.Epoch)
+			log.Debug("uptime-trace: WriteBlockWithState", "blocknum", block.NumberU64(), "epoch", epochNum)
```
asaj: Trace?  Or remove?

core/rawdb/accessors_chain.go
asaj: Trace or remove?

core/rawdb/accessors_chain.go
asaj: Ditto

core/rawdb/accessors_chain.go
```diff
@@ -253,6 +254,43 @@ func ReadTd(db DatabaseReader, hash common.Hash, number uint64) *big.Int {
...
+
+	return *uptime
+}
+
+// WriteAccumulatedEpochUptime updates the accumulated uptime array for the validators of the specified epoch
+func WriteAccumulatedEpochUptime(db DatabaseWriter, epoch uint64, uptime []istanbul.Uptime) {
+	log.Debug("uptime-trace: WriteAccumulatedEpochUptime", "epoch", epoch, "uptime", uptime)
```
asaj: Trace or remove

core/rawdb/accessors_chain_test.go
```diff
@@ -183,6 +185,42 @@ func TestPartialBlockStorage(t *testing.T) {
...
+		Score:           8,
+		LastSignedBlock: 8,
+	}
+	WriteAccumulatedEpochUptime(db, epoch, uptime)
+	if entry := ReadAccumulatedEpochUptime(db, epoch); entry == nil {
+		t.Fatalf("Stored uptime not found")
+	} else if !reflect.DeepEqual(entry, uptime) {
+		t.Fatalf("Retrieved uptime mismatch: have %v, want %v", entry, uptime)
+	}
+	// Delete the TD and verify the execution
```
asaj: nit: s/TD/uptime

eth/backend.go
```diff
@@ -259,6 +259,12 @@ func CreateConsensusEngine(ctx *node.ServiceContext, chainConfig *params.ChainCo
 		if chainConfig.Istanbul.Epoch != 0 {
 			config.Istanbul.Epoch = chainConfig.Istanbul.Epoch
 		}
+		if chainConfig.Istanbul.LookbackWindow != 0 {
+			if chainConfig.Istanbul.LookbackWindow > chainConfig.Istanbul.Epoch {
```
asaj: Should this be `if chainConfig.Istanbul.LookbackWindow >= chainConfig.Istanbul.Epoch`? Since we can't count the first block of the epoch? Or even `if chainConfig.Istanbul.LookbackWindow >= chainConfig.Istanbul.Epoch - 1`, since we don't count the last block eitehr?

consensus/istanbul/backend/pos.go
kevjue: function is removed

consensus/istanbul/backend/pos.go
kevjue: function is removed

consensus/istanbul/backend/pos.go
kevjue: function is removed

core/blockchain_test.go
asaj: I think the way around this is to make the datastructure be an array whose length is equal to the highest order bit that we've seen in the bitmap thus far. But I don't think we should be in a hurry to fix this

PR #619, opened 2019-11-20
Move state inside RoundState.

mcortesi:
### Description

- c.state now lives inside RoundState
- RoundState has method for every state transition
- Use RoundState methods to Add Prepare/Commit/ParentCommit msgs
- Cleanup everywhere
- Remove unncessary construction arguments for RoundState
- Change where backlog processing is dispatched (at the beginning of a
round)
### Tested

ci/unit

### DIFF
--- a/consensus/istanbul/core/backlog_test.go
+++ b/consensus/istanbul/core/backlog_test.go
@@ -38,12 +38,11 @@ func TestCheckMessage(t *testing.T) {
 	}
 	c := &core{
 		logger:  testLogger,
-		state:   StateAcceptRequest,
 		backend: backend,
 		current: newRoundState(&istanbul.View{
 			Sequence: big.NewInt(2),
 			Round:    big.NewInt(2),
-		}, newTestValidatorSet(4), nil, nil, istanbul.EmptyPreparedCertificate(), nil),
+		}, newTestValidatorSet(4), nil, istanbul.EmptyPreparedCertificate(), nil),
 	}
 
 	// invalid view format
@@ -62,7 +61,7 @@ func TestCheckMessage(t *testing.T) {
 		Round: big.NewInt(1),
 	}
 	for i := 0; i < len(testStates); i++ {
-		c.state = testStates[i]
+		c.current.(*roundStateImpl).state = testStates[i]
 		for j := 0; j < len(testCode); j++ {
 			err := c.checkMessage(testCode[j], v)
 			if testCode[j] == istanbul.MsgCommit {
@@ -84,7 +83,7 @@ func TestCheckMessage(t *testing.T) {
 		Round: big.NewInt(0),
 	}
 	for i := 0; i < len(testStates); i++ {
-		c.state = testStates[i]
+		c.current.(*roundStateImpl).state = testStates[i]
 		for j := 0; j < len(testCode); j++ {
 			err := c.checkMessage(testCode[j], v)
 			if err != errOldMessage {
@@ -99,7 +98,7 @@ func TestCheckMessage(t *testing.T) {
 		Round:    big.NewInt(0),
 	}
 	for i := 0; i < len(testStates); i++ {
-		c.state = testStates[i]
+		c.current.(*roundStateImpl).state = testStates[i]
 		for j := 0; j < len(testCode); j++ {
 			err := c.checkMessage(testCode[j], v)
 			if err != errOldMessage {
@@ -114,7 +113,7 @@ func TestCheckMessage(t *testing.T) {
 		Round:    big.NewInt(0),
 	}
 	for i := 0; i < len(testStates); i++ {
-		c.state = testStates[i]
+		c.current.(*roundStateImpl).state = testStates[i]
 		for j := 0; j < len(testCode); j++ {
 			err := c.checkMessage(testCode[j], v)
 			if err != errFutureMessage {
@@ -129,7 +128,7 @@ func TestCheckMessage(t *testing.T) {
 		Round:    big.NewInt(3),
 	}
 	for i := 0; i < len(testStates); i++ {
-		c.state = testStates[i]
+		c.current.(*roundStateImpl).state = testStates[i]
 		for j := 0; j < len(testCode); j++ {
 			err := c.checkMessage(testCode[j], v)
 			if testCode[j] == istanbul.MsgRoundChange {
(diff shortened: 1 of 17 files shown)


consensus/istanbul/core/core.go
```diff
@@ -97,16 +95,17 @@ type core struct {
...
 		seq = c.current.Sequence()
 		round = c.current.Round()
 	} else {
 		seq = common.Big0
 		round = big.NewInt(-1)
 	}
-	tmp := c.logger.New(ctx...)
-	return tmp.New("cur_seq", seq, "cur_round", round, "state", state, "address", c.address)
+	logger := c.logger.New(ctx...)
+	return logger.New("cur_seq", seq, "cur_round", round, "state", state)
```
kevjue: Why remove `address`?

mcortesi: not intented.. i'll add it again

consensus/istanbul/core/testbackend_test.go
```diff
@@ -360,7 +360,7 @@ func NewTestSystemWithBackendAndCurrentRoundState(n, f uint64, getRoundState fun
 		backend.blsKey = blsKeys[i]
 
 		core := New(backend, config).(*core)
-		core.state = StateAcceptRequest
+		// core.state = StateAcceptRequest
```
nambrot: Remove?

consensus/istanbul/core/commit.go
```diff
@@ -113,11 +113,17 @@ func (c *core) handleCheckedCommitForPreviousSequence(msg *istanbul.Message, com
 		logger.Debug("Received a commit message for the previous sequence with an unexpected hash", "expected", headBlock.Hash().String(), "received", commit.Subject.Digest.String())
 		return errInconsistentSubject
 	}
-	return c.acceptParentCommit(msg, commit.Subject.View)
+
+	// Add the ParentCommit to current round state
+	if err := c.current.AddParentCommit(msg); err != nil {
```
nambrot: In `acceptParentCommit` we used to log the view of the parent commit, I assume we  no longer care about that?

consensus/istanbul/core/core.go
```diff
@@ -357,38 +361,41 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
+	c.sendRoundChange(r)
 }
 
-func (c *core) updateRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, roundChange bool) {
+func (c *core) resetRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, roundChange bool) {
 	// TODO(Joshua): Include desired round here.
 	if c.current != nil {
 		if roundChange {
-			c.current = newRoundState(view, validatorSet, nil, c.current.PendingRequest(), c.current.PreparedCertificate(), c.current.ParentCommits())
+			c.current = newRoundState(view, validatorSet, c.current.PendingRequest(), c.current.PreparedCertificate(), c.current.ParentCommits())
```
nambrot: Since you added locks everywhere, don't we technically have to have locks here as well to assemble this new roundState? Or are we suffciently protected by the core mutex?

mcortesi: actually i'm not actually sure that roundState needs locks tbh, i've added them since they were already there for other methods.

the core engine runs on a single thread, commanded the handleEvents go routine.

So i assume it's safe to skip them here.

consensus/istanbul/core/commit.go
```diff
@@ -142,16 +152,19 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
-	} else if c.current.GetPrepareOrCommitSize() >= minQuorumSize && c.state.Cmp(StatePrepared) < 0 {
-		if err := c.current.CreateAndSetPreparedCertificate(minQuorumSize); err != nil {
+	} else if c.current.GetPrepareOrCommitSize() >= minQuorumSize && c.current.State().Cmp(StatePrepared) < 0 {
+		err := c.current.TransitionToPrepared(minQuorumSize)
+		if err != nil {
 			logger.Error("Failed to create and set preprared certificate", "err", err)
 			return err
 		}
+		// Process Backlog Messages
+		c.processBacklog()
```
asaj: Why `processBacklog` separately rather than inside `TransitionTo*`? Wouldn't the latter be more consistent with the current behavior?

consensus/istanbul/core/commit.go
mcortesi: Done

Now using:
```go
logger := c.newLogger("func", "handleCheckedCommitForPreviousSequence", "tag", "handleMsg", "msg_view", commit.Subject.View)
```

consensus/istanbul/core/commit.go
mcortesi: `RoundState` doesn't have access to `core` for now. I agree it's not the best design, I'll try to figure it out on a later PR

consensus/istanbul/core/backlog_test.go
```diff
@@ -408,12 +406,12 @@ func testProcessBacklog(t *testing.T, msg *istanbul.Message) {
...
 		backlogsMu: new(sync.Mutex),
 		backend:    backend,
-		state:      State(msg.Code),
 		current: newRoundState(&istanbul.View{
 			Sequence: big.NewInt(1),
 			Round:    big.NewInt(0),
-		}, newTestValidatorSet(4), nil, nil, istanbul.EmptyPreparedCertificate(), nil),
+		}, newTestValidatorSet(4), nil, istanbul.EmptyPreparedCertificate(), nil),
 	}
+	c.current.(*roundStateImpl).state = State(msg.Code)
```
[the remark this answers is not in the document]
mcortesi: The roundStateImpl casting or the `State(msg.Code)` one? 
I'm only guilty of the first one :P

consensus/istanbul/core/commit.go
nambrot: Agree with Mariano, I had the same reaction at first, but unless we expose all roundState methods as an interface on core, i prefer this

consensus/istanbul/core/roundstate.go
```diff
@@ -160,62 +170,37 @@ func (s *roundStateImpl) Proposal() istanbul.Proposal {
...
-	s.mu.RLock()
-	defer s.mu.RUnlock()
-
-	return s.sequence
+	s.desiredRound = new(big.Int).Set(r)
+	s.state = StateWaitingForNewRound
 }
 
-func (s *roundStateImpl) CreateAndSetPreparedCertificate(quorumSize int) error {
+func (s *roundStateImpl) TransitionToPrepared(quorumSize int) error {
```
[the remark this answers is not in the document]
mcortesi: We should move quorumSize inside. In a later PR valSet lives inside, so makes lot more sense to do it then

PR #622, opened 2019-11-20
Ethstats with proxy and proxied validator (client)

### LINKED ISSUE #482: Ethstats SBAT support validator sentry setup

diminator:
### Description

Enable the proxy to send stats by delegating signed messages to the proxied validator

also see https://github.com/celo-org/ethstats-server/pull/7

### Related issues

- Fixes #482 

### Usage

- The `proxied validator` needs to have the account unlocked for signing stats
- Both `proxy` and `proxied validator` need to have the ethstats flag enabled in `geth`
- The ethstats operator needs to whitelist the `proxied validator`


consensus/istanbul/backend/backend.go
```diff
@@ -186,6 +186,21 @@ type Backend struct {
...
+	delegateSignFeed event.Feed
+}
+
+// This is for ethstats communication
+func (sb *Backend) ProxiedPeer() consensus.Peer {
+	return sb.proxiedPeer
+}
+
+// This is for ethstats communication
+func (sb *Backend) ProxyNode() consensus.Peer {
```
tkporter: nit: probably best to rename this ProxyPeer or something similar

consensus/istanbul/backend/handler.go
```diff
@@ -72,6 +75,18 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, peer consensus.Pe
 			return true, errDecodeFailed
 		}
 
+		if msg.Code == istanbulDelegateSign {
+			if sb.config.Proxy {
```
tkporter: Seems like the same thing is being done, maybe condense this into `if (this is a proxy || this is proxied)` and return an error if neither

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
 				}
-
 			// node stopped
 			case <-txSub.Err():
 				break HandleLoop
 			case <-headSub.Err():
 				break HandleLoop
+			case delegateSignMsg := <-istDelegateSignCh:
+				var statsPayload StatsPayload
+				err := json.Unmarshal(delegateSignMsg.Payload, &statsPayload)
```
tkporter: generally the pattern in go is
```
if err != nil {
  // do something, like return an error, or in this case maybe just continue HandleLoop and print an error
}
```

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+				if err == nil {
+					if s.backend.ProxyNode() != nil {
```
tkporter: both if statements seem to be doing the same thing, so again you could condense them into one

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
 		}
 		close(quitCh)
 	}()
+
 	// Loop reporting until termination
 	for {
-		// Resolve the URL, defaulting to TLS, but falling back to none too
-		path := fmt.Sprintf("%s/api", s.host)
-		urls := []string{path}
+		if s.backend.ProxyNode() != nil {
```
tkporter: I find myself getting confused what this is saying, maybe a comment saying "if this is a proxied validator"

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+			messageToSign := <-signCh
```
tkporter: will this hang if there is nothing in signCh?

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+			if err = s.report(conn, sendCh); err != nil {
+				log.Warn("Initial stats report failed", "err", err)
+				conn.Close()
 				continue
 			}
-			conf.Dialer = &net.Dialer{Timeout: 5 * time.Second}
-			if conn, err = websocket.DialConfig(conf); err == nil {
-				break
+			// Keep sending status updates until the connection breaks
+			fullReport := time.NewTicker(15 * time.Second)
```
tkporter: lots of constants, like here & the time.Sleep calls, consider putting them in a const at the top of the file

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+	infos := s.server.NodeInfo()
+
+	var (
+		etherBase common.Address
+		network   string
+		protocol  string
+	)
+	p := s.engine.Protocol()
+	if info := infos.Protocols[p.Name]; info != nil {
+		network = fmt.Sprintf("%d", info.(*eth.NodeInfo).Network)
```
tkporter: Could this type assertion fail? We should check an err:
```
ethInfo, ok := info.(*eth.NodeInfo)
if !ok {
  // return error
}
```

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+		protocol = fmt.Sprintf("%s/%d", p.Name, p.Versions[0])
+	} else {
+		network = fmt.Sprintf("%d", infos.Protocols["les"].(*les.NodeInfo).Network)
```
tkporter: we should check that `infos.Protocols["les"]` exists:
```
val, ok := infos.Protocols["les"]
if !ok {
  // return error
}
```
and should likewise check that the type assertion doesn't fail

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+		protocol = fmt.Sprintf("les/%d", les.ClientProtocolVersions[0])
+	}
+	if s.eth != nil {
+		etherBase, _ = s.eth.Etherbase()
```
tkporter: check err

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
 			}
+		case <-time.After(5 * time.Second):
+			// Login timeout, abort
+			return errors.New("login timed out")
+		}
+	}
+
+	// Retrieve the remote ack or connection termination
+	var ack map[string][]string
+	if err := websocket.JSON.Receive(conn, &ack); err != nil || len(ack["emit"]) != 1 || ack["emit"][0] != "ready" {
```
tkporter: I think we should separate out his error check into its own if, and also check if `ack["emit"]` exists before checking things on it

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+
+	if err == nil {
+		signedMessage := &StatsPayload{
+			Action: messageToSign.Action,
+			Stats:  signedStats,
 		}
-		// Make sure the connection is closed
-		conn.Close()
+		msg, _ := json.Marshal(signedMessage)
+		return s.backend.ProxyNode().Send(0x15, msg)
```
tkporter: If we do sending in here, should we do it as a goroutine? We also shouldn't hardcode `0x15`

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+	signedStats, err := s.signStats(messageToSign.Stats)
+
+	if err == nil {
+		signedMessage := &StatsPayload{
+			Action: messageToSign.Action,
+			Stats:  signedStats,
 		}
-		// Make sure the connection is closed
-		conn.Close()
+		msg, _ := json.Marshal(signedMessage)
```
tkporter: check err

ethstats/ethstats.go
```diff
@@ -211,93 +253,215 @@ func (s *Service) loop() {
...
+
+func (s *Service) handleDelegateSign(messageToSign StatsPayload) error {
+
+	if s.backend.ProxyNode() == nil {
+		return errors.New("No Proxy found")
+	}
+
+	signedStats, err := s.signStats(messageToSign.Stats)
+
+	if err == nil {
```
tkporter: I think we should try to keep with the `if err != nil { return err }` golang pattern

ethstats/ethstats.go
```diff
@@ -485,6 +579,14 @@ func (s *Service) reportLatency(conn *websocket.Conn) error {
 	if err := s.sendStats(conn, actionNodePing, ping); err != nil {
 		return err
 	}
+	// Proxy needs a delegate send here to get ACK
+	if s.backend.ProxiedPeer() != nil {
```
tkporter: I wonder if for basic checks like this to determine if this is a proxy or a proxied validator, we can instead use a boolean that's exposed from backend. It would help make this more readable too

ethstats/ethstats.go
```diff
@@ -583,18 +691,31 @@ func (s *Service) sendStats(conn *websocket.Conn, action string, stats interface
...
+	return signedStats, nil
+}
+
+func (s *Service) sendStats(conn *websocket.Conn, action string, stats interface{}) error {
+	if s.backend.ProxiedPeer() != nil {
+		statsWithAction := map[string]interface{}{
+			"stats":  stats,
+			"action": action,
+		}
+		msg, _ := json.Marshal(statsWithAction)
```
tkporter: check err

ethstats/ethstats.go
```diff
@@ -583,18 +691,31 @@ func (s *Service) sendStats(conn *websocket.Conn, action string, stats interface
...
+		go s.backend.ProxiedPeer().Send(0x15, msg)
```
tkporter: instead of calling `s.backend.ProxiedPeer()` here and before in the if condition, we can pull it out into a variable. I also don't think we should hardcode the literal `0x15`

ethstats/ethstats.go
```diff
@@ -485,6 +579,14 @@ func (s *Service) reportLatency(conn *websocket.Conn) error {
 	if err := s.sendStats(conn, actionNodePing, ping); err != nil {
 		return err
 	}
+	// Proxy needs a delegate send here to get ACK
+	if s.backend.ProxiedPeer() != nil {
+		signedMessage := <-sendCh
```
tkporter: More just thinking out loud, now this latency will include the time it takes for the proxy to send a message to the validator and receive from it :(

consensus/istanbul/backend/handler.go
```diff
@@ -166,6 +181,11 @@ func (sb *Backend) handleFwdMsg(peer consensus.Peer, payload []byte) error {
 	return nil
 }
 
+// SubscribeNewDelegateSignEvent subscribes a channel to any new delegate sign messages
+func (sb *Backend) SubscribeNewDelegateSignEvent(ch chan<- istanbul.MessageEvent) event.Subscription {
+	return sb.delegateSignFeed.Subscribe(ch)
```
tkporter: I think we should also use an `event.SubscriptionScope` similar to here: https://github.com/celo-org/celo-blockchain/blob/master/core/tx_pool.go#L495

so that we can unsubscribe channels https://github.com/celo-org/celo-blockchain/blob/master/core/tx_pool.go#L480 when we close here: https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/backend.go#L209

consensus/istanbul/backend/backend.go
```diff
@@ -186,6 +186,21 @@ type Backend struct {
 	proxiedPeer consensus.Peer
 
 	newEpochCh chan struct{}
+
+	delegateSignFeed event.Feed
+}
+
+// This is for ethstats communication
```
kevjue: I notice that you have alot of checks similar to:

`if ...ProxiedPeer() != nil`

to figure out if this node is a proxy or is being proxied.

I think it will be more readable if you create two additional methods that are more explicit:

`IsProxy`
`IsProxiedValidator`

it will be easier to understand the code.

ethstats/ethstats.go
kevjue: Perhaps do something like this?

```
var channel chan StatsPayload
if s.backend.IsProxy()l {
     channel = sendCh
} else if s.backend.IsProxiedValidator() {
     channel = signCh
}
channel <- statsPayload
```

ethstats/ethstats.go
```diff
@@ -181,11 +217,17 @@ func (s *Service) loop() {
...
+	istDelegateSignCh := make(chan istanbul.MessageEvent, istDelegateSignChanSize)
+	istDelegateSignSub := s.backend.SubscribeNewDelegateSignEvent(istDelegateSignCh)
+	defer istDelegateSignSub.Unsubscribe()
+
 	// Start a goroutine that exhausts the subsciptions to avoid events piling up
 	var (
 		quitCh = make(chan struct{})
 		headCh = make(chan *types.Block, 1)
 		txCh   = make(chan struct{}, 1)
+		signCh = make(chan StatsPayload, istDelegateSignChanSize)
```
kevjue: Should this be of type `chan *StatsPayload`?

I suppose it's not a huge deal, but would save a bit on memory usage by passing by reference vs. passing by value.

ethstats/ethstats.go
tkporter: Oops didn't realize they were actually different

ethstats/ethstats.go
diminator: I know, it's hard to remove that, unless we 
- measure the proxy <> validator roundtrip and substract that
- ask serverTime at the validator and only have a one way trip from validator to proxy added

ethstats/ethstats.go
diminator: The Proxied Validator should only sign in ethstats, so probably this is not a problem?

consensus/istanbul/backend/handler.go
```diff
@@ -166,6 +178,16 @@ func (sb *Backend) handleFwdMsg(peer consensus.Peer, payload []byte) error {
 	return nil
 }
 
+func (sb *Backend) shouldHandleDelegateSign() bool {
+	return sb.IsProxy() || sb.IsProxiedValidator()
+}
+
+// SubscribeNewDelegateSignEvent subscribes a channel to any new delegate sign messages
+func (sb *Backend) SubscribeNewDelegateSignEvent(ch chan<- istanbul.MessageEvent) event.Subscription {
+	// return sb.delegateSignFeed.Subscribe(ch)
```
tkporter: you can delete this

ethstats/ethstats.go
```diff
@@ -61,7 +61,15 @@ const (
 	txChanSize = 4096
 	// chainHeadChanSize is the size of channel listening to ChainHeadEvent.
 	chainHeadChanSize = 10
-
+	// istDelegateSignChanSize is the size of the channel listening to DelegateSignEvent
+	istDelegateSignChanSize = 5
+
+	// connectionTimeOut waits for the websocket connection to be established
+	connectionTimeOut = 10
```
tkporter: nit: TimeOut -> Timeout to keep with the pattern of the variable right below

ethstats/ethstats.go
```diff
@@ -211,93 +259,229 @@ func (s *Service) loop() {
...
+		if s.backend.IsProxiedValidator() {
+			messageToSign := <-signCh
+			if err := s.handleDelegateSign(messageToSign); err != nil {
+				log.Warn("Delegate sign failed", "err", err)
+			}
+		} else {
+			conn, err := s.getConnection()
+			if err != nil {
+				log.Warn("Stats server unreachable", "err", err)
+				time.Sleep(connectionTimeOut * time.Second)
```
tkporter: I'm confused why we sleep here and line 305

ethstats/ethstats.go
```diff
@@ -211,93 +259,229 @@ func (s *Service) loop() {
...
+	}
 
-		// Send the initial stats so our node looks decent from the get go
-		if err = s.report(conn); err != nil {
-			log.Warn("Initial stats report failed", "err", err)
-			conn.Close()
-			continue
+	// Retrieve the remote ack or connection termination
+	var ack map[string][]string
+	if err := websocket.JSON.Receive(conn, &ack); err != nil {
```
tkporter: I think there's an issue here where you're doing things if there was an error, instead of doing things if there wasn't an error

maybe instead do
```
if err := websocket.JSON.Receive(conn, &ack); err != nil {
  return err
}
// everything else
```

timmoreton: checked this out and it's correct. The stuff in the branch are other error conditions.

PR #623, opened 2019-11-20
Onboarding feedback

jfoutts-celo:
### Description

Small updates from my onboarding notes to slightly improve the onboarding process.

### Tested

Tested that the celo-blockchain COMMIT_SHA is stored to /version.txt in the geth container, and that if COMMIT_SHA is not specified the docker build still succeeds with a blank /version.txt.

While we already tag the container with the COMMIT_SHA, later we derive additional containers FROM it, and this can obscure the version being packaged.  This PR produces evidence to confidently settle any question of the packaged version.


README.md
```diff
@@ -14,6 +14,8 @@ Once the dependencies are installed, run
 or, to build the full suite of utilities:
 
     make all
+    
+Note that when changing branches `make clean` may be required.
```
timmoreton: or `make clean-geth` ?

jfoutts-celo: To answer your question I tested `make clean-geth` and it's insufficient.  This leads directly to the problem / solution.  Unlike for mobile builds, libbls_zexe.a source changes aren't tracked.  The library is never updated.  Marking libbls_exe.a PHONY in the Makefile fixes the behavior.

PR #624, opened 2019-11-20
Fix for backlogging of commit messages

kevjue:
### Description

The format of the commit message's payload changed recently.  The backlog was still parsing the commit msg payload using the old format.

### Tested

Unit tests.

### Backwards compatibility

Yes


asaj:
Can you create an issue to add a test that would have caught this and add it to betanet+2?

kevjue:
> Can you create an issue to add a test that would have caught this and add it to betanet+2?

> How did our tests not catch this? Should we add some here, seems like we would not have been able to process future commit messages?

The backlog test originally was creating commit messages using the old format, so when they were processed through the backlog, it was succeeding.

I think that we should modify the Encode function https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/core/types.go#L78 to accept a message code, and this function verifies that the payload being encoded is the correct format.  I'll create an issue for that.

PR #626, opened 2019-11-21
Add StartNewRound/Sequence to RoundState

mcortesi:
### Description

Instead of replacing c.current, we now call
c.current.StartNewRound/Sequence() to transation round or sequence.

### Tested

ci/unit

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -376,26 +376,28 @@ func (c *core) waitForDesiredRound(r *big.Int) {
 
 func (c *core) resetRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, roundChange bool) {
 	// TODO(Joshua): Include desired round here.
-	if c.current != nil {
-		if roundChange {
-			c.current = newRoundState(view, validatorSet, c.current.PendingRequest(), c.current.PreparedCertificate(), c.current.ParentCommits())
-		} else {
-			lastSubject, err := c.backend.LastSubject()
-			if err != nil && c.current.Proposal() != nil && c.current.Proposal().Hash() == lastSubject.Digest && c.current.Round().Cmp(lastSubject.View.Round) == 0 {
-				// When changing sequences, if our current Commit messages match the latest block in the chain
-				// (i.e. they're for the same block hash and round), we use this sequence's commits as the ParentCommits field
-				// in the next round.
-				c.current = newRoundState(view, validatorSet, nil, istanbul.EmptyPreparedCertificate(), c.current.Commits())
-			} else {
-				headBlock := c.backend.GetCurrentHeadBlock()
-				// Otherwise, we will initialize an empty ParentCommits field with the validator set of the last proposal.
-				c.current = newRoundState(view, validatorSet, nil, istanbul.EmptyPreparedCertificate(), newMessageSet(c.backend.ParentBlockValidators(headBlock)))
-			}
-		}
-	} else {
+	if c.current == nil {
 		// When the current round is nil, we must start with the current validator set in the parent commits
 		// either `validatorSet` or `backend.Validators(lastProposal)` works here
-		c.current = newRoundState(view, validatorSet, nil, istanbul.EmptyPreparedCertificate(), newMessageSet(validatorSet))
+		c.current = newRoundState(view, validatorSet)
+	} else if roundChange {
+		c.current.StartNewRound(view.Round, validatorSet)
+	} else {
+		// sequence change
+
+		var newParentCommits MessageSet
+		lastSubject, err := c.backend.LastSubject()
+		if err != nil && c.current.Proposal() != nil && c.current.Proposal().Hash() == lastSubject.Digest && c.current.Round().Cmp(lastSubject.View.Round) == 0 {
+			// When changing sequences, if our current Commit messages match the latest block in the chain
+			// (i.e. they're for the same block hash and round), we use this sequence's commits as the ParentCommits field
+			// in the next round.
+			newParentCommits = c.current.Commits()
+		} else {
+			// Otherwise, we will initialize an empty ParentCommits field with the validator set of the last proposal.
+			headBlock := c.backend.GetCurrentHeadBlock()
+			newParentCommits = newMessageSet(c.backend.ParentBlockValidators(headBlock))
+		}
+		c.current.StartNewSequence(view, validatorSet, newParentCommits)
 	}
 }
 
--- a/consensus/istanbul/core/roundstate.go
+++ b/consensus/istanbul/core/roundstate.go
@@ -31,8 +31,7 @@ var (
 	errFailedCreatePreparedCertificate = errors.New("failed to create PREPARED certficate")
 )
 
-// newRoundState creates a new roundState instance with the given view and validatorSet
-func newRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, pendingRequest *istanbul.Request, preparedCertificate istanbul.PreparedCertificate, parentCommits MessageSet) RoundState {
+func newRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet) RoundState {
 	return &roundStateImpl{
 		state:        StateAcceptRequest,
 		round:        view.Round,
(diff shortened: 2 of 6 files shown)


consensus/istanbul/core/roundstate.go
```diff
@@ -177,6 +182,37 @@ func (s *roundStateImpl) Round() *big.Int {
...
 
+func (s *roundStateImpl) changeRound(nextRound *big.Int, validatorSet istanbul.ValidatorSet) {
+	s.state = StateAcceptRequest
+	s.round = nextRound
+	s.desiredRound = nextRound
+
+	s.prepares = newMessageSet(validatorSet)
+	s.commits = newMessageSet(validatorSet)
+
+	// ??
```
nambrot: ?

consensus/istanbul/core/core.go
```diff
@@ -376,26 +376,28 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
+	} else if roundChange {
+		c.current.StartNewRound(view.Round, validatorSet)
+	} else {
+		// sequence change
+
+		var newParentCommits MessageSet
+		lastSubject, err := c.backend.LastSubject()
+		if err != nil && c.current.Proposal() != nil && c.current.Proposal().Hash() == lastSubject.Digest && c.current.Round().Cmp(lastSubject.View.Round) == 0 {
+			// When changing sequences, if our current Commit messages match the latest block in the chain
+			// (i.e. they're for the same block hash and round), we use this sequence's commits as the ParentCommits field
```
asaj: nit: "this round's commits"

consensus/istanbul/core/core.go
```diff
@@ -376,26 +376,28 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
-			}
-		}
-	} else {
+	if c.current == nil {
 		// When the current round is nil, we must start with the current validator set in the parent commits
 		// either `validatorSet` or `backend.Validators(lastProposal)` works here
-		c.current = newRoundState(view, validatorSet, nil, istanbul.EmptyPreparedCertificate(), newMessageSet(validatorSet))
+		c.current = newRoundState(view, validatorSet)
+	} else if roundChange {
+		c.current.StartNewRound(view.Round, validatorSet)
```
nambrot: Previously, we would have passed `c.current.PendingRequest()`  to the round state, but it now looks like we always set it to nil after a round change?

consensus/istanbul/core/roundstate.go
```diff
@@ -177,6 +182,37 @@ func (s *roundStateImpl) Round() *big.Int {
 	return s.round
 }
 
+func (s *roundStateImpl) changeRound(nextRound *big.Int, validatorSet istanbul.ValidatorSet) {
+	s.state = StateAcceptRequest
+	s.round = nextRound
+	s.desiredRound = nextRound
```
asaj: Are you sure it's correct to overwrite `desiredRound` here?

consensus/istanbul/core/roundstate.go
asaj: Are you sure this is correct?

consensus/istanbul/core/roundstate.go
mcortesi: `StartNewRound()` and `StartNewSequence` are replacements for `c.current := newRoundState()`. So basically in this impl i'm making sure i do the same things the constructor did.

Wether that is correct or not, not sure. But the impl should be equal.

There's a TODO from @trianglesphere about keeping the desired round on new rounds.

consensus/istanbul/core/core.go
mcortesi: before we would be overriding c.current with a new instance. now we mutate it, changing only the necessary fields.

So `StartNewRound` should behave exactly as calling the old `newRoundState` with the old parameters. Since the only new information is the view and validatorSet, the other parameters are not needs, and are not changed on StartNewRound

consensus/istanbul/core/roundstate.go
mcortesi: it seems strange, but it's the same behaviour we had before

consensus/istanbul/core/core.go
```diff
@@ -376,26 +376,28 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
-				c.current = newRoundState(view, validatorSet, nil, istanbul.EmptyPreparedCertificate(), newMessageSet(c.backend.ParentBlockValidators(headBlock)))
-			}
-		}
-	} else {
+	if c.current == nil {
 		// When the current round is nil, we must start with the current validator set in the parent commits
 		// either `validatorSet` or `backend.Validators(lastProposal)` works here
-		c.current = newRoundState(view, validatorSet, nil, istanbul.EmptyPreparedCertificate(), newMessageSet(validatorSet))
+		c.current = newRoundState(view, validatorSet)
+	} else if roundChange {
```
kevjue: When there is a round change or sequence change, the `current` variable gets mutated.  Originally, a new one was created.

You could argue that there should be a new one for every round (it is a `roundState` instead of a `sequenceState`).  Also, I feel it may introduce more bugs that we aren't aware of with a change that is this significant.  WDYT?

mcortesi: I agree that is a change that should be done/reviewed carefully. Still i think it's for the better; and if we can/want to do it, we should do it sooner (before the audit) than later. If not, it will never be a good time to clean up this code.

There's actual a reason for the change, besides cleaning up. I wanted to leave all mutation to `c.current` behind a single interface (now `RoundState`) so i could easily intercept all those mutations and trigger roundState persistence just after the mutation. Of course, you can try do the same, but then you would need to make `Core` persistence aware, and you will now 2 different places where mutations happens

 1. When you create a new RoundState (eg `c.current = ...`)
 2. When you mutate the current one (eg `c.current.SomeMutationMethod()`)

Now for the actual review of the change, a few things that i have considered:

* `c.current` is only ever used inside core, and core event handling is single threaded (as we can see in `Start()`. 
* In fact there was not lock when overwriting `c.current`. It makes sense since there's only one user, and that user uses a single thread.
* There were 3 cases to set `c.current`:
     * When it was unset (only on `Start()`)
     * When we are on `roundChange`
     * When we are at sequence change
* In all these situations, we called `newRoundState()` and decide which fields we wanted to leave with the old `c.current` information. What i did was to verify that the resulting fields after calling each `newRoundChange` on all 3 different scenarios, matches with all the resulting fields after replacing those with methods. That of course, it will be good to have a second pairs of eyes. But; i believe that's the only important consideration we should do, since having the same or a different instance of RoundState doesn't generate a change in behaviour from my review.

In general, I agree with you that we should be careful about this kind of changes. But; i don't think it should stop us from doing them, since that would deny us the possibility of refactoring for the better (if we agree it reads/follows better). And it's better to do those sooner than later.

Again, i might be mistaken; and this generates a change in behaviour that i didn't forsee. Do you have a concrete idea of how that could be happening?

kevjue: My response is inline:

> I agree that is a change that should be done/reviewed carefully. Still i think it's for the better; and if we can/want to do it, we should do it sooner (before the audit) than later. If not, it will never be a good time to clean up this code.
> 
> There's actual a reason for the change, besides cleaning up. I wanted to leave all mutation to `c.current` behind a single interface (now `RoundState`) so i could easily intercept all those mutations and trigger roundState persistence just after the mutation. Of course, you can try do the same, but then you would need to make `Core` persistence aware, and you will now 2 different places where mutations happens
> 
> 1. When you create a new RoundState (eg `c.current = ...`)
> 2. When you mutate the current one (eg `c.current.SomeMutationMethod()`)
> 

Keep in mind that other mutations could be happening to the state.  E.g. the `prepares` message set can at added to when a `prepare` message is received.  And I think that we would need to persist that mutation as well.

> Now for the actual review of the change, a few things that i have considered:
> 
> * `c.current` is only ever used inside core, and core event handling is single threaded (as we can see in `Start()`.
> * In fact there was not lock when overwriting `c.current`. It makes sense since there's only one user, and that user uses a single thread.
> * There were 3 cases to set `c.current`:
>   
>   * When it was unset (only on `Start()`)
>   * When we are on `roundChange`
>   * When we are at sequence change
> * In all these situations, we called `newRoundState()` and decide which fields we wanted to leave with the old `c.current` information. What i did was to verify that the resulting fields after calling each `newRoundChange` on all 3 different scenarios, matches with all the resulting fields after replacing those with methods. That of course, it will be good to have a second pairs of eyes. But; i believe that's the only important consideration we should do, since having the same or a different instance of RoundState doesn't generate a change in behaviour from my review.
> 
> In general, I agree with you that we should be careful about this kind of changes. But; i don't think it should stop us from doing them, since that would deny us the possibility of refactoring for the better (if we agree it reads/follows better). And it's better to do those sooner than later.
> 
> Again, i might be mistaken; and this generates a change in behaviour that i didn't forsee. Do you have a concrete idea of how that could be happening?

Thoughts on renaming it to `SequenceState`?  If there is only one instance, then seems like that's more accurate.

kevjue:
As discussed over slack, I think that after all your changes to round state, it would be great to add unit tests to check that invariants are not broken during any round state mutation.

PR #628, opened 2019-11-21
Move Proposer outside ValidatorSet

mcortesi:
### Description

Now ValidatorSet is inmutable, meaning that remains the same for the
whole epoch. While, before ValidatorSet changed every roundChange since
it had the proposer.

Also, by doing this, we eliminate a way of changing data inside the
RoundState without calling any RoundState function.

RounState has a MessageSet which has a ValidatorSet

And on some scenarios the ValidatorSet proposer was being changed
without notifying the RoundState

### Tested

unit

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -26,6 +26,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
+	"github.com/ethereum/go-ethereum/consensus/istanbul/validator"
 
 	"github.com/ethereum/go-ethereum/core/types"
 	blscrypto "github.com/ethereum/go-ethereum/crypto/bls"
@@ -39,8 +40,9 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 	c := &core{
 		config:             config,
 		address:            backend.Address(),
-		handlerWg:          new(sync.WaitGroup),
 		logger:             log.New("address", backend.Address()),
+		selectProposer:     validator.GetProposerSelector(config.ProposerPolicy),
+		handlerWg:          new(sync.WaitGroup),
 		backend:            backend,
 		backlogs:           make(map[istanbul.Validator]*prque.Prque),
 		backlogsMu:         new(sync.Mutex),
@@ -58,9 +60,10 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 // ----------------------------------------------------------------------------
 
 type core struct {
-	config  *istanbul.Config
-	address common.Address
-	logger  log.Logger
+	config         *istanbul.Config
+	address        common.Address
+	logger         log.Logger
+	selectProposer istanbul.ProposerSelector
 
 	backend               istanbul.Backend
 	events                *event.TypeMuxSubscription
@@ -68,7 +71,6 @@ type core struct {
 	timeoutSub            *event.TypeMuxSubscription
 	futurePreprepareTimer *time.Timer
 
-	valSet     istanbul.ValidatorSet
 	validateFn func([]byte, []byte) (common.Address, error)
 
 	backlogs   map[istanbul.Validator]*prque.Prque
@@ -140,7 +142,8 @@ func (c *core) broadcast(msg *istanbul.Message) {
 	}
 
 	// Broadcast payload
-	if err := c.backend.BroadcastConsensusMsg(istanbul.GetAddressesFromValidatorList(c.valSet.FilteredList()), payload); err != nil {
+	validators := istanbul.GetAddressesFromValidatorList(c.current.ValidatorSet().FilteredList())
+	if err := c.backend.BroadcastConsensusMsg(validators, payload); err != nil {
 		logger.Error("Failed to broadcast message", "msg", msg, "err", err)
 		return
 	}
@@ -240,7 +243,7 @@ func UnionOfSeals(aggregatedSignature types.IstanbulAggregatedSeal, seals Messag
 
 // Generates the next preprepare request and associated round change certificate
 func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbul.Request, istanbul.RoundChangeCertificate, error) {
-	roundChangeCertificate, err := c.roundChangeSet.getCertificate(round, c.valSet.MinQuorumSize())
+	roundChangeCertificate, err := c.roundChangeSet.getCertificate(round, c.current.ValidatorSet().MinQuorumSize())
 	if err != nil {
 		return &istanbul.Request{}, istanbul.RoundChangeCertificate{}, err
 	}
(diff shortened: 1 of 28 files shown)


consensus/istanbul/core/core.go
```diff
@@ -304,6 +308,12 @@ func (c *core) startNewRound(round *big.Int) {
 	var newView *istanbul.View
 	var roundChangeCertificate istanbul.RoundChangeCertificate
 	var request *istanbul.Request
+
+	var valSet istanbul.ValidatorSet
```
kevjue: How does the proposer get selected  if `valSet` is nil?  (e.g. when `c.current == nil`).

consensus/istanbul/core/core.go
```diff
@@ -304,6 +308,12 @@ func (c *core) startNewRound(round *big.Int) {
...
+	if c.current != nil {
```
kevjue: When is `c.current` == nil?

consensus/istanbul/core/core.go
```diff
@@ -342,7 +356,7 @@ func (c *core) startNewRound(round *big.Int) {
 	}
 	c.newRoundChangeTimer()
 
-	logger.Debug("New round", "new_round", newView.Round, "new_seq", newView.Sequence, "new_proposer", c.valSet.GetProposer(), "valSet", c.valSet.List(), "size", c.valSet.Size(), "isProposer", c.isProposer())
+	logger.Debug("New round", "new_round", newView.Round, "new_seq", newView.Sequence, "new_proposer", c.current.Proposer(), "valSet", c.current.ValidatorSet().List(), "size", c.current.ValidatorSet().Size(), "isProposer", c.isProposer())
```
kevjue: Can `c.current` be nil at this point?  If so, there would be a seg fault as it's methods are used multiple points in the log message.

consensus/istanbul/core/preprepare.go
```diff
@@ -86,13 +86,14 @@ func (c *core) handlePreprepare(msg *istanbul.Message) error {
 	if err := c.checkMessage(istanbul.MsgPreprepare, preprepare.View); err != nil {
 		if err == errOldMessage {
 			// Get validator set for the given proposal
-			valSet := c.backend.ParentBlockValidators(preprepare.Proposal).Copy()
-			previousProposer := c.backend.AuthorForBlock(preprepare.Proposal.Number().Uint64() - 1)
-			valSet.CalcProposer(previousProposer, preprepare.View.Round.Uint64())
+			valSet := c.backend.ParentBlockValidators(preprepare.Proposal)
```
kevjue: Why remove the `Copy`?

consensus/istanbul/core/roundstate.go
```diff
@@ -182,30 +220,34 @@ func (s *roundStateImpl) Round() *big.Int {
...
 
-func (s *roundStateImpl) StartNewSequence(view *istanbul.View, validatorSet istanbul.ValidatorSet, parentCommits MessageSet) {
+func (s *roundStateImpl) StartNewSequence(view *istanbul.View, validatorSet istanbul.ValidatorSet, nextProposer istanbul.Validator, parentCommits MessageSet) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	s.changeRound(view.Round, validatorSet)
+	s.validatorSet = validatorSet
+
+	s.changeRound(view.Round, validatorSet, nextProposer)
```
kevjue: Should `view.Round` always be 0 here?  If so, perhaps checking for that here and log and error and return an error if it's not.

consensus/istanbul/validator.go
```diff
@@ -85,16 +85,10 @@ type Validators []Validator
...
-	CalcProposer(lastProposer common.Address, round uint64)
-	// Get current proposer
-	GetProposer() Validator
-	// Check whether the validator with given address is the current proposer
-	IsProposer(address common.Address) bool
-	// Policy by which this selector chooses proposers
-	Policy() ProposerPolicy
 	// Sets the randomness for use in the proposer policy
 	SetRandomness(seed common.Hash)
+	// Sets the randomness for use in the proposer policy
```
kevjue: Is there a specific reason that `randomness` is left within the valset?    Should it be within round state?

consensus/istanbul/core/core.go
mcortesi: The first time we call `startNewRound()` which happens inside `Start()`. Actually this happens before even we start listening to events. 

After that `c.current` is never nil (until we call `Stop()`

consensus/istanbul/core/core.go
mcortesi: The code is hard to follow since i didn't want to refactor this method, and it's quite a mess...

But is as follows:

1. When c.current is nil, then we enter on the first if (line 280); thus `roundChange = false`
2. Since roundChange is false, we enter the `else` on line 329, and there on line 337 valSet is set

consensus/istanbul/core/core.go
mcortesi: No because on line 348 on this same method we call `resetRoundState` that set `c.current`

consensus/istanbul/core/preprepare.go
mcortesi: Copy was there because `CalcProposer` mutates the valSet. Now a valSet is immutable, so we don't need to copy it

consensus/istanbul/validator.go
mcortesi: tbh, it was easier this way. `randomness` is set on `getOrderedValidators()` but not everytime we get a `ValidatorSet`. To change it, involved a larger refactor

One can always argue that valSet and randomness are constant within an epoch, and can be contained in the same object. But surely that object should not be called ValidatorSet.

consensus/istanbul/core/roundstate.go
mcortesi: It is always 0, as you can see here: https://github.com/celo-org/celo-blockchain/blob/83b6a40c2526803d64d75db4e8e1d02938453104/consensus/istanbul/core/core.go#L335

Again, the problem is that `startNewRound` is a very confusing function, that should be refactor. But even if i was tempted, i felt it wouldn't get me much love from my reviewers... so i left that for a later PR ;)

kevjue: As we discussed over slack, the plan is to modify `StartNewSequence` to accept a sequence number instead of a view number, and to modify `resetRoundState` to check that the round number is zero here:  https://github.com/celo-org/celo-blockchain/pull/628/files/83b6a40c2526803d64d75db4e8e1d02938453104#diff-7d85e6de59ad4e84300834cf03eb41d1R400

consensus/istanbul/validator.go
kevjue: Can you include a comment mentioning that randomness is injected into the valset.

Also, I think your 2nd paragraph in the response makes a lot of sense and would be really helpful if you add a comment in the validatorset struct that states the same.

mcortesi: done

consensus/istanbul/core/roundstate.go
mcortesi: done

PR #632, opened 2019-11-22
Persist RoundState on every transition

mcortesi:
### Description

Creates a RoundStatePersistence object that will save state to levelDB on each change.

- Now all transition method returns errors on are handled by clients
- RLP encoding for some missing parts

**TODO** RLP encoding for MessageSet. Not sure how to do it for a `map`. @kevjue can you help?

### Tested

ci/test

### DIFF
--- a/consensus/istanbul/config.go
+++ b/consensus/istanbul/config.go
@@ -36,6 +36,7 @@ type Config struct {
 	Epoch                uint64         `toml:",omitempty"` // The number of blocks after which to checkpoint and reset the pending votes
 	LookbackWindow       uint64         `toml:",omitempty"` // The window of blocks in which a validator is forgived from voting
 	ValidatorEnodeDBPath string         `toml:",omitempty"` // The location for the validator enodes DB
+	RoundStateDBPath     string         `toml:",omitempty"` // The location for the round states DB
 
 	// Proxy Configs
 	Proxy                   bool           `toml:",omitempty"` // Specifies if this node is a proxy
@@ -54,6 +55,7 @@ var DefaultConfig = &Config{
 	Epoch:                30000,
 	LookbackWindow:       12,
 	ValidatorEnodeDBPath: "validatorenodes",
+	RoundStateDBPath:     "roundstates",
 	Proxy:                false,
 	Proxied:              false,
 }
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -149,8 +149,11 @@ func (c *core) broadcast(msg *istanbul.Message) {
 	}
 }
 
-func (c *core) commit() {
-	c.current.TransitionToCommited()
+func (c *core) commit() error {
+	err := c.current.TransitionToCommited()
+	if err != nil {
+		return err
+	}
 
 	// Process Backlog Messages
 	c.processBacklog()
@@ -160,13 +163,14 @@ func (c *core) commit() {
 		aggregatedSeal, err := GetAggregatedSeal(c.current.Commits(), c.current.Round())
 		if err != nil {
 			c.sendNextRoundChange()
-			return
+			return nil
 		}
 		if err := c.backend.Commit(proposal, aggregatedSeal); err != nil {
 			c.sendNextRoundChange()
-			return
+			return nil
 		}
 	}
+	return nil
 }
 
 // GetAggregatedSeal aggregates all the given seals for a given message set to a bls aggregated
@@ -270,16 +274,14 @@ func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbu
 }
 
 // startNewRound starts a new round. if round equals to 0, it means to starts a new sequence
-func (c *core) startNewRound(round *big.Int) {
+func (c *core) startNewRound(round *big.Int) error {
 	logger := c.newLogger("func", "startNewRound", "tag", "stateTransition")
 
 	roundChange := false
 	// Try to get last proposal
 	headBlock, headAuthor := c.backend.GetCurrentHeadBlockAndAuthor()
 
-	if c.current == nil {
-		logger.Trace("Start the initial round")
-	} else if headBlock.Number().Cmp(c.current.Sequence()) >= 0 {
+	if headBlock.Number().Cmp(c.current.Sequence()) >= 0 {
 		// Want to be working on the block 1 beyond the last committed block.
 		diff := new(big.Int).Sub(headBlock.Number(), c.current.Sequence())
 		c.sequenceMeter.Mark(new(big.Int).Add(diff, common.Big1).Int64())
@@ -293,15 +295,15 @@ func (c *core) startNewRound(round *big.Int) {
 		// Working on the block immediately after the last committed block.
 		if round.Cmp(c.current.Round()) == 0 {
 			logger.Trace("Already in the desired round.")
-			return
+			return nil
 		} else if round.Cmp(c.current.Round()) < 0 {
 			logger.Warn("New round should not be smaller than current round", "lastBlockNumber", headBlock.Number().Int64(), "new_round", round)
-			return
+			return nil
 		}
 		roundChange = true
 	} else {
 		logger.Warn("New sequence should be larger than current sequence", "new_seq", headBlock.Number().Int64())
-		return
+		return nil
 	}
 
 	// Generate next view and pre-prepare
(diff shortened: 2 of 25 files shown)


consensus/istanbul/core/core.go
```diff
@@ -387,33 +388,30 @@ func (c *core) waitForDesiredRound(r *big.Int) {
...
-	} else if roundChange {
-		c.current.StartNewRound(view.Round, validatorSet, nextProposer)
+	if roundChange {
+		return c.current.StartNewRound(view.Round, validatorSet, nextProposer)
+	}
+
+	// sequence change
+	var newParentCommits MessageSet
+	lastSubject, err := c.backend.LastSubject()
+	if err != nil && c.current.Proposal() != nil && c.current.Proposal().Hash() == lastSubject.Digest && c.current.Round().Cmp(lastSubject.View.Round) == 0 {
```
asaj: if `err == nil`...

consensus/istanbul/core/handler.go
```diff
@@ -32,8 +32,23 @@ func (c *core) ParentCommits() MessageSet {
 
 // Start implements core.Engine.Start
 func (c *core) Start() error {
-	// Start a new round from last sequence + 1
-	c.startNewRound(common.Big0)
+	headBlock, headAuthor := c.backend.GetCurrentHeadBlockAndAuthor()
```
asaj: Does it matter that we may not be synced by the time we call `Start()`?

consensus/istanbul/core/handler.go
```diff
@@ -32,8 +32,23 @@ func (c *core) ParentCommits() MessageSet {
...
+	valSet := c.backend.Validators(headBlock)
+
+	nextBlock := new(big.Int).Add(headBlock.Number(), common.Big1)
+	nextProposer := c.selectProposer(valSet, headAuthor, 0)
```
asaj: What if we shut down in the middle sequence N, and restart when the network is still on sequence N? Wouldn't we want to load our latest round state?

consensus/istanbul/core/roundstates/CURRENT
```diff
@@ -0,0 +1 @@
+MANIFEST-000000
```
asaj: Add these files to .gitignore

consensus/istanbul/core/core.go
mcortesi: I didn't change this line, only the indent changed:

The original is `err != nil`
```
if err != nil && c.current.Proposal() != nil && c.current.Proposal().Hash() == lastSubject.Digest && c.current.Round().Cmp(lastSubject.View.Round) == 0 {
```

consensus/istanbul/core/handler.go
mcortesi: I haven't change the behavior.

Before we called `startNewRound(common.Big0)`. Now, instead, i copied all the significant lines that would have been executed on that precise scenario.

So, not sure how IBFT is supposed to work in this scenario.

consensus/istanbul/core/handler.go
mcortesi: On line 40, we call `newRoundStateWithPersistence()` which will check which is the last saved View. If the stored view >= the computed view here, we will ignore all the parameters and use the stored version.

So, in essence calling `newRoundStateWithPersistence()` is like saying "try restoring a newer version of roundState and if not, create a new one with this parameters"

consensus/istanbul/core/core.go
asaj: The original is wrong 😄

consensus/istanbul/core/roundstate_persist.go
```diff
@@ -0,0 +1,352 @@
...
+		if err == leveldb.ErrNotFound {
+			logger.Info("Creating new RoundState", "reason", "No storedView found")
+		} else {
+			logger.Info("Creating new RoundState", "reason", "old view", "stored_view", lastStoredView, "requested_view", view)
+		}
+		return &roundStatePersistence{
+			db:       db,
+			delegate: newRoundState(view, validatorSet, proposer),
+		}, nil
+	}
```
kevjue: nit:  I think that having an `else {` here will be more readable.

consensus/istanbul/core/roundstate_persist.go
```diff
@@ -0,0 +1,352 @@
...
+
+	lastStoredView, err := getLastStoredView(db)
+
+	if err != nil && err != leveldb.ErrNotFound {
+		logger.Error("Failed to fetch lastStoredView", "err", err)
+		db.Close()
+		return nil, err
+	}
+
+	if err == leveldb.ErrNotFound || lastStoredView.Cmp(view) < 0 {
```
kevjue: I think adding a comment here stating that this function will create a new round state object if the last stored view is old will be helpful.

consensus/istanbul/core/roundstate_persist.go
```diff
@@ -0,0 +1,351 @@
...
+func (rsp *roundStatePersistence) PendingRequest() *istanbul.Request {
+	return rsp.delegate.PendingRequest()
+}
+
+// PreparedCertificate implements RoundState.PreparedCertificate
+func (rsp *roundStatePersistence) PreparedCertificate() istanbul.PreparedCertificate {
+	return rsp.delegate.PreparedCertificate()
+}
+
+func (rsp *roundStatePersistence) storeRoundState() error {
```
kevjue: I think that adding a comment stating that the reason for storing at every view be helpful (e.g. saving evidence to show that a validator did a "valid" double signing).

consensus/istanbul/core/roundstate_persist.go
```diff
@@ -0,0 +1,351 @@
...
+			logger.Info("Creating new RoundState", "reason", "old view", "stored_view", lastStoredView, "requested_view", view)
+		}
+		return &roundStatePersistence{
+			db:       db,
+			delegate: newRoundState(view, validatorSet, proposer),
+		}, nil
+	}
+
+	logger.Info("Retrieving stored RoundState", "stored_view", lastStoredView, "requested_view", view)
+	lastStoredRoundState, err := getStoredRoundState(db, lastStoredView)
```
kevjue: `getLastStoredView` and `getStoredRoundState` are called separately.  Is there a chance of a race condition?   It seems that writing these objects is done atomically.

consensus/istanbul/config.go
```diff
@@ -54,6 +55,7 @@ var DefaultConfig = &Config{
 	Epoch:                30000,
 	LookbackWindow:       12,
 	ValidatorEnodeDBPath: "validatorenodes",
+	RoundStateDBPath:     "roundstates",
```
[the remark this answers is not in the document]
mcortesi: mmm... any name works for me, but to be clear, there are many roundstates... or we save the roundstate (with a differente key) with each view change

consensus/istanbul/core/roundstate_persist.go
mcortesi: I sometimes thinks the same, but I'm trying to follow go code guidelines (the linter complains)

Basically, if the `if` has a `return` the `else` is omitted

https://golang.org/doc/effective_go.html#if
> In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

consensus/istanbul/core/roundstate_persist.go
mcortesi: Added a comment and renamed the function to `createOrRestoreRoundState()` so that the name makes it clearer

consensus/istanbul/core/roundstate_persist.go
mcortesi: I don't think so. Since:

1. You can only have only one active levelDB instance on the process. So no 2 roundStatePersistence can exist at the same time
2. Read happens only once, at the beginning, before there's even a RoundState interface; so no store is possible at that time

consensus/istanbul/core/handler.go
mcortesi: I've changed to:

```go
	roundState, err := c.createRoundState()
	if err != nil {
		return err
	}

	c.current = roundState
	c.roundChangeSet = newRoundChangeSet(c.current.ValidatorSet())
```

Also, fixed a bug where i was using a maybe old valSet to create the `RoundChangeSet` :D

consensus/istanbul/config.go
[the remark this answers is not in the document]
kevjue: I agree with the original name.  At the time of my comment, I didn't realize it was storing historical round states.

PR #639, opened 2019-11-24
Always run the election after block finalization

asaj:
### Description

This PR fixes a discrepancy between block proposal creation and block proposal verification. In the former, the block was finalized after the election was run, while in the latter, the block was finalized before the election was run. When block finalization affects the results of the election (as it sometimes will via rewards to voters), this would lead to proposals that would fail to be verified.
  
### Tested

By running the end-to-end tests with 10 validators multiple times.

### Other changes

- None

### Backwards compatibility

Not backwards compatible


asaj:
End-to-end tests are broken in the monorepo due to https://github.com/celo-org/celo-monorepo/pull/1838, merging anyway to unblock the integration deployment.

PR #640, opened 2019-11-24
Separate public key and signature in proof-of-possession command

asaj:
### Description

Needed now that these are supplied separately when registering a validator.

### Tested

- Ran `account new` and `account proof-of-possession`.


asaj:
End to end tests are broken in monorepo, merging as this should not affect them.

PR #645, opened 2019-11-25
Move init/close valenodedb to Engine start/stop

mcortesi:
### Description

To avoid light clients and others create an unnecessary level db table.

### Tested

e2e tests


mcortesi:
We double checked on this and it's not a good solution (doesn't work for proxy)

Issue #607, opened 2019-11-18
val enode table manager SNBAT explicitly manage validator connections


kevjue:
### Expected Behavior

The management of the val enode table and the validator connections should be decoupled.  

To that end, @mcortesi  came up with a good idea in that there is a separate goroutine in the istanbul package that will take new announce/val_enode_share updates and will do the necessary changes to the val_enode_table and the validator connections.  This goroutine will also do periodic consistency checks between the val enode table and the validator connections.

### Current Behavior

The val_enode_table manager effectively manages the validator connections.

nategraf:
@kevjue is this issue still relevant?

kevjue:
> @kevjue is this issue still relevant?

Unfortunately yes.  Didn't get time to address this under the push to get RC1 done.

PR #665, opened 2019-11-27
create validator connections on upsert

### LINKED ISSUE #662: Proxy SBAT always establish validator connections on reboot
### Expected Behavior

Proxy SBAT establish a validator connection on reboot.

### Current Behavior

Proxy doesn't establish a validator connection on reboot, if the enodeURLs are already in the level db.

kevjue:
### Description

This is a temporary fix to handle the case when a proxy or validator starts up and it's val_enode_table is persisted.  It originally wouldn't establish validator connections because the announce/val_enode_share entries are already on disk.  

This change will establish those val connections even if the entries are already on disk.

The more permanent fix is to have a separate validator conn manager goroutine that will monitor the val connections with respect to the val_enode_table (https://github.com/celo-org/celo-blockchain/issues/607).

### Tested

Unit tests

### Other changes

None

### Related issues

- Fixes #662

### Backwards compatibility

Is backwards compatible.

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -700,7 +700,7 @@ func (sb *Backend) RefreshValPeers(valset istanbul.ValidatorSet) {
 		return
 	}
 
-	sb.valEnodeTable.RefreshValPeers(valset, sb.Address())
+	sb.valEnodeTable.RefreshValPeers(valset, sb.ValidatorAddress())
 }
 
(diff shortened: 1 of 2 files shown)


asaj:
Is calling `AddPeer` on an already connected peer a no-op?

kevjue:
> Is calling `AddPeer` on an already connected peer a no-op?

Yes, it is.

PR #667, opened 2019-11-28
Fix non-fatal panic on forwarding transactions without gateway fee

nategraf:
### Description

If a light client is connected to a full-node who does not specify an etherbase, a panic will prevent the light client from forwarding them transactions. This panic is nonfatal, as it is recovered from in a resonable place.

### Tested

Tested with a local testnet where no etherbase is specified on a full-node and transactions are forwarded.

### Backwards compatibility

Fully backwards compatible and only the light client needs to be changed.

### DIFF
--- a/les/peer.go
+++ b/les/peer.go
@@ -607,17 +607,20 @@ func (ps *peerSet) randomPeerEtherbase() common.Address {
 	return r
 }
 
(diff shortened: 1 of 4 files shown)


kevjue:
Agreed!  Nice!

PR #668, opened 2019-11-28
Print public key in ECDSA account proof-of-possession

asaj:
### Description

This is needed so that validators can register with validator signers. Accompanying monorepo PR:
https://github.com/celo-org/celo-monorepo/pull/1952

### Tested

- Used to register a validator on baklavastaging.

_Brief explanation of why these changes are/are not backwards compatible. Feel free to remove for trivial, backwards compatible changes._

### DIFF
--- a/cmd/geth/accountcmd.go
+++ b/cmd/geth/accountcmd.go
@@ -240,7 +240,9 @@ func accountProofOfPossession(ctx *cli.Context) error {
 	var key []byte
 	var pop []byte
 	var err error
+	keyType := "ECDSA"
 	if ctx.IsSet(blsFlag.Name) {
+		keyType = "BLS"
 		key, pop, err = ks.GenerateProofOfPossessionBLS(account, message)
 	} else {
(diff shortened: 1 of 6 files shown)


cmd/geth/accountcmd.go
```diff
@@ -237,19 +237,18 @@ func accountProofOfPossession(ctx *cli.Context) error {
...
-		if err != nil {
-			return err
-		}
-		fmt.Printf("Account {%x}:\n  Signature: %s\n", account.Address, hex.EncodeToString(pop))
+		key, pop, err = ks.GenerateProofOfPossession(account, message)
+	}
+	if err != nil {
+		return err
 	}
+	fmt.Printf("Account {%x}:\n  Signature: %s\n  Public Key: %s\n", account.Address, hex.EncodeToString(pop), hex.EncodeToString(key))
```
timmoreton: useful to distinguish BLS vs ECDSA in the output ?

PR #669, opened 2019-11-28
 Only reward voters for each group that elected validators once

asaj:
### Description

Fixes a small bug in which we would push the group to a list even if we had already.


contract_comm/election/election.go
```diff
@@ -135,7 +135,7 @@ func getTotalVotesForEligibleValidatorGroups(header *types.Header, state vm.Stat
 	var values []*big.Int
 	_, err := contract_comm.MakeStaticCall(params.ElectionRegistryId, electionABI, "getTotalVotesForEligibleValidatorGroups", []interface{}{}, &[]interface{}{&groups, &values}, params.MaxGasForGetEligibleValidatorGroupsVoteTotals, header, state)
 	if err != nil {
-		log.Error("Error calling getTotalVotesForEligibleValidatorGroups", "err", err)
+		return nil, nil
```
tkporter: Should this return `nil, err`?

PR #670, opened 2019-11-28
Wait T before committing new work so that the round is updated

### LINKED ISSUE #649: Race condition when creating union of seals
Occasionally `miner/worker.go` will call `Backend.Prepare()` before `core.updateRoundState` is called. When this happens, the `roundState` is one sequence out of date, and the `ParentCommits` that we use to make a union of seals will not match the `ParentAggregatedSeal` that we pull from the most recent block.

This will only happen when there was a round change in the previous sequence. Otherwise, the `Sleep` in `Backend.Prepare` prevents this from happening.

```
INFO [11-26|05:10:25.695] Check round change message failed        address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a state="Waiting for new round" from=0xDC2942F83A049239F257a11168eDb5e7d1A64CB3 cur_round=0 cur_seq=15151 func=handleRoundChange              tag=handleMsg       err="old message"
INFO [11-26|05:10:25.722] Check round change message failed        address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a state="Waiting for new round" from=0x8dB714d5f45F28e20CEcF01171A466fc440f9Cb0 cur_round=0 cur_seq=15151 func=handleRoundChange              tag=handleMsg       err="old message"
INFO [11-26|05:10:25.805] Got pre-prepare message                  address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a func=handlePreprepare               tag=handleMsg       from=0x1303C42C35fa8b716Aa3DFCCF71aE54a7B1Aa61b cur_seq=15151 cur_round=0 state="Waiting for new round" address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a msg="{Code: 0, Address: 0x1303C42C35fa8b716Aa3DFCCF71aE54a7B1Aa61b}"
WARN [11-26|05:10:25.806] Invalid round change certificate with preprepare. address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a func=handlePreprepare               tag=handleMsg       from=0x1303C42C35fa8b716Aa3DFCCF71aE54a7B1Aa61b cur_seq=15151 cur_round=0 state="Waiting for new round" address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a err="message in ROUND CHANGE certificate for wrong view"
INFO [11-26|05:10:26.017] Check round change message failed        address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a state="Waiting for new round" from=0x2d25F0740FCae0839855a2803093E1036a88a980 cur_round=0 cur_seq=15151 func=handleRoundChange              tag=handleMsg       err="old message"
INFO [11-26|05:10:26.972] Imported new chain segment               blocks=1  txs=10  mgas=0.294  elapsed=36.734ms  mgasps=7.993  number=15151 hash=529767…681469 cache=0.00B
INFO [11-26|05:10:26.974] Combining additional seals with the parent aggregated seal func=Backend.Prepare()              additionalSeals="[<21> 0xf087eFe81554B63e7B23B720c3A9F28a1770B621, 0x12db4bD12CCeC62aDCbf946553742D17ff1E680D, 0x9D579002017cf1F26Cf75EC1184e0D683595E339, 0x05622C4aE18Ac192A55d9457F44cC38e29Bc981b, 0x6074000342722d5A83c5cacB3cCabd62863efbf2, 0x2d25F0740FCae0839855a2803093E1036a88a980, 0x01abBfC23D6eC84B94F236fBE96e4786e0802E57, 0xcDC18A1D6CE330Fa76C15760bEb4568c8aF2201A, 0x34B7BAE0C0b48F5674d19bb716105de215CBF1D2, 0x64068Dce7C24C45BF6510FEF00c6a3541C66Cc10, 0x6377fAf01648e8f8E4f9D30712795847095DB0c6, 0x8dB714d5f45F28e20CEcF01171A466fc440f9Cb0, 0x3c584B5657978c377db0a0805E4B9e6Fb2530B0C, 0xeEae3C863D02Ad111FC453e9eC14F00154436d3f, 0x67850188C88709f3066356E328aeF30ea93f875B, 0xe2E2E7A2a108aAFe9BeF4A697316828E2A31eD50, 0x4e86AE526d9F4495722d083d10CB1B73aeb283Ea, 0x7059f5038a97CB70f37c73FaDAE11B83Dca982A8, 0xDC2942F83A049239F257a11168eDb5e7d1A64CB3, 0x7FA5081E107a8fa979aA080Ba0A88575b9EccCb9, 0x1303C42C35fa8b716Aa3DFCCF71aE54a7B1Aa61b]" num=15152 parentAggregatedSeal="{round: 1, bitmap: 1010010001001000001101110011100110111111101111111111111111111111111111, signature: 681abd8a4e2476621dd378b269d5cce489ce8b0787d401ff980f7020d329d50ed01a1ace9e65897523e2c5cf6510a601}"
INFO [11-26|05:10:26.975] Discarding commit messages and not using as parent commits address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a func=updateRoundState               tag=stateTransition cur_seq=15151 cur_round=0 state="Waiting for new round" address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a err=nil
INFO [11-26|05:10:26.976] No current proposal                      address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a func=updateRoundState               tag=stateTransition cur_seq=15151 cur_round=0 state="Waiting for new round" address=0x5ABc094B0498f92f43fCf0Ce45b877733fbc811a
ERROR[11-26|05:10:27.015] Unable to verify aggregated signature    func=Backend.verifyAggregatedSeal() err="Not verified"
ERROR[11-26|05:10:27.015] Failed to combine additional seals with parent aggregated seal. func=Backend.Prepare()              err="invalid signature"
```

From our Slack:

> asa 7:30 AM
> It seems to always happen after round changes
> I believe it's a concurrency issue. Basically, miner/worker.go calls Backend.Prepare before we update the round state object in core.go
> This means our ParentCommits are actually 2 blocks old, rather than just one, and when we try to take the union of seals, we fail
> 
> One (very hacky) option here would be to, in Backend.Prepare, check both the Commits and the ParentCommits and see which match the last proposal.
> But very open to other suggestions as I'm not particularly happy with that solution

gakonst:
### Description

We need to guarantee the backend will always receive the `FinalCommittedEvent` event to change the sequence, before the worker calls `backend.Prepare`. The event gets emitted via the handler's [NewWork](https://github.com/celo-org/celo-blockchain/blob/87e07a2cd532222baeae876c3d284adc506669b4/consensus/istanbul/backend/handler.go#L208) function. 

I suspect that `commitNewWork` gets called "too quickly", which results in `ParentCommits` being called before the event has been captured by the engine

Approaches:
- Add a tiny sleep in Line 450 in worker.go
- A sustainable fix would be to wait for the FinalCommittedEvent on the worker and call commitNewWork after it has been emitted.

### Tested

I manually ran all E2E tests with approach (1) and there is no occurrence of the error from Line https://github.com/celo-org/celo-blockchain/blob/10faac32afbf49b92f26e83f2a3949bf75e4813b/consensus/istanbul/backend/engine.go#L421

It might be the case that 100ms delay is too much and we can fix the issue with less. 

Ideally we want to deploy another cluster with this fix and see if it fixes the issue 

### Related issues

- Fixes #649 

### Backwards compatibility

OK

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -447,6 +447,10 @@ func (w *worker) mainLoop() {
 		case req := <-w.newWorkCh:
 			if h, ok := w.engine.(consensus.Handler); ok {
 				h.NewWork()
+				// Wait a minimal amount of time so that the FinalizeCommittedEvent gets picked up by
+				// the engine and a new sequence gets started, so that we get the correct
+				// `ParentCommits` in `backend.Prepare`
+				time.Sleep(100 * time.Millisecond)
 			}
 			w.commitNewWork(req.interrupt, req.noempty, req.timestamp)
 
--- a/miner/worker_test.go
+++ b/miner/worker_test.go
@@ -261,7 +261,7 @@ func testPendingStateAndBlock(t *testing.T, chainConfig *params.ChainConfig, eng
 	defer w.close()
 
 	// Ensure snapshot has been updated.
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(200 * time.Millisecond)
 	block, state := w.pending()
 	if block.NumberU64() != 1 {
 		t.Errorf("block number mismatch: have %d, want %d", block.NumberU64(), 1)
@@ -272,7 +272,7 @@ func testPendingStateAndBlock(t *testing.T, chainConfig *params.ChainConfig, eng
 	b.txPool.AddLocals(newTxs)
 
 	// Ensure the new tx events has been processed
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(200 * time.Millisecond)
 	block, state = w.pending()
 	if balance := state.GetBalance(testUserAddress); balance.Cmp(big.NewInt(2000)) != 0 {
 		t.Errorf("account balance mismatch: have %d, want %d", balance, 2000)


gakonst:
(unit tests failing seems pretty arbitrary, I cannot reproduce that in my local, and it's odd because there is no change made to the `core` module in this PR)

PR #672, opened 2019-11-29
Buffer future messages (NEW)

### LINKED ISSUE #519: Validators SNBAT DDOS each other with future sequence messages

mcortesi:
### Description

Aims to replace https://github.com/celo-org/celo-blockchain/pull/517

Basically:
  * Rebases #517 into latest IBFT branch changes
  * the 1st commit is a squash of #517 
  * The 2nd commit implements changes

Backlog Changes:

 - Decouple Backlog from Core. Make it it's own type
 - Make MsgBacklog a interface + a default implementation
 - Backlog: extract common logic into `extractMessageView()` fn
 - Backlog: separate between `cleanBacklogForSeq` and
 `processBacklogForSeq`
 - `processBacklogForSeq` will stop iteration when process() returns
 false (optimization)
 - Rename `processBacklog()` to `updateState()`. Processing the backlog
 is a sideeffect of changing the currentView + State
 - Clean up backlog  changes. Add inner tests.

 Other changes:

  - cleanup checkMessage with a "first check the view, then the msgCode"
  strategy
  - checkMessage no longer err with errTooFarIntoTheFuture, instead
  backloag will check that, and don't store those messages
  - errTooFarIntoTheFuture is no longer an error
  - checkMessage doesn't have a case for Old commit messages. Instead
  handleCommit will check if checkMEssage failed with oldMessage and do
  the check there.
  - FIX: error when processing RoundChangeMessages

### Tested

unit / ci

### Related issues

- Fixes #519

### DIFF
--- a/consensus/istanbul/core/backlog.go
+++ b/consensus/istanbul/core/backlog.go
@@ -274,11 +274,12 @@ func (c *msgBacklogImpl) processBacklog() {
 			// Current sequence. Process all in order.
 			c.processBacklogForSeq(seq, func(msg *istanbul.Message) bool {
 				view, err := extractMessageView(msg)
-
-				// TODO check the error, not the view
+				if err != nil {
+					logger.Error("Error decoding msg", "msg", msg, "err", err)
+					return false
+				}
 				if view == nil {
 					logger.Error("Nil view", "msg", msg)
-					// continue
 					return false
 				}
 
--- a/consensus/istanbul/core/backlog.go
+++ b/consensus/istanbul/core/backlog.go
@@ -17,10 +17,14 @@
 package core
 
 import (
-	"github.com/ethereum/go-ethereum/common/prque"
-	"github.com/ethereum/go-ethereum/consensus/istanbul"
 	"math/big"
 	"sort"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/prque"
+	"github.com/ethereum/go-ethereum/consensus/istanbul"
+	"github.com/ethereum/go-ethereum/log"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


consensus/istanbul/core/backlog.go
```diff
@@ -29,189 +35,313 @@ var (
...
+
+		if seq < c.currentView.Sequence.Uint64() {
+			// Earlier sequence. Prune all messages.
+			c.clearBacklogForSeq(seq)
+		} else if seq == c.currentView.Sequence.Uint64() {
+			// Current sequence. Process all in order.
+			c.processBacklogForSeq(seq, func(msg *istanbul.Message) bool {
+				view, err := extractMessageView(msg)
+
+				// TODO check the error, not the view
```
timmoreton: we ignore `err` here - `gosec` would complain

consensus/istanbul/core/backlog.go
```diff
@@ -29,189 +35,314 @@ var (
...
+			})
 		}
 	}
 }
 
 func toPriority(msgCode uint64, view *istanbul.View) int64 {
 	if msgCode == istanbul.MsgRoundChange {
-		// For istanbul.MsgRoundChange, set the message priority based on its sequence
-		return -int64(view.Sequence.Uint64() * 1000)
+		// msgRoundChange comes first
```
nambrot: This seems like a departure?

mcortesi: Oh, no. It's not a departure actually. Basically the formula was:
```
priorityRoundChange = -1 * ( 1000 * sequence)
prioritiyMsg = -1 * (1000 * sequence + 10 * round + msgCodePriority)
```
but since we have one queure per sequence, we can remove that term, so we now have:
```
priorityRoundChange = 0
prioritiyMsg = -1 * (10 * round + msgCodePriority)
```
The order relation is still the same

nambrot: 👍

PR #673, opened 2019-11-30
Use validator account addresses rather than signers in celostats

asaj:
### Description

### Tested

Not tested


asaj:
Dupe

PR #675, opened 2019-11-30
retrieve validators and signers

diminator:
Description

Tested
Tested locally and with a tx node against baklavastaging


diminator:
closing in favor of #677 (not pull from fork for build)

PR #677, opened 2019-12-02
retrieve validators and signers

diminator:
Description

Tested
Tested locally and with a tx node against baklavastaging

### DIFF
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -69,9 +69,9 @@ const (
 	// delegateSendTimeout waits for the proxy to sign a message
 	delegateSendTimeout = 5
 	// statusUpdateInterval is the frequency of sending full node reports
-	statusUpdateInterval = 15
+	statusUpdateInterval = 13
 	// valSetInterval is the frequency in blocks to send the validator set
-	valSetInterval = 10
+	valSetInterval = 11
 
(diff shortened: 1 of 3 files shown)


ethstats/ethstats.go
```diff
@@ -96,14 +96,14 @@ type blockChain interface {
...
-	backend *istanbulBackend.Backend // Istanbul consensus backend
-
-	node string // Name of the node to display on the monitoring page
-	host string // Remote address of the monitoring service
+	server    *p2p.Server              // Peer-to-peer server to retrieve networking infos
+	eth       *eth.Ethereum            // Full Ethereum service if monitoring a full node
+	les       *les.LightEthereum       // Light Ethereum service if monitoring a light node
+	engine    consensus.Engine         // Consensus engine to retrieve variadic block fields
+	backend   *istanbulBackend.Backend // Istanbul consensus backend
+	etherBase common.Address
```
mrsmkl: Add comment?
Also cannot see where this variable is used...

diminator: it's used here: https://github.com/celo-org/celo-blockchain/pull/677/files/71e60d3c806c1555d566bd7a4e3f212c721b894d#diff-9d0b059013dd700f399bf16ba37421afR123

PR #679, opened 2019-12-02
expose validator.signer address and warn on latency error

diminator:
- expose the validator.signer field for account coupling in the operator, see https://github.com/celo-org/celo-monorepo/pull/1997/ for the contract side

- `reportLatency` often errors because of handshaking with `ethstats-server` (hello-ack, ping-pong). This might throw and close the connection on initialReport as both login and initialReport are waiting for responses. This is a patch that might give a glitch in latency, but avoids new connections being made upon every login try

### Tested
Not tested, best tested in staging with >50 nodes

### Logs
Connections are being made and not closed due to a ping-pong hello-ack conflict:
image

### DIFF
--- a/contract_comm/validators/validators.go
+++ b/contract_comm/validators/validators.go
@@ -104,6 +104,10 @@ const validatorsABIString string = `[
         {
           "name": "score",
           "type": "uint256"
+        },
+        {
+          "name": "signer",
+          "type": "address"
         }
       ],
       "payable": false,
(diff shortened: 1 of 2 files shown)


tkporter:
Yeah: https://github.com/celo-org/celo-monorepo/pull/1997

diminator:
yeah, we need to implement the signer - validator mapping into ethstats - that's coming up

PR #685, opened 2019-12-04
Use inmemory valnodedb for mobile geth

mcortesi:
### Description

When running geth on mobile we set validatorEnodeDbPath to `""` so we use the inmemory version of it.

### Tested

Manual

### DIFF
--- a/mobile/geth.go
+++ b/mobile/geth.go
@@ -191,6 +191,8 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		ethConf.SyncMode = getSyncMode(config.SyncMode)
 		ethConf.NetworkId = uint64(config.EthereumNetworkID)
 		ethConf.DatabaseCache = config.EthereumDatabaseCache
+		// Use an in memory DB for validatorEnode table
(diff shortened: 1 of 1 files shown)


mobile/geth.go
```diff
@@ -119,6 +119,7 @@ var defaultNodeConfig = &NodeConfig{
 // NewNodeConfig creates a new node option set, initialized to the default values.
 func NewNodeConfig() *NodeConfig {
 	config := *defaultNodeConfig
+
```
kevjue: Please remove the extra new line.

mobile/geth.go
```diff
@@ -191,6 +192,7 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		ethConf.SyncMode = getSyncMode(config.SyncMode)
 		ethConf.NetworkId = uint64(config.EthereumNetworkID)
 		ethConf.DatabaseCache = config.EthereumDatabaseCache
+		ethConf.Istanbul.ValidatorEnodeDBPath = ""
```
kevjue: Please a comment stating that "" will mean that it will use an in memory db.

jmrossy:
Confirmed works on mobile

PR #700, opened 2019-12-06
Patch crashes on mobile from rounded state db

jmrossy:
### Description

Patch crashes on mobile from rounded state db write access denial

### Tested

Synced on mobile

### Backwards compatibility

Yes

### DIFF
--- a/mobile/geth.go
+++ b/mobile/geth.go
@@ -193,6 +193,7 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		ethConf.DatabaseCache = config.EthereumDatabaseCache
 		// Use an in memory DB for validatorEnode table
 		ethConf.Istanbul.ValidatorEnodeDBPath = ""
+		// Use an in memory DB for roundState table
 		ethConf.Istanbul.RoundStateDBPath = ""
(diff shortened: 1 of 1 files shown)


mobile/geth.go
```diff
@@ -193,6 +193,7 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		ethConf.DatabaseCache = config.EthereumDatabaseCache
 		// Use an in memory DB for validatorEnode table
 		ethConf.Istanbul.ValidatorEnodeDBPath = ""
+		ethConf.Istanbul.RoundStateDBPath = ""
```
kevjue: Please add a comment stating that this will mean that the DB will be in-memory.

mobile/geth.go
```diff
@@ -193,6 +193,8 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		ethConf.DatabaseCache = config.EthereumDatabaseCache
 		// Use an in memory DB for validatorEnode table
 		ethConf.Istanbul.ValidatorEnodeDBPath = ""
+		// Use an in memory DB for roundState table
```
asaj: nit: `Use an in memory DB for IBFT round state`

PR #707, opened 2019-12-06
Exp RC timer backoff, reply to old RC with our own RC

timmoreton:
### Description

* Remove the cap on exponential backoff for IBFT round change messages. 
* If we receive a valid RC message for current seq but round earlier than our desired round, send a RC message back to the sender to speed recipent collecting F+1 and fast-forwarding to our desired round. 
* Merge fix: `Storage for `message` loop variable is reused`

### Tested

TODO

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -26,7 +26,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
-
 	"github.com/ethereum/go-ethereum/core/types"
 	blscrypto "github.com/ethereum/go-ethereum/crypto/bls"
 	"github.com/ethereum/go-ethereum/event"
@@ -435,8 +434,8 @@ func (c *core) newRoundChangeTimerForView(view *istanbul.View) {
 		// timeout for first round takes into account expected block period
 		timeout += time.Duration(c.config.BlockPeriod) * time.Second
 	} else {
-		// timeout for subsequent rounds adds an exponential backup, capped at 2**5 = 32s
-		timeout += time.Duration(math.Pow(2, math.Min(float64(round), 5.))) * time.Second
+		// timeout for subsequent rounds adds an exponential backup.
+		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
 	}
 
 	c.roundChangeTimer = time.AfterFunc(timeout, func() {
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -130,7 +130,17 @@ func (c *core) finalizeMessage(msg *istanbul.Message) ([]byte, error) {
 	return payload, nil
 }
 
+// Send message to all current validators
 func (c *core) broadcast(msg *istanbul.Message) {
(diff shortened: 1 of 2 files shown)


consensus/istanbul/core/core.go
```diff
@@ -435,8 +434,8 @@ func (c *core) newRoundChangeTimerForView(view *istanbul.View) {
 		// timeout for first round takes into account expected block period
 		timeout += time.Duration(c.config.BlockPeriod) * time.Second
 	} else {
-		// timeout for subsequent rounds adds an exponential backup, capped at 2**5 = 32s
-		timeout += time.Duration(math.Pow(2, math.Min(float64(round), 5.))) * time.Second
+		// timeout for subsequent rounds adds an exponential backup.
+		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
```
nambrot: It seems like a very aggresivve exponentiation, should we try a slightly less aggressive one?

consensus/istanbul/core/core.go
```diff
@@ -435,8 +444,8 @@ func (c *core) newRoundChangeTimerForView(view *istanbul.View) {
 		// timeout for first round takes into account expected block period
 		timeout += time.Duration(c.config.BlockPeriod) * time.Second
 	} else {
-		// timeout for subsequent rounds adds an exponential backup, capped at 2**5 = 32s
-		timeout += time.Duration(math.Pow(2, math.Min(float64(round), 5.))) * time.Second
+		// timeout for subsequent rounds adds an exponential backup.
```
asaj: nit: "backoff"?

consensus/istanbul/core/roundchange.go
```diff
@@ -157,12 +175,20 @@ func (c *core) handleRoundChange(msg *istanbul.Message) error {
...
+		logger.Info("Failed to decode ROUND CHANGE", "err", err)
 		return errInvalidMessage
 	}
 
 	// Must be same sequence and future round.
-	if err := c.checkMessage(istanbul.MsgRoundChange, rc.View); err != nil {
+	err := c.checkMessage(istanbul.MsgRoundChange, rc.View)
+
+	// If the RC message is for the current sequence but a prior round, help the sender fast forward
+	// by sending back to it (not broadcasting) a round change message for our current/desired TODO round.
```
asaj: "for our desired round."

consensus/istanbul/core/roundchange.go
```diff
@@ -157,12 +175,20 @@ func (c *core) handleRoundChange(msg *istanbul.Message) error {
...
+	if err == errOldMessage && rc.View.Sequence.Cmp(c.current.Sequence()) == 0 {
+		logger.Trace("Sending round change for current round to node in previous round", "msg_round", rc.View.Round)
```
asaj: "for desired round to node with previous desired round"

asaj:
Was this supposed to be against master?

consensus/istanbul/core/roundchange.go
timmoreton: raises an interesting question. Right now, we only send if we get `errOldMessage` which relates to rounds < our current round.  Rounds that are >= current but < desired round currently don't get sent the message.

timmoreton:
going to make another PR against master (in case we did want to consider this today)

PR #745, opened 2019-12-11
Announce timestamp

kevjue:
### Description

Replacing the `View` field in the announce message with `Timestamp` (which will have a value of unix timestamp).

In hindsight, the announce/val_enode_table/val_enode_share mechanism shouldn't use any IBFT concepts.  The whole point of that mechanism is to bootstrap a mesh network so that IBFT can be run on it.

### Tested

1) Ran unit tests.
2) Ran proxy e2e tests.
3) Changed the valEnodeDb version, and verified that it was recreated.

### Other changes

None

### Backwards compatibility

No

### DIFF
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -25,6 +25,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/syndtr/goleveldb/leveldb"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
 	vet "github.com/ethereum/go-ethereum/consensus/istanbul/backend/internal/enodes"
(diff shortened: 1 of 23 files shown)


consensus/istanbul/backend/announce.go
```diff
@@ -256,10 +255,10 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
 		return err
 	}
 
-	logger = logger.New("msgAddress", msg.Address, "msg_round", announceData.View.Round, "msg_seq", announceData.View.Sequence)
+	logger = logger.New("msgAddress", msg.Address, "msg_timestamp", announceData.Timestamp)
 
-	if view, err := sb.valEnodeTable.GetViewFromAddress(msg.Address); err == nil && announceData.View.Cmp(view) <= 0 {
-		logger.Trace("Received an old announce message", "senderAddr", msg.Address, "messageView", announceData.View, "currentEntryView", view)
+	if currentEntryTimestamp, err := sb.valEnodeTable.GetTimestampFromAddress(msg.Address); err == nil && announceData.Timestamp < currentEntryTimestamp {
```
asaj: What if err != nil here?

consensus/istanbul/backend/announce.go
```diff
@@ -256,10 +255,10 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
+		logger.Trace("Received an old announce message", "senderAddr", msg.Address, "messageTimestamp", announceData.Timestamp, "currentEntryTimestamp", currentEntryTimestamp)
```
asaj: nit: `senderAddr` and `messageTimestamp` are redundant since we add them in the logger on line 258

consensus/istanbul/backend/announce.go
kevjue: It will mean that the entry is not in the valEnodeTable.  I'll add a comment to that effect.

kevjue:
> It seems we finally moved to a local measurement of time!

@mcortesi - It's just temporary for now!  My worry is that validator's local timestamp is not guaranteed to be monotonic.  E.g. let's say a validator starts up a computer that is a year ahead, and then an announce gets sent, and then the computer updates it's time using NTP.  That validator's announce messages will be ignored for a year.

I think that a more ideal solution is to have a separate counter that is persisted in your val_enode_db (perhaps we should rename it to `validator_network_db`) and increment that on every announce message.  Or we finally implement the push/pull gossip network that I always wanted to do, instead of sending a push announce message every minute.

I'm going to create an issue for that.

PR #746, opened 2019-12-11
Check that announced IP address is not an internal IP

asaj:
### Description

This PR checks that proxied validators do not gossip internal IP addresses.

### Tested

New unit tests, and by various command line tests. 

### Other changes

- Improve readability of IBFT logs

### DIFF
--- a/p2p/enode/urlv4_test.go
+++ b/p2p/enode/urlv4_test.go
@@ -86,6 +86,16 @@ var parseNodeTests = []struct {
 		wantPrivate: true,
 	},
 	{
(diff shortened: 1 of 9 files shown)


p2p/enode/urlv4_test.go
```diff
@@ -71,6 +83,7 @@ var parseNodeTests = []struct {
 			52150,
 			52150,
 		),
+		wantPrivate: true,
```
asaj: Can we add a unit test with something like `172.168.1.1`?

kevjue:
There are cases when using the loopback address for the announced IP address is valuable.  E.g. the e2e proxy tests uses that address.

One way to address this is to add an --override-announced-external-ip-check flag.  Anybody has concerns with adding this?

Issue #726, opened 2019-12-08
Proxy and validators SBAT output the valenode table via RPC


kevjue:
### Expected Behavior

Proxy and validators SBAT output the valenode table via RPC

### Current Behavior

Proxy and validators currently is not accessible via RPC.

PR #748, opened 2019-12-11
getValEnodeTable RPC

kevjue:
### Description

Added the rpc `istanbul.getValEnodeTable`.

### Tested

Ran unit tests and e2e tests and verified the rpc worked.

### Other changes

None

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/726

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/backend/internal/enodes/val_enode_db.go
+++ b/consensus/istanbul/backend/internal/enodes/val_enode_db.go
@@ -436,3 +436,22 @@ func (vet *ValidatorEnodeDB) iterateOverAddressEntries(onEntry func(common.Addre
 	}
 	return iter.Error()
 }
+
+type ValEnodeEntryInfo struct {
+	Enode     string `json:"enode"`
+	Timestamp uint   `json:"timestamp"`
+}
+
+func (vet *ValidatorEnodeDB) ValEnodeTableInfo() (map[string]*ValEnodeEntryInfo, error) {
+	valEnodeTableInfo := make(map[string]*ValEnodeEntryInfo)
+
+	valEnodeTable, err := vet.GetAllValEnodes()
+	if err == nil {
+		for address, valEnodeEntry := range valEnodeTable {
+			valEnodeTableInfo[address.Hex()] = &ValEnodeEntryInfo{Enode: valEnodeEntry.Node.String(),
+				Timestamp: valEnodeEntry.Timestamp}
+		}
+	}
+
+	return valEnodeTableInfo, err
+}
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -792,6 +792,11 @@ web3._extend({
 			call: 'istanbul_removeProxy',
 			params: 1
 		}),
+		new web3._extend.Method({
+			name: 'getValEnodeTableInfo',
+			call: 'istanbul_getValEnodeTable',
+			params: 0
+		}),		
 		new web3._extend.Property({
 			name: 'proxyInfo',
 			getter: 'istanbul_proxyInfo'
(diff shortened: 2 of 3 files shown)


internal/web3ext/web3ext.go
```diff
@@ -792,6 +792,11 @@ web3._extend({
 			call: 'istanbul_removeProxy',
 			params: 1
 		}),
+		new web3._extend.Method({
```
tkporter: nit: Is there a reason why this is a method and not a property? Imo the properties are nice because you can just see them without calling a function

kevjue: No particular reason.  I can change it to a property.

consensus/istanbul/backend/internal/enodes/val_enode_db.go
```diff
@@ -436,3 +436,22 @@ func (vet *ValidatorEnodeDB) iterateOverAddressEntries(onEntry func(common.Addre
...
+
+type ValEnodeEntryInfo struct {
+	Enode     string `json:"enode"`
+	Timestamp uint   `json:"timestamp"`
+}
+
+func (vet *ValidatorEnodeDB) ValEnodeTableInfo() (map[string]*ValEnodeEntryInfo, error) {
+	valEnodeTableInfo := make(map[string]*ValEnodeEntryInfo)
+
+	valEnodeTable, err := vet.GetAllValEnodes()
```
mcortesi: i don't remember, but i suppose getAllVAlEnodes does the locking here.

internal/web3ext/web3ext.go
```diff
@@ -792,6 +792,10 @@ web3._extend({
 			call: 'istanbul_removeProxy',
 			params: 1
 		}),
+		new web3._extend.Property({
```
mcortesi: not sure who this is done, but as we have `web.eth` `web.ssh` and other... maybe it would be nice to have `web.ibft` here

mcortesi: Or some other namespace. Not istanbul, since that's also the name for eth2.0

consensus/istanbul/backend/internal/enodes/val_enode_db.go
kevjue: `GetAllValEnodes` does do the locking.

Note that this new function does go against your initial pattern of having all public functions acquire the lock.   Do you see an issue with that?

internal/web3ext/web3ext.go
kevjue: Agreed that we should rename the `istanbul` rpc namespace to something different.   In fact, I think we should rename the geth istanbul package to that new name (btw, we really should come up with a name for our POS/Consensus mechanism.  All other blockchain companies did that).

IMO, that would be a much larger ticket.

consensus/istanbul/backend/internal/enodes/val_enode_db.go
mcortesi: not at all, since it call another public function that acquires the lock.

btw, i'm seeing the initial patter on geth1.9

internal/web3ext/web3ext.go
mcortesi: just the package rename is happening just after 1.9 or even there... since it already has "istanbul" as a thing

kevjue: Oh really?  That's interesting.  So sounds like IBFT will be a "first class" engine.

kevjue: Actually, I think I may have misread your comment.  I thought that you meant that geth 1.9 already has the istanbul package.   But after re-reading, I think when you say `1.9`, you mean to refer to the task of merging 1.9 into our implementation.

PR #749, opened 2019-12-12
validators SBAT handle clock drift for announce messages

kevjue:
### Description

When the announce message is sent, it uses the sender's computer's local timestamp.  That timestamp may not necessarily always increase (e.g. NTP could make it earlier).

This change will modify the validators so that they only accept announce messages with timestamps that are within a time range of it's own local timestamp.

### Tested

* Ran e2e proxy tests
* Ran geth unit tests

### Other changes

- Added the flag `proxy.override-announced-ip-check` to be used for the proxy e2e test
- Converted the timestamp field to be type big.Int instead of uint

### Backwards compatibility

Yes

### DIFF
--- a/cmd/geth/usage.go
+++ b/cmd/geth/usage.go
@@ -249,7 +249,7 @@ var AppHelpFlagGroups = []flagGroup{
 			utils.ProxiedValidatorAddressFlag,
 			utils.ProxiedFlag,
 			utils.ProxyEnodeURLPairFlag,
-			utils.ProxyOverrideAnnounceIPCheckFlag,
+			utils.ProxyAllowExternalPrivateIPFlag,
 		},
 	},
 	{
--- a/cmd/geth/usage.go
+++ b/cmd/geth/usage.go
@@ -249,6 +249,7 @@ var AppHelpFlagGroups = []flagGroup{
 			utils.ProxiedValidatorAddressFlag,
 			utils.ProxiedFlag,
 			utils.ProxyEnodeURLPairFlag,
+			utils.ProxyOverrideAnnounceIPCheckFlag,
 		},
 	},
 	{
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -38,6 +38,10 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
+const (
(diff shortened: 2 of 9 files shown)


cmd/geth/usage.go
```diff
@@ -249,6 +249,7 @@ var AppHelpFlagGroups = []flagGroup{
 			utils.ProxiedValidatorAddressFlag,
 			utils.ProxiedFlag,
 			utils.ProxyEnodeURLPairFlag,
+			utils.ProxyOverrideAnnounceIPCheckFlag,
```
timmoreton: naming nit: how about `ProxyAllowExternalPrivateIPFlag` ? we don't call it the announce ip elsewhere do we?

consensus/istanbul/backend/announce.go
```diff
@@ -37,6 +38,10 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
+const (
+	timestampDiffThreshold = 20 * time.Second
```
timmoreton: thinking it would be ok if this was a bit bigger? what would be the effect of making this the same as the announce interval?

consensus/istanbul/backend/announce.go
```diff
@@ -260,8 +264,15 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
 
 	logger = logger.New("msgAddress", msg.Address, "msg_timestamp", announceData.Timestamp)
 
+	// Check to see if the msg's timestamp is within 20 seconds of this computer's timestamp (This is the same clock threshold check that the p2p layer uses)
+	now := time.Now()
+	if int64(announceData.Timestamp.Uint64()) < now.Add(-timestampDiffThreshold).Unix() || int64(announceData.Timestamp.Uint64()) > now.Add(timestampDiffThreshold).Unix() {
+		logger.Warn("Ignoring announce messages since it's timestamp diff is over the threshold", "now", now, "announce timestamp", announceData.Timestamp.Uint64())
```
timmoreton: nits: maybe `Info`? also `it's` -> `its` and maybe more useful to log the diff in seconds so you can e.g graph them or add them to metrics more easily

consensus/istanbul/backend/announce.go
kevjue: I don't see any major issues with making this bigger.

PR #757, opened 2019-12-14
Make issue reporting better

tkporter:
### Description

Made some changes to help us deal with reported issues in the future.

* Asks for reproduction steps, logs, & system info when creating issues
* Updates `geth bug` command to open an issue with the correct system info
* Updates `geth version` to have more system information

### Tested

Ran `geth version`, `geth bug`

### Other changes

n/a

### Related issues

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,8 @@ Please make sure your contributions adhere to our coding guidelines:
 
 ### Submitting an issue
 
-If this is a bug, you can run `geth bug` to
+If you come across a bug, you can run `geth bug` to open a GitHub issue draft
+with pre-filled system information.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -69,6 +69,11 @@ Please make sure your contributions adhere to our coding guidelines:
  * Commit messages should be prefixed with the package(s) they modify.
    * E.g. "eth, rpc: make trace configs optional"
 
+### Submitting an issue
+
+If you come across a bug, you can run `geth bug` to open a GitHub issue draft
```
nategraf: It may be worth clarifying that if they were running the node with `docker run` they should run this command the same way

PR #767, opened 2019-12-17
Jfoutts/slashing precompiles

jfoutts-celo:
### Description

Add `hashHeader` and `getBlockNumberFromHeader` precompiled contracts.

The arguments are malleable.   Append a 0 byte to the input, and you get the same output.
Is this a problem for us? 

### Tested

Added unit tests using the block header constants from core/types/block_test TestBlockEncoding.


core/vm/contracts.go
```diff
@@ -669,3 +676,50 @@ func (c *epochSize) Run(input []byte, caller common.Address, evm *EVM, gas uint6
...
+}
+
+func (c *blockNumberFromHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
+	gas, err := debitRequiredGas(c, input, gas)
+	if err != nil {
+		return nil, gas, err
+	}
+
+	var header types.Header
+	err = rlp.Decode(bytes.NewReader(input), &header)
```
nategraf: No need to create the bytes reader https://godoc.org/github.com/ethereum/go-ethereum/rlp#DecodeBytes
```suggestion
	err = rlp.DecodeBytes(input, &header)
```

core/vm/contracts.go
```diff
@@ -669,3 +676,50 @@ func (c *epochSize) Run(input []byte, caller common.Address, evm *EVM, gas uint6
...
+}
+
+func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
+	gas, err := debitRequiredGas(c, input, gas)
+	if err != nil {
+		return nil, gas, err
+	}
+
+	var header types.Header
+	err = rlp.Decode(bytes.NewReader(input), &header)
```
nategraf: Same as above
```suggestion
	err = rlp.DecodeBytes(input, &header)
```

core/vm/contracts_test.go
```diff
@@ -365,6 +365,34 @@ var proofOfPossessionTests = []precompiledTest{
...
+	},
+	{
+		input:         "f901f9a07285abd5b24742f184ad676e31f6054663b3529bc35ea2fcad8a3e0f642a46f7a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347948888f1f195afa192cfee860698584c030f4c9db1a0ecc60e00b3fe5ce9f6e1a10e5469764daf51f1fe93c22ec3f9a7583a80357217a0d35d334d87c0cc0a202e3756bf81fae08b1575f286c7ee7a3f8df4f0f3afc55da056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b90100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008302000001832fefd8825208845c47775c80a0000000000000000000000000000000000000000000000000000000000000000088000000000000000",
+		expected:      "rlp: value size exceeds available input length",
+		name:          "input_invalid_removed_byte",
+		errorExpected: true,
+	},
+}
+
```
nategraf: Could you add empty input tests to both of these? I would expect an error, but I want to make sure there is not a panic.

jfoutts-celo: Good call. Added.

asaj:
> The arguments are malleable. Append a 0 byte to the input, and you get the same output.
Is this a problem for us?

I think this will be a problem if the input args are used to determine whether or not a user has been slashed for a particular offense. You shouldn't be able to be slashed twice for double signing the same block.

asaj:
Maybe we should require no leading zeros?

core/vm/contracts.go
```diff
@@ -669,3 +675,50 @@ func (c *epochSize) Run(input []byte, caller common.Address, evm *EVM, gas uint6
...
+		return nil, gas, err
+	}
+
+	var header types.Header
+	err = rlp.DecodeBytes(input, &header)
+	if err != nil {
+		return nil, gas, err
+	}
+
+	hashBytes := header.Hash().Bytes()
```
asaj: Assuming this is guaranteed to be 32 bytes?

asaj:
Actually I don't think it's an issue, we can (and should) use the block hash to record whether or not someone has been slashed for a particular block

params/protocol_params.go
```diff
@@ -95,9 +95,11 @@ const (
 	// TODO: make this cost variable- https://github.com/celo-org/geth/issues/250
 	FractionMulExpGas uint64 = 1050 // Cost of performing multiplication and exponentiation of fractions to an exponent of up to 10^3.
 	// TODO(kobigurk):  Figure out what the actual gas cost of this contract should be.
-	ProofOfPossessionGas uint64 = 50000 // Cost of verifying a BLS proof of possession.
-	GetValidatorGas      uint64 = 5000  // Cost of reading a validator's address.
-	GetEpochSizeGas      uint64 = 1000  // Cost of querying the number of blocks in an epoch.
+	ProofOfPossessionGas        uint64 = 50000 // Cost of verifying a BLS proof of possession.
+	GetValidatorGas             uint64 = 5000  // Cost of reading a validator's address.
+	GetEpochSizeGas             uint64 = 1000  // Cost of querying the number of blocks in an epoch.
+	GetBlockNumberFromHeaderGas uint64 = 10000 // Cost of decoding a block header.
```
asaj: How were these decided?

core/vm/contracts.go
jfoutts-celo: Yes, I believe so. I see no path (after rlp.DecodeBytes with nil error) that doesn't return 32 bytes. But I tested that the unit tests run regardless of the return value.  And I'll also test what Solidity would do with a non-32-byte return value.

params/protocol_params.go
jfoutts-celo: I used the GetEpochSizeGas = 1000 as a rule of thumb, and said that if (returning a constant) costs x, then parsing a header costs 10x, and parsing then hashing a header is 20x.  But now that you mention it, it's not consistent with Sha256BaseGas and Sha256PerWordGas. Which would indicate hashing a header of 508 bytes costs only 252 gas, not 10,000.

PR #770, opened 2019-12-18
Loop until IBFT core catches up so that signatures can be added to block

### LINKED ISSUE #758: Validators are unable to aggregate signatures from all commit messages
### Expected Behavior

Validators can combine signatures from the latest block in their chain with any others they collected since proposing

### Current Behavior

Occasionally get the following errors:
```
err: "invalid signature"   
  func: "Backend.Prepare()"   
  msg: "Failed to combine additional seals with parent aggregated seal." 
```

This is likely due to a race condition in which the round state has yet to be updated, so `ParentCommits` represents commits from a different block than the one we are attempting to combine seals for.

asaj:
### Description

This PR addresses a race condition that rarely occurs after blocks are created in round > 0, in which `Backend.Prepare()` is called before the IBFT core sequence changes, resulting in an attempt to aggregate signatures from two different blocks.

### Tested

Added a sleep of 200ms and 1s in `RoundState.StartNewSequence()` and confirmed that we saw the error message in master but not in this implementation, where we succeeded and timed out, respectively.

 
### Other changes
- Revert previous attempt to fix this race condition (https://github.com/celo-org/celo-blockchain/pull/670)
- Minor changes to logging

### Related issues

- Fixes #758 

### Backwards compatibility

Compatible

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -417,7 +417,8 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 						return seq
 					}
 				case <-timeout:
-					log.Warn("Timed out while waiting for core to sequence change, unable to combine commit messages with ParentAggregatedSeal", "cur_seq", sb.core.Sequence())
+					// TODO(asa): Why is this logged by full nodes?
+					log.Trace("Timed out while waiting for core to sequence change, unable to combine commit messages with ParentAggregatedSeal", "cur_seq", sb.core.Sequence())
 					return nil
 				}
 			}
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -404,20 +404,26 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 		// This typically happens in round > 0, since round 0 typically hits the "time.Sleep()"
 		// above.
 		// When this happens, loop until sb.core moves to the next sequence, with a limit of 500ms.
-		var seq *big.Int
-		for stay, timeout := true, time.After(500*time.Millisecond); stay; {
-			select {
-			case <-timeout:
-				log.Warn("Timed out while waiting for core to sequence change, unable to combine commit messages with ParentAggregatedSeal", "cur_seq", sb.core.Sequence())
-				stay = false
-			default:
-				seq = sb.core.Sequence()
-				stay = seq == nil || seq.Cmp(header.Number) < 0
-				if !stay {
-					logger.Trace("Current sequence matches header", "cur_seq", seq)
+		waitForSequenceChange := func() *big.Int {
+			timeout := time.After(500 * time.Millisecond)
+			ticker := time.NewTicker(10 * time.Millisecond)
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ticker.C:
+					seq := sb.core.Sequence()
+					if seq != nil && seq.Cmp(header.Number) == 0 {
+						logger.Trace("Current sequence matches header", "cur_seq", seq)
+						return seq
+					}
+				case <-timeout:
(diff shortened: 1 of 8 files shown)


lucasege:
Assuming tests are flakes and coverage failure is acknowledged, the code looks good - can't speak to the logic in context

consensus/istanbul/backend/engine.go
```diff
@@ -395,39 +394,71 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
-			} else {
-				// need to pass the previous block from the parent to get the parent's validators
-				// (otherwise we'd be getting the validators for the current block)
-				parentValidators := sb.getValidators(parent.Number.Uint64()-1, parent.ParentHash)
-				// only update to use the union if we indeed provided a valid aggregate signature for this block
-				if err := sb.verifyAggregatedSeal(parent.Hash(), parentValidators, unionAggregatedSeal); err != nil {
-					logger.Error("Failed to combine additional seals with parent aggregated seal.", "err", err)
+		logger = logger.New("parentAggregatedSeal", parentAggregatedSeal.String())
+		logger.Trace("Got ParentAggregatedSeal")
+		if seq != nil && seq.Cmp(header.Number) == 0 {
```
mcortesi: It seems to me, that if we couldn't get a seq (`== nil`) or if it's no the seq we want... then there's something wrong and we should return an error.

consensus/istanbul/backend/engine.go
```diff
@@ -395,39 +394,71 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
+			parentCommits := sb.core.ParentCommits()
+			logger = logger.New("cur_seq", sb.core.Sequence())
+			if parentCommits != nil && parentCommits.Size() != 0 {
+				logger = logger.New("numParentCommits", parentCommits.Size())
+				logger.Trace("Found commit messages from previous sequence to combine with ParentAggregatedSeal")
+				// if we had any seals gossiped to us, proceed to add them to the
+				// already aggregated signature
+				if unionAggregatedSeal, err := istanbulCore.UnionOfSeals(parentExtra.AggregatedSeal, parentCommits); err != nil {
+					logger.Error("Failed to combine commit messages with ParentAggregatedSeal", "err", err)
```
mcortesi: same here, why not return an error?

consensus/istanbul/backend/engine.go
```diff
@@ -395,39 +394,71 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
...
 				} else {
-					parentAggregatedSeal = unionAggregatedSeal
-					logger.Debug("Succeeded in combining additional seals with parent aggregated seal", "combinedAggregatedSeal", parentAggregatedSeal.String())
+					// need to pass the previous block from the parent to get the parent's validators
+					// (otherwise we'd be getting the validators for the current block)
+					parentValidators := sb.getValidators(parent.Number.Uint64()-1, parent.ParentHash)
+					// only update to use the union if we indeed provided a valid aggregate signature for this block
+					if err := sb.verifyAggregatedSeal(parent.Hash(), parentValidators, unionAggregatedSeal); err != nil {
+						logger.Error("Failed to verify combined ParentAggregatedSeal", "err", err)
```
mcortesi: same

consensus/istanbul/backend/engine.go
asaj: It won't strictly stop us from proposing using the 2f+1 signatures from the block in our chain, so I think we should continue to try to make a block

consensus/istanbul/backend/engine.go
asaj: Same answer, we can get by with just using the signatures from the block in our chain

consensus/istanbul/backend/engine.go
asaj: Same

PR #771, opened 2019-12-18
Add les chain reader for istanbul

cmcewen:
### Description

We saw an occasional crash in the mobile app during the selectIssuers transaction due to trying to call this function: https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/engine.go#L703 with a nil chain. This creates a chain reader for istanbul when running les.

### Tested

Did a local build and was able to run the selectIssuers transaction without a crash


cmcewen:
chatted with @mcortesi and we agreed to leave the initialization logic where it is:
```My impression is that engine was created outside this call, in some process initializer function…. and that’s the place where  i would do this wiring (setting the lightChainreader).
But i’ve checked, and it’s created here https://github.com/celo-org/celo-blockchain/commit/4391ffd5fa6d96365848efdd7962e2088745090e#diff-d4bc153f1d5fe77ce41504e98b557213R119
So, nothing to do…```

Also the test failures seem unrelated to my change as far as I can tell

Issue #696, opened 2019-12-05
Precompile to detect double signing for slashing


asaj:
### Expected Behavior

A precompile that takes two RLP encoded blocks, and a signer, and requires that the two block hashes are different, have the same height, have a quorum of signatures, and that `signer` was part of the quorum. Returns the block number of the provided blocks.

### Current Behavior

No such precompile.

Issue #695, opened 2019-12-05
Precompile to detect downtime for slashing


asaj:
### Expected Behavior

A new precompile that iterates over block headers from `startBlock` to `startBlock + numBlocks` and returns the sum of parentAggregatedSeal.Bitmap[signerIndex[0|1]], i.e. the number of signatures present by `signer`.

### Current Behavior

No such precompile

PR #772, opened 2019-12-18
Add precompiles to support signature checking on provided and historic headers

nategraf:
### Description

This PR modifies the `getValidatorAddress` and `numberValidators` precompiles accept a block number to query a historic validator set, and adds the 

### Tested

Unit tests

### Other changes

* Change an instance of the name "infrastructure fund" to "community fund"
* Change the unused `VerifySeal` function on the Istanbul consensus engine to have a more useful verification of the seal. (The previous definition was an incomplete implementation of verification and was almost certainly implemented to satisfy the `consensus.Engine` interface.)
* Modified a few comments to describe more accurately the implementation.
* Move `core/evm.go` to `core/vm/context.go` to allow code sharing with `contract_comm/evm.go` and testing in `core/vm/contracts_test.go`
* Add helper function to calculate celo precompile addresses

### Related issues

- Related https://github.com/celo-org/celo-blockchain/issues/695
- Related https://github.com/celo-org/celo-blockchain/issues/696

### Backwards compatibility

Not backwards compatible due to a change in the validator set precompiles.


core/vm/contracts.go
```diff
@@ -722,3 +754,81 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+	// input is comprised of a single argument:
+	//   blockNumber: 32 byte integer representing the block number to access
+	if len(input) < 32 {
+		return nil, gas, ErrInputLength
+	}
+
+	blockNumber := new(big.Int).SetBytes(input[0:32])
+
+	// Ensure the request is for information from a previously sealed block.
+	if blockNumber.Cmp(evm.Context.BlockNumber) >= 0 {
```
asaj: What does the parent seal bitmap look like for block 0?

core/vm/contracts.go
```diff
@@ -722,3 +754,81 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+	}
+
+	// Ensure the request is for a sufficiently recent block to limit state expansion.
+	historyLimit := new(big.Int).SetUint64(evm.Context.Engine.EpochSize() * 4)
+	if blockNumber.Cmp(new(big.Int).Sub(evm.Context.BlockNumber, historyLimit)) <= 0 {
+		return nil, gas, ErrBlockNumberOutOfBounds
+	}
+
+	bitmap := evm.Context.GetParentSealBitmap(blockNumber.Uint64())
+	if bitmap == nil {
```
asaj: This should be invalid, right?

contract_comm/evm.go
```diff
@@ -124,6 +130,64 @@ func Transfer(db vm.StateDB, sender, recipient common.Address, amount *big.Int)
...
+			istanbulExtra, err := types.ExtractIstanbulExtra(header)
+			if err == nil {
+				bitmap = istanbulExtra.AggregatedSeal.Bitmap
+				cache[header.Number.Uint64()] = bitmap
+			}
+			if n == header.Number.Uint64() {
+				return bitmap
+			}
+		}
+		return nil
```
asaj: Should we ever get here?

params/protocol_params.go
```diff
@@ -100,6 +100,8 @@ const (
 	GetEpochSizeGas             uint64 = 1000  // Cost of querying the number of blocks in an epoch.
 	GetBlockNumberFromHeaderGas uint64 = 10000 // Cost of decoding a block header.
 	HashHeaderGas               uint64 = 20000 // Cost of hashing a block header.
+	GetParentSealBitmapGas      uint64 = 5000  // Cost of reading the parent seal bitmap from the chain.
```
asaj: How were these values decided upon?

nategraf: `GetParentSealBitmapGas` I had very little intuition for but I've now set to get equal to the gas got the `BLOCKHASH` opcode * 12 (because it may retrieve 12 times as many blocks)
`GetVerifiedSealBitmapGas` I set to `ProofOfPossessionGas + GetValidatorGas` because those are its main elements.
I wasn't as worries about it because I figured we'd revise these latter.

contract_comm/evm.go
nategraf: Only if something goes very wrong.

core/vm/contracts.go
nategraf: Yes, something went quite wrong if this happens. Do you think there is a more reasonable return value?

core/vm/contracts.go
nategraf: It's invalid to request from block 0. I've added an error.

core/vm/contracts.go
```diff
@@ -642,17 +661,30 @@ func (c *numberValidators) RequiredGas(input []byte) uint64 {
...
+	blockNumber := new(big.Int).SetBytes(input[0:32])
+	if blockNumber.Cmp(common.Big0) == 0 {
+		// Genesis validator set is empty. Return 0.
+		return make([]byte, 32), gas, nil
+	}
+	if blockNumber.Cmp(evm.Context.BlockNumber) > 0 {
+		return nil, gas, ErrBlockNumberOutOfBounds
+	}
+
+	validators := evm.Context.Engine.GetValidators(new(big.Int).Sub(blockNumber, common.Big1), common.Hash{})
```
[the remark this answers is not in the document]
nategraf: Same (Mainly because we don't bound the the blockNumber parameter, but also because it isn't actually used for much.)

core/vm/contracts.go
```diff
@@ -642,17 +661,30 @@ func (c *numberValidators) RequiredGas(input []byte) uint64 {
...
+	// input is comprised of a single argument:
+	//   blockNumber: 32 byte integer representing the block number to access
+	if len(input) < 32 {
 		return nil, gas, ErrInputLength
 	}
 
-	validators := evm.Context.Engine.GetValidators(big.NewInt(evm.Context.BlockNumber.Int64()-1), evm.Context.GetHash(evm.Context.BlockNumber.Uint64()-1))
+	blockNumber := new(big.Int).SetBytes(input[0:32])
+	if blockNumber.Cmp(common.Big0) == 0 {
+		// Genesis validator set is empty. Return 0.
```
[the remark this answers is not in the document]
nategraf: It returns an error in the other precompile because when the validator set is empty, as in the case of the genesis block, every index is out of bounds. The other precompile was returning the wrong error, so I fixed that and added a comment.

core/vm/contracts.go
```diff
@@ -604,25 +614,34 @@ func (c *getValidator) RequiredGas(input []byte) uint64 {
...
-	index := (&big.Int{}).SetBytes(input[0:32])
+	index := new(big.Int).SetBytes(input[0:32])
 
-	validators := evm.Context.Engine.GetValidators(big.NewInt(evm.Context.BlockNumber.Int64()-1), evm.Context.GetHash(evm.Context.BlockNumber.Uint64()-1))
+	blockNumber := new(big.Int).SetBytes(input[32:64])
+	if blockNumber.Cmp(common.Big0) == 0 || blockNumber.Cmp(evm.Context.BlockNumber) > 0 {
+		return nil, gas, ErrBlockNumberOutOfBounds
+	}
 
+	validators := evm.Context.Engine.GetValidators(new(big.Int).Sub(blockNumber, common.Big1), common.Hash{})
```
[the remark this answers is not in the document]
nategraf: Mainly because we don't bound the the blockNumber parameter, but also because it isn't actually used for much.

contract_comm/evm.go
```diff
@@ -124,6 +130,64 @@ func Transfer(db vm.StateDB, sender, recipient common.Address, amount *big.Int)
 	db.AddBalance(recipient, amount)
 }
 
+// GetParentSealFn returns a GetParentSeal function that returns the aggregated parent seal at the given block number.
+// Note: Unless previously cached, retrieves every block between the reference block and n.
+func GetParentSealBitmapFn(ref *types.Header, chain ChainContext) func(uint64) *big.Int {
```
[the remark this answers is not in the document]
nategraf: Yes, but the downside will be that caching would be more expensive (the cache is going to get useful for downtime calculation where we expect to get many bitmaps)

contract_comm/evm.go
```diff
@@ -124,6 +130,64 @@ func Transfer(db vm.StateDB, sender, recipient common.Address, amount *big.Int)
 	db.AddBalance(recipient, amount)
 }
 
+// GetParentSealFn returns a GetParentSeal function that returns the aggregated parent seal at the given block number.
+// Note: Unless previously cached, retrieves every block between the reference block and n.
```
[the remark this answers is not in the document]
nategraf: Because retrieving by just the block number is only available on the canonical chain.

contract_comm/evm.go
```diff
@@ -70,24 +70,30 @@ func NewEVMContext(msg types.Message, header *types.Header, chain ChainContext,
 		beneficiary = *author
 	}
 
+	var engine consensus.Engine
+	if chain != nil {
```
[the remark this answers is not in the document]
nategraf: Alright, I've brought this in line with the duplicate of this code

contract_comm/evm.go
```diff
@@ -41,12 +41,12 @@ var (
...
 
 // ChainContext supports retrieving chain data and consensus parameters
-// from the blockchain to be used during transaction processing.
+// from the block chain to be used during transaction processing.
 type ChainContext interface {
 	// Engine retrieves the blockchain's consensus engine.
 	Engine() consensus.Engine
 
-	// GetHeader returns the current header.
+	// GetHeader returns the hash corresponding to the given hash and number.
```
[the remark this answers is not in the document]
nategraf: Done

contract_comm/evm.go
```diff
@@ -41,12 +41,12 @@ var (
 // TODO(kevjue) - Figure out a way to not have duplicated code between this file and core/evm.go
 
 // ChainContext supports retrieving chain data and consensus parameters
-// from the blockchain to be used during transaction processing.
+// from the block chain to be used during transaction processing.
```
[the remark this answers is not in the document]
nategraf: Done

consensus/istanbul/backend/engine.go
```diff
@@ -350,17 +349,24 @@ func (sb *Backend) verifyAggregatedSeal(headerHash common.Hash, validators istan
...
 	}
-	return sb.verifySigner(chain, header, nil)
+
+	// Acquire the validator set whose signatures will be verified.
+	// FIXME: Based on the current implemenation of validator set construction, only validator sets
+	// from the canonical chain will be used. This means that if the provided header is a valid
+	// member of a non-canonical chain, seal verification will only succeed if the validator set
+	// happens to be the same as the canonical chain at the same block number (as would be the case
+	// for a fork from the canonical chain which does not cross an epoch boundary)
+	valSet := sb.getValidators(header.Number.Uint64()-1, common.Hash{})
```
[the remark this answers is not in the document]
nategraf: Yes

params/protocol_params.go
nategraf: Actually I realized `GetParentSealBitmap` has about 12 times the history length as `BLOCKHASH` on baklava where the epoch size is 720. On mainnet it will be 24 times longer than that. Meaning that if I am still using the `BLOCKHASH` gas cost as a guide it's `20 * 12 * 24 = 6000`. Adjusted it once more.

contract_comm/evm.go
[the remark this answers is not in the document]
asaj: I think that's okay for now, since doing it this way is an extremely effective DoS vector, since I can cause everyone to iterate over every single header in the chain for cheap. This sort of caching only works in getBlockHash because the limit is 256 blocks, but here our limit is so large that we can't afford to cache.

contract_comm/evm.go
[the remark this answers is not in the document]
asaj: See previous comment about the safety (or lack thereof) of caching the way we're currently doing it

core/vm/contracts.go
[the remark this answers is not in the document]
asaj: Why does not bounding the block number mean we can't include the hash?

core/vm/contracts.go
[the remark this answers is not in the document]
asaj: Maybe I should rephrase, can you add your reply as a comment in the code?

core/vm/contracts.go
asaj: Will have to revisit, but certainly we should log an error

contract_comm/evm.go
asaj: Should we log and return an error instead?

core/vm/contracts.go
asaj: And probably return one, and the call should fail

params/protocol_params.go
asaj: I'm probably missing something obvious, but why is it `12 * 24` times as expensive? Should we make the cost a function of the epoch length then, rather than hard code an assumed epoch length?

core/vm/contracts.go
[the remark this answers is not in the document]
nategraf: Because we don't have any way to access arbitrary block hashes here.

core/vm/contracts.go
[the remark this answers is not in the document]
nategraf: I've added a note here 👍

contract_comm/evm.go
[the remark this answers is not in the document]
nategraf: As discussed this morning, this is changed to get get the block by number and I've created #775 to start talking about efforts to make that operation safe and fast.

contract_comm/evm.go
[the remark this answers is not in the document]
nategraf: Eliminated this function

core/vm/contracts.go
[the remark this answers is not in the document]
nategraf: I've added the method retrieve arbitrary block hashes, and this could be used here, but it isn't necessary so I'd still say we should just pass empty hash. your thoughts?

core/vm/contracts.go
nategraf: No longer relevant.

consensus/istanbul/backend/engine.go
```diff
@@ -350,17 +349,24 @@ func (sb *Backend) verifyAggregatedSeal(headerHash common.Hash, validators istan
...
 func (sb *Backend) VerifySeal(chain consensus.ChainReader, header *types.Header) error {
-	// get parent header and ensure the signer is in parent's validator set
-	number := header.Number.Uint64()
-	if number == 0 {
+	// Ensure the block number is greater than zero, but less or equal to than max uint64.
+	if header.Number.Cmp(common.Big0) <= 0 || !header.Number.IsUint64() {
 		return errUnknownBlock
 	}
 
-	// ensure that the difficulty equals to defaultDifficulty
```
asaj: Why remove this?

consensus/istanbul/backend/engine.go
```diff
@@ -840,7 +846,9 @@ func (sb *Backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 	returnSnap := snap.copy()
 
 	returnSnap.Number = number
-	returnSnap.Hash = hash
+	if hash != (common.Hash{}) {
```
asaj: Why is this necessary?

core/vm/context.go
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
-package core
+package vm
```
asaj: Just flagging that this change will make pulling in future changes more difficult

core/vm/contracts.go
```diff
@@ -691,7 +729,7 @@ func (c *blockNumberFromHeader) Run(input []byte, caller common.Address, evm *EV
 	var header types.Header
 	err = rlp.DecodeBytes(input, &header)
 	if err != nil {
-		return nil, gas, err
+		return nil, gas, ErrInputDecode
```
asaj: I'm worried that returning a different error will mean we won't log the more informative error

core/vm/contracts.go
```diff
@@ -715,10 +753,99 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+	// input is comprised of a single argument:
+	//   blockNumber: 32 byte integer representing the block number to access
+	if len(input) < 32 {
+		return nil, gas, ErrInputLength
+	}
+
+	blockNumber := new(big.Int).SetBytes(input[0:32])
+
+	// Ensure the request is for information from a previously sealed block.
+	if blockNumber.Cmp(common.Big0) == 0 {
```
asaj: nit: combine into `if blockNumber.Cmp(common.Big0) == 0 || blockNumber.Cmp(evm.Context.BlockNumber) >= 0`...

core/vm/contracts.go
```diff
@@ -715,10 +753,99 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+		return nil, gas, ErrBlockNumberOutOfBounds
+	}
+	if blockNumber.Cmp(evm.Context.BlockNumber) >= 0 {
```
asaj: Why is it invalid to ask for the parentseal bitmap of the current block?

core/vm/contracts.go
```diff
@@ -715,10 +753,99 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+
+	// Ensure the request is for a sufficiently recent block to limit state expansion.
+	historyLimit := new(big.Int).SetUint64(evm.Context.Engine.EpochSize() * 4)
+	if blockNumber.Cmp(new(big.Int).Sub(evm.Context.BlockNumber, historyLimit)) <= 0 {
+		return nil, gas, ErrBlockNumberOutOfBounds
+	}
+
+	header := evm.Context.GetHeaderByNumber(blockNumber.Uint64())
+	if header == nil {
+		// Under sane circumstances, this error should never be reached.
```
asaj: Let's log an error message here

core/vm/contracts.go
```diff
@@ -715,10 +753,99 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+		return nil, gas, ErrUnexpected
+	}
+
+	extra, err := types.ExtractIstanbulExtra(header)
+	if err != nil {
+		// Header is from a non-Istanbul engine.
```
asaj: Ditto, we should never get here

core/vm/contracts.go
```diff
@@ -715,10 +753,99 @@ func (c *hashHeader) Run(input []byte, caller common.Address, evm *EVM, gas uint
...
+	}
+	if !evm.Context.VerifySeal(header) {
+		return nil, gas, ErrInputValidation
+	}
+
+	// Extract the verified seal from the header.
+	extra, err := types.ExtractIstanbulExtra(header)
+	if err != nil {
+		// Seal verified by a non-Istanbul engine. Return an error.
+		return nil, gas, ErrEngineIncompatible
```
asaj: Unlike the previous precompile, this is possible to reach since the user can put any data they want. They should maybe be different errors

core/vm/contracts_test.go
```diff
@@ -368,13 +368,13 @@ var proofOfPossessionTests = []precompiledTest{
...
 		name:          "input_invalid_empty",
 		errorExpected: true,
 	},
 	{
 		input:         "f901f9a07285abd5b24742f184ad676e31f6054663b3529bc35ea2fcad8a3e0f642a46f7a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347948888f1f195afa192cfee860698584c030f4c9db1a0ecc60e00b3fe5ce9f6e1a10e5469764daf51f1fe93c22ec3f9a7583a80357217a0d35d334d87c0cc0a202e3756bf81fae08b1575f286c7ee7a3f8df4f0f3afc55da056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b90100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008302000001832fefd8825208845c47775c80a0000000000000000000000000000000000000000000000000000000000000000088000000000000000",
-		expected:      "rlp: value size exceeds available input length",
```
asaj: IMO this is a more informative error than "unable to decode input"

core/vm/contracts.go
nategraf: Logged

core/vm/contracts.go
nategraf: It's not possible to hit as long as we are using the Istanbul engine because seal verification uses the Istanbul extra data (namely the aggregated seal)

core/vm/contracts_test.go
nategraf: See above comment about propagating errors from RLP

core/vm/context.go
asaj: I agree!

core/vm/contracts.go
asaj: Interesting, why do they need to be deterministic?

core/vm/contracts.go
asaj: But a user could pass invalid istanbul extradata, right?

core/vm/contracts.go
nategraf: I had just assumed error messages to be included under the umbrella of EVM determinism, but maybe they aren't. If not, I'm happy to put the rlp errors back.

core/vm/contracts.go
nategraf: I was able to confirm that the parent seal is added in the block prepare step, which occurs before transactions are added to the block. https://github.com/celo-org/celo-blockchain/blob/fef30781c9215b3bba4422074dd9d38ed92aed92/miner/worker.go#L918

nategraf: So, I've changed it to allow getting the chain head parent seal.

core/vm/context.go
mcortesi: At least you could wait until i finish 1.9 merge :P

mcortesi: Is there a way to just "link" things from the original package?

PR #773, opened 2019-12-18
Add garbage collection for RoundStateDB

### LINKED ISSUE #764: Too many round states are persisted
### Expected Behavior

Less states persisted long-term, smaller directory

### Current Behavior
After just one week of validator 0 on baklava:

```
~/.celo/geth# du -h
1.3G	./chaindata
92K	./lightchaindata
92K	./ultralightchaindata
20K	./nodes
2.0M	./validatorenodes
4.5G	./roundstates
5.7G	.
```

mcortesi:
### Description

To implement this we had to change the key format on levelDB; so that
Range queries over istanbul.View are possible.

To do so, a View is serialized to a byte[] as 2 uint64 (pair sequence,
round)

Also, garbage collection is optional and is turned on using constructor
options within the newRoundStateDB()

Finally, to implement a recurrent task, a new common/task package was
created. This makes it easy to test the recurrent task logic and it can
be also used in many other places since it's a quite common patter
within geth.

**Notes for the Reviewers**: 
  * pls check the view2key() serialization. It's limited to uint64 and will completely break if negative number for round/seq are used.
  * on calling task.stop() this won't interrupt the current run if it's already running, nor stop() will wait for that run to finish... probably i need a WaitGroup for that.

### Tested

 * unit tests

### Other changes

* Adds common/task package to launch recurring tasks

### Related issues

- Fixes #764 

### Backwards compatibility

New rounddbVersion. It will automatically delete and create a new roundstateDB

### DIFF
--- a/common/task/task.go
+++ b/common/task/task.go
@@ -5,7 +5,7 @@ import "time"
 type StopFn func()
 
 func RunTaskRepeateadly(task func(), period time.Duration) StopFn {
-	// Setup the ticket and the channel to signal
+	// Setup the ticker and the channel to signal
 	// the ending of the interval
 	ticker := time.NewTicker(period)
 	stop := make(chan struct{})
@@ -23,6 +23,7 @@ func RunTaskRepeateadly(task func(), period time.Duration) StopFn {
 	}()
 
 	return func() {
+		// stop <- struct{}{}
 		close(stop)
 	}
 }
--- a/common/task/task.go
+++ b/common/task/task.go
@@ -1,0 +1,28 @@
+package task
+
+import "time"
+
+type StopFn func()
+
+func RunTaskRepeateadly(task func(), period time.Duration) StopFn {
+	// Setup the ticket and the channel to signal
+	// the ending of the interval
+	ticker := time.NewTicker(period)
+	stop := make(chan struct{})
+
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				task()
+			case <-stop:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+
+	return func() {
+		close(stop)
+	}
+}
--- a/common/task/task.go
+++ b/common/task/task.go
@@ -23,7 +23,6 @@ func RunTaskRepeateadly(task func(), period time.Duration) StopFn {
 	}()
 
 	return func() {
-		// stop <- struct{}{}
-		close(stop)
+		stop <- struct{}{}
 	}
 }
--- a/consensus/istanbul/core/roundstate_db.go
+++ b/consensus/istanbul/core/roundstate_db.go
@@ -233,7 +233,7 @@ func (rsdb *roundStateDBImpl) Close() error {
 }
 
 func (rsdb *roundStateDBImpl) garbageCollectEntries() {
-	logger := rsdb.logger.New("func", "StartGarbageCollector")
+	logger := rsdb.logger.New("func", "garbageCollectEntries")
 
 	oldestValidView, err := rsdb.GetOldestValidView()
 	if err != nil {
@@ -241,14 +241,14 @@ func (rsdb *roundStateDBImpl) garbageCollectEntries() {
 		return
 	}
 
-	logger.Debug("Prunning entries from old views", "oldestValidView", oldestValidView)
+	logger.Debug("Pruning entries from old views", "oldestValidView", oldestValidView)
 	count, err := rsdb.deleteEntriesOlderThan(oldestValidView)
 	if err != nil {
 		logger.Error("Aborting RoundStateDB GarbageCollect: Failed to remove entries", "entries_removed", count, "err", err)
 		return
 	}
 
-	logger.Info("Finished RoundStateDB GarbageCollect", "removed_entries", count)
+	logger.Debug("Finished RoundStateDB GarbageCollect", "removed_entries", count)
 }
 
 func (rsdb *roundStateDBImpl) deleteEntriesOlderThan(lastView *istanbul.View) (int, error) {
@@ -273,6 +273,12 @@ func (rsdb *roundStateDBImpl) deleteEntriesOlderThan(lastView *istanbul.View) (i
 // view2Key will encode a view in binary format
 // so that the binary format maintains the sort order for the view
 func view2Key(view *istanbul.View) []byte {
+	// leveldb sorts entries by key
+	// keys are sorted with their binary representation, so we need a binary representation
+	// that mantains the key order
+	// The key format is [ prefix . BigEndian(Sequence) . BigEndian(Round)]
+	// We use BigEndian so to maintain order in binary format
+	// And we want to sort by (seq, round); since seq had higher precedence than round
 	prefix := []byte(rsKey)
 	buff := make([]byte, len(prefix)+16)
 
--- a/consensus/istanbul/core/roundstate_db.go
+++ b/consensus/istanbul/core/roundstate_db.go
@@ -19,8 +19,12 @@ package core
 import (
 	"bytes"
 	"encoding/binary"
+	"math/big"
 	"os"
+	"time"
 
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/task"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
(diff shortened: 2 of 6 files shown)


common/task/task.go
```diff
@@ -0,0 +1,28 @@
...
+				task()
+			case <-stop:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+
+	return func() {
+		close(stop)
```
tkporter: I think calling this function won't ever cause the `ticker.Stop()` block to be run, I played around with this here: https://play.golang.org/p/DmOIKBLPLFR

consensus/istanbul/core/roundstate_db.go
```diff
@@ -146,19 +191,28 @@ func (rsdb *roundStateDBImpl) GetLastView() (*istanbul.View, error) {
...
+	} else if err != nil {
 		return nil, err
 	}
 
+	oldestValidSequence := new(big.Int).Sub(lastView.Sequence, new(big.Int).SetUint64(rsdb.opts.sequencesToSave))
+	if oldestValidSequence.Cmp(common.Big0) < 0 {
+		oldestValidSequence = common.Big0
+	}
+
+	return &istanbul.View{Sequence: oldestValidSequence, Round: common.Big0}, nil
```
tkporter: is the round as 0 intentional?

consensus/istanbul/core/roundstate_db.go
```diff
@@ -172,5 +226,70 @@ func (rsdb *roundStateDBImpl) GetRoundStateFor(view *istanbul.View) (RoundState,
...
+func (rsdb *roundStateDBImpl) garbageCollectEntries() {
+	logger := rsdb.logger.New("func", "StartGarbageCollector")
+
+	oldestValidView, err := rsdb.GetOldestValidView()
+	if err != nil {
+		logger.Error("Aborting RoundStateDB GarbageCollect: Failed to fetch oldestValidView", "err", err)
+		return
+	}
+
+	logger.Debug("Prunning entries from old views", "oldestValidView", oldestValidView)
```
tkporter: nit: prunning -> pruning

consensus/istanbul/core/roundstate_db.go
```diff
@@ -172,5 +226,70 @@ func (rsdb *roundStateDBImpl) GetRoundStateFor(view *istanbul.View) (RoundState,
...
+	}
+
+	logger.Info("Finished RoundStateDB GarbageCollect", "removed_entries", count)
+}
+
+func (rsdb *roundStateDBImpl) deleteEntriesOlderThan(lastView *istanbul.View) (int, error) {
+	fromViewKey := view2Key(&istanbul.View{Sequence: common.Big0, Round: common.Big0})
+	toViewKey := view2Key(lastView)
+
+	iter := rsdb.db.NewIterator(&util.Range{Start: fromViewKey, Limit: toViewKey}, nil)
```
tkporter: Very cool that leveldb can do this

consensus/istanbul/core/roundstate_db.go
```diff
@@ -172,5 +226,70 @@ func (rsdb *roundStateDBImpl) GetRoundStateFor(view *istanbul.View) (RoundState,
...
+
+// view2Key will encode a view in binary format
+// so that the binary format maintains the sort order for the view
+func view2Key(view *istanbul.View) []byte {
+	prefix := []byte(rsKey)
+	buff := make([]byte, len(prefix)+16)
+
+	copy(buff, prefix)
+	// TODO (mcortesi) Support Seq/Round bigger than 64bits
+	binary.BigEndian.PutUint64(buff[len(prefix):], view.Sequence.Uint64())
```
tkporter: This took me a little while to picture in my head & understand this order, I think a comment that leveldb evaluates ranges of `[]byte` as one big big-endian string of bits is why we have our key structured as `[big endian sequence . big endian round ]` would be helpful to a reader in the future

We never set the sequence or round to be negative for whatever reason do we? Personally I think Uint64 is huge enough that we won't have to realistically worry about getting close to the max value

consensus/istanbul/core/roundstate_db.go
```diff
@@ -19,35 +19,75 @@ package core
...
-	db *leveldb.DB
+	db                   *leveldb.DB
+	stopGarbageCollector task.StopFn
+	opts                 RoundStateDBOptions
+	logger               log.Logger
+}
+
+var defaultRoundStateDBOptions = RoundStateDBOptions{
+	withGarbageCollector:   true,
+	sequencesToSave:        100,
```
asaj: For the future, it will need to save sequences where a validator committed multiple blocks for far longer, namely the duration of the statute of limitations. I think it's fine to save this for when that type of slashing gets implemented.

consensus/istanbul/core/roundstate_db.go
```diff
@@ -19,35 +19,75 @@ package core
...
+	garbageCollectorPeriod: 2 * time.Minute,
 }
 
-func newRoundStateDB(path string) (RoundStateDB, error) {
+func coerceOptions(opts *RoundStateDBOptions) RoundStateDBOptions {
```
asaj: nit: elsewhere in the codebase I believe this is `setDefaults`, but only in one place so I'm not sure if it's a pattern

consensus/istanbul/core/roundstate_db.go
```diff
@@ -172,5 +226,70 @@ func (rsdb *roundStateDBImpl) GetRoundStateFor(view *istanbul.View) (RoundState,
...
 
 func (rsdb *roundStateDBImpl) Close() error {
+	if rsdb.opts.withGarbageCollector {
+		rsdb.stopGarbageCollector()
+	}
 	return rsdb.db.Close()
 }
+
+func (rsdb *roundStateDBImpl) garbageCollectEntries() {
+	logger := rsdb.logger.New("func", "StartGarbageCollector")
```
asaj: nit: should this be `"garbageCollectEntries"`?

consensus/istanbul/core/roundstate_db.go
asaj: nit: "Pruning"

consensus/istanbul/core/roundstate_db.go
```diff
@@ -172,5 +226,70 @@ func (rsdb *roundStateDBImpl) GetRoundStateFor(view *istanbul.View) (RoundState,
...
+	}
+
+	logger.Debug("Prunning entries from old views", "oldestValidView", oldestValidView)
+	count, err := rsdb.deleteEntriesOlderThan(oldestValidView)
+	if err != nil {
+		logger.Error("Aborting RoundStateDB GarbageCollect: Failed to remove entries", "entries_removed", count, "err", err)
+		return
+	}
+
+	logger.Info("Finished RoundStateDB GarbageCollect", "removed_entries", count)
```
asaj: Debug?

consensus/istanbul/core/roundstate_db.go
```diff
@@ -172,5 +226,70 @@ func (rsdb *roundStateDBImpl) GetRoundStateFor(view *istanbul.View) (RoundState,
...
+}
+
+// view2Key will encode a view in binary format
+// so that the binary format maintains the sort order for the view
+func view2Key(view *istanbul.View) []byte {
+	prefix := []byte(rsKey)
+	buff := make([]byte, len(prefix)+16)
+
+	copy(buff, prefix)
+	// TODO (mcortesi) Support Seq/Round bigger than 64bits
```
asaj: At 5s blocks, and 2^n second rounds, 64 bits feels sufficient 😄

consensus/istanbul/core/roundstate_db_test.go
```diff
@@ -47,3 +50,114 @@ func TestRoundStateDB(t *testing.T) {
...
+	if count != 25 {
+		t.Fatalf("Expected 10 deleted entries but got %d", count)
+	}
+
+}
+
+func TestRSDBKeyEncodingOrder(t *testing.T) {
+	iterations := 1000
+
+	t.Run("ViewKey Enconding should decode the same view", func(t *testing.T) {
```
asaj: s/Enconding/encoding

consensus/istanbul/core/roundstate_db.go
mcortesi: Yes, basically I'm want to say that all round state beginning on seq = `oldestValidSequence` are valid. So, round `0` and above

consensus/istanbul/core/roundstate_db.go
mcortesi: Yes! key are sorted, and you can define the comparator... but soring happens at byte level, so i had to think about sort order when the key is serialized

consensus/istanbul/core/roundstate_db.go
mcortesi: I agree this deserves a better comment!

common/task/task.go
mcortesi: you're right...
closing a channel is a way to stop things but not here!
Thanks!

PR #778, opened 2019-12-26
IBFT: Add ParentSeal on Finalize (not in Prepare)

### LINKED ISSUE #777: Generate ParentCommit seal on Engine.Finalize()
### Expected Behavior

ParentCommit should be created as one of the final things when a block is assembled. That's the finalize step, not the prepare. So we should refactor code to include it in finalize.

### Current Behavior

It's assembled on Prepare step

mcortesi:
### Description

Refactor istanbul engine so that the "ParentSeal" is added to the header
in the Finalize Step, instead of the Prepare step.

It extracts logic to it's own function to make it more readable, and
switched to a style of `if err => return` instead of relying on nested
ifs.

### Tested

 * CI

### Other changes

 * Refactor code

### Related issues

- Fixes #777

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -400,7 +400,7 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 	delay := time.Unix(header.Time.Int64(), 0).Sub(now())
 	time.Sleep(delay)
 
-	return nil
+	return sb.addParentSeal(chain, header)
 }
 
 // UpdateValSetDiff will update the validator set diff in the header, if the mined header is the last block of the epoch
@@ -446,13 +446,8 @@ func (sb *Backend) LookbackWindow() uint64 {
 // consensus rules that happen at finalization (e.g. block rewards).
 func (sb *Backend) Finalize(chain consensus.ChainReader, header *types.Header, state *state.StateDB, txs []*types.Transaction, uncles []*types.Header, receipts []*types.Receipt, randomness *types.Randomness) (*types.Block, error) {
 
-	err := sb.addParentSeal(chain, header)
-	if err != nil {
-		return nil, err
-	}
-
 	snapshot := state.Snapshot()
-	err = sb.setInitialGoldTokenTotalSupplyIfUnset(header, state)
+	err := sb.setInitialGoldTokenTotalSupplyIfUnset(header, state)
 	if err != nil {
 		state.RevertToSnapshot(snapshot)
 	}
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -400,73 +400,6 @@ func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 	delay := time.Unix(header.Time.Int64(), 0).Sub(now())
 	time.Sleep(delay)
 
-	logger := sb.logger.New("func", "Backend.Prepare()", "number", number)
-	// modify the block header to include all the ParentCommits
-	// only do this for blocks which start with block 1 as a parent
-	if number > 1 {
-
-		// In some cases, "Prepare" may be called before sb.core has moved to the next sequence,
-		// preventing signature aggregation.
-		// This typically happens in round > 0, since round 0 typically hits the "time.Sleep()"
-		// above.
-		// When this happens, loop until sb.core moves to the next sequence, with a limit of 500ms.
-		waitForSequenceChange := func() *big.Int {
-			timeout := time.After(500 * time.Millisecond)
-			ticker := time.NewTicker(10 * time.Millisecond)
-			defer ticker.Stop()
-			for {
-				select {
-				case <-ticker.C:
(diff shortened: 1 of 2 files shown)


consensus/istanbul/backend/engine.go
```diff
@@ -882,6 +820,62 @@ func (sb *Backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+func (sb *Backend) addParentSeal(chain consensus.ChainReader, header *types.Header) error {
+	number := header.Number.Uint64()
+	logger := sb.logger.New("func", "Backend.addParentSeal()", "number", number)
+
+	// only do this for blocks which start with block 1 as a parent
+	if number <= 1 {
+		return nil
+	}
+
+	// In some cases, "Prepare" may be called before sb.core has moved to the next sequence,
```
trianglesphere: nit: `addParentSeal` is now called from "Finalize"

consensus/istanbul/backend/engine.go
```diff
@@ -882,6 +820,62 @@ func (sb *Backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+	// preventing signature aggregation.
+	// This typically happens in round > 0, since round 0 typically hits the "time.Sleep()"
+	// above.
+	// When this happens, loop until sb.core moves to the next sequence, with a limit of 500ms.
+	seq := waitCoreToReachSequence(sb.core, header.Number)
```
trianglesphere: Is this still strictly necessary? I could be wrong, but finalize should be called once the block is committed and the validator moves onto the next sequence.

Note: this probably doesn't hurt to keep it in.

consensus/istanbul/backend/engine.go
```diff
@@ -882,6 +820,62 @@ func (sb *Backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 	return returnSnap, nil
 }
 
+func (sb *Backend) addParentSeal(chain consensus.ChainReader, header *types.Header) error {
+	number := header.Number.Uint64()
+	logger := sb.logger.New("func", "Backend.addParentSeal()", "number", number)
+
+	// only do this for blocks which start with block 1 as a parent
+	if number <= 1 {
+		return nil
```
trianglesphere: Confirming that if this fails we want to return `nil` and not include the parent seal in the block.

mcortesi: Well, not completely sure about the answer here.

For starters, the PR only does code rearrange for this, so it was not being written before. 

Now, wether it should be written, idk. In fact, that's not the only case...

Error handling strategy is not particularly consistent here...

This what we do on errors/checks:

 1. number <= 1 => do nothing
 2. can't find seq => do nothing
 3. can't access parentBlock.seal => do nothing
 4. failed or empty core.ParentCommits => do nothing (before was "write parentBlock.seal")
 5. failure on call UnionOfSeals  => do nothing (before was "write parentBlock.seal")
 6. verification failure of union => do nothing (before was "write parentBlock.seal")

I've intruded a change on 4,5,6 which seemed consistent, but not completely sure. My guess was that we should write one for EVERY case; but that wasn't the case.

trianglesphere: Seems reasonable for now.

asaj: Blocks 0 and 1 cannot have parent seals, as there's no block -1, and the genesis block isn't signed.

Essentially, the block always needs 2F+1 signatures in the parent seal, which is what `3` is supposed to guarantee. If the 2F+1 signatures from the previous block are available via `3`, they should always be written to the block header, whether or not they're able to be combined with the (possibly) additional `ParentCommits` from `core`.

It's fine for failure on `2`, `4`, `5`, and `6` to return `nil` so long as those 2F+1 signatures from fetching the `parentExtra.AggregatedSeal` are added to the header being finalized.

It looks like the pattern here is to return early, it be easier to assign a `parentAggregatedSeal` variable via `3` and then re-assign to the union if `2`, `4`, `5`, and `6` are successful.

mcortesi: Ok, so i've changed the code to reflect 4 different outputs for addParentSeal

 1. block number <=1 => do nothing
 2. failed to fetch parentHeader.AggregatedSeal => return error
 3. failed to generate union of seals => use parentHeader.AggregatedSeal
 as ParentSeal
 4. All ok => use UnionOfSeal as ParentSeal

Basically, we do our best effort to always guarantee an AggregatedSeal to be written for the header.ParentSeal; wether is the original parentHeader.AggregatedSeal or a unionOfSeals. The only case where de don't do that, is when (a) block number <-=1 or (b) failed to fetch parentHeader.AggregatedSeal.

consensus/istanbul/backend/engine.go
```diff
@@ -882,6 +820,68 @@ func (sb *Backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+	}
+
+	// Get parent's extra to fetch it's AggregatedSeal
```
asaj: nit: "its"

PR #780, opened 2020-01-03
Adds SNARK-friendly epoch encoding

kobigurk:
### Description

This adds the SNARK-friendly epoch encoding signed by a super-majority, required to create proofs to compress a big chunk of history. 
It's added as part of the block so it's implicitly verified by validators, and not by full nodes or light client, allowing for easier upgardes of it.

The SNARK-friendly encoding is used by provers to create a short proof attesting to the validity of a large chunk of the header chain - it receives as input a validator set at block X and outputs the validator set at block X + N, by verifying the validator set evolution inside a SNARK.

### Other changes

Makes BLSPublicKey fixed size.

### Backwards compatibility

Not backwards compatible.


kevjue:
> It's added as part of the block so it's implicitly verified by validators, and not by full nodes or light client, allowing for easier upgardes of it.

Out of curiosity, if a full node modifies that that encoding and then starts to share it, would it be easy to detect?

kobigurk:
Yes, the signature won’t verify (it’s also signed by a majority).

consensus/istanbul/core/commit.go
```diff
@@ -45,7 +45,33 @@ func (c *core) generateCommittedSeal(sub *istanbul.Subject) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return committedSeal, nil
+	return committedSeal[:], nil
```
kevjue: Should the `generateCommittedSeal` function also return a `blscrypto.SerializedSignature` type, as opposed to a `[]byte` type?

consensus/istanbul/backend/backend.go
```diff
@@ -441,7 +445,7 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
 
 	err = sb.VerifyHeader(sb.chain, block.Header(), false)
 
-	// ignore errEmptyCommittedSeals error because we don't have the committed seals yet
+	// ignore errEmptyAggregatedSeal error because we don't have the committed seals yet
```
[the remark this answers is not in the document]
kobigurk: Yep, just the name changed a while ago

consensus/istanbul/backend/backend.go
```diff
@@ -549,15 +553,26 @@ func (sb *Backend) Sign(data []byte) ([]byte, error) {
...
 }
 
+func (sb *Backend) SignEpochSnarkData(data []byte) (blscrypto.SerializedSignature, error) {
+	if sb.signMessageBLSFn == nil {
+		return blscrypto.SerializedSignature{}, errInvalidSigningFn
+	}
+	sb.signFnMu.RLock()
+	defer sb.signFnMu.RUnlock()
+	// Currently, ExtraData is unused. In the future, it could include data that could be used to introduce
+	// "firmware-level" protection
```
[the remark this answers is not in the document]
kobigurk: Yes, of course!

consensus/istanbul/backend/backend_test.go
```diff
@@ -138,7 +140,8 @@ func TestCommit(t *testing.T) {
 		{
 			// invalid signature
 			errInvalidAggregatedSeal,
-			nil,
+			[]byte{},
```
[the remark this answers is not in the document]
kobigurk: Relic from when I used fixed size array, thanks

crypto/bls/bls.go
```diff
@@ -12,11 +12,14 @@ import (
...
 )
 
-var (
+const (
 	PUBLICKEYBYTES = bls.PUBLICKEYBYTES
 	SIGNATUREBYTES = bls.SIGNATUREBYTES
 )
 
+type SerializedPublicKey [PUBLICKEYBYTES]byte
+type SerializedSignature [SIGNATUREBYTES]byte
```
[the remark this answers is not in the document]
kobigurk: I want to differentiate the PublicKey and Signature structs from the byte arrays that represent them

consensus/istanbul/core/commit.go
```diff
@@ -45,7 +45,33 @@ func (c *core) generateCommittedSeal(sub *istanbul.Subject) ([]byte, error) {
...
+}
+
+func (c *core) generateEpochSeal() ([]byte, error) {
+	// TODO(kobi): is this the correct handling?
+	if c.current.Proposal() == nil {
+		return []byte{}, nil
+	}
+	currentBlockNumber := c.current.Proposal().Number().Uint64()
+	if istanbul.IsLastBlockOfEpoch(currentBlockNumber, c.config.Epoch) {
```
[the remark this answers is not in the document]
kobigurk: Right!

crypto/bls/bls.go
```diff
@@ -158,3 +164,23 @@ func VerifySignature(publicKey []byte, message []byte, extraData []byte, signatu
...
+
+		pubKeys = append(pubKeys, publicKeyObj)
+	}
+	apk, err := bls.AggregatePublicKeys(pubKeys)
+	if err != nil {
+		return nil, err
+	}
+	defer apk.Destroy()
+
+	return bls.EncodeEpochToBytes(epochIndex, maximumNonSignersPlusOne, apk, pubKeys)
```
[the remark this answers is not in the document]
kobigurk: Yes, I'll change the inner function, thanks

core/types/istanbul.go
```diff
@@ -50,6 +50,11 @@ type IstanbulAggregatedSeal struct {
 	Round *big.Int
 }
 
+type IstanbulAggregatedEpochSeal struct {
```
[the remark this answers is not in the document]
kobigurk: True! I'll add a comment

consensus/istanbul/backend/engine_test.go
```diff
@@ -812,20 +821,4 @@ func TestWriteAggregatedSeal(t *testing.T) {
...
-	}
-	err = writeAggregatedSeal(h, invalidAggregatedSeal, false)
-	if err != errInvalidAggregatedSeal {
-		t.Errorf("error mismatch: have %v, want %v", err, errInvalidAggregatedSeal)
-	}
-
-	err = writeAggregatedSeal(h, invalidAggregatedSeal, true)
-	if err != errInvalidAggregatedSeal {
-		t.Errorf("error mismatch: have %v, want %v", err, errInvalidAggregatedSeal)
-	}
```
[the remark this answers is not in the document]
kobigurk: Oh, it is :/ I removed it when I tried using fixed size array. Thanks for noticing.

consensus/istanbul/backend.go
```diff
@@ -53,15 +58,16 @@ type Backend interface {
...
 
 	// Verify verifies the proposal. If a consensus.ErrFutureBlock error is returned,
 	// the time difference of the proposal and current time is also returned.
 	Verify(Proposal) (time.Duration, error)
 
 	// Sign signs input data with the backend's private key
 	Sign([]byte) ([]byte, error)
-	SignBlockHeader([]byte) ([]byte, error)
+	SignBlockHeader([]byte) (blscrypto.SerializedSignature, error)
+	SignEpochSnarkData([]byte) (blscrypto.SerializedSignature, error)
```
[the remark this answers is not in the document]
kobigurk: Good idea. It should convey that it's a special hash, but not specifically epoch snark data.

consensus/istanbul/core/commit.go
```diff
@@ -39,15 +39,40 @@ func (c *core) sendCommitForOldBlock(view *istanbul.View, digest common.Hash) {
...
+	currentBlockNumber := c.current.Proposal().Number().Uint64()
+	if !istanbul.IsLastBlockOfEpoch(currentBlockNumber, c.config.Epoch) {
+    return blscrypto.SerializedSignature{}, nil
+  }
+  valSet := c.current.ValidatorSet()
+  blsPubKeys := []blscrypto.SerializedPublicKey{}
+  for _, v := range valSet.List() {
+    blsPubKeys = append(blsPubKeys, v.BLSPublicKey())
+  }
+  epochData, err := blscrypto.EncodeEpochSnarkData(blsPubKeys, uint32(valSet.F()+1), uint16(currentBlockNumber/c.config.Epoch))
```
kevjue: You can use this function to calculate the current epoch number:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/utils.go#L90

kobigurk: Ah, useful!

consensus/istanbul/core/core.go
```diff
@@ -230,6 +237,28 @@ func GetAggregatedSeal(seals MessageSet, round *big.Int) (types.IstanbulAggregat
 	return types.IstanbulAggregatedSeal{Bitmap: bitmap, Signature: asig, Round: round}, nil
 }
 
+// GetAggregatedEpochSeal aggregates all the given seals for a the SNARK-friendly epoch encoding
+// to a bls aggregated signature and bitmap
+func GetAggregatedEpochSeal(seals MessageSet, round *big.Int) (types.IstanbulAggregatedEpochSeal, error) {
```
asaj: What is the round used for here?

In the regular aggregated seal, validators sign the block hash + round. There, it's useful for signatures to be specific to a certain view (round and sequence) for detecting double signing.

contract_comm/validators/validators.go
```diff
@@ -237,13 +237,17 @@ func GetValidatorData(header *types.Header, state vm.StateDB, validatorAddresses
 		if err != nil {
 			return nil, err
 		}
+
 		if len(blsKey) != blscrypto.PUBLICKEYBYTES {
-			return nil, fmt.Errorf("length of BLS key incorrect. Expected %d, got %d", blscrypto.PUBLICKEYBYTES, len(blsKey))
+			return nil, fmt.Errorf("length of bls public key incorrect. Expected %d, got %d", blscrypto.PUBLICKEYBYTES, len(blsKey))
 		}
-		validatorData = append(validatorData, istanbul.ValidatorData{
+		blsKeyFixed := blscrypto.SerializedPublicKey{}
```
asaj: Why "Fixed"?

crypto/bls/bls.go
```diff
@@ -158,3 +164,32 @@ func VerifySignature(publicKey []byte, message []byte, extraData []byte, signatu
 	err = publicKeyObj.VerifySignature(message, extraData, signatureObj, shouldUseCompositeHasher)
 	return err
 }
+
+func EncodeEpochSnarkData(newValSet []SerializedPublicKey, maximumNonSignersPlusOne uint32, epochIndex uint16) ([]byte, error) {
```
asaj: What is `maximumNonSignersPlusOne` used for? Note that in some cases, `F+1 != ValSet.Size() - MinQuorumSize() + 1`, e.g. a val set of size 9 will have a min quorum size of 6 but an F of 2.

consensus/istanbul/core/commit.go
```diff
@@ -30,15 +30,36 @@ func (c *core) sendCommit() {
...
 	}
 	return committedSeal, nil
 }
 
+func (c *core) generateEpochSeal() (blscrypto.SerializedSignature, error) {
+	currentBlockNumber := c.current.Proposal().Number().Uint64()
+	if !istanbul.IsLastBlockOfEpoch(currentBlockNumber, c.config.Epoch) {
+		return blscrypto.SerializedSignature{}, nil
+	}
+	valSet := c.current.ValidatorSet()
```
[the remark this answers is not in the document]
kobigurk: I'm pretty sure it isn't :/

contract_comm/validators/validators.go
kobigurk: Fixed size

crypto/bls/bls.go
kobigurk: The SNARK needs it, but your way remind me it was the wrong way to do it. Thank.

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
kobigurk: Fixed!

consensus/istanbul/core/commit.go
```diff
@@ -30,15 +31,43 @@ func (c *core) sendCommit() {
...
 	seal := PrepareCommittedSeal(sub.Digest, sub.View.Round)
 	committedSeal, err := c.backend.SignBlockHeader(seal)
 	if err != nil {
-		return nil, err
+		return blscrypto.SerializedSignature{}, err
 	}
 	return committedSeal, nil
 }
 
+func (c *core) generateEpochValidatorSetSeal() (blscrypto.SerializedSignature, error) {
```
asaj: We need a `verifyEpochValidatorSetSeal` just like `verifyCommittedSeal`. Otherwise, a single validator would be able to sabotage consensus by sending a commit with an invalid `EpochValidatorSetSeal`, which would be admitted and cause `GetAggregatedEpochSeal` to fail.

consensus/istanbul/core/core.go
```diff
@@ -191,7 +191,14 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
-		if err := c.backend.Commit(proposal, aggregatedSeal); err != nil {
+		aggregatedEpochSeal, err := GetAggregatedEpochSeal(c.current.Commits())
```
asaj: `EpochValidatorSetSeal` here and everywhere for consistency

consensus/istanbul/core/commit.go
kobigurk: I'm pretty sure we had a similar discussion about it in the past, where we also added this to normal signatures. 
I agree we should add it, but I don't think it would fail right now more than it would fail with the added check, since what we do now is basically just add points.

consensus/istanbul/backend/backend.go
```diff
@@ -255,6 +256,30 @@ func (sb *Backend) ParentBlockValidators(proposal istanbul.Proposal) istanbul.Va
...
+	if err != nil {
+		return nil, err
+	}
+
+	// There was no change
+	if len(istExtra.AddedValidators) == 0 && istExtra.RemovedValidators.BitLen() == 0 {
+		return sb.ParentBlockValidators(proposal), nil
+	}
+
+	newValSet, err := sb.getNewValidatorSet(proposal.Header(), state)
```
nategraf: It seems that this code already implicitly trusts the header input, so it may be better to use the snapshot functions to construct the validator set here. Namely by calling `snap := sb.snapshot(...)` to get the snapshot for the parent of the proposal and using `snap.apply(...)` to add the validator set from the given proposal header. `getNewValidatorSet` calls into a very expensive EVM call (it uses more than the block gas limit).

https://github.com/celo-org/celo-blockchain/blob/4ea7566836aee9cdbaa6c4561c3f41b9d79037c7/consensus/istanbul/backend/engine.go#L673
https://github.com/celo-org/celo-blockchain/blob/4ea7566836aee9cdbaa6c4561c3f41b9d79037c7/consensus/istanbul/backend/snapshot.go#L94
https://github.com/celo-org/celo-blockchain/blob/4ea7566836aee9cdbaa6c4561c3f41b9d79037c7/params/protocol_params.go#L164

consensus/istanbul/core/commit.go
```diff
@@ -30,15 +31,43 @@ func (c *core) sendCommit() {
...
+	newValSet, err := c.backend.NextBlockValidators(c.current.Proposal())
+	if err != nil {
+		return blscrypto.SerializedSignature{}, err
+	}
+	blsPubKeys := []blscrypto.SerializedPublicKey{}
+	for _, v := range newValSet.List() {
+		blsPubKeys = append(blsPubKeys, v.BLSPublicKey())
+	}
+	fmt.Printf("HERR current valset: %d, %d, %d, %v\n", valSet.Size(), valSet.MinQuorumSize(), uint32(valSet.Size()-valSet.MinQuorumSize()+1), valSet.List())
+	maxNonSignersPlusOne := uint32(newValSet.Size() - newValSet.MinQuorumSize() + 1)
```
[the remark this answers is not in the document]
nategraf: Just saw https://github.com/celo-org/celo-blockchain/pull/803/files which answers so many questions!

crypto/bls/bls.go
```diff
@@ -120,7 +126,7 @@ func VerifyAggregatedSignature(publicKeys [][]byte, message []byte, extraData []
 func AggregateSignatures(signatures [][]byte) ([]byte, error) {
 	signatureObjs := []*bls.Signature{}
 	for _, signature := range signatures {
-		signatureObj, err := bls.DeserializeSignature(signature)
+		signatureObj, err := bls.DeserializeSignature(signature[:])
```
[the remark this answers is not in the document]
kobigurk: Nope, just a relic

consensus/istanbul/backend/backend.go
kobigurk: `apply` wasn't usable for me since it requires an `Epoch` long jump.
I did utilize the method I learnt from it though!

consensus/istanbul/core/commit.go
asaj: I'm not sure I understand

kobigurk: FYI, I already added these checks!
I can see some problematic scenarios without them.

consensus/istanbul/core/commit.go
```diff
@@ -48,9 +65,30 @@ func (c *core) broadcastCommit(sub *istanbul.Subject) {
 		return
 	}
 
+	currentBlockNumber := c.current.Proposal().Number().Uint64()
+	newValSet, err := c.backend.NextBlockValidators(c.current.Proposal())
+	if err != nil {
+		logger.Error("Failed to commit epoch validator set seal", "err", err)
```
asaj: nit: "Failed to get next block's validators"

consensus/istanbul/core/core.go
```diff
@@ -306,6 +313,28 @@ func (c *core) commit() error {
 	return nil
 }
 
+// GetAggregatedEpochValidatorSetSeal aggregates all the given seals for a the SNARK-friendly epoch encoding
+// to a bls aggregated signature and bitmap
```
asaj: We don't actually use a bitmap here, do we?

consensus/istanbul/core/commit.go
```diff
@@ -30,15 +30,32 @@ func (c *core) sendCommit() {
...
 
+func (c *core) generateEpochValidatorSetData(blockNumber uint64, newValSet istanbul.ValidatorSet) ([]byte, error) {
+	if !istanbul.IsLastBlockOfEpoch(blockNumber, c.config.Epoch) {
+		return nil, errNotLastBlockInEpoch
+	}
+	blsPubKeys := []blscrypto.SerializedPublicKey{}
+	for _, v := range newValSet.List() {
+		blsPubKeys = append(blsPubKeys, v.BLSPublicKey())
+	}
+	maxNonSignersPlusOne := uint32(newValSet.Size() - newValSet.MinQuorumSize() + 1)
```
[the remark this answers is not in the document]
kobigurk: I just need the number that tells me the threshold of the number of non-signers below which it's an acceptable signature. It's a different way of presenting the threshold of signers, which is something I need later on (and I don't have easy access to the ValSet size in the snark). I *think* this is the correct way to get it after @nategraf's comment, though would be happy for input.

consensus/istanbul/core/core.go
kobigurk: You're right!

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
asaj: IIUC you'd actually want newValSet.Size() - newValSet.MinQuorumSize(), without the +1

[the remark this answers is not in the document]
asaj: Discussed offline, sounds like the +1 is necessary as the circuit checks that the number of 0's in the bitmap is less than this number

PR #784, opened 2020-01-04
cache proposal verify result

kevjue:
### Description

This PR contains changes in the istanbul.core module to cache the proposal verify result, as a proposal could be verified multiple times within a round (from prepare certificates on the round change message).

### Tested

- Added new tests for the proposal verification cache
- Ran unit tests
- Ran e2e tests

### Other changes

* Modified some of the logging messages
* Fixed spelling for `persitOnNoError`
* Fixed flaky preprepare unit tests

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

is backwards compatible

### DIFF
--- a/consensus/istanbul/backend/pos.go
+++ b/consensus/istanbul/backend/pos.go
@@ -94,7 +94,7 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
 func (sb *Backend) updateValidatorScores(header *types.Header, state *state.StateDB, valSet []istanbul.Validator) ([]*big.Int, error) {
 	epoch := istanbul.GetEpochNumber(header.Number.Uint64(), sb.EpochSize())
 	logger := sb.logger.New("func", "Backend.updateValidatorScores", "blocknum", header.Number.Uint64(), "epoch", epoch, "epochsize", sb.EpochSize(), "window", sb.LookbackWindow())
-	sb.logger.Trace("Updating validator scores")
+	logger.Trace("Updating validator scores")
 
 	// The denominator is the (last block - first block + 1) of the val score tally window
 	denominator := istanbul.GetValScoreTallyLastBlockNumber(epoch, sb.EpochSize()) - istanbul.GetValScoreTallyFirstBlockNumber(epoch, sb.EpochSize(), sb.LookbackWindow()) + 1
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -555,8 +555,8 @@ func (c *core) Sequence() *big.Int {
 }
 
 func (c *core) verifyProposal(proposal istanbul.Proposal) (time.Duration, error) {
-	logger := c.newLogger("func", "verifyProposal", "proposal", proposal)
-	if verificationStatus, isChecked := c.current.GetProposalVerificationStatus(proposal.Hash()); isChecked {
+	logger := c.newLogger("func", "verifyProposal", "proposal", proposal.Hash())
+	if verificationStatus, isCached := c.current.GetProposalVerificationStatus(proposal.Hash()); isCached {
 		logger.Trace("verification status cache hit", "verificationStatus", verificationStatus)
 		return 0, verificationStatus
 	} else {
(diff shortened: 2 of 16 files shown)


consensus/istanbul/backend/pos.go
```diff
@@ -94,7 +94,7 @@ func (sb *Backend) distributeEpochPaymentsAndRewards(header *types.Header, state
 func (sb *Backend) updateValidatorScores(header *types.Header, state *state.StateDB, valSet []istanbul.Validator) ([]*big.Int, error) {
 	epoch := istanbul.GetEpochNumber(header.Number.Uint64(), sb.EpochSize())
 	logger := sb.logger.New("func", "Backend.updateValidatorScores", "blocknum", header.Number.Uint64(), "epoch", epoch, "epochsize", sb.EpochSize(), "window", sb.LookbackWindow())
-	sb.logger.Trace("Updating validator scores")
+	logger.Trace("Updating validator scores")
```
mcortesi: why use the root logger?

consensus/istanbul/core/roundstate.go
```diff
@@ -378,6 +387,31 @@ func (s *roundStateImpl) PreparedCertificate() istanbul.PreparedCertificate {
 	return s.preparedCertificate
 }
 
+func (s *roundStateImpl) SetProposalVerificationStatus(proposalHash common.Hash, verificationStatus error) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.proposalVerificationStatus == nil {
```
mcortesi: maybe it's simpler to initialize it on `New()` and then reset it on `StartNewSequence`

consensus/istanbul/core/roundstate.go
```diff
@@ -89,6 +91,12 @@ type roundStateImpl struct {
 	pendingRequest      *istanbul.Request
 	preparedCertificate istanbul.PreparedCertificate
 
+	// Verification status for proposals seen in this round
```
asaj: Round, view, or sequence?

consensus/istanbul/core/roundstate_save_decorator.go
```diff
@@ -37,7 +37,7 @@ type rsSaveDecorator struct {
 	db RoundStateDB
 }
 
-func (rsp *rsSaveDecorator) persitOnNoError(err error) error {
+func (rsp *rsSaveDecorator) persistOnNoError(err error) error {
```
mcortesi: hehe... my english

consensus/istanbul/core/core.go
```diff
@@ -552,3 +553,23 @@ func (c *core) Sequence() *big.Int {
 	}
 	return c.current.Sequence()
 }
+
+func (c *core) verifyProposal(proposal istanbul.Proposal) (time.Duration, error) {
+	logger := c.newLogger("func", "verifyProposal", "proposal", proposal)
```
asaj: Consider including the proposal hash rather than the full proposal?

consensus/istanbul/core/roundstate.go
```diff
@@ -378,6 +387,31 @@ func (s *roundStateImpl) PreparedCertificate() istanbul.PreparedCertificate {
 	return s.preparedCertificate
 }
 
+func (s *roundStateImpl) SetProposalVerificationStatus(proposalHash common.Hash, verificationStatus error) error {
```
mcortesi: You don't need to return `error`. That was done to allow persistence to fail, but we are not persisting this

consensus/istanbul/core/roundstate.go
mcortesi: And on `deserialize`

consensus/istanbul/backend/pos.go
kevjue: It's using the logger created in the line above that contains all of the contextual key value pairs.

consensus/istanbul/core/roundstate.go
kevjue: Do you have a strong opinion on doing it that way?

I actually initially was going to do that, but decided to create it on demand, since it would mean only needing to create in one place (as opposed to 3 different places).

mcortesi:
No strong opinion.
In an OO world i would prefer doing on initialize. But here it seems easy
to forget when creating the struct. And there are 2 places for that for now

PR #788, opened 2020-01-07
add istanbul_getProposer rpc method and remove a few others

nategraf:
### Description

This PR makes the following changes to the Istanbul RPC API to make accessible new information about the proposer for each sequence. Additionally methods for retrieving validator set and snapshot information by hash are removed because they are equivalent to looking up the block by hash then calling the respective method with the block number because the hash is not actually used in snapshot retrieval.

* Add `GetProposer(sequence *rpc.BlockNumber, round *uint64)`
* Remove `GetValidatorsAtHash` and `GetSnapshotAtHash`
* Change `GetValidators` to return the validators that must sign the given block.

### Tested

Manually testing on a testnet which `geth attach` and executing the changes RPC methods.

### Other changes

* Remove a mutability bug in `consensus/istanbul/core/core.go` which could result in building a block proposal with invalid parent seal.
* Fix an outdated comment in `consensus/istanbul/backend/backend.go`

### Backwards compatibility

Breaks usage of the `istanbul_getSnapshotAtHash` and `istanbul_getValidatorsAtHash` RPC methods, which are not known to be in use.

### DIFF
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
 	vet "github.com/ethereum/go-ethereum/consensus/istanbul/backend/internal/enodes"
+	"github.com/ethereum/go-ethereum/consensus/istanbul/validator"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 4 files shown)


consensus/istanbul/backend/api.go
```diff
@@ -35,66 +36,78 @@ type API struct {
...
-// GetValidatorsAtHash retrieves the state snapshot at a given block.
-func (api *API) GetValidatorsAtHash(hash common.Hash) ([]common.Address, error) {
-	header := api.chain.GetHeaderByHash(hash)
-	if header == nil {
-		return nil, errUnknownBlock
+// GetProposer retrieves the proposer for a given block number (i.e. sequence) and round.
+func (api *API) GetProposer(sequence *rpc.BlockNumber, round *uint64) (common.Address, error) {
+	header, err := api.getParentHeaderByNumber(sequence)
+	if err != nil {
+		return common.Address{}, err
```
mcortesi: there's `common.ZeroAddress`. I'm never sure what would be a good null value...
i guess ZeroAddress but not sure

nategraf: Empty value for the address I think is the best null value. It also happens to be equal to ZeroAddress. Honestly not sure why they have that constant.

PR #789, opened 2020-01-08
Increase MasGas for getRegisteredValidatorSigners

nambrot:
### Description

With the number of registered validators increased, our original `MaxGasForGetRegisteredValidators` no longer is sufficient. In this PR, I doubled it.

### Tested

- [x] On my validator


timmoreton:
would be reasonable to increase it further, but fine for now!

PR #797, opened 2020-01-08
changed announce message frequency

### LINKED ISSUE #790: Geth SBAT use less network bandwidth for announce mechanism
### Expected Behavior

Geth SBAT to use less network bandwidth for announce mechanism, since some existing validators' ISP is stopping their internet service because of too much network usage.  Specifically, registered validators should send the existing announce message at the following events:

1)  Once every 10 minutes
2)  Istanbul engine start (This is on the assumption that this gets started when the node becomes synced.  If that assumption is wrong, then will need to figure out how to send it when the node is synced).
3) On epoch transition

### Current Behavior

The announce message is sent once every minute.

Run a registered validator node.

### Logs

In stackdriver.

### System information
Geth
Version: 1.8.23-stable
Git Commit: 600d14f60e5739beb8bc9d8545b925fd88876f6e
Architecture: amd64
Network Id: 1
Go Version: go1.11.13
Operating System: linux
GOPATH=
GOROOT=/usr/local/go

kevjue:
### Description

This PR changes the frequency of when announce messages are sent.  Also, it will now send an announce message on istanbul engine start.

The exact times when the message is sent is described here:  https://github.com/celo-org/celo-blockchain/issues/790

### Tested

Created a testnet and verified that the announce messages are only sent on the intended events.

### Other changes

None

### Related issues

- Fixes #790 

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -109,7 +109,10 @@ func (sb *Backend) sendAnnounceMsgs() {
 	sb.announceWg.Add(1)
 	defer sb.announceWg.Done()
 
-	ticker := time.NewTicker(time.Minute)
+	// Send out an announce message when this thread starts
+	go sb.sendIstAnnounce()
+
+	ticker := time.NewTicker(10 * time.Minute)
 
 	for {
 		select {
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -103,8 +103,31 @@ func (ad *announceData) DecodeRLP(s *rlp.Stream) error {
 	return nil
 }
 
-// This function is meant to be run as a goroutine.  It will periodically gossip announce messages
-// to the rest of the registered validators to communicate it's enodeURL to them.
+// ==============================================
+//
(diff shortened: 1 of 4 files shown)


consensus/istanbul/backend/announce.go
```diff
@@ -109,16 +109,67 @@ func (sb *Backend) sendAnnounceMsgs() {
 	sb.announceWg.Add(1)
 	defer sb.announceWg.Done()
 
-	ticker := time.NewTicker(time.Minute)
+	// Send out an announce message when this thread starts
+	go sb.sendIstAnnounce()
+
+	const (
+		// In this state, send out an announce message every 1 minute until the first peer is established
```
nategraf: Constants are generally still camelCase (e.g. https://golang.org/pkg/time/#pkg-constants), also it may be better to move these constants to the package level.
You might also make them a new type, e.g. `type AnnounceFrequencyState int`

nategraf: Feel free to skip this if we need to submit this PR. I understand we will rework this soon anyway

PR #798, opened 2020-01-08
Istanbul Core Cleanup & Testing

mcortesi:
### Description

Side work while adding the RPC for Istanbul RoundState.

#### Validator

* Add `AsData()` and Json Encode/Decode
* Cleanup Methods
* Reorder methods in file

#### Istanbul Core Types

In  `istanbul/core/types.go`

 * Add test for RLP encode/decode for each type
 * Remove unnecessary custom RLP encode/decode functions
 * Comments to separate types / interface implementations

#### Cleanup on istanbul/core. CurrentRoundState

 * Add `CurrentRoundState()` to engine (will be used for the RPC)
 * Remove `Sequence()` form engine. Replace with existing `CurrentView()` call
 * Move getter method from `handler.go` to `core.go` where all others live
 * Reorder methods in `core.go`. First public methods, then private methods.

### Tested

Unit tests, CI

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -37,41 +37,6 @@ import (
 	"github.com/syndtr/goleveldb/leveldb"
 )
 
-// New creates an Istanbul consensus core
-func New(backend istanbul.Backend, config *istanbul.Config) Engine {
-	rsdb, err := newRoundStateDB(config.RoundStateDBPath, nil)
-	if err != nil {
-		log.Crit("Failed to open RoundStateDB", "err", err)
-	}
-
-	c := &core{
-		config:             config,
-		address:            backend.Address(),
-		logger:             log.New(),
-		selectProposer:     validator.GetProposerSelector(config.ProposerPolicy),
-		handlerWg:          new(sync.WaitGroup),
-		backend:            backend,
-		pendingRequests:    prque.New(nil),
-		pendingRequestsMu:  new(sync.Mutex),
-		consensusTimestamp: time.Time{},
-		rsdb:               rsdb,
-		roundMeter:         metrics.NewRegisteredMeter("consensus/istanbul/core/round", nil),
-		sequenceMeter:      metrics.NewRegisteredMeter("consensus/istanbul/core/sequence", nil),
-		consensusTimer:     metrics.NewRegisteredTimer("consensus/istanbul/core/consensus", nil),
-	}
-	msgBacklog := newMsgBacklog(
-		func(msg *istanbul.Message) {
-			c.sendEvent(backlogEvent{
-				msg: msg,
-			})
-		}, c.checkMessage)
-	c.backlog = msgBacklog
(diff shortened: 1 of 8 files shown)


consensus/istanbul/validator.go
```diff
@@ -71,6 +71,9 @@ type Validator interface {
 	// Serialize returns binary reprenstation of the Validator
 	// can be use used to instantiate a validator with DeserializeValidator()
 	Serialize() ([]byte, error)
+
+	// AsData returns Validator representation as ValidatorData
+	AsData() *ValidatorData
```
nategraf: I don't think we should be adding functions to these interfaces unless it accomplishes something that we can't do otherwise. These are already too bloated for the intended uses of interfaces and we really should add to that.

consensus/istanbul/validator/default.go
```diff
@@ -32,42 +33,52 @@ type defaultValidator struct {
 	blsPublicKey []byte
 }
 
-func (val *defaultValidator) Address() common.Address {
-	return val.address
+func NewValidatorFromData(data *istanbul.ValidatorData) istanbul.Validator {
```
nategraf: I don't think this should be public.

consensus/istanbul/validator/default.go
```diff
@@ -81,52 +92,29 @@ type defaultSet struct {
...
-	valSet := &defaultSet{}
-
-	// init validators
-	valSet.validators = make([]istanbul.Validator, len(validators))
-	for i, validator := range validators {
-		valSet.validators[i] = New(validator.Address, validator.BLSPublicKey)
+func newValidatorSetFromData(data *istanbul.ValidatorSetData) *defaultSet {
+	return &defaultSet{
+		validators: mapDataToValidators(data.Validators),
+		randomness: data.Randomness,
```
nategraf: I don't think we should create this function, especially because it is unused.

consensus/istanbul/core/core.go
```diff
@@ -37,6 +37,41 @@ import (
 	"github.com/syndtr/goleveldb/leveldb"
 )
 
+type core struct {
```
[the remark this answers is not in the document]
mcortesi: The addition of `CurrentRoundState()`. 

`CurrentView()` is not an addition, but it lived in `handler.go` before, I've just moved it there. I removed `Sequence()` since it seemed redundant...

consensus/istanbul/validator/default.go
mcortesi: removed

consensus/istanbul/validator.go
mcortesi: I've removed the `AsData()` from the ValidatorSet inteface... but for `Validator` is need to map Validators to Data

consensus/istanbul/validator/default.go
mcortesi: Removed!

PR #799, opened 2020-01-08
Log all levels if round > 1

lucasege:
### Description

Implements new logger wrapper to log all messages if `round > 1` , specifically high verbosity `trace` and `debug`.

### Tested

Running CI

### Other changes

Changed instances within `istanbul.backend.backend.go` and `istanbul.core.core.go` that were using `log.{log_level}` rather than `sb.log.{log_level}` or `c.log.{log_level}`

### Related Issues 
* Fixes #799

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -43,20 +43,11 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 	if err != nil {
 		log.Crit("Failed to open RoundStateDB", "err", err)
 	}
-	istanbulLogger := istanbul.NewIstLogger(
-		func() *big.Int {
-			view, err := rsdb.GetLastView()
-			if err != nil {
-				log.Crit("Failed to fetch last view")
-			}
-			return view.Round
-		},
-	)
 
 	c := &core{
 		config:             config,
 		address:            backend.Address(),
-		logger:             istanbulLogger,
+		logger:             log.New(),
 		selectProposer:     validator.GetProposerSelector(config.ProposerPolicy),
 		handlerWg:          new(sync.WaitGroup),
 		backend:            backend,
@@ -76,6 +67,14 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 		}, c.checkMessage)
 	c.backlog = msgBacklog
 	c.validateFn = c.checkValidatorSignature
+	c.logger = istanbul.NewIstLogger(
+		func() *big.Int {
+			if c != nil {
+				return c.current.Round()
+			}
+			return big.NewInt(0)
+		},
+	)
 	return c
 }
 
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -50,7 +50,8 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 				log.Crit("Failed to fetch last view")
 			}
 			return view.Round
-		})
+		},
+	)
 
 	c := &core{
 		config:             config,
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -43,7 +43,7 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 	if err != nil {
 		log.Crit("Failed to open RoundStateDB", "err", err)
 	}
-	istanbulLogger, err := istanbul.NewIstLogger(
+	istanbulLogger := istanbul.NewIstLogger(
 		func() *big.Int {
 			view, err := rsdb.GetLastView()
 			if err != nil {
@@ -51,9 +51,6 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 			}
 			return view.Round
 		})
-	if err != nil {
-		log.Crit("Failed to instantiate Istanbul Logger", "err", err)
-	}
 
 	c := &core{
 		config:             config,
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -69,7 +69,7 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 	c.validateFn = c.checkValidatorSignature
 	c.logger = istanbul.NewIstLogger(
 		func() *big.Int {
-			if c != nil {
+			if c != nil && c.current != nil {
 				return c.current.Round()
 			}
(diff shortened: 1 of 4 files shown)


kevjue:
One manual test that I think is worth doing is to start up a testnet with at least 2 validators, and then take one down so that the others start round change, and then making sure that you start seeing some Trace and Debug statements that you didn't before.

consensus/istanbul/core/core.go
```diff
@@ -43,11 +43,20 @@ func New(backend istanbul.Backend, config *istanbul.Config) Engine {
 	if err != nil {
 		log.Crit("Failed to open RoundStateDB", "err", err)
 	}
+	istanbulLogger := istanbul.NewIstLogger(
+		func() *big.Int {
+			view, err := rsdb.GetLastView()
```
asaj: Instead of reading from the DB, can you instead read from `c.current`?

consensus/istanbul/logger.go
```diff
@@ -0,0 +1,80 @@
...
+	childLogger := l.logger.New(ctx)
+	return &istLogger{logger: childLogger, round: l.round}
+}
+
+func (l *istLogger) Trace(msg string, ctx ...interface{}) {
+	// If the current round > 1, then upgrade this message to Info
+	if l.round().Cmp(common.Big1) > 0 {
+		l.Info(msg, ctx...)
+	} else {
+		l.Trace(msg, ctx)
```
timmoreton: `ctx...` ?

consensus/istanbul/core/core.go
lucasege: Yeah but it involves a bit of weird flow

lucasege:
> One manual test that I think is worth doing is to start up a testnet with at least 2 validators, and then take one down so that the others start round changing, and then making sure that you start seeing some Trace and Debug statements that you didn't see before.

So I looked into this, but ran into some issues trying to force a round change - `kubectl` does not like you stopping containers.
What I did instead (please let me know if this is sufficient) is ran tests from `roundchange_test.go` specifically `TestCommitsBlocksAfterRoundChange` which forced the issue. I had to change the test suite slightly since it normally overrides core's logger and I had to change the logger to only check `round >= 1` not `round > 1` since rounds 2 or more don't happen in these tests. But, after this I was able to produce `Info` logs from `Trace` calls, specifically `verification status cache hit`.

One slightly confusing thing I couldn't figure out: this trace message `verification status cache hit` is also normally produced (without the istLogger) with `debug` severity, despite only being called from `trace`. I assume this is something related to the test logger, but haven't understood it fully yet.

kevjue:
> > One manual test that I think is worth doing is to start up a testnet with at least 2 validators, and then take one down so that the others start round changing, and then making sure that you start seeing some Trace and Debug statements that you didn't see before.
> 
> So I looked into this, but ran into some issues trying to force a round change - `kubectl` does not like you stopping containers.
> What I did instead (please let me know if this is sufficient) is ran tests from `roundchange_test.go` specifically `TestCommitsBlocksAfterRoundChange` which forced the issue. I had to change the test suite slightly since it normally overrides core's logger and I had to change the logger to only check `round >= 1` not `round > 1` since rounds 2 or more don't happen in these tests. But, after this I was able to produce `Info` logs from `Trace` calls, specifically `verification status cache hit`.
> 
> One slightly confusing thing I couldn't figure out: this trace message `verification status cache hit` is also normally produced (without the istLogger) with `debug` severity, despite only being called from `trace`. I assume this is something related to the test logger, but haven't understood it fully yet.

So one trick that Tim does to test "fail stop" nodes with a testnet is that he'll scale down the number of validators within it:

Screen Shot 2020-01-10 at 12 07 15 PM

So one thing you may want to try is to scale down to half of the original number of validators.

lucasege:
> > > One manual test that I think is worth doing is to start up a testnet with at least 2 validators, and then take one down so that the others start round changing, and then making sure that you start seeing some Trace and Debug statements that you didn't see before.
> > 
> > 
> > So I looked into this, but ran into some issues trying to force a round change - `kubectl` does not like you stopping containers.
> > What I did instead (please let me know if this is sufficient) is ran tests from `roundchange_test.go` specifically `TestCommitsBlocksAfterRoundChange` which forced the issue. I had to change the test suite slightly since it normally overrides core's logger and I had to change the logger to only check `round >= 1` not `round > 1` since rounds 2 or more don't happen in these tests. But, after this I was able to produce `Info` logs from `Trace` calls, specifically `verification status cache hit`.
> > One slightly confusing thing I couldn't figure out: this trace message `verification status cache hit` is also normally produced (without the istLogger) with `debug` severity, despite only being called from `trace`. I assume this is something related to the test logger, but haven't understood it fully yet.
> 
> So one trick that Tim does to emulate "fail stop" nodes with a testnet is that he'll scale down the number of validators within it:
> 
> Screen Shot 2020-01-10 at 12 07 15 PM
> 
> So one thing you may want to try is to scale down to half of the original number of validators.

Was able to produce logs RE `Round change` and `validator status cache` at INFO level when called at TRACE level.

consensus/istanbul/logger.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (l *istLogger) New(ctx ...interface{}) log.Logger {
+	childLogger := l.logger.New(ctx...)
+	return &istLogger{logger: childLogger, round: l.round}
+}
+
+func (l *istLogger) Trace(msg string, ctx ...interface{}) {
+	// If the current round > 1, then upgrade this message to Info
+	if l.round().Cmp(common.Big1) > 0 {
```
timmoreton: could you add a check to gracefully handle the case where `l.round` is `nil`

consensus/istanbul/logger.go
```diff
@@ -0,0 +1,80 @@
...
+		l.Info(msg, ctx...)
+	} else {
+		l.logger.Trace(msg, ctx...)
+	}
+}
+
+func (l *istLogger) Debug(msg string, ctx ...interface{}) {
+	// If the current round > 1, then upgrade this message to Info
+	if l.round().Cmp(common.Big1) > 0 {
```
timmoreton: ditto

consensus/istanbul/logger.go
lucasege: Done, good call

piersy:
Hi @timmoreton & @asaj I'm wondering if you guys can fill me in on why this logger was added?

PR #800, opened 2020-01-09
Periodically resend RoundChanges for same round 

timmoreton:
### Description

Separates the IBFT round change timer into two: one that moves into `StateWaitingForRoundChange` and advances to to the next desired round; and one that while in  in `StateWaitingForRoundChange` periodically resends RoundChange messages for the same current desired round if the other timeout exceeds `MinResendRoundChangeTimeout`. 

Add hard-coded config options to control timeout parameters:
* `TimeoutBackoffFactor` - Timeout at subsequent rounds is: `RequestTimeout + 2**round * TimeoutBackoffFactor` (in milliseconds)
* `MinResendRoundChangeTimeout` - Minimum interval with which to resend RoundChange messages for same round
* `MaxResendRoundChangeTimeout` -Maximum interval with which to resend RoundChange messages for same round

### Tested

New test in `core_test.go`.

TODO - on a fault-injection network.

### Other changes

* Fix logging typo. 
* Speed up IBFT tests by setting lower timeouts.

### Backwards compatibility

For networks with same timeouts, additional `RoundChange` messages do not break backwards-compatibility.

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -542,7 +542,7 @@ func (c *core) getRoundChangeTimeout() time.Duration {
 		return baseTimeout + time.Duration(c.config.BlockPeriod)*time.Second
 	} else {
 		// timeout for subsequent rounds adds an exponential backoff.
-		return baseTimeout + time.Duration(math.Pow(2, float64(round)))*time.Second
+		return baseTimeout + time.Duration(math.Pow(2, float64(round)))*time.Duration(c.config.TimeoutBackoffFactor)*time.Millisecond
 	}
 }
 
@@ -552,20 +552,30 @@ func (c *core) resetRoundChangeTimer() {
 	c.stopTimers()
 
 	view := &istanbul.View{Sequence: c.current.Sequence(), Round: c.current.DesiredRound()}
-	c.roundChangeTimer = time.AfterFunc(c.getRoundChangeTimeout(), func() {
+	timeout := c.getRoundChangeTimeout()
+	c.roundChangeTimer = time.AfterFunc(timeout, func() {
 		c.sendEvent(timeoutAndMoveToNextRoundEvent{view})
 	})
+
+	if c.current.DesiredRound().Cmp(common.Big1) > 0 {
+		logger := c.newLogger("func", "resetRoundChangeTimer")
+		logger.Info("Reset round change timer", "timeout_ms", timeout/time.Millisecond)
+	}
 
 	c.resetResendRoundChangeTimer()
 }
 
-// Reset then, if in StateWaitingForNewRound and on a large enough round, set a timer
-// that causes a resendRoundChangeEvent to be processed.
+// Reset then, if in StateWaitingForNewRound and on round whose timeout is greater than MinResendRoundChangeTimeout,
+// set a timer that is at most MaxResendRoundChangeTimeout that causes a resendRoundChangeEvent to be processed.
 func (c *core) resetResendRoundChangeTimer() {
 	c.stopResendRoundChangeTimer()
-	if c.current.State() == StateWaitingForNewRound && c.current.DesiredRound().Cmp(big.NewInt(4)) > 0 {
-		const maxResendTimeout = time.Duration(5) * time.Minute
+	if c.current.State() == StateWaitingForNewRound {
+		minResendTimeout := time.Duration(c.config.MinResendRoundChangeTimeout) * time.Millisecond
 		resendTimeout := c.getRoundChangeTimeout() / 2
+		if resendTimeout < minResendTimeout {
+			return
+		}
+		maxResendTimeout := time.Duration(c.config.MaxResendRoundChangeTimeout) * time.Millisecond
 		if resendTimeout > maxResendTimeout {
 			resendTimeout = maxResendTimeout
 		}
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -528,7 +528,7 @@ func (c *core) stopResendRoundChangeTimer() {
 	}
 }
 
-func (c *core) stopTimers() {
+func (c *core) stopAllTimers() {
 	c.stopFuturePreprepareTimer()
 	c.stopRoundChangeTimer()
 	c.stopResendRoundChangeTimer()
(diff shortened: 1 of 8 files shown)


consensus/istanbul/core/core.go
```diff
@@ -512,36 +510,88 @@ func (c *core) isProposer() bool {
...
+	view := &istanbul.View{Sequence: c.current.Sequence(), Round: c.current.DesiredRound()}
+	timeout := c.getRoundChangeTimeout()
 	c.roundChangeTimer = time.AfterFunc(timeout, func() {
-		c.sendEvent(timeoutEvent{&istanbul.View{Sequence: view.Sequence, Round: view.Round}})
+		c.sendEvent(timeoutAndMoveToNextRoundEvent{view})
 	})
+
+	if c.current.DesiredRound().Cmp(common.Big1) > 0 {
+		logger := c.newLogger("func", "resetRoundChangeTimer")
+		logger.Info("Reset round change timer", "timeout_ms", timeout/time.Millisecond)
```
nategraf: Sequence and round may be useful additions to this log message.
```suggestion
		logger.Info("Reset round change timer", "timeout_ms", timeout/time.Millisecond, "sequence", view.Sequence, "round", view.Round)
```

consensus/istanbul/core/handler.go
```diff
@@ -219,3 +224,15 @@ func (c *core) handleTimeoutMsg(desiredView *istanbul.View) error {
 	nextRound := new(big.Int).Add(desiredView.Round, common.Big1)
 	return c.waitForDesiredRound(nextRound)
 }
+
+func (c *core) handleResendRoundChangeEvent(desiredView *istanbul.View) error {
+	logger := c.newLogger("func", "handleResendRoundChangeEvent", "set_at_seq", desiredView.Sequence, "set_at_desiredRound", desiredView.Round)
+
+	if c.current.Sequence().Cmp(desiredView.Sequence) != 0 || c.current.DesiredRound().Cmp(desiredView.Round) != 0 {
```
nategraf: If `currentRound == desiredRound`, doesn't that mean we have advanced to the desired round and shouldn't send a round change message? I would assume we would send a round change here iff `currentRound < desiredRound`

consensus/istanbul/core/core.go
```diff
@@ -512,36 +510,88 @@ func (c *core) isProposer() bool {
...
+			c.sendEvent(resendRoundChangeEvent{view})
+		})
+	}
+}
+
+// Broadcast round change message for current desired round and reset timer.
+func (c *core) resendRoundChangeMessage() {
+	if c.current.State() == StateWaitingForNewRound {
+		c.sendRoundChange()
+	}
```
nategraf: It seems like this function should be unconditional. I would move this check into `handleResendRoundChangeEvent` instead.

consensus/istanbul/core/roundchange.go
```diff
@@ -28,15 +28,10 @@ import (
 )
 
 // sendRoundChange sends the ROUND CHANGE message with the given round
```
nategraf:
```suggestion
 // sendRoundChange sends the ROUND CHANGE message with the desired round
```

consensus/istanbul/core/roundchange_test.go
```diff
@@ -561,3 +561,85 @@ func TestPreparedCertificatePersistsThroughRoundChanges(t *testing.T) {
...
+			}
+			if expectedCommitted.Hash() != committed.Hash() {
+				t.Errorf("Backend %v got committed block with unexpected hash: expected %v, got %v", i, expectedCommitted.Hash(), committed.Hash())
+			}
+		}
+	}
+
+	// Manually open and close b/c hijacking sys.listen
+	for _, b := range sys.backends {
+		b.engine.Stop() // start Istanbul core
```
nategraf:
```suggestion
		b.engine.Stop() // stop Istanbul core
```

consensus/istanbul/core/core.go
```diff
@@ -512,36 +510,88 @@ func (c *core) isProposer() bool {
...
 		// timeout for subsequent rounds adds an exponential backoff.
-		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
+		return baseTimeout + time.Duration(math.Pow(2, float64(round)))*time.Duration(c.config.TimeoutBackoffFactor)*time.Millisecond
 	}
+}
+
+// Reset then set the timer that causes a timeoutAndMoveToNextRoundEvent to be processed.
+// This may also reset the timer for the next resendRoundChangeEvent.
+func (c *core) resetRoundChangeTimer() {
+	c.stopTimers()
```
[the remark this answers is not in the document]
timmoreton: yes it should - will add a comment

consensus/istanbul/core/core.go
timmoreton: already added by `c.newLogger`

consensus/istanbul/core/handler.go
timmoreton: This is specifically to catch race condition of event being executed but view having changed in the mean time. I will add a comment.

timmoreton: Added a comment.  This check is only to prevent races since other messages could  be processed between enqueueing this timeout and processing the event. Equality with both means there has not been a race with updating the sequence or advancing the actual or desired view elsewhere - in those cases this event should be suppressed and we shouldn't send a RoundChange as a result.

consensus/istanbul/core/core.go
timmoreton: It's arguable we should check in the handler too, but I sort of kept the handler as dealing with the (otherwise unneeded) view in the event and everything else here. If we moved it, we'd surely want an assert here, which is not really an approach we have used yet here.

consensus/istanbul/core/core.go
```diff
@@ -512,36 +510,92 @@ func (c *core) isProposer() bool {
...
 	}
+}
+
+// Reset then set the timer that causes a timeoutAndMoveToNextRoundEvent to be processed.
+// This may also reset the timer for the next resendRoundChangeEvent.
+func (c *core) resetRoundChangeTimer() {
+	// Stop all timers here since all 'resends' happen within the interval of a round's timeout.
+	// (Races are handled anyway by checking the seq and desired round haven't changed between
+	// submitting and processing events).
+	c.stopAllTimers()
```
asaj: Why do we stop the future preprepare timer here?

PR #803, opened 2020-01-09
Explain Q and F

timmoreton:
### Description

Add a comment to explain `MinQuorumSize` and `F`.

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/validator/default.go
+++ b/consensus/istanbul/validator/default.go
@@ -257,16 +257,16 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 //
 // 1 2 3 4 5 6 7 8 9
 // x x x x x x
-// 			 y y y y y y
-// 			 F F H
+//       y y y y y y
+//       F F H
 //
 // For N=10, F=3, Q=7. Any two sets of Q=7 nodes from N=10 must overlap by >4 nodes.
 // At least 4-F=4-3=1 must be honest.
 //
 // 1 2 3 4 5 6 7 8 9 10
 // x x x x x x x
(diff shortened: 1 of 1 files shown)


consensus/istanbul/validator/default.go
```diff
@@ -247,6 +247,27 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 	return newValSet
 }
 
+// We include the optimization described at https://arxiv.org/pdf/1901.07160.pdf as “PM-6” and
+// discussed in Lemma 22. For values of N=3F for integer F=1,2,3,.. we can tolerate a quorum
```
nategraf: Should this be `N>=3F`?
```suggestion
// discussed in Lemma 22. For values of N>=3F for integer F=1,2,3,.. we can tolerate a quorum
```

consensus/istanbul/validator/default.go
```diff
@@ -247,6 +247,27 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
...
+// size one smaller than anticipated by Q = N - F. The intersection of any two sets of Q
+// nodes of N=3F must contain an honest validator.
```
nategraf: Same comment as above
```suggestion
// nodes of N>=3F must contain an honest validator.
```

consensus/istanbul/validator/default.go
timmoreton: I think no, just specifically those `N` that are exact multiples of `3F` ?

nategraf: Ah I see now. You are discussing this specific case in this paragraph.

PR #804, opened 2020-01-09
Add RPC to force round timeout

mcortesi:
### Description

Add RPC to force round timeout

### Tested

CI

### DIFF
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -150,7 +150,7 @@ func (api *API) GetValEnodeTable() (map[string]*vet.ValEnodeEntryInfo, error) {
 
 // GetCurrentRoundState retrieves the current IBFT RoundState
 func (api *API) GetCurrentRoundState() (*core.RoundStateSummary, error) {
-	if api.istanbul.coreStarted {
+	if !api.istanbul.coreStarted {
 		return nil, istanbul.ErrStoppedEngine
 	}
 	return api.istanbul.core.CurrentRoundState().Summary(), nil
@@ -158,7 +158,7 @@ func (api *API) GetCurrentRoundState() (*core.RoundStateSummary, error) {
 
 // GetCurrentRoundState retrieves the current IBFT RoundState
 func (api *API) ForceRoundChange() (bool, error) {
-	if api.istanbul.coreStarted {
+	if !api.istanbul.coreStarted {
 		return false, istanbul.ErrStoppedEngine
 	}
 	api.istanbul.core.ForceRoundChange()
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -150,11 +150,17 @@ func (api *API) GetValEnodeTable() (map[string]*vet.ValEnodeEntryInfo, error) {
 
 // GetCurrentRoundState retrieves the current IBFT RoundState
 func (api *API) GetCurrentRoundState() (*core.RoundStateSummary, error) {
+	if api.istanbul.coreStarted {
+		return nil, istanbul.ErrStoppedEngine
+	}
 	return api.istanbul.core.CurrentRoundState().Summary(), nil
 }
 
 // GetCurrentRoundState retrieves the current IBFT RoundState
 func (api *API) ForceRoundChange() (bool, error) {
+	if api.istanbul.coreStarted {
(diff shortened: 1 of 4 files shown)


consensus/istanbul/backend/api.go
```diff
@@ -153,6 +153,12 @@ func (api *API) GetCurrentRoundState() (*core.RoundStateSummary, error) {
 	return api.istanbul.core.CurrentRoundState().Summary(), nil
 }
 
+// GetCurrentRoundState retrieves the current IBFT RoundState
+func (api *API) ForceRoundChange() (bool, error) {
+	api.istanbul.core.ForceRoundChange()
```
timmoreton: check if engine is started here?

consensus/istanbul/core/handler.go
```diff
@@ -211,17 +211,17 @@ func (c *core) handleCheckedMsg(msg *istanbul.Message, src istanbul.Validator) e
 	return errInvalidMessage
 }
 
-func (c *core) handleTimeoutAndMoveToNextRound(desiredView *istanbul.View) error {
-	logger := c.newLogger("func", "handleTimeoutAndMoveToNextRound", "set_at_seq", desiredView.Sequence, "set_at_desiredRound", desiredView.Round)
+func (c *core) handleTimeoutAndMoveToNextRound(timedOutView *istanbul.View) error {
+	logger := c.newLogger("func", "handleTimeoutAndMoveToNextRound", "set_at_seq", timedOutView.Sequence, "set_at_desiredRound", timedOutView.Round)
```
timmoreton: maybe `"timed_out_seq"` etc too?

consensus/istanbul/backend/api.go
```diff
@@ -150,9 +150,21 @@ func (api *API) GetValEnodeTable() (map[string]*vet.ValEnodeEntryInfo, error) {
 
 // GetCurrentRoundState retrieves the current IBFT RoundState
 func (api *API) GetCurrentRoundState() (*core.RoundStateSummary, error) {
+	if api.istanbul.coreStarted {
+		return nil, istanbul.ErrStoppedEngine
+	}
```
lucasege: Is this correct? Seems to be like there should be a `not` here, as in if the core is *not* started, return `ErrStoppedEngine`.
Maybe I'm missing something, but if so maybe a comment would be helpful.

consensus/istanbul/core/types.go
```diff
@@ -25,11 +25,15 @@ import (
...
 	Start() error
 	Stop() error
+	// CurrentView returns the current view or nil if none
 	CurrentView() *istanbul.View
+	// CurrentRoundState returns the current roundState or nil if none
 	CurrentRoundState() RoundState
 	SetAddress(common.Address)
 	// Validator -> CommittedSeal from Parent Block
 	ParentCommits() MessageSet
+	// ForceRoundChange will force round change to the current deriredRound + 1
```
lucasege: nit: `deriredRound` -> `desiredRound`

consensus/istanbul/backend/api.go
```diff
@@ -150,9 +150,21 @@ func (api *API) GetValEnodeTable() (map[string]*vet.ValEnodeEntryInfo, error) {
...
 	return api.istanbul.core.CurrentRoundState().Summary(), nil
 }
 
+// GetCurrentRoundState retrieves the current IBFT RoundState
```
nategraf: Comment should be updated

consensus/istanbul/backend/api.go
mcortesi: heh... you're right!

consensus/istanbul/backend/api.go
```diff
@@ -150,9 +150,21 @@ func (api *API) GetValEnodeTable() (map[string]*vet.ValEnodeEntryInfo, error) {
...
+func (api *API) ForceRoundChange() (bool, error) {
+	if !api.istanbul.coreStarted {
+		return false, istanbul.ErrStoppedEngine
+	}
+	api.istanbul.core.ForceRoundChange()
```
nategraf: What happens when this is called while the engine is already waiting for the next round?

timmoreton: it will be as though that timeout got sped up, and we then move to the next round.

PR #807, opened 2020-01-13
Separate Etherbase flag by responsibility

### LINKED ISSUE #738: Separate flags for tx fee recipient and consensus message signer
### Expected Behavior

`--validator` and `--tx-fee-recipient`
### Current Behavior

`--etherbase` for both

lucasege:
### Description

Separating `etherbase` into two flags: 
`--tx-fee-recipient` for the recipient of tx-fee's (coinbase/gatewayfee), and `--miner.validator` which signs consensus messages.

This also adds a function to `blockchain_parameters` to allow for future "hard-fork" actions, an example of which appears in this PR: proposers will now utilize a block header's `coinbase` field for their `txFeeRecipient` address, so we are rolling this out in phases to accommodate this change. Blockchain version 1.1.0 should introduce this PR that starts to set/check `header.Coinbase`, but won't actually allow the value of `header.Coinbase` to differ from `Author(block.header)` until the `minClientVersion` has been bumped to `1.1.0` so we know users will read the correct addresses.

Sister PR (tests): https://github.com/celo-org/celo-monorepo/pull/2529

### Tested

Tested in e2e_transfer by verifying blocks are still signed by the singular private key (validator)
Also tested that tx fees are properly received in the right address.

### Other changes

Small typo in tests

### Related issues

- Fixes #738 

### Backwards compatibility

No, the new `minClientVersion` check introduces a "pseudo" hard-fork (handled in code rather than by block number). 
The flags themselves are backwards compatible, i.e. if `miner.validator` and `txFeeRecipient` are not provided then the behavior will remain as is currently implemented (`etherbase` covers both `tx-fee-receipt` and `validator`)


eth/backend.go
```diff
@@ -395,25 +420,25 @@ func (s *Ethereum) BLSbase() (eb common.Address, err error) {
...
 	author, err := s.engine.Author(block.Header())
 	if err != nil {
 		log.Warn("Failed to retrieve block author", "number", block.NumberU64(), "hash", block.Hash(), "err", err)
 		return false
 	}
-	// Check whether the given address is etherbase.
+	// Check whether the given address is configured validator.
 	s.lock.RLock()
-	etherbase := s.etherbase
+	validator := s.validator
```
lucasege: Question for reviewer: is this correct? I assume "local miner account" in this context refers to the message signer, but maybe I'm wrong.

kevjue: Looks like what you did is correct.  The `s.engine.Author` function retrieves the signer of the block header:  

https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/engine.go#L109

https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/engine.go#L900

kevjue: Perhaps renaming the variable `author` to `signer` will make it a bit more clear?

cmd/utils/flags.go
```diff
@@ -198,9 +198,19 @@ var (
...
+	}
+	TxFeeRecipientFlag = cli.StringFlag{
+		Name:  "tx-fee-recipient",
+		Usage: "Public address for block mining rewards (default = first account)",
+		Value: "0",
+	}
+	EtherbaseLegacyFlag = cli.StringFlag{
 		Name:  "etherbase",
-		Usage: "Public address for transaction broadcasting and block mining rewards (default = first account)",
+		Usage: "Public address for block mining rewards (default = first account, deprecated, use --tax-fee-recipient)",
```
asaj: tx

cmd/utils/flags.go
```diff
@@ -198,9 +198,19 @@ var (
...
 	}
-	EtherbaseFlag = cli.StringFlag{
+	ValidatorFlag = cli.StringFlag{
+		Name:  "validator",
+		Usage: "Public address for transaction broadcasting (default = first account)",
+		Value: "0",
+	}
+	TxFeeRecipientFlag = cli.StringFlag{
+		Name:  "tx-fee-recipient",
+		Usage: "Public address for block mining rewards (default = first account)",
```
asaj: for block transaction fees

cmd/utils/flags.go
```diff
@@ -198,9 +198,19 @@ var (
 		Name:  "whitelist",
 		Usage: "Comma separated block number-to-hash mappings to enforce (<number>=<hash>)",
 	}
-	EtherbaseFlag = cli.StringFlag{
+	ValidatorFlag = cli.StringFlag{
+		Name:  "validator",
+		Usage: "Public address for transaction broadcasting (default = first account)",
```
asaj: Is this right?

cmd/utils/flags.go
```diff
@@ -947,13 +957,32 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
 	return accs[index], nil
 }
 
+// setValidator retrieves the validator address either from the directly specified
```
asaj: Can we include comments above each of these functions describing the difference between validator, etherbase, and blsbase?

eth/backend.go
```diff
@@ -490,6 +515,13 @@ func (s *Ethereum) StartMining(threads int) error {
 			log.Error("Cannot start mining without etherbase", "err", err)
 			return fmt.Errorf("etherbase missing: %v", err)
 		}
+
+		val, err := s.Validator()
+		if err != nil {
```
asaj: Is this necessary? Isn't this what `			log.Info("Validator automatically configured", "address", validator)` does already?

eth/config.go
```diff
@@ -94,6 +94,7 @@ type Config struct {
 	GatewayFee *big.Int `toml:",omitempty"`
 	// Etherbase is the GatewayFeeRecipient light clients need to specify in order for their transactions to be accepted by this node.
 	// Also the coinbase used for mining.
+	Validator common.Address `toml:",omitempty"`
```
asaj: Comment needs updating

cmd/utils/flags.go
lucasege: Let me know if the added comments are sufficient

cmd/utils/flags.go
lucasege: Changed

lucasege:
Comments describing validator usage are different due to different context (messages for devs vs. general users), let me know if that is acceptable or not. Tracing the use of these flags is difficult since their use is ultimately determined by runtime parameters (hash functions of the provided wallet address), so I've done my best to suss out the usage and differences.

Also added small change to tests https://github.com/celo-org/celo-monorepo/pull/2529

cmd/geth/main.go
```diff
@@ -268,6 +270,7 @@ func main() {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+
```
asaj: Remove extra line

eth/backend.go
```diff
@@ -518,8 +518,8 @@ func (s *Ethereum) StartMining(threads int) error {
 
 		val, err := s.Validator()
 		if err != nil {
-			log.Trace("Validator address not provided, defaulting to provided etherbase address")
-			val = eb
+			log.Error("Cannot start mining without validator", "err", err)
```
asaj: nit; validating

cmd/utils/flags.go
```diff
@@ -996,6 +998,7 @@ func setEtherbase(ctx *cli.Context, ks *keystore.KeyStore, cfg *eth.Config) {
 
 // setBLSbase retrieves the blsbase either from the directly specified
 // command line flags or from the keystore if CLI indexed.
+// `BLSbase` is the public address used for block mining BLS signatures.
```
asaj: nit: used for block finalization in consensus

cmd/utils/flags.go
```diff
@@ -959,6 +959,7 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
 
 // setValidator retrieves the validator address either from the directly specified
 // command line flags or from the keystore if CLI indexed.
+// `Validator` determines the address used to sign hashes using ECDSA.
```
asaj: // `Validator` is the address used to sign consensus messages

asaj:
Please refrain from merging until https://github.com/celo-org/celo-monorepo/pull/2529 is approved

lucasege:
So I've investigated an issue from this change, if you set `validator` and `tfFeeRecipient` to separate addresses, blocks are no longer mined. I've traced this to the sending of the "gas fee tip" which calls on the statedb to add balance. Somehow this is "corrupting" the state such that geth prints "invalid merkle root" errors when trying to verify state transitions from istanbul messages. Still trying to determine why exactly this is happening.

`addBalance` creates a new state object because the `txFeeRecipient` being passed in is an arbitrary address, so my current assumption is that the issue is related to this new object not being committed to the new stateDB.

lucasege:
Please review the overloading of `etherbase`, worried if it is clear the distinction between `etherbase`, `validator` and `txFeeRecipient` everywhere. 
I made some progress on removing instances of `etherbase` but there are quite a lot of uses of it and I'm not sure if it's better to remove these instances or leave them since people are familiar with it.

lucasege:
Release process for this change is still in flux, but the tentative plan is:
Release blockchain 1.1 with reading `txFeeRecipient` from `header.Coinbase` but with a constraint that users cannot supply two different addresses for `validator` and `txFeeRecipient` to allow for the network to upgrade to this new expectation.
Then with release 1.2 we verify that network participants have upgraded and we will remove that constraint, and enforce these changes via `minimum_client_version`

cmd/utils/flags.go
```diff
@@ -218,7 +218,17 @@ var (
 	}
 	EtherbaseFlag = cli.StringFlag{
 		Name:  "etherbase",
-		Usage: "Public address for transaction broadcasting and block mining rewards (default = first account)",
+		Usage: "Public address for transaction broadcasting and block mining rewards (default = first account, deprecated, use --tx-fee-recipient and --validator instead)",
+		Value: "0",
+	}
+	ValidatorFlag = cli.StringFlag{
+		Name:  "validator",
```
asaj, line 225: Should this be `miner.validator` instead?

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,66 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
-			account, err := MakeAddress(ks, etherbase)
-			if err != nil {
-				Fatalf("Invalid miner etherbase: %v", err)
-			}
-			cfg.Miner.Etherbase = account.Address
-			cfg.Etherbase = account.Address
-		} else {
-			Fatalf("No etherbase configured")
+	if ctx.GlobalIsSet(ValidatorFlag.Name) {
+		validator = ctx.GlobalString(ValidatorFlag.Name)
```
asaj, line 1088: Should this flag be mutually exclusive with the etherbase flags? i.e. you can set `validator/tx-fee-recipient` or `etherbase`, but not both?

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,66 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
+	if ctx.GlobalIsSet(TxFeeRecipientFlag.Name) {
+		txFeeRecipient = ctx.GlobalString(TxFeeRecipientFlag.Name)
+	}
+	// Convert the txFeeRecipient into an address and configure it
+	if txFeeRecipient != "" {
+		account, err := MakeAddress(ks, txFeeRecipient)
+		if err != nil {
+			Fatalf("Invalid txFeeRecipient: %v", err)
 		}
+		cfg.Miner.TxFeeRecipient = account.Address
```
asaj, line 1121: Should this be confined to the `Miner` config? I would imagine this flag would cover gateway fee recipient for full nodes as well

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,66 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
+	// Extract the current txFeeRecipient, new flag overriding legacy etherbase
+	var txFeeRecipient string
+	if ctx.GlobalIsSet(MinerLegacyEtherbaseFlag.Name) {
+		txFeeRecipient = ctx.GlobalString(MinerLegacyEtherbaseFlag.Name)
+	}
+	if ctx.GlobalIsSet(MinerEtherbaseFlag.Name) {
+		txFeeRecipient = ctx.GlobalString(MinerEtherbaseFlag.Name)
+	}
+	if ctx.GlobalIsSet(TxFeeRecipientFlag.Name) {
+		txFeeRecipient = ctx.GlobalString(TxFeeRecipientFlag.Name)
```
asaj, line 1113: Ditto on mutual exclusivity

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,66 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
+		if err != nil {
+			Fatalf("Invalid txFeeRecipient: %v", err)
 		}
+		cfg.Miner.TxFeeRecipient = account.Address
 	}
 }
 
 // setBLSbase retrieves the blsbase either from the directly specified
 // command line flags or from the keystore if CLI indexed.
+// `BLSbase` is the public address used for block finalization in consensus.
```
asaj, line 1127: nit: `BLSbase` is the address derived from the BLS private key used for block finalization in consensus.

consensus/istanbul/backend/backend.go
```diff
@@ -469,17 +469,7 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
-	if err != nil {
-		sb.logger.Error("Could not recover orignal author of the block to verify the randomness", "err", err, "func", "Verify")
-		return 0, errInvalidProposal
-	} else if addr != block.Header().Coinbase {
-		sb.logger.Error("Original author of the block does not match the coinbase", "addr", addr, "coinbase", block.Header().Coinbase, "func", "Verify")
-		return 0, errInvalidCoinbase
-	}
-
-	err = sb.VerifyHeader(sb.chain, block.Header(), false)
+	err := sb.VerifyHeader(sb.chain, block.Header(), false)
```
asaj, line 472: I guess the conditional around min client version should go here?

cmd/utils/flags.go
asaj: As a follow up, how does this PR address gateway fee recipient for full nodes that serve light clients?

lucasege: Yup you're right it shouldn't be in `miner`.
Gateway fee recipient is handled here: ``
```
les/server.go:92
srv.handler = newServerHandler(srv, e.BlockChain(), e.ChainDb(), e.TxPool(), e.Synced, config.TxFeeRecipient, config.GatewayFee)`
```
And is the `txFeeRecipient`

consensus/istanbul/backend/backend.go
lucasege: Yup I've added it (will push after finishing testing), had to do a little bit of awkward error handling while the registry/version contract is not deployed, but otherwise looks good.

lucasege:
Added the branching based on versioning in `istanbul/backend/backend.go`. Some error handling is a bit wonky, but functionally good.
Tested by trying to split on this version and hit the error, then changing my local geth build to version 1.2, and the minclientVersion on my contract to 1.2 and it allowed me to split, as expected.
This messes with testing on https://github.com/celo-org/celo-monorepo/pull/2529 so I think we may just need to wait on adding those tests until geth 1.2 - don't think there's a way to change the geth versions created in the e2e tests (to avoid getting stopped by `minClientVersion`)

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,72 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
-		if ks != nil {
-			account, err := MakeAddress(ks, etherbase)
-			if err != nil {
-				Fatalf("Invalid miner etherbase: %v", err)
-			}
-			cfg.Miner.Etherbase = account.Address
-			cfg.Etherbase = account.Address
-		} else {
-			Fatalf("No etherbase configured")
+	if validator != "" && ctx.GlobalIsSet(MinerValidatorFlag.Name) {
```
asaj, line 1087: nit: Is this cleaner?

```
if ctx.GlobalIsSet(MinerValidatorFlag.Name) {
  if validator != "" {
    Fatalf("...")
  }
  validator = ctx.Global...
}
```

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,72 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
+func setTxFeeRecipient(ctx *cli.Context, ks *keystore.KeyStore, cfg *eth.Config) {
+	// Extract the current txFeeRecipient, new flag overriding legacy etherbase
+	var txFeeRecipient string
+	if ctx.GlobalIsSet(MinerLegacyEtherbaseFlag.Name) {
+		txFeeRecipient = ctx.GlobalString(MinerLegacyEtherbaseFlag.Name)
+	}
+	if ctx.GlobalIsSet(MinerEtherbaseFlag.Name) {
+		txFeeRecipient = ctx.GlobalString(MinerEtherbaseFlag.Name)
+	}
+	if txFeeRecipient != "" && ctx.GlobalIsSet(TxFeeRecipientFlag.Name) {
```
asaj, line 1115: Ditto

consensus/istanbul/backend/backend.go
```diff
@@ -468,15 +471,24 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
 	if txnHash != block.Header().TxHash {
 		return 0, errMismatchTxhashes
 	}
-
-	// The author should be the first person to propose the block to ensure that randomness matches up.
-	addr, err := sb.Author(block.Header())
+	minVersion := &params.VersionInfo{Major: 1, Minor: 1, Patch: 0}
+	version, err := blockchain_parameters.GetMinimumVersion(nil, nil)
```
asaj, line 475: I expect this will be a pattern we use frequently, so consider adding something that allows you to do:
```
if blockchain_parameters.IsMinimumVersionAtLeast(1, 1, 0) {
  addr, err := sb.Author(block.header())
}
```

consensus/istanbul/backend/backend.go
```diff
@@ -468,15 +471,24 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
 	if err != nil {
-		sb.logger.Error("Could not recover orignal author of the block to verify the randomness", "err", err, "func", "Verify")
-		return 0, errInvalidProposal
-	} else if addr != block.Header().Coinbase {
-		sb.logger.Error("Original author of the block does not match the coinbase", "addr", addr, "coinbase", block.Header().Coinbase, "func", "Verify")
-		return 0, errInvalidCoinbase
+		if err != comm_errors.ErrRegistryContractNotDeployed && err != comm_errors.ErrSmartContractNotDeployed {
+			log.Debug("Error checking client version", "err", err, "contract", hexutil.Encode(params.BlockchainParametersRegistryId[:]))
+			return 0, err
```
asaj, line 479: What are the implications of returning an error here if we can't read the minimum client version? Should we fail to verify proposals as a result?

consensus/istanbul/backend/backend.go
```diff
@@ -468,15 +471,24 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
+		}
+	} else {
+		if version.Cmp(minVersion) <= 0 {
```
asaj, line 482: Do we expect this PR to be included in 1.1.0? If so, if the minimum client version was 1.1.0, wouldn't we want to allow the author to differ from the coinbase?

consensus/istanbul/backend/backend.go
```diff
@@ -468,15 +471,24 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
+			addr, err := sb.Author(block.Header())
+			if err != nil {
+				sb.logger.Error("Could not recover orignal author of the block to verify the randomness", "err", err, "func", "Verify")
```
asaj, line 485: How does this check relate to randomness?

consensus/istanbul/backend/backend.go
```diff
@@ -468,15 +471,24 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
+				return 0, errInvalidProposal
+			} else if addr != block.Header().Coinbase {
+				sb.logger.Error("Support for differing validator and txFeeRecipient is forthcoming")
```
asaj, line 488: nit: Consider `Block proposal author and coinbase must be the same when the minimum client version is less than 1.1.0`

consensus/istanbul/backend/engine.go
```diff
@@ -335,7 +335,9 @@ func (sb *Backend) VerifySeal(chain consensus.ChainReader, header *types.Header)
 // rules of a particular engine. The changes are executed inline.
 func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) error {
 	// unused fields, force to set to empty
-	header.Coinbase = sb.address
+	if header.Coinbase == (common.Address{}) {
```
asaj, line 338: Can you explain to me why this needs to change?

consensus/istanbul/backend/backend.go
lucasege: I wasn't sure, I think the answer should be yes, although I'm not sure what the alternative would be? We can't have them continue to verify proposals, but maybe there is a middle ground?

consensus/istanbul/backend/backend.go
lucasege: My understanding was no we wouldn't want that: we want version 1.1 to enforce using/reading from the `header.Coinbase` but not writing to it.
Then version 1.1+ will allow them to change it.

Is this not what we discussed?

consensus/istanbul/backend/engine.go
lucasege: Without this in place if a proposer tries to set their own `header.Coinbase` with `txFeeRecipient` it will be overwritten here. This enforces that if the proposer has set `header.Coinbase` intentionally, do not overwrite.

consensus/istanbul/backend/backend.go
lucasege: Resolved via slack, we indeed want to use < 1.1 as asa described

cmd/geth/main.go
```diff
@@ -95,6 +95,8 @@ var (
 		utils.UltraLightOnlyAnnounceFlag,
 		utils.WhitelistFlag,
 		utils.EtherbaseFlag,
+		utils.MinerValidatorFlag,
```
nategraf, line 98: Nit: Could we move this flag to be grouped with the miner flags?

cmd/utils/flags.go
```diff
@@ -218,7 +218,17 @@ var (
 	}
 	EtherbaseFlag = cli.StringFlag{
 		Name:  "etherbase",
-		Usage: "Public address for transaction broadcasting and block mining rewards (default = first account)",
+		Usage: "Public address for transaction broadcasting and block mining rewards (default = first account, deprecated, use --tx-fee-recipient and --validator instead)",
+		Value: "0",
+	}
+	MinerValidatorFlag = cli.StringFlag{
```
nategraf, line 224: Nit: Same comment about moving this to group with the mining flags.

cmd/utils/flags.go
```diff
@@ -218,7 +218,17 @@ var (
 	}
 	EtherbaseFlag = cli.StringFlag{
```
nategraf, line 219: How does this flag differ from `MinerLegacyEtherbaseFlag`? It seems to me like there might have been a bad merge here. I dont see this flag used in this file.

cmd/utils/flags.go
```diff
@@ -218,7 +218,17 @@ var (
...
+		Value: "0",
+	}
+	MinerValidatorFlag = cli.StringFlag{
+		Name:  "miner.validator",
+		Usage: "Public address for participation in consensus (default = first account)",
+		Value: "0",
+	}
+	TxFeeRecipientFlag = cli.StringFlag{
+		Name:  "tx-fee-recipient",
+		Usage: "Public address for block transaction fees (default = first account)",
```
nategraf, line 231: This address also receives gateway fees, which should be mentioned in this comment.

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,72 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
+		if err != nil {
+			Fatalf("Invalid txFeeRecipient: %v", err)
 		}
+		cfg.TxFeeRecipient = account.Address
 	}
 }
 
 // setBLSbase retrieves the blsbase either from the directly specified
 // command line flags or from the keystore if CLI indexed.
+// `BLSbase` is the address derived from the BLS private key used for block finalization in consensus.
```
nategraf, line 1133: If I understand correctly, this flag is the Ethereum address which identifies an ECDSA key from which the BLS key is derived.

cmd/utils/flags.go
```diff
@@ -1062,41 +1072,72 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
-			if err != nil {
-				Fatalf("Invalid miner etherbase: %v", err)
-			}
-			cfg.Miner.Etherbase = account.Address
-			cfg.Etherbase = account.Address
-		} else {
-			Fatalf("No etherbase configured")
+	if ctx.GlobalIsSet(MinerValidatorFlag.Name) {
+		if validator != "" {
+			Fatalf("`etherbase` and `validator` flag should not be used together. `validator` and `txFeeRecipient` constitute both of `etherbase`' functions")
```
nategraf, line 1089: Nit: Please change this to read like the user with fee the flags.

cmd/geth/usage.go
```diff
@@ -82,6 +82,8 @@ var AppHelpFlagGroups = []flagGroup{
 			utils.LightKDFFlag,
 			utils.WhitelistFlag,
 			utils.EtherbaseFlag,
+			utils.MinerValidatorFlag,
```
nategraf, line 85: Nit: Could we group this flag with the other miner flags?

consensus/istanbul/backend/engine.go
```diff
@@ -335,7 +335,9 @@ func (sb *Backend) VerifySeal(chain consensus.ChainReader, header *types.Header)
 // rules of a particular engine. The changes are executed inline.
 func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) error {
 	// unused fields, force to set to empty
```
nategraf, line 337: Nit: This comment seems to be inaccurate or at least misleading

core/state_processor.go
```diff
@@ -67,7 +67,8 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	}
 
 	if random.IsRunning() {
-		err := random.RevealAndCommit(block.Randomness().Revealed, block.Randomness().Committed, header.Coinbase, header, statedb)
+		author, _ := p.bc.Engine().Author(header) // Ignore error, we're past header validation
```
nategraf, line 70: We should at least log any error that comes up here.

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -114,6 +114,19 @@ func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionI
...
 
+func IsMinimumVersionAtLeast(major uint64, minor uint64, patch uint64) (bool, error) {
+	desiredVersion := &params.VersionInfo{Major: major, Minor: minor, Patch: patch}
+	minVersion, err := GetMinimumVersion(nil, nil)
+	if err != nil {
+		if err != errors.ErrRegistryContractNotDeployed && err != errors.ErrSmartContractNotDeployed {
+			log.Debug("Error checking client version", "err", err, "contract", hexutil.Encode(params.BlockchainParametersRegistryId[:]))
+			return false, err
+		}
+		return true, nil
```
nategraf, line 125: Returning true here make an assumption about the behavior of the caller, specifically that the check should be defualt pass. This was surprising to me when I first saw the call site. I agree that we should handle the "contract not deployed error here", but could this be refactored to give the caller a choice about what to do when that case arises? Returning `(*bool, error)` or `(gte bool, ok bool, error)` or a specific error that can be checked would be some ways to do that, in which case the return value indicates when the minimum client version is "undefined".

eth/backend.go
```diff
@@ -374,27 +375,54 @@ func (s *Ethereum) ResetWithGenesisBlock(gb *types.Block) {
...
+			log.Info("TxFeeRecipient automatically configured", "address", txFeeRecipient)
+			return txFeeRecipient, nil
 		}
 	}
-	return common.Address{}, fmt.Errorf("etherbase must be explicitly specified")
+	return common.Address{}, fmt.Errorf("txFeeRecipient must be explicitly specified")
+}
+
+func (s *Ethereum) Etherbase() (common.Address, error) {
+	return s.Validator()
```
nategraf, line 425: In `eth/api.go` the `Etherebase` function routes to the tx fee recipient. This is inconsistent with that, and it seems like changing this to tx fee recipient would have better sematics.

cmd/utils/flags.go
lucasege: I think you're right, removed.

eth/backend.go
lucasege: It appears `ethstats` uses this address to sign data, do you know what address would be idiomatic to use here?
I would assume `validator`, but agree this isn't great semantics.
If you agree that the signer should be `validator` maybe I'll change this from `etherbase` to `validator()`

nategraf: It uses the validator signer address, so making this `Validator` would work well

PR #809, opened 2020-01-15
Update EstimateGas to work for accounts w small balances

annakaz:
### Description

Previously, the EstimateGas binary search assumed that if a tx fails to execute it was due to gas being too low, so it increased gas. But sometimes a tx fails to execute because gas is too high, as the sender cannot afford to pay the gas. This updates the binary search to distinguish between those cases

### Tested

Ran a full node and sent an estimate gas rpc call
Redeemed an invite using the light client on a pixel 2

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/2429

### Backwards compatibility

Yes


core/state_transition.go
```diff
@@ -318,7 +318,8 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 
 	err = st.payFees()
 	if err != nil {
-		log.Error("Transaction failed to buy gas", "err", err, "gas", gas)
+		// Expected error when called by EstimateGas() for accounts with balance < binary search hi
+		log.Warn("Transaction failed to buy gas. This is expected when invoked by EstimateGas() for accounts with small balances", "err", err, "gas", gas)
```
nategraf: So this is an error because if a transaction is being processed here, it is assumed that it was already validated by the transaction pool, in which case this should never happen. We are breaking that assumption in the `EstimateGas` function. I think the best way to handle this would be to run the transaction through transaction pool validation, either in its entirety or just to validate that the transaction can pay its fees, before applying the state transition.
Here is the `validateTx` function on `TxPool`, which you obviously can't call directly as it stands https://github.com/celo-org/celo-blockchain/blob/f86d6733274bb43377584cb9475a2fe46e17f368/core/tx_pool.go#L607
And here is the part that validates balances to ensure fees can be paid https://github.com/celo-org/celo-blockchain/blob/f86d6733274bb43377584cb9475a2fe46e17f368/core/tx_pool.go#L1201
Two courses of action would either be to export the former, or just call the latter if that turns out to be inappropriate. The call should probably occur here-ish https://github.com/celo-org/celo-blockchain/blob/f86d6733274bb43377584cb9475a2fe46e17f368/internal/ethapi/api.go#L740

If this seems onerous, please leave this log statement as is and copy the above into an issue to address later. And feel free to assign me to it.

internal/ethapi/api.go
```diff
@@ -778,27 +778,35 @@ func (s *PublicBlockChainAPI) EstimateGas(ctx context.Context, args CallArgs) (h
 	cap = hi
 
 	// Create a helper to check if a gas allowance results in an executable transaction
-	executable := func(gas uint64) bool {
+	executable := func(gas uint64) (bool, error) {
```
nategraf: Just return the error, because you can just check `err == nil` instead of having this bool.

internal/ethapi/api.go
```diff
@@ -778,27 +778,35 @@ func (s *PublicBlockChainAPI) EstimateGas(ctx context.Context, args CallArgs) (h
...
 	}
 	// Execute the binary search and hone in on an executable gas limit
 	for lo+1 < hi {
 		mid := (hi + lo) / 2
-		if !executable(mid) {
-			lo = mid
+		exec, execError := executable(mid)
+		if !exec {
+			if execError == vm.ErrOutOfGas { // Not enough gas
+				log.Debug("Ran out of gas, will increase lo to mid", "gas", mid)
```
nategraf: I would make this trace or remove it entirely.

internal/ethapi/api.go
```diff
@@ -778,27 +778,35 @@ func (s *PublicBlockChainAPI) EstimateGas(ctx context.Context, args CallArgs) (h
...
+				lo = mid
+			} else { // Gas too high
+				log.Debug("Failed because gas too high, lowering hi to mid", "gas", mid)
+				hi = mid
+			}
 		} else {
 			hi = mid
 		}
 	}
 	// Reject the transaction as invalid if it still fails at the highest allowance
```
nategraf:
```suggestion
	// Reject the transaction as invalid if it still runs out of gas at the highest allowance
```

internal/ethapi/api.go
```diff
@@ -778,27 +778,35 @@ func (s *PublicBlockChainAPI) EstimateGas(ctx context.Context, args CallArgs) (h
...
 	if hi == cap {
-		if !executable(hi) {
+		exec, _ := executable(hi)
+		if !exec {
 			return 0, fmt.Errorf("gas required exceeds allowance or always failing transaction")
```
nategraf: We can change this comment to "gas required for transaction exceeds allowance" because we will now only reach this point on consistent "out of gas" errors.

internal/ethapi/api.go
annakaz: This follows the pattern of returning (bool, error) throughout the celo-blockchain codebase- any reason not to be consistent?

internal/ethapi/api.go
annakaz: Updated to trace

internal/ethapi/api.go
annakaz: Good catch, updated comment

core/state_transition.go
annakaz: Set back to original logging and created the issue!

internal/ethapi/api.go
```diff
@@ -790,18 +790,32 @@ func (s *PublicBlockChainAPI) EstimateGas(ctx context.Context, args CallArgs) (h
...
+		lo = params.TxGas - 1 // Re initialize to lower limit
+		for lo + 1 < hi {
+			mid := (hi + lo) / 2
+			if !executable(mid) { // Assume failure from insufficent balance
+				hi = mid
+			} else {
+				lo = mid
+			}
 		}
+		hi = lo // Hi is one too high to execute, set to executable value
```
annakaz: Not sure what geth storage costs are like, can make another var `gas` if cheap and set `gas = hi` in case above, `gas = lo` here to make more readable

nategraf: You don't need to worry about EVM storage costs here, if that's what you are talking about. This code is only run as an RPC to request a fee estimate, not as part of the consensus.

PR #815, opened 2020-01-20
Android and iOS build on CircleCI

jeanregisser:
### Description

This PR enables automatic building and publishing of Geth for both Android and iOS using CircleCI.
- [x] Build Geth for Android and iOS using CircleCI
- [x] Publish to npm (secret NPM token needed)
- [x] Publish only when building master

### Tested

Tested that it can build Geth for both Android and iOS.

### Other changes

- Removed Android specific cloudbuild
- Use `tar` instead of `zip` to compress the built iOS framework. This is to prevent postinstall issues on linux systems without the `unzip` command, for people using `celo-monorepo`.

### Related issues

- https://github.com/celo-org/celo-monorepo/issues/1154
- https://github.com/celo-org/celo-monorepo/issues/1166

### Backwards compatibility

Yes


scripts/publish-mobile-client.sh
```diff
@@ -0,0 +1,28 @@
...
+a=( ${version//./ } )
+((a[2]++))
+new_version="${a[0]}.${a[1]}.${a[2]}"
+
+# Add npm token if provided
+if [ -n "$npm_token" ]; then
+  echo "//registry.npmjs.org/:_authToken=$npm_token" > ~/.npmrc
+fi
+
+rm README.md
```
jmrossy: Why remove the README?

jeanregisser: Not sure, I copied this from the previous script.
@cmcewen do you know why we did this?

jmrossy: I'm guessing it's why we see this error here: https://www.npmjs.com/package/@celo/client

cmcewen: Yeah that's just because the readme wasn't up to date and is for the blockchain - thought we should make it clear that this is the mobile build

jeanregisser: Agreed, I've added a TODO for now.

PR #816, opened 2020-01-20
Announce protocol v2

### LINKED ISSUE #811: Reduce bandwidth of announce messages

kevjue:
### Description

This PR makes changes to the Announce protocol so that it will drastically reduce the amount of network bandwidth.  Also, it will reduce the amount of time for a new node to learn all of the validator enodeURLs.

# Announce Protocol v2 Specification

## High level design

The main difference between v1 and v2 is that ALL full nodes and validators in the Celo p2p network will store a cache of the latest versioned announce message from each validator (it will only cache the current registered/elected validator set announce messages).  The version of an announce message will be represented by the announce's originating node's local timestamp.  Note that the cache in this PR is in-memory.  There should be a follow up PR to persist that cache.

As with v1, each node with --mine command line option will periodically generate a new 
 IstanbulAnnounce message (see comments in the code of why it's not just registered/elected validators that do the gossip.   Note that only messages from registered/elected validators will be propagated throughout the p2p network).  When a node generates it's announce message, it will also cache it's own message since a peer may later request it using the protocol described below.  Once a separate node receives one of these gossiped announce message, it will update it's cache, then re-gossip that message to it's peers.

Also, when two nodes peer with each other, they will request the other node's announce cache's version set (each element of the set will contain a (originating validator address, version) tuple) via the IstanbulGetAnnounceVersions message.  The nodes will reply with the IstanbulAnnounceVersions with those versions sets.  The nodes will then determine if any of the other node's announce message are a later version than it's own, and then request those later-versioned announce messages from that node via a IstanbulGetAnnounces message.

Periodically, a node will check for later-versioned announce message by running the protocol described in the paragraph above with all of it's peers.  Also, that protocol will always be run on the start of a new epoch.

Once it's established that the announce msg propagation via the version set compare protocol is reliable and performant, then the periodic gossiping of announce message can be removed.  The only time a node will gossip an announce message is when it starts up or if it's signing key or ip address is rotated.

## Message types

Msg Name | Msg Description | Msg Code | Msg Payload
-----------|------------|-----------| --------------
istanbulGetAnnouncesMsg | This message will request the cached announce messages from another peer |  0x16| RLP encoded array of validator signing addresses of the requested announce messages
istanbulAnnounceMsg  | This message will contain an announce message from the sender node.  This message type will be the response of a istanbulGetAnnouncesMsg or will be gossiped from the announce message's originating node periodically |  0x12 | RLP encoded announce msg
istanbulGetAnnounceVersionsMsg | This message will request the set of cached announce message from a connected peer | 0x17 | nil
istanbulAnnounceVersionsMsg | This message will contain the set of cached of announce version numbers from the sender node | 0x18 | RLP encoded array of announce versions

### Tested

TODO

### Other changes

* Refactored istanbul Backend's Gossip and BroadcastConsensusMsg.
* Utilized the istanbul backend's recents and known cache for ONLY announce messages (since it's the only message that is gossiped), and renamed the caches to be more accurate.
* Moved the function `retrieveActiveAndRegisteredValidators` from announce.go to backend.go

### Related issues

- Fixes #811

### Backwards compatibility

No.  The announce protocol has changed significantly and is not currently designed to be backwards compatible.


consensus/istanbul/backend/announce.go
```diff
@@ -127,114 +58,230 @@ const (
...
+	var currentAnnounceGossipTickerDuration time.Duration
+	var numGossipedMsgsInHighFreqAfterFirstPeerState int
 
 	for {
 		select {
-		case <-sb.newEpochCh:
-			go sb.sendIstAnnounce()
+		case <-coreStartedTicker.C:
+			sb.coreMu.RLock()
+			if sb.coreStarted && announceGossipTickerCh == nil {
```
kevjue: Check if it's a registered/elected validator as well.

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +390,92 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
 	}
 	logger.Trace("Handling an IstanbulAnnounce message", "from", msg.Address)
 
-	// If the message is originally from this node, then ignore it
-	if msg.Address == sb.Address() {
-		logger.Trace("Received an IstanbulAnnounce message originating from this node. Ignoring it.")
-		return nil
-	}
```
nategraf: This is gone because the caching take over that function, or no?

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +390,92 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
-	logger = logger.New("msgAddress", msg.Address, "msg_timestamp", announceData.Timestamp)
-
-	if currentEntryTimestamp, err := sb.valEnodeTable.GetTimestampFromAddress(msg.Address); err == nil {
-		if announceData.Timestamp < currentEntryTimestamp {
-			logger.Trace("Received an old announce message", "currentEntryTimestamp", currentEntryTimestamp)
-			return errOldAnnounceMessage
-		}
-	} else if err != leveldb.ErrNotFound {
-		logger.Warn("Error when retrieving timestamp for entry in the ValEnodeTable", "err", err)
+	if !regAndActiveVals[msg.Address] {
```
nategraf: We chatted about having a signature check. Is this where that should be?

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +390,92 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
+	err = rlp.DecodeBytes(msg.Msg, &announcePayload)
 	if err != nil {
+		logger.Warn("Error in decoding received Istanbul Announce message content", "err", err, "IstanbulMsg", msg.String())
 		return err
 	}
 
-	if !regAndActiveVals[msg.Address] {
-		logger.Debug("Received an IstanbulAnnounce message from a non registered validator. Ignoring it.", "AnnounceMsg", msg.String(), "err", err)
-		return errUnauthorizedAnnounceMessage
+	logger = logger.New("msgAddress", msg.Address, "msg_timestamp", announcePayload.Timestamp)
```
nategraf: Nit: Can we make these two named log field consistent in style :P

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +390,92 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
-	var processedAddresses = make(map[common.Address]bool)
+	// If this is a registered or elected validator, then process the announce message
+	var destAddresses = make([]string, 0, len(announcePayload.EncryptedEnodes))
 	var msgHasDupsOrIrrelevantEntries bool = false
-	for _, announceRecord := range announceData.AnnounceRecords {
-		// Don't process duplicate entries or entries that are not in the regAndActive valset
-		if !regAndActiveVals[announceRecord.DestAddress] || processedAddresses[announceRecord.DestAddress] {
-			msgHasDupsOrIrrelevantEntries = true
-			continue
+	if sb.coreStarted && regAndActiveVals[sb.Address()] {
```
nategraf: Does `sb.coreStarted` need to be accessed under a mutex? I noticed that was done elsewhere but it unclear to me why.
https://github.com/celo-org/celo-blockchain/blob/417be0b1925ad71ea06c70fff511bc0b68f3acd0/consensus/istanbul/backend/announce.go#L90-L105

nategraf: If this validator is not registered at the time it receives an announce, then becomes registered later, will it need to wait for a new announce message to place the enode entry into its table? If so, would it be better to simply check `sb.coreStarted` here because if this not is running Istanbul that means they intend to participate in validation, or no?

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +390,92 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
-		if err := sb.valEnodeTable.Upsert(map[common.Address]*vet.AddressEntry{msg.Address: {Node: node, Timestamp: announceData.Timestamp}}); err != nil {
-			logger.Warn("Error in upserting a valenode entry", "AnnounceData", announceData.String(), "error", err)
-			return err
-		}
 	}
 
 	if !msgHasDupsOrIrrelevantEntries {
-		if err = sb.regossipIstAnnounce(msg, payload, announceData, regAndActiveVals, destAddresses); err != nil {
+		// Cache this announce message
+		sb.cachedAnnounceMsgs[msg.Address] = &announceMsgCachedEntry{MsgTimestamp: announcePayload.Timestamp,
```
nategraf: Nit: Weird formatting

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +390,92 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
+			MsgPayload: payload}
+
+		// Prune the announce cache for entries that are not in the current registered/elected validator set
+		for cachedValAddress := range sb.cachedAnnounceMsgs {
+			if !regAndActiveVals[cachedValAddress] {
+				delete(sb.cachedAnnounceMsgs, cachedValAddress)
+			}
+		}
```
nategraf: Why is this skipped when the message has duplicate or irrelevant entries. Also, although it is fine to handle this here, it also seems like something that could be handled out-of-band. I would suggest adding a comment explaining that this is here for a particular reason, or that it is arbitrary that it be handled here.

consensus/istanbul/backend/announce.go
```diff
@@ -419,3 +516,125 @@ func (sb *Backend) regossipIstAnnounce(msg *istanbul.Message, payload []byte, an
...
+// ===============================================================
+//
+// define the AnnounceVersion messge format, the getAnnounceVersions send and handle function, and the announceVersions handle function
+
+type announceVersion struct {
+	ValAddress           common.Address
+	AnnounceMsgTimestamp uint
+}
+
+// EncodeRLP serializes av into the Ethereum RLP format.
```
nategraf: Nit: Took me a few seconds to figure out what `av` was.
```suggestion
// EncodeRLP serializes announceVersion into the Ethereum RLP format.
```

consensus/istanbul/backend/announce.go
```diff
@@ -419,3 +516,125 @@ func (sb *Backend) regossipIstAnnounce(msg *istanbul.Message, payload []byte, an
...
+func (av *announceVersion) EncodeRLP(w io.Writer) error {
+	return rlp.Encode(w, []interface{}{av.ValAddress, av.AnnounceMsgTimestamp})
+}
+
+// DecodeRLP implements rlp.Decoder, and load the ee fields from a RLP stream.
```
nategraf: Should `ee` be `announceVersion`?
```suggestion
// DecodeRLP implements rlp.Decoder, and load the announceVersion fields from a RLP stream.
```

consensus/istanbul/backend/announce.go
```diff
@@ -419,3 +516,125 @@ func (sb *Backend) regossipIstAnnounce(msg *istanbul.Message, payload []byte, an
...
+	}
+
+	if err := s.Decode(&msg); err != nil {
+		return err
+	}
+	av.ValAddress, av.AnnounceMsgTimestamp = msg.ValAddress, msg.AnnounceMsgTimestamp
+	return nil
+}
+
+// The stringer function for the announceVersion type
```
nategraf: Nite: We should try to follow the Effective Go comment guidelines. Readers who know what `Stringer` is won't need explanation that `String` satisfied it's interface.
```suggestion
// String returns the string representation of announceVersion.
```

consensus/istanbul/backend/announce.go
```diff
@@ -419,3 +516,125 @@ func (sb *Backend) regossipIstAnnounce(msg *istanbul.Message, payload []byte, an
...
+func (av *announceVersion) String() string {
+	return fmt.Sprintf("{ValAddress: %s, AnnounceMsgTimstamp: %d}", av.ValAddress.String(), av.AnnounceMsgTimestamp)
+}
+
+// This function will send a GetAnnounceVersions message to a specific peer to request it's announceVersion set
```
nategraf: Nit: The following functions should replace "this function" with the function name according to Effective Go.

consensus/istanbul/backend/announce.go
```diff
@@ -419,3 +516,125 @@ func (sb *Backend) regossipIstAnnounce(msg *istanbul.Message, payload []byte, an
...
+		logger.Trace("Going to send a GetAnnounces", "announcesToRequest", common.ConvertToStringSlice(announcesToRequest), "peer", peer)
+
+		go peer.Send(istanbulGetAnnouncesMsg, announcesToRequestBytes)
+	}
+
+	return nil
+}
+
+func (sb *Backend) checkPeersAnnounceVersions() {
+	peers := sb.broadcaster.FindPeers(nil, p2p.AnyPurpose)
```
nategraf: This will not include light clients, correct?

consensus/consensus.go
```diff
@@ -170,13 +170,19 @@ type Istanbul interface {
...
-	// Stop stops the engine
-	Stop() error
+	// StopValidating stops the validating engine
+	StopValidating() error
+
+	// StartAnnouncing starts the announcing
+	StartAnnouncing() error
+
+	// StopAnnouncing stops the announcing
+	StopAnnouncing() error
```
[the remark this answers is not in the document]
kevjue: Yes, that's correct.

consensus/istanbul/backend/announce.go
```diff
@@ -127,114 +58,230 @@ const (
...
+	return fmt.Sprintf("{DecrypterAddress: %s, EncryptedEnodeURL length: %d}", ee.DecrypterAddress.String(), len(ee.EncryptedEnodeURL))
+}
+
+// The set of encrypted enodes of ValAddress's enodeURL.  There is an encrypted enode for each registered validator.
+// The RLP encoding of this struct is the payload for the IstanbulAnnounce message.
+type valEncryptedEnodes struct {
+	ValAddress      common.Address
+	EncryptedEnodes []*encryptedEnode
+	EnodeURLHash    common.Hash
+	Timestamp       uint
```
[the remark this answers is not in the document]
kevjue: Good point.

consensus/istanbul/backend/announce.go
kevjue: Yes, that is correct.

consensus/istanbul/backend/announce.go
kevjue: Nice catch!  That check is just currently checking if the signer of the message is in the current valset and is not the original message sender.

consensus/istanbul/backend/announce.go
kevjue: That's correct.  Only the full nodes/validators.

consensus/istanbul/backend/announce.go
kevjue: Actually, the code already does check for that here:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/types.go#L375.

consensus/istanbul/backend/announce.go
kevjue: It's not really needed to be protected within a read lock.  Will remove it.

kevjue: In that scenario, that registered will periodically ask it's neighbors for their announce version set (https://github.com/celo-org/celo-blockchain/pull/816/files#diff-49675a1e6ab38a575ed8c1e9a5f31145R170) and will discover that it's missing the announce in question and then request it from it's neighbor.

consensus/istanbul/backend/val_enodes_share.go
```diff
@@ -169,7 +170,7 @@ func (sb *Backend) sendValEnodesShareMsg() error {
 	return nil
 }
 
-func (sb *Backend) handleValEnodesShareMsg(payload []byte) error {
+func (sb *Backend) handleValEnodesShareMsg(peer consensus.Peer, payload []byte) error {
```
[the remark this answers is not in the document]
kevjue: Yes.  All the handler functions for announce related messages need to follow this format:

https://github.com/celo-org/celo-blockchain/pull/816/files#diff-6e58dd6edf43343da874fb58a00f6727R209

kevjue:
> Would it be possible or worthwhile to split `consensus/istanbul/backend/announce.go` off into it's own package and possibly break up this file. I can understand if this is not worth doing now.

Yes!  I *really* want to do this as well, but for time constraint reasons haven't got to it.  I'll create a tech debt issue for it.

consensus/istanbul/backend/announce.go
```diff
@@ -127,114 +58,230 @@ const (
...
 			}
 
-			go sb.sendIstAnnounce()
+			go sb.gossipAnnounce()
+
+		case <-announceVersionsCheckTicker.C:
+			go sb.checkPeersAnnounceVersions()
+
+		case <-sb.newEpochCh:
+			if sb.coreStarted {
```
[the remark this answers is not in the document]
kevjue: This channel has been removed altogether.

consensus/istanbul/backend/announce.go
```diff
@@ -127,114 +58,230 @@ const (
...
+// This function will generate the lastest announce msg from this node (as opposed to retrieving from the announceMsgCache) and then broadcast it to it's peers,
+// which should then gossip the announce msg message throughout the p2p network (since this announce msg's timestamp should be the latest among all of this
+// validator's previous announce msgs).
+// Note that ANY node with the command line option --mine enabled, will send an announce message.  Ideally, only registered or elected validators should
+// send this message.  However, this gossip message is infrequently sent (once an hour), and the time between registration and the first announce message
+// could be up to 1 hour.  At each epoch transition, nodes will exchange their announceVersions, and if this node is now registered or elected, then
+// this node's enode will be gossiped throughout the network.
+// Ideally, this node should listen for registration smart contract events and send out the gossip message then (and stop gossiping
+// once an hour on unregistration event).  Note that when a non registered/elected validators sends a gossip message, it's neighbors will not regossip
+// those messages.
```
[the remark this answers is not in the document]
kevjue: It's removed.

consensus/istanbul/config.go
```diff
@@ -49,19 +49,25 @@ type Config struct {
 	Proxied                 bool        `toml:",omitempty"` // Specifies if this node is proxied
 	ProxyInternalFacingNode *enode.Node `toml:",omitempty"` // The internal facing node of the proxy that this proxied validator will contect to
 	ProxyExternalFacingNode *enode.Node `toml:",omitempty"` // The external facing node of the proxy that the proxied validator will broadcast via the announce message
+
+	// Announce Configs
+	AnnounceGossipPeriod                 uint64 `toml:",omitemptry"` // Time duration (in seconds) between gossiped announce messages
```
tkporter: `omitemptry` -> `omitempty`

kevjue:
> Looking super good! I'm curious what level of bandwidth savings you've seen when testing it

I'm curious too :).  I'll be running more rigorous tests in google cloud, and do some bandwidth comparisons and post here.

consensus/istanbul/backend.go
```diff
@@ -48,8 +48,10 @@ type Backend interface {
 	// BroadcastConsensusMsg sends a message to all validators (include self)
 	BroadcastConsensusMsg(validators []common.Address, payload []byte) error
 
-	// Gossip sends a message to all validators (exclude self)
-	Gossip(validators []common.Address, payload []byte, ethMsgCode uint64, ignoreCache bool) error
```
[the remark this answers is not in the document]
kevjue: Great question.  I had a meeting with @timmoreton to chat about this.  The main purpose of removal of this is that only gossiped messages will now be cached.

All direct messages will not be since they are not meant to be duplicated (which gossiped messages could be).  E.g. Tim made changes so that istanbul will periodically resend the same round change message within the same round.  Caching those types of messages will prevent those intended resends.

consensus/istanbul/backend/announce.go
```diff
@@ -122,119 +53,252 @@ const (
...
+				announceGossipTicker.Stop()
+				announceGossipTickerCh = nil
+				logger.Trace("Disabled periodic gossiping of announce message")
+			}
+
+		case <-announceGossipTickerCh: // If this is nil (when sb.coreStarted == false), this channel will never receive an event
+			logger.Trace("Going to gossip an announce message", "announceGossipFrequencyState", announceGossipFrequencyState, "numGossipedMsgsInHighFreqAfterFirstPeerState", numGossipedMsgsInHighFreqAfterFirstPeerState)
+			switch announceGossipFrequencyState {
 			case HighFreqBeforeFirstPeerState:
 				{
```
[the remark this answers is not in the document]
kevjue: Good point.

consensus/istanbul/backend/announce.go
```diff
@@ -302,83 +405,98 @@ func (sb *Backend) handleIstAnnounce(payload []byte) error {
...
-	}
-	// Save in the valEnodeTable if mining
-	if sb.coreStarted && node != nil {
-		if err := sb.valEnodeTable.Upsert(map[common.Address]*vet.AddressEntry{msg.Address: {Node: node, Timestamp: announceData.Timestamp}}); err != nil {
-			logger.Warn("Error in upserting a valenode entry", "AnnounceData", announceData.String(), "error", err)
-			return err
-		}
 	}
 
 	if !msgHasDupsOrIrrelevantEntries {
```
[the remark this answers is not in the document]
kevjue: I would argue that is what we would want to happen.  E.g. an announce msg propagation is stopped the instance it's realized that it's not relevant.

consensus/istanbul/backend/announce.go
```diff
@@ -245,52 +316,84 @@ func (sb *Backend) sendIstAnnounce() error {
...
+		encryptedEnodes = append(encryptedEnodes, &encryptedEnode{DecrypterAddress: addr, EncryptedEnodeURL: []byte(enodeUrl)})
 	}
 
-	for _, address := range registeredValidators {
-		validatorsSet[address] = true
+	announcePayload := &validatorEncryptedEnodes{
+		ValAddress:      sb.Address(),
+		EncryptedEnodes: encryptedEnodes,
+		EnodeURLHash:    istanbul.RLPHash(enodeUrl),
+		// Unix() returns a int64, but we need a uint for the golang rlp encoding implmentation. Warning: This timestamp value will be truncated in 2106.
```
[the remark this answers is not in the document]
kevjue: Comment is not relevant anymore.  I removed it.

consensus/istanbul/backend.go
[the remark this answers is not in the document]
kevjue: Let me know if you have any concerns with that change.

consensus/istanbul/backend/announce.go
```diff
@@ -122,119 +53,252 @@ const (
...
+	}
 
-	return msg, nil
+	return nil
 }
 
-func (sb *Backend) sendIstAnnounce() error {
-	istMsg, err := sb.generateIstAnnounce()
+// This function will generate the lastest announce msg from this node (as opposed to retrieving from the announceMsgCache) and then broadcast it to it's peers,
+// which should then gossip the announce msg message throughout the p2p network, since this announce msg's timestamp should be the latest among all of this
```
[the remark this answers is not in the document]
kevjue: Does this comment here work for you:  https://github.com/celo-org/celo-blockchain/pull/816/files#diff-49675a1e6ab38a575ed8c1e9a5f31145R482 ?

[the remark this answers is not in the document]
tkporter: Ah yeah that's fine

consensus/istanbul/backend/announce.go
kevjue: I added a function `validateAnnounce` that will explain the reasononing.

Basically, we want to force the validators to construct as succint announce message as possible to minimize any network usage.

p2p/server.go
```diff
@@ -568,17 +567,17 @@ func (srv *Server) setupLocalNode() error {
 	srv.localnode.SetFallbackIP(net.IP{127, 0, 0, 1})
 	srv.localnode.Set(capsByNameAndVersion(srv.ourHandshake.Caps))
 	// TODO: check conflicts
-	var primaries []string
+	primaries := make(map[string]bool)
 	for _, p := range srv.Protocols {
 		if p.Primary {
-			primaries = append(primaries, p.Name)
+			primaries[p.Name] = true
```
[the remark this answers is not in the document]
kevjue: Yes, it's ok.

cmd/utils/flags.go
```diff
@@ -672,6 +672,17 @@ var (
...
+	AnnounceGossipPeriodFlag = cli.Uint64Flag{
+		Name:  "announce.gossipperiod",
+		Usage: "Time duration (in seconds) between gossiped announce messages",
+		Value: eth.DefaultConfig.Istanbul.AnnounceGossipPeriod,
+	}
+	AnnounceAggressiveGossipOnEnablementFlag = cli.BoolFlag{
+		Name:  "announce.aggressivegossiponenablement",
+		Usage: "Specifies if this node should do aggressive gossip on announce enablement",
+	}
+
```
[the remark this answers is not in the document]
kevjue: I actually argue that is it not a detail of Istanbul, and that it just happens that all the code is in the istanbul package.   And that the announce protocol is a layer underneath the istanbul consensus layer.

E.g. none of the announce data structes and messages are dependent on any of the consensus message.

wdty?

consensus/istanbul/backend/backend.go
```diff
@@ -794,3 +818,44 @@ func (sb *Backend) ConnectToVals() {
 		sb.RefreshValPeers(valset)
 	}
 }
+
+func (sb *Backend) retrieveRegisteredAndElectedValidators() (map[common.Address]bool, error) {
```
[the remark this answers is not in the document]
kevjue: I knew you were going to mention this :).

I was a bit worried about inconsistent reads.  e.g. if the reg/elected valset is cached from a few blocks ago, what's the impact of that to user of that cache.  I'll do some thinking about this.

p2p/server.go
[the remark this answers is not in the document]
kevjue: To provide a bit more detail, what's going to happen is that in the handshake, the peers will figure out the highest version within the "protocol name" they have in common and use that.

cmd/utils/flags.go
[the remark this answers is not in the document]
nategraf: If a node were running PoW, would it ever use the announce messages? No, right or am I missing something? Essentially I would lump it in with Istanbul because it's only relevant if you are running Istanbul. (In my view flag naming is communicating functionality groups to the reader moreso than the way the code is structured)

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
+				logger.Trace("Enabled periodic gossiping of announce message")
+			} else if !shouldAnnounce && announceGossipTickerCh != nil {
+				// Disable periodic gossiping by setting announceGossipTickerCh to nil
+				announceGossipTicker.Stop()
+				announceGossipTickerCh = nil
+				logger.Trace("Disabled periodic gossiping of announce message")
+			}
+
+			// Use this timer to also prune all announce related data structures.
+			if err := sb.pruneAnnounceDataStructures(); err != nil {
```
nategraf: Pruning announce data structures could be expensive due to the need for EVM data (and less so because it loops over all the cached data.). It may not be wise to do that every 5 seconds. I might suggest pruning the table as part of the send process, but maybe that is too infrequent.

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
+				logger.Warn("Error is pruning announce data structures", "err", err)
+			}
+
+		case <-announceGossipTickerCh: // If this is nil (when shouldAnnounce was most recently false), this channel will never receive an event
+			logger.Trace("Going to gossip an announce message", "announceGossipFrequencyState", announceGossipFrequencyState, "numGossipedMsgsInHighFreqAfterFirstPeerState", numGossipedMsgsInHighFreqAfterFirstPeerState)
+			switch announceGossipFrequencyState {
+			case HighFreqBeforeFirstPeerState:
+				if len(sb.broadcaster.FindPeers(nil, p2p.AnyPurpose)) > 0 {
```
nategraf: What happens when `sb.gossipAnnounce` is called when `len(sb.broadcaster.FindPeers(nil, p2p.AnyPurpose)) == 0`. Based on a quick reading, it generates and caches the message, then doesn't send it, right? Is there any reason we'd want to do that?

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
+					announceGossipFrequencyState = HighFreqAfterFirstPeerState
+				}
+
+			case HighFreqAfterFirstPeerState:
+				if numGossipedMsgsInHighFreqAfterFirstPeerState >= 10 {
+					announceGossipFrequencyState = LowFreqState
+				}
+				numGossipedMsgsInHighFreqAfterFirstPeerState += 1
+
+			case LowFreqState:
```
nategraf: How this is implemented right now results in a slightly odd edge case in that 12 announce messages will be sent at high frequency before actually slowing down, when it seems the intention may be for 10. +1 because it checks `numGossipedMsgsInHighFreqAfterFirstPeerState` before incrementing it, and +1 because the ticker remains at high frequency for one more iteration. It's also a little odd in that the loop invariant of `currentAnnounceGossipTickerDuration` matching `announceGossipFrequencyState` is broken for one iteration. Would it make sense to move the code from the `LowFreqState` into the if statement of the one above?

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
+	return fmt.Sprintf("{DecrypterAddress: %s, EncryptedEnodeURL length: %d}", ee.DecrypterAddress.String(), len(ee.EncryptedEnodeURL))
+}
+
+// The set of encrypted enodes of ValAddress's enodeURL.  There is an encrypted enode for each registered validator.
+// The RLP encoding of this struct is the payload for the IstanbulAnnounce message.
+type validatorEncryptedEnodes struct {
+	ValAddress      common.Address
+	EncryptedEnodes []*encryptedEnode
+	EnodeURLHash    common.Hash
+	Version         uint64 // This fields first version will have a value of the ValAddress's local timestamp at the time it sent it's announce message
```
nategraf: I was a bit confused about what this comment meant on first reading. It could be simplified to 'Local Unix timestamp is used for the version number' or no?

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
 
-	// Set the initial states
-	announceThreadState := HighFreqBeforeFirstPeerState
-	currentTickerDuration := HighFreqTickerDuration
-	ticker := time.NewTicker(currentTickerDuration)
-	numSentMsgsInHighFreqAfterFirstPeerState := 0
+// This function will generate the lastest announce msg from this node (as opposed to retrieving from the announceMsgCache) and then broadcast it to it's peers,
+// which should then gossip the announce msg message throughout the p2p network, since this announce msg's timestamp should be the latest among all of this
+// validator's previous announce msgs.
+func (sb *Backend) gossipAnnounce() error {
```
nategraf: It might be worth renaming this to `generateAndGossipAnnounce` to make it clear that a fresh announce will be generated and cached.

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
-						// Reset the ticker
-						currentTickerDuration = LowFreqTickerDuration
-						ticker.Stop()
-						ticker = time.NewTicker(currentTickerDuration)
-					}
-				}
-			}
+	// Add the generated announce message to this node's cache
+	var announcePayload validatorEncryptedEnodes
+	rlp.DecodeBytes(istMsg.Msg, &announcePayload)
```
nategraf: Can this be restructured to avoid doing the decode here seeing as we just generated that message?

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
+	sb.cachedAnnounceMsgs[announcePayload.ValAddress] = &announceMsgCachedEntry{MsgVersion: announcePayload.Version,
+		MsgPayload: payload}
 
-			go sb.sendIstAnnounce()
+	sb.Multicast(nil, payload, istanbulAnnounceMsg)
```
nategraf: Even though we know, based on the current implementation, `sb.Multicast` will return a `nil` error, we should still handle it. One reason for this is that if someone updates it to return non-nil in the future they likely won't think to update the call sites. (This comment applied to all the call sites for `sb.Multicast` in this file)

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
 
-	announceBytes, err := rlp.EncodeToBytes(announceData)
+	announceBytes, err := rlp.EncodeToBytes(announcePayload)
 	if err != nil {
-		sb.logger.Error("Error encoding announce content in an Istanbul Validator Enode Share message", "AnnounceData", announceData.String(), "err", err)
+		logger.Error("Error encoding announce payload for an Announce message", "AnnouncePayload", announcePayload.String(), "err", err)
 		return nil, err
 	}
 
+	// This message should be signed before it's sent out, or else the other nodes will reject it.
```
nategraf: What is the logic behind not simply signing it here?

consensus/istanbul/backend/announce.go
```diff
@@ -17,405 +17,671 @@
...
+// create as succint message as possible, and prevent any possible network DOS attacks
+// via extremely large announce message.
+func (sb *Backend) validateAnnounce(announcePayload *validatorEncryptedEnodes) (bool, error) {
+	logger := sb.logger.New("func", "validateAnnounce")
+
+	// Check if there are any duplicates in the announce message
+	var encounteredAddresses = make(map[common.Address]bool)
+	for _, encryptedEnode := range announcePayload.EncryptedEnodes {
+		if encounteredAddresses[encryptedEnode.DecrypterAddress] {
+			logger.Info("announce message has duplicate entries", "address", encryptedEnode.DecrypterAddress)
```
nategraf: Nit: Log messages star with capital letters

eth/backend.go
```diff
@@ -209,9 +215,8 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
 				select {
 				case chainHeadEvent := <-chainHeadCh:
 					istanbul.NewChainHead(chainHeadEvent.Block)
-				case <-chainHeadSub.Err():
+				case err := <-chainHeadSub.Err():
```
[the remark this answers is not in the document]
kevjue: Nice catch.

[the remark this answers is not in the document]
kevjue: I decided to exit the goroutine, as that is the same that that the mining worker does when it gets an error:

https://github.com/celo-org/celo-blockchain/blob/master/miner/worker.go#L525

consensus/istanbul/backend/announce.go
kevjue: The reason it caches is that when peers connect to it, this node will send to them it's cached entry (and then those nodes will subsequently gossip it out to to the rest of the network).

consensus/istanbul/backend/announce.go
kevjue: I placed it in the "send" timer.  This pruning can happen relatively infrequently.

consensus/istanbul/backend/announce.go
kevjue: do you have a strong opinion about this code block?  This code is mostly mean for backward's compatibility between phase 2 and phase 3 nodes, and the plan is to remove this altogether for phase 4.

nategraf: I didn't really think about that. I wouldn't worry too much about it perfecting it then.

consensus/istanbul/backend/announce.go
nategraf: 👍 That makes sense.

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
kevjue: The node is now caching the reg/elected valset such that it's retrieved at most once a minute.  Here's the commit:  https://github.com/celo-org/celo-blockchain/pull/816/files/5f32145a33146ce1b79c123a4a542146596cce02..ab4fdb2526f16a22611261847372a885088fbafb

I also made changes so that the first gossiped announce message will be delayed by 1 minute.  This will be so that all of the receivers of that message will not use a stale valset (which may cause them to not process the gossiped message).

cmd/utils/flags.go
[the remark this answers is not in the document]
kevjue: That is correct.  However, PoW may not be a great example w.r.t. my argument, as there is no "valset", since any node could be a "validator" without an election.

However, it wouldn't be too hard to get the announce protocol to work for geth's PoA implementation.  Or any other consensus algorithm that requires a subnetwork among a designated set of nodes.  So my thinking that "in principle" it is independent of the consensus algorithm, but it just happens to be currently tightly coupled with it.

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
nategraf: Ok, thanks for making this change!

kevjue:
> Looks good to me. I left one more comment. Thanks for getting all this done!

Thanks for your thorough review and great feedback!

consensus/istanbul/backend/backend.go
```diff
@@ -794,3 +769,48 @@ func (sb *Backend) ConnectToVals() {
...
+	sb.cachedRegisteredElectedValSetMu.Lock()
+	defer sb.cachedRegisteredElectedValSetMu.Unlock()
+
+	// Check to see if there is a cached registered/elected validator set, and if it's for the current block
+	if sb.cachedRegisteredElectedValSet != nil && time.Since(sb.cachedRegisteredElectedValSetTimestamp) <= 1*time.Minute {
+		return sb.cachedRegisteredElectedValSet, nil
```
tkporter: Returning this map isn't quite thread safe is it? If we want to keep things as safe as possible we should make a copy of this before returning

consensus/istanbul/backend/handler.go
```diff
@@ -29,40 +29,53 @@ import (
...
+	istanbulGetAnnouncesMsg        = 0x16
+	istanbulGetAnnounceVersionsMsg = 0x17
+	istanbulAnnounceVersionsMsg    = 0x18
 )
 
-var (
-	// errDecodeFailed is returned when decode message fails
-	errDecodeFailed = errors.New("fail to decode istanbul message")
+func (sb *Backend) isIstanbulMsg(msg p2p.Msg) bool {
+	return msg.Code >= istanbulConsensusMsg && msg.Code <= istanbulAnnounceVersionsMsg
```
tkporter: 👏way better

consensus/istanbul/backend/announce.go
```diff
@@ -17,208 +17,386 @@
...
-	Timestamp       uint
+func (sb *Backend) shouldGenerateAndProcessAnnounce() (bool, error) {
+
+	// Check if this node is in the registered/elected validator set
+	regAndElectedVals, err := sb.retrieveRegisteredAndElectedValidators()
+	if err != nil {
+		return false, err
+	}
+
+	return sb.coreStarted && regAndElectedVals[sb.Address()], nil
```
tkporter: is the core mutex needed here?

consensus/istanbul/backend/backend.go
```diff
@@ -371,26 +339,49 @@ func (sb *Backend) Gossip(destAddresses []common.Address, payload []byte, ethMsg
...
+	go sb.istanbulEventMux.Post(msg)
+	return nil
+}
+
+// Multicast implements istanbul.Backend.Multicast
+// Multicast will send the eth message (with the message's payload and msgCode field set to the params
+// payload and ethMsgCode respectively) to the nodes with the signing address in the destAddresses param.
+// If the destAddresses param is set to nil, then this function will send the message to all connected
+// peers.
+func (sb *Backend) Multicast(destAddresses []common.Address, payload []byte, ethMsgCode uint64) error {
```
tkporter: This function will always return nil, so we can get rid of its returned error

consensus/istanbul/backend/announce.go
```diff
@@ -17,208 +17,386 @@
...
+	// Convert to payload
+	payload, err := istMsg.Payload()
+	if err != nil {
+		logger.Error("Error in converting Istanbul Announce Message to payload", "AnnounceMsg", istMsg.String(), "err", err)
+		return err
+	}
 
-			go sb.sendIstAnnounce()
+	// Add the generated announce message to this node's cache
+	sb.cachedAnnounceMsgs[announceAddress] = &announceMsgCachedEntry{MsgVersion: announceVersion, MsgPayload: payload}
```
tkporter: Because `cachedAnnounceMsgs` can be read/written outside of the main event loop we should be locking here

consensus/istanbul/backend/announce.go
```diff
@@ -228,194 +406,286 @@ func (sb *Backend) generateIstAnnounce() (*istanbul.Message, error) {
...
-	sb.lastAnnounceGossiped[msg.Address] = &AnnounceGossipTimestamp{enodeURLHash: announceData.EnodeURLHash, timestamp: time.Now(), destAddressesHash: destAddressesHash}
-
-	// prune non registered validator entries in the valEnodeTable, reverseValEnodeTable, and lastAnnounceGossiped tables about 5% of the times that an announce msg is handled
-	if (mrand.Int() % 100) <= 5 {
-		for remoteAddress := range sb.lastAnnounceGossiped {
-			if !regAndActiveVals[remoteAddress] {
-				logger.Trace("Deleting entry from the lastAnnounceGossiped table", "address", remoteAddress, "gossip timestamp", sb.lastAnnounceGossiped[remoteAddress])
-				delete(sb.lastAnnounceGossiped, remoteAddress)
-			}
+		if cachedEntry, ok := sb.cachedAnnounceMsgs[announceVersion.ValAddress]; !ok || cachedEntry.MsgVersion < announceVersion.AnnounceMsgVersion {
```
tkporter: We should be locking here as well

consensus/istanbul/backend/announce.go
```diff
@@ -17,208 +17,386 @@
...
+	logger := sb.logger.New("func", "pruneAnnounceDataStructures")
+
+	// retrieve the registered/elected validator set
+	regAndElectedVals, err := sb.retrieveRegisteredAndElectedValidators()
+	if err != nil {
+		return err
+	}
+
+	// Prune the announce cache for entries that are not in the current registered/elected validator set
+	for cachedValAddress := range sb.cachedAnnounceMsgs {
```
tkporter: again given we can be modifying `cachedAnnounceMsgs` in separate goroutines we should lock

consensus/istanbul/backend/announce.go
```diff
@@ -228,194 +406,286 @@ func (sb *Backend) generateIstAnnounce() (*istanbul.Message, error) {
...
 	}
 
 	return nil
 }
+
+func (sb *Backend) checkPeersAnnounceVersions() {
+	peers := sb.broadcaster.FindPeers(nil, p2p.AnyPurpose)
+
+	for _, peer := range peers {
+		if peer.Version() == Istanbul65 {
```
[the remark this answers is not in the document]
kevjue: Good point.  Made the change.

consensus/istanbul/backend/backend.go
```diff
@@ -371,26 +339,49 @@ func (sb *Backend) Gossip(destAddresses []common.Address, payload []byte, ethMsg
...
+	// Get peers to send.
 	peers := sb.getPeersForMessage(destAddresses)
 
+	// Only cache for the announceMsg, as that is the only message that is gossiped.
 	var hash common.Hash
-	if !ignoreCache {
+	if ethMsgCode == istanbulAnnounceMsg {
 		hash = istanbul.RLPHash(payload)
-		sb.knownMessages.Add(hash, true)
+		sb.selfRecentMessages.Add(hash, true)
```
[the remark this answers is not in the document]
kevjue: I actually think it's worth caching it.  Once this node starts peering with other nodes, they will discover the cached announce message, and then once gossiping of that message is started, it's entirely possible that the message will end up sent to the node.   Caching it would prevent this node from processing it.

consensus/istanbul/backend/announce.go
kevjue: I actually think it's not, because the thread that uses it runs this once every 5 minutes.  If it's "off" during one of those runs, then the thread will correct it's understanding very quickly.  And it's not strictly necessary that this variable is consistent w/ the announce protocol.  Worst case is that a node will send out an extra announce message when it's not needed.

consensus/istanbul/backend/announce.go
kevjue: Nice catch.

kevjue:
Closing this PR as it's merged into PR #511 and that PR will be merged into master.

PR #817, opened 2020-01-21
Get gas price suggestion from the correct block

### LINKED ISSUE #638: In RPC, if no gas price is specified, make sure that suitable gas price is selected
### Expected Behavior

RPC calls will work when gas price is not set (at least `eth_call`)

### Current Behavior

Had some problems with doing `eth_call` with old blocks, will investigate if can reproduce.

mrsmkl:
### Description

Get the gas price for the block specified in the RPC call so that it doesn't fail transaction pre-check.

### Tested

Only manual testing for now. Perhaps an e2e test could be added.

### Related issues

- Fixes #638


internal/ethapi/api.go
```diff
@@ -708,7 +708,8 @@ func (s *PublicBlockChainAPI) doCall(ctx context.Context, args CallArgs, blockNr
 	// which will always be in Gold. This allows the default price to be set for the proper currency.
 	// TODO(asa): Remove this once this is handled in the Provider.
 	if gasPrice.Sign() == 0 || gasPrice.Cmp(big.NewInt(0)) == 0 {
-		gasPrice, err = s.b.SuggestPriceInCurrency(ctx, args.FeeCurrency)
+		gasPrice, err = s.b.SuggestPriceInCurrency(ctx, args.FeeCurrency, header, state)
```
lucasege: Handle `err` here

PR #832, opened 2020-01-24
Proxies should ignore consensus messages from non validators

### LINKED ISSUE #651: Proxies should not forward all consensus messages to the validator
### Expected Behavior

Proxies filter consensus messages from non-validators

### Current Behavior

All messages forwarded

lucasege:
### Description

Proxies that receive a consensus message from a non validator now return an error. 

### Tested

Added unit testing to verify that messages not signed by a validator are ignored, whereas those that are signed are properly forwarded to the proxies' validator.

### Other changes

Added comments to present more context on proxy setup (will be removed with multiproxy).

### Related issues

- Fixes #651

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -133,6 +133,11 @@ func (sb *Backend) handleConsensusMsg(peer consensus.Peer, payload []byte) error
 			return nil
 		}
 
+		if _, err := sb.valEnodeTable.GetAddressFromNodeID(peer.Node().ID()); err != nil {
+			sb.logger.Trace("Got a consensus message from a non validator.  Ignoring it")
+			return nil
+		}
+
 		// Need to forward the message to the proxied validator
 		sb.logger.Trace("Forwarding consensus message to proxied validator")
 		if sb.proxiedPeer != nil {
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -133,7 +133,15 @@ func (sb *Backend) handleConsensusMsg(peer consensus.Peer, payload []byte) error
 			return nil
 		}
 
-		if _, err := sb.valEnodeTable.GetAddressFromNodeID(peer.Node().ID()); err != nil {
+		msg := new(istanbul.Message)
+
+		// Verify that this message is coming from a legitimate validator before forwarding.
+		checkValidatorSignature := func(data []byte, sig []byte) (common.Address, error) {
+			block := sb.currentBlock()
+			valSet := sb.getValidators(block.Number().Uint64(), block.Hash())
+			return istanbul.CheckValidatorSignature(valSet, data, sig)
+		}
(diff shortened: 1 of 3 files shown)


consensus/istanbul/backend/handler.go
```diff
@@ -133,6 +133,11 @@ func (sb *Backend) handleConsensusMsg(peer consensus.Peer, payload []byte) error
 			return nil
 		}
 
+		if _, err := sb.valEnodeTable.GetAddressFromNodeID(peer.Node().ID()); err != nil {
```
kevjue: This is a great first pass.

However, the valEnodeTable will contain entries for all registered and elected validators.  And the contents of this table is not the most suitable for the query "get current elected validators".  Specifically this table's purpose to do map a validator signing address to that validator's enode url (akin to a DNS table).

The best way to get the current elected validators is via the blockchain header's.  You may already know this, but at each epoch's last block, there will be an elected validator set diff (with the initial set stored in the genesis block).  So to get the current elected validators, you would need to retrieve those blocks, and calculate the current set.  This function does that:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/backend.go#L275

consensus/istanbul/backend/handler.go
```diff
@@ -133,6 +133,19 @@ func (sb *Backend) handleConsensusMsg(peer consensus.Peer, payload []byte) error
...
 		}
 
+		msg := new(istanbul.Message)
+
+		// Verify that this message is coming from a legitimate validator before forwarding.
+		checkValidatorSignature := func(data []byte, sig []byte) (common.Address, error) {
+			block := sb.currentBlock()
+			valSet := sb.getValidators(block.Number().Uint64(), block.Hash())
+			return istanbul.CheckValidatorSignature(valSet, data, sig)
+		}
```
lucasege: For reviewer: this is functionally a copy of https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/core/core.go#L640 , which is private.
Does it make sense to copy or is it safe to expose `checkValidatorSignature` on `core.go`?

kevjue: This looks good to me!

kevjue:
The code changes LGTM!  Please add a test case here:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/handler_test.go

consensus/istanbul/backend/handler.go
```diff
@@ -133,6 +133,19 @@ func (sb *Backend) handleConsensusMsg(peer consensus.Peer, payload []byte) error
...
+		if err := msg.FromPayload(payload, checkValidatorSignature); err != nil {
+			sb.logger.Trace("Got a consensus message from a non validator.  Ignoring it")
+			return nil
```
kevjue: Please have this return a non nil error.

PR #840, opened 2020-01-31
Prevent ethstats from blocking ChainHead publishing

nambrot:
### Description

This was a gnarly one.

**Problem**.

I have a testnet with 3 validators, validator0 is proxied. Due to 1 second block times, when validator0 is just behind, it wouldn't be able to participate in consensus unless it imports blocks that brings it back to the most recent sequence. Eventually however, it would stall and be stuck at a block. Sometimes it would recover only to stop again.

**Investigation**:
I first tried to figure out why I would no longer import blocks. Turns out that I eventually hit [this line](https://github.com/celo-org/celo-blockchain/blob/master/eth/fetcher/fetcher.go#L599) that prevents DOS from our peers.

It turns out that something prevented the completion of an insert which let the queue accumulate. After a lot of log lines, it turns out that sending [`ChainHeadEvents`](https://github.com/celo-org/celo-blockchain/blob/master/core/blockchain.go#L1642) was blocking. This can happen if one of the subscriptions channels is not accepting any messages.

After digging deeper, I realized that the [ethstats subscription](https://github.com/celo-org/geth/blob/master/ethstats/ethstats.go#L218) is the culprit. The channel has a capacity of 10, but it wasn't getting pulled from.

`IsProxiedValidator` actually checks whether we currently have a proxy peer. If that is not the case, we end up stuck in this [else branch](https://github.com/celo-org/geth/blob/master/ethstats/ethstats.go#L297) which itself contains a for loop out of which we never get out until there is a reporting failure.

Because we do not land in the other branch, we will never consume from `signCh`. Unluckily, that is exactly where the `HandleLoop` is trying to [publish to](https://github.com/celo-org/celo-blockchain/blob/master/ethstats/ethstats.go#L280).

By being blocked there, we no longer have a goroutine consuming from the ChainHeadEvent channel, which causes the sending of the ChainHeadEvent to block, which causes all subsequent imports of blocks to eventually stop (after 64 of them), which causes the istanbul engine to no longer be able to get to the most recent sequence.

**Solution:**
The ethstats code in general isn't quite pretty and since I'm focused on the 1.9 merge, I just made the publishing to the `signCh` non blocking, since stats paylods should be easily ignored. I also moved the defer to close the subscription into the goroutine that is actually listening in case we are breaking out of the `HandleLoop` for any reason.

### Tested

- [x] Tested the patch on `namgeth19repro`

### Related issues

- Relates to #808

It exhibits somewhat similar symptoms as #808, however not quite exactly. We do suspect that the blocking of subscriptions to be at least part of the cause.

### Backwards compatibility

- This is backwards compatible

### DIFF
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -216,7 +216,6 @@ func (s *Service) loop() {
 
 	chainHeadCh := make(chan core.ChainHeadEvent, chainHeadChanSize)
 	headSub := blockchain.SubscribeChainHeadEvent(chainHeadCh)
-	defer headSub.Unsubscribe()
 
 	txEventCh := make(chan core.NewTxsEvent, txChanSize)
 	txSub := txpool.SubscribeNewTxsEvent(txEventCh)
@@ -236,6 +235,7 @@ func (s *Service) loop() {
 	)
 	go func() {
 		var lastTx mclock.AbsTime
+		defer headSub.Unsubscribe()
 
 	HandleLoop:
 		for {
@@ -277,7 +277,11 @@ func (s *Service) loop() {
 					// proxied validator should sign
 					channel = signCh
 				}
-				channel <- &statsPayload
+				select {
+				case channel <- &statsPayload:
+				default:
+				}
+
 			}
 		}
 		close(quitCh)
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -214,9 +214,6 @@ func (s *Service) loop() {
 		txpool = s.les.TxPool()
 	}
 
-	chainHeadCh := make(chan core.ChainHeadEvent, chainHeadChanSize)
-	headSub := blockchain.SubscribeChainHeadEvent(chainHeadCh)
-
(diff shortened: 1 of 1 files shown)


ethstats/ethstats.go
```diff
@@ -236,6 +235,7 @@ func (s *Service) loop() {
 	)
 	go func() {
 		var lastTx mclock.AbsTime
+		defer headSub.Unsubscribe()
```
mcortesi: why not move the subscribe also?

ethstats/ethstats.go
```diff
@@ -277,7 +277,11 @@ func (s *Service) loop() {
 					// proxied validator should sign
 					channel = signCh
 				}
-				channel <- &statsPayload
+				select {
```
mcortesi: add a TODO here. to signal this is a patch, a and we should do a proper clean up of this code

nambrot:
Created the issue at https://github.com/celo-org/celo-blockchain/issues/841

PR #844, opened 2020-02-04
Use contract historical randomness

nambrot:
### Description

Currently, we are trying to fetch the randomness of the last epoch block to determine the validator set order (to know who is supposed to propose). However, we only store the last 128 tries by default, which causes us not to be able to retrieve the state of the epoch block, and thus not be able to fetch the randomness.

This PR suggests one solution to this problem. It actually uses the state of the current block and takes advantage of that fact that we are storing historical randomness on the contract itself. The window on baklava is 720 blocks right now, and the epoch size is 720. To avoid off-by-one error, we'll probably need to run a governance proposal to set it to something higher.

The alternate solution to this would be to increase the `TriesInMemory` default to something larger than the epoch size.

### Tested

- Deployed it to `nammissingtrie`, so the error pop up only until historical randomness on the contract was no longer available with the current block (since epoch lengths are 1000 by default on testnets)

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -648,15 +648,12 @@ func (sb *Backend) validatorRandomnessAtBlockNumber(number uint64, hash common.H
 	if number > 0 {
 		lastBlockInPreviousEpoch = number - istanbul.GetNumberWithinEpoch(number, sb.config.Epoch)
 	}
-	header := sb.chain.GetHeaderByNumber(lastBlockInPreviousEpoch)
-	if header == nil {
-		return common.Hash{}, errNoBlockHeader
-	}
+	header := sb.chain.CurrentHeader()
 	state, err := sb.stateAt(header.Hash())
 	if err != nil {
 		return common.Hash{}, err
 	}
-	return random.Random(header, state)
+	return random.HistoricalRandom(header, state, lastBlockInPreviousEpoch)
 }
 
 func (sb *Backend) getOrderedValidators(number uint64, hash common.Hash) istanbul.ValidatorSet {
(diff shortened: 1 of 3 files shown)


contract_comm/random/random.go
```diff
@@ -199,3 +223,9 @@ func Random(header *types.Header, state vm.StateDB) (common.Hash, error) {
 	_, err := contract_comm.MakeStaticCall(params.RandomRegistryId, randomFuncABI, "random", []interface{}{}, &randomness, params.MaxGasForComputeCommitment, header, state)
 	return randomness, err
 }
+
+func HistoricalRandom(header *types.Header, state vm.StateDB, blockNumber uint64) (common.Hash, error) {
```
asaj: nit: "BlockRandomness"?

contract_comm/random/random.go
```diff
@@ -199,3 +223,9 @@ func Random(header *types.Header, state vm.StateDB) (common.Hash, error) {
...
+	randomness := common.Hash{}
+	_, err := contract_comm.MakeStaticCall(params.RandomRegistryId, historicRandomFuncABI, "getBlockRandomness", []interface{}{big.NewInt(int64(blockNumber))}, &randomness, params.MaxGasForComputeCommitment, header, state)
```
asaj: Is this the right amount of gas to provide?

consensus/istanbul/backend/backend.go
```diff
@@ -648,15 +648,12 @@ func (sb *Backend) validatorRandomnessAtBlockNumber(number uint64, hash common.H
 	if number > 0 {
 		lastBlockInPreviousEpoch = number - istanbul.GetNumberWithinEpoch(number, sb.config.Epoch)
 	}
-	header := sb.chain.GetHeaderByNumber(lastBlockInPreviousEpoch)
-	if header == nil {
```
asaj: Why remove this?

nambrot: I assume `sb.chain.CurrentHeader` can never be nil, but you are right, I might as well leave the check in there

contract_comm/random/random.go
nambrot: I added another constant called `MaxGasForBlockRandomness`, mostly just copied it from the existing implementation

contract_comm/random/random.go
```diff
@@ -101,13 +102,36 @@ const (
       "type": "function"
     }
 ]`
+
+	historicRandomAbi = `[
```
asaj: `blockRandomnessAbi`?

PR #845, opened 2020-02-04
Higher trie limit

nambrot:
### Description

I'm seeing a weird off-by-one error in https://github.com/celo-org/celo-blockchain/pull/844, so to move on, this PR just increases the TrieInMemory Limit to exceed the epoch size for now. This is meant as a temporary solution and should be fixed with https://github.com/celo-org/celo-monorepo/issues/2625

### Tested

- [x] on https://nammissingtrie-celostats.celo-networks-dev.org/


asaj:
Please create an issue to revert this so we don't forget

nambrot:
Created https://github.com/celo-org/celo-blockchain/issues/854

PR #856, opened 2020-02-06
Validate Preprepares with checkMessage 

timmoreton:
### Description

* Fix `checkMessage` to match the desired state machine transitions. Now in `handlePreprepare` we can call it before we check the proposer, meaning we'll no longer see the erroneous message `Ignore preprepare message from non-proposer` (actually the proposal is for the subsequent sequence, and from the valid proposer)
* Fix IBFT tests. Many tests were polluting the results of others by not shutting down nodes and not clearing backlogs, and causing flakiness.
* Improve logging to help track down conditions where consensus and block signing stops temporarily.

### Tested

Integration and e2e tests.

### Backwards compatibility

Yes.


package.json
```diff
@@ -17,5 +17,8 @@
   ],
   "scripts": {
     "postinstall": "rm -rf build/bin/Geth.framework && tar -xvf build/bin/Geth.framework.tgz -C build/bin && touch Empty.m && ln -sf build/bin/Geth.framework/Versions/A/Geth libGeth.a"
+  },
+  "devDependencies": {
+    "husky": "^4.2.1"
```
asaj: Why does this need to be added?

PR #864, opened 2020-02-13
Increase MaxPeers to be greater than light client MaxPeers

tkporter:
### Description

Folks on Discord were running into an issue when starting the node where the node would refuse to start with the default MaxPeers (50) being less than the default light client MaxPeers (99).

### Tested

Ran geth with default values & node started up

### Other changes

n/a

### Related issues

n/a

### Backwards compatibility

Yes :)


H34D:
https://github.com/celo-org/celo-blockchain/pull/511#discussion_r376269715

Issue #848, opened 2020-02-04
Improve celo-blockchain tooling (linting, gosec, CI)


mcortesi:
Get1.9 uses [golangci-lint](https://github.com/golangci/golangci-lint) which provides several linting tools we can use to improve our code.

Things to do:

 - [x] Investigate linters available and activate them
 - [x] Document/Share with team how to activate linter with vscode (& other editors)
 - [x] Fix linting issues
 - [x] Investigate security related linters
 - [x] Check Trail of Bits audit doc for suggested linters

PR #869, opened 2020-02-18
Add gosec as a linter

mcortesi:
### Description

Adds gosec as a linter. For now we ignore all current issues seems they don't seem important for our current usage.

Still is useful to add it, so we can't check future code

### Tested

* CI

### Related issues

- Related to #848 

### Backwards compatibility

Yes

### DIFF
--- a/cmd/geth/retesteth.go
+++ b/cmd/geth/retesteth.go
@@ -505,7 +505,6 @@ func (api *RetestethAPI) mineBlock() error {
 	txCount := 0
 	var txs []*types.Transaction
 	var receipts []*types.Receipt
-	var coalescedLogs []*types.Log
 	var blockFull = gasPool.Gas() < params.TxGas
 	for address := range api.txSenders {
 		if blockFull {
@@ -533,7 +532,6 @@ func (api *RetestethAPI) mineBlock() error {
 				}
 				txs = append(txs, tx)
 				receipts = append(receipts, receipt)
-				coalescedLogs = append(coalescedLogs, receipt.Logs...)
 				delete(m, nonce)
 				if len(m) == 0 {
 					// Last tx for the sender
@@ -694,7 +692,7 @@ func (api *RetestethAPI) AccountRange(ctx context.Context,
 		if preimage := accountTrie.GetKey(it.Key); preimage != nil {
 			result.AddressMap[common.BytesToHash(it.Key)] = common.BytesToAddress(preimage)
 			//fmt.Printf("%x: %x\n", it.Key, preimage)
-		} else {
+			// } else {
 			//fmt.Printf("could not find preimage for %x\n", it.Key)
 		}
 	}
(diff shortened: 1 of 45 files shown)


cmd/geth/retesteth.go
```diff
@@ -900,7 +898,8 @@ func retesteth(ctx *cli.Context) error {
 		log.Info("HTTP endpoint closed", "url", httpEndpoint)
 	}()
 
-	abortChan := make(chan os.Signal)
+	// the channel used with signal.Notify should be buffered
+	abortChan := make(chan os.Signal, 5)
```
nategraf: This change could be non-trivial if blocking is the _expected_ result of a send or recv. Are you sure this is safe?

internal/build/download.go
```diff
@@ -82,6 +82,7 @@ func (db *ChecksumDB) DownloadFile(url, dstPath string) error {
 	fmt.Printf("%s is stale\n", dstPath)
 	fmt.Printf("downloading from %s\n", url)
 
+	// #nosec (internal tool nothing to worry about)
```
nategraf: What is the warning? As a reason, "internal tool" is tricky because if it exposes the dev machines, that could be very bad.

p2p/discv5/ticket.go
```diff
@@ -697,6 +697,7 @@ func globalRandRead(b []byte) {
 	val := 0
 	for n := 0; n < len(b); n++ {
 		if pos == 0 {
+			// #nosec TODO?
```
nategraf: Could you file an issue with the warning. Otherwise, this will be immediately lost.

cmd/geth/retesteth.go
mcortesi: I did this because of https://staticcheck.io/docs/checks#SA1017
which seems reasonable

wdyt?

internal/build/download.go
mcortesi: The warning was about doing `http.Get(url)` where `url` is not sanitized and comes from the user.

The rule is: **G107: Url provided to HTTP request as taint input** ([rules here](https://github.com/securego/gosec#available-rules))

p2p/discv5/ticket.go
mcortesi: The rule is: **G404: Insecure random number source (rand)** ([rules here](https://github.com/securego/gosec#available-rules))

mcortesi: Done #877 !

cmd/geth/retesteth.go
nategraf: 👍

internal/build/download.go
nategraf: Ok, that makes sense 👌 . A comment that is more specific would be helpful.
```suggestion
	// #nosec (Ignoring sanitization warning. Intended to be an arbitrary url.)
```

internal/build/util.go
```diff
@@ -129,6 +129,7 @@ func render(tpl *template.Template, outputFile string, outputPerm os.FileMode, x
 // because runtime.Version checks on the host should match the tools that are run.
 func GoTool(tool string, args ...string) *exec.Cmd {
 	args = append([]string{tool}, args...)
+	// #nosec (Internal tool. nothing to worry about)
```
nategraf:
```suggestion
	// #nosec (Ignoring sanitization warning. Build tool is intended to take arbitrary arguments.)
```

PR #873, opened 2020-02-19
Validator proof in handshake to bypass max peer checks

tkporter:
### Description

This PR adds a handshake to the istanbul package. In this handshake, either of the two newly peered nodes can "prove" to the other that they are a validator. This is done by sending a signed message containing the enodeURL of the node claiming to be the validator.

### Implementation

A new struct is introduced, `versionedEnode`, which contains an unencrypted enodeURL and its version. This is sent in an `istanbul.Message` during the handshake if the node chooses to prove itself as a validator.

```
type versionedEnode struct {
	Node    string
	Version uint
}
```

During the istanbul handshake, a node will only reveal its identity to the new peer if:
1. The self node is a registered or active validator
2. The peer's enodeURL is in the self node's validator enode table

This is to prevent a node from identifying itself as a validator node to non-validators. If a node chooses not to identify itself, an empty and unsigned `istanbul.Message` will be sent to the peer so that the handshake is still completed. This handshake message is called an `istanbulValidatorProofMsg` to avoid confusing a validator proof message (which can be empty) from being just a `versionedEnode` message.

To solve the issue of a proxy sending a validator proof message to new peers on behalf of the validator, the proxied validator will send the proxy an `istanbulVersionedEnodeMsg`, a signed `istanbul.Message` with a `versionedEnode` containing the publicly facing enodeURL of the proxy. The validator sends a versionedEnode when first connected to its proxy and along with any subsequently generated announce messages to ensure the version of the proxy versionedEnode is the same as the gossiped announce.

There are two major changes to the existing p2p implementation:

1. Max peer checks are left to the protocols, rather than occurring at the p2p level. There used to be a check in the p2p package as well as the `eth`/`les` packages, but now the p2p check is removed
2. Max inbound connection checks are similarly left to the protocols. The inbound connection count is still maintained at the p2p level, but the enforcement happens in `eth`/`les`

For context, the full setup process with a new peer is:

1. (p2p layer) Do encryption handshake
2. (p2p layer) Do protocol handshake to learn which protocols the nodes have in common
3. (p2p layer) Run protocol with peer
4. (eth protocol layer) ethereum handshake, exchange block number etc
5. (eth/istanbul layer) new istanbul handshake
6. (eth layer) max peer check

### Tested

e2e tests, local tests.

e2e test PR: https://github.com/celo-org/celo-monorepo/pull/2827

### Other changes

I added the node ID to be shown for each entry in `valEnodeTableInfo` RPC

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/2667

### Backwards compatibility

NOT backwards compatible-- nodes will expect the handshake to occur


consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,171 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
 		}
 	}
 }
+
+// Handshake allows this node to identify itself to the peer as a validator and vice versa
+func (sb *Backend) Handshake(peer consensus.Peer) (bool, error) {
+	errCh := make(chan error, 4)
```
asaj: Nice 👍

consensus/istanbul/backend/handler.go
```diff
@@ -48,10 +52,14 @@ const (
 	istanbulGetAnnouncesMsg        = 0x16
 	istanbulGetAnnounceVersionsMsg = 0x17
 	istanbulAnnounceVersionsMsg    = 0x18
+	istanbulDirectAnnounceMsg      = 0x19
+	istanbulValidatorProofMsg      = 0x1a
```
asaj: Could be "SignedVersionedEnode"

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,171 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	return <-isValidatorCh, nil
+}
+
+// generateValidatorProofMessage will create a directAnnounce message that serves
+// as a proof to the peer during the handshake that this node is a validator.
+// If this node is not a validator or this node does not believe the peer
+// is a validator, no proof is generated to hide the address of this node.
+// If this node is a proxy, it will use the most recent directAnnounce this
+// node has received from the proxied validator.
+func (sb *Backend) generateValidatorProofMessage(peer consensus.Peer) (*istanbul.Message, error) {
```
asaj: nit: I think the naming is a bit confusing here, as `generateValidatorProofMessage` returns the return value of `generateDirectAnnounce`. Consider changing the naming to be more consistent

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,171 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	regAndActiveVals, err := sb.retrieveRegisteredAndElectedValidators()
+	if err != nil {
+		return false, err
+	}
+	if !regAndActiveVals[validatorAddress] {
+		return false, nil
+	}
+	// If the peer is not a known validator, we do not want to give up extra
+	// information about ourselves
+	if _, err := sb.valEnodeTable.GetAddressFromNodeID(peer.Node().ID()); err != nil {
```
asaj: What if we don't have an entry in the val enode table for this peer?

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,171 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// readValidatorProofMessage reads a validator proof message as a part of the
+// istanbul handshake. Returns if the peer is a validator or if an error occurred.
+func (sb *Backend) readValidatorProofMessage(peer consensus.Peer) (bool, error) {
+	peerMsg, err := peer.ReadMsg()
```
asaj: Is the peer guaranteed to send a validator proof message here? Could we accidentally read a different message?

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,171 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	}
+
+	// Check if the peer is within the registered/elected valset
+	regAndActiveVals, err := sb.retrieveRegisteredAndElectedValidators()
+	if err != nil {
+		sb.logger.Trace("Error in retrieving registered/elected valset", "err", err)
+		return false, err
+	}
+	if !regAndActiveVals[msg.Address] {
+		return false, nil
```
asaj: nit: Should we return an error here? Or log something? IIUC, with everyone running the proper code, we should never, or almost never, reach this line

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,171 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	}
+
+	// If the directAnnounce message is too old, we don't count this proof as valid
+	// An error is given if the entry doesn't exist, so we ignore the error.
+	knownVersion, err := sb.valEnodeTable.GetVersionFromAddress(msg.Address)
```
asaj: Do we need to worry about re-used direct announce messages? e.g. if you send a direct announce to me, could I use that to impersonate you when trying to connect to another node?

les/client_handler.go
```diff
@@ -99,8 +99,12 @@ func (h *clientHandler) runPeer(version uint, p *p2p.Peer, rw p2p.MsgReadWriter)
 }
 
 func (h *clientHandler) handle(p *peer) error {
-	if h.backend.peers.Len() >= h.backend.config.LightPeers && !p.Peer.Info().Network.Trusted {
-		return p2p.DiscTooManyPeers
+	if !p.Peer.Info().Network.Trusted {
+		if h.backend.peers.Len() >= h.backend.config.LightPeers {
+			return p2p.DiscTooManyPeers
+		} else if p.Peer.Server.InboundCount() >= p.Peer.Server.MaxInboundConns() {
+			return p2p.DiscTooManyInboundPeers
```
asaj: Why is this change needed?

consensus/istanbul/backend/announce.go
```diff
@@ -700,3 +708,153 @@ func (sb *Backend) checkPeersAnnounceVersions() {
 		}
 	}
 }
+
+type directAnnounce struct {
```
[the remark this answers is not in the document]
tkporter: ended up renaming it to this

tkporter:
@asaj I ended up changing the `directAnnounce` struct to `versionedEnode` and the `istanbulDirectAnnounceMsg` to be `istanbulVersionedEnodeMsg`.

I've still kept the `istanbulValidatorProofMsg` to emphasize that the proof message isn't strictly the same thing as a versioned enode message because it can be empty, and I added a couple comments to hopefully make it clearer.

I also addressed the other comments so I think it's ready for another pass

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,177 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+// and an empty message is created.
+// If this node is a proxy, it will use the most recent versioned enode message this
+// node has received from the proxied validator.
+func (sb *Backend) generateValidatorProofMessage(peer consensus.Peer) (*istanbul.Message, error) {
+	shouldSend, err := sb.shouldSendValidatorProof(peer)
+	if err != nil {
+		return nil, err
+	}
+	if shouldSend {
+		msg, err := sb.getSelfVersionedEnodeMsg(getCurrentAnnounceVersion())
```
nategraf: 🤔

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,177 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+		if err != nil {
+			return nil, err
+		}
+		if msg != nil {
+			return msg, nil
+		}
+	}
+	// Even if we decide not to identify ourselves,
+	// send an empty message to complete the handshake
+	return &istanbul.Message{}, nil
```
nategraf: Why can't we skip that part of the handshake? Based on my reading of the code above, this formulation adds at least one full RTT to the connection process (or more, see blow), even between non-validating fullnodes (but maybe I missed something). Is it possible to avoid this?

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,177 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+// validator proof revealing its address to a peer
+func (sb *Backend) shouldSendValidatorProof(peer consensus.Peer) (bool, error) {
+	var validatorAddress common.Address
+	if sb.config.Proxy {
+		validatorAddress = sb.config.ProxiedValidatorAddress
+	} else {
+		validatorAddress = sb.Address()
+	}
+	// Check to see if this node is a validator
+	regAndActiveVals, err := sb.retrieveRegisteredAndElectedValidators()
```
nategraf: ⚠️ This function can potentially block for multiple seconds. If that happens on the path of a handshake, the connection could timeout. ⚠️

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,177 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	}
+
+	// Ensure the node in the versionedEnode matches the peer node
+	if node.ID() != peer.Node().ID() {
+		logger.Warn("Peer provided incorrect node in versionedEnode", "versionedEnode node", versionedEnode.Node, "peer node", peer.Node().URLv4())
+		return false, errors.New("Incorrect node in versionedEnode")
+	}
+
+	// Check if the peer is within the registered/elected valset
+	regAndActiveVals, err := sb.retrieveRegisteredAndElectedValidators()
```
nategraf: Same concern as above

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,177 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+// verifyValidatorProofMessage allows messages that are not signed to still be
+// decoded in case the peer has decided not to identify itself with the validator proof
+// message
+func (sb *Backend) verifyValidatorProofMessage(data []byte, sig []byte) (common.Address, error) {
+	// If the message was not signed, allow it to still be decoded.
+	// A later check will verify if the address is in the validator set, which
+	// will fail.
+	if len(sig) == 0 {
+		return common.ZeroAddress, nil
+	}
```
nategraf: Why not just return an error and fail here? The output of this function could be misinterpreted as signed message from address 0

consensus/istanbul/backend/announce.go
```diff
@@ -700,3 +719,180 @@ func (sb *Backend) checkPeersAnnounceVersions() {
...
+}
+
+// getSelfVersionedEnodeMsg gets the most recent self versioned enode message to send
+// and generates a new one with `version` if one does not exist.
+// New versioned enode messages are not always generated to ensure the version
+// of a versioned enode message is not greater than a recently gossiped announce
+// version.
+// May be nil if this is a proxy that does not have a versioned enode message
+// from its proxy.
+func (sb *Backend) getSelfVersionedEnodeMsg(version uint) (*istanbul.Message, error) {
```
nategraf: I find this interface somewhat misleading because it is titled as a `get` function, but it can end up modifying the Backend state. It could be considered a cache, but it is not the only place this is set, and the result of the calculation is dependent on time. Namely if this node have not yet gossiped out any announce messages, this function will break the invariant that is stated above that the version should not be newer than the gossiped announce. I also thought when first read the function that the `version` arg is used in retrieval (i.e. that the returned message version would be the same as the version arg) but that is not the case.

consensus/istanbul/backend/handler.go
```diff
@@ -48,10 +52,14 @@ const (
 	istanbulGetAnnouncesMsg        = 0x16
 	istanbulGetAnnounceVersionsMsg = 0x17
 	istanbulAnnounceVersionsMsg    = 0x18
+	istanbulVersionedEnodeMsg      = 0x19
+	istanbulValidatorProofMsg      = 0x1a
```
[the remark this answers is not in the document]
tkporter: I also shared this concern, as it does feel weird. The eth handshake message is msg code `0x00` so I thought maybe of putting this as the first in the istanbul package and excluding it from the istanbul message range. What do you think of that? I was hesitant because it will require reassigning other msg codes but if we aren't worried about backward compatibility it should work

consensus/istanbul/backend/handler.go
tkporter: The issue is that the eth handler that calls this handshake will wait for the handshake to be completed before continuing on with the rest of the setup (like registering the peer with `eth` and `istanbul` as ready for use). If I'm a node and my new peer doesn't send any message at all then I will end up having to wait the 5 second timeout to wait for a response. The 5 second wait is undesirable & currently the peer gets dropped if the handshake fails with an error (which timing out will result in)

consensus/istanbul/backend/handler.go
tkporter: I don't think there's a straightforward way around this.. I'm going to up the handshake timeout from 5 to 10 seconds, but if you have any ideas I'd love to hear them.

consensus/istanbul/backend/handler.go
tkporter: Returning an error and failing will cause the whole handshake to fail, which causes the peer to be dropped.

I'm going to change https://github.com/celo-org/celo-blockchain/pull/873/files#diff-991be8cbcd11f71d672bc2301b763c4cR415 to check if the signature is empty instead of the `Msg` and make it clearer in the comments in the function

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,177 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+
+// generateValidatorProofMessage will create a message that contains a proof
+// to the peer during the handshake that this node is a validator, which is just
+// a versioned enode message.
+// If this node is not a validator or this node does not believe the peer
+// is a validator, no proof is generated to hide the address of this node
+// and an empty message is created.
+// If this node is a proxy, it will use the most recent versioned enode message this
+// node has received from the proxied validator.
+func (sb *Backend) generateValidatorProofMessage(peer consensus.Peer) (*istanbul.Message, error) {
```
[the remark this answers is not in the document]
tkporter: I'm hesitant to bring out the `shouldSendValidatorProof` call because then it'd either involve an awkward boolean `shouldSendProof` or it would mean removing this function and instead putting the logic in the call site.

The content of the message is really dependent on the peer and I wanted to separate the logic for deciding what content to send in its own function, following the pattern of the other `generate` functions in `announce.go`. Is there a way I could beef up the comments to make this more clear?

consensus/istanbul/backend/announce.go
tkporter: Yeah, I agree this is pretty confusing. For the time being I renamed `getSelfVersionedEnodeMsg` -> `retrieveSelfVersionedEnodeMsg` to make it feel like less of a getter, and renamed `version` to `fallbackVersion`. I also clarified the invariant you mentioned
I couldn't think of a good way to refactor, but I could spend more time if you don't think those small changes are enough

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,181 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
 		}
 	}
 }
+
+// Handshake allows this node to identify itself to the peer as a validator and vice versa
+func (sb *Backend) Handshake(peer consensus.Peer) (bool, error) {
```
kevjue: IIUC, both peers will be sending the validatorproofmessage.  Is that necessary.  Can only the peer initiating the connection send it?

tkporter: Great point, you're totally right. I'll fix this

tkporter: https://github.com/celo-org/celo-blockchain/pull/873/commits/85644cb256f61dc4558dd503e4f0193497cc359a

p2p/server.go
```diff
@@ -984,12 +1002,10 @@ running:
 	}
 }
 
-func (srv *Server) postHandshakeChecks(peers map[enode.ID]*Peer, inboundCount int, c *conn) error {
+// postHandshakeChecks performs basic checks on a new peer after a handshake.
+// Max peer checks are left to the protocols.
+func (srv *Server) postHandshakeChecks(peers map[enode.ID]*Peer, c *conn) error {
 	switch {
-	case !c.is(trustedConn|staticDialedConn) && len(peers) >= srv.MaxPeers:
```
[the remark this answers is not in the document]
tkporter: I looked more at the code and it became more apparent-- the p2p code is common between `les` and `eth`, so it performs a check early on in the handshake process to make sure that the total peer count (normal peers + light peers) isn't exceeded. If that check passes, then `eth`/`les` perform their own check to ensure that the # of peers allowed for those protocols isn't at its cap. The eth cap is `maxPeers - lightMaxPeers` and the les cap is just `lightMaxPeers`. It's possible for the p2p check to pass but for one of the protocol checks to fail (if for example the # of light peers a node is serving is at its capacity, but the number of normal peers isn't it its cap)

`eth` check: https://github.com/celo-org/celo-blockchain/blob/9ee355daadc8f2e1712f6d64e6c9983f225aef38/eth/backend.go#L641

`les` check: https://github.com/celo-org/celo-blockchain/blob/9ee355daadc8f2e1712f6d64e6c9983f225aef38/les/client_handler.go#L102

consensus/istanbul/backend/handler.go
tkporter: Fixed with: https://github.com/celo-org/celo-blockchain/compare/c3d137bd8cff0f965c41b1b4d4db738645149eec...806841c246c5127997c248fe9d8410ec0b1318b7?expand=1

Now the initiating peer only sends a message and the other peer just listens

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,191 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
 		}
 	}
 }
+
+// Handshake allows the initiating peer to identify itself as a validator
+func (sb *Backend) Handshake(peer consensus.Peer, peerIsInbound bool) (bool, error) {
```
kevjue: What do you think of using `peer.Inbound()` instead adding a new parameter?  E.g. add that `Inbound` function to the Peer interface.

kevjue: I'm wondering if this handshake can be done only if it's an attempted validator connection.

consensus/istanbul/backend/handler.go
```diff
@@ -290,3 +298,191 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+		}
+	}
+	// Even if we decide not to identify ourselves,
+	// send an empty message to complete the handshake
+	return &istanbul.Message{}, false, nil
+}
+
+// shouldSendValidatorProof determines if this node should send a
+// validator proof revealing its address to a peer
+func (sb *Backend) shouldSendValidatorProof(peer consensus.Peer) (bool, error) {
```
kevjue: If you can figure out a way to do this:  https://github.com/celo-org/celo-blockchain/pull/873/files#r385459074, then this check could become unnecessary.

consensus/istanbul/backend/announce.go
```diff
@@ -700,3 +719,175 @@ func (sb *Backend) checkPeersAnnounceVersions() {
...
+
+	payload, err := versionedEnodeMsg.Payload()
+	if err != nil {
+		logger.Error("Error getting payload of versioned enode message", "err", err)
+		return err
+	}
+	return peer.Send(istanbulVersionedEnodeMsg, payload)
+}
+
+func getCurrentAnnounceVersion() uint {
```
kevjue: Should the announce version always be updated here?

consensus/istanbul/backend/announce.go
```diff
@@ -349,6 +349,26 @@ func (sb *Backend) generateAndGossipAnnounce() error {
 		return nil
 	}
 
+	// Generate a new versioned enode message with the updated version
+	versionedEnodeMsg, err := sb.generateVersionedEnodeMsg(announceVersion)
+	if err != nil {
+		return err
+	}
+	if versionedEnodeMsg != nil {
+		sb.selfVersionedEnodeMsgMu.Lock()
```
codyborn: Would it be useful to define these three lines as a helper function?

consensus/istanbul/backend/handler.go
tkporter: If I'm a validator and the new peer initiated the connection & wants to prove itself as a validator, I have no idea if this new peer is a full node or a validator, so I'd have to listen for a message from them to decide what kind of node they are. If the peer never sent a message the handshake would timeout (which is right now considered an error)

I don't think there's a straightforward way around this, let me know if you think otherwise

tkporter: https://github.com/celo-org/celo-blockchain/pull/873/commits/6dff0ee185d9c5408dce55537c18894913e71d05

consensus/istanbul/backend/announce.go
tkporter: Good idea https://github.com/celo-org/celo-blockchain/pull/873/commits/97d7ef05988a92748c57bb556e6b723b7d891b40

consensus/istanbul/backend/announce.go
tkporter: I just refactored a little to remove the notion of a "fallback version", so now every call to `getCurrentAnnounceVersion` is when the announce version is being updated. I'm a little confused what you mean here

consensus/istanbul/backend/announce.go
tkporter: Just refactored in a better way and removed the need for the fallbackVersion

consensus/istanbul/backend/handler.go
kevjue: Can non validators send a "no-op" message?

consensus/istanbul/backend/handler.go
kevjue: Also, keep in mind that the proof should be sent only when a validator connection (which is created in the istanbul package and you should be able to figure out that it's a validator connection) is initiated.

I also echo @nategraf 's concern in the comment below, so I think it would be a big value-add if that can be avoided all together.

consensus/istanbul/backend/announce.go
```diff
@@ -700,3 +716,169 @@ func (sb *Backend) checkPeersAnnounceVersions() {
...
+func (sb *Backend) retrieveSelfVersionedEnodeMsg() (*istanbul.Message, error) {
+	sb.selfVersionedEnodeMsgMu.Lock()
+	defer sb.selfVersionedEnodeMsgMu.Unlock()
+	if sb.selfVersionedEnodeMsg == nil {
+		// Proxies cannot generate a versioned enode message because are not
+		// able to sign on behalf of the proxied validator.
+		if sb.config.Proxy {
+			return nil, nil
+		}
+		versionedEnodeMsg, err := sb.generateVersionedEnodeMsg(getCurrentAnnounceVersion())
```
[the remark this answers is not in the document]
tkporter: When the node first starts up there's at least a 1 minute period before the first call to `generateAndGossipAnnounce` (that's how I'm reading the code, correct me if I'm wrong)
I think that first minute is likely to be a time period where this handshake would be very useful (ie new validator connections from a new node)
I tried adding this to `announceThread` when it's first called, but by that time the istanbul signing function is not set yet.

consensus/istanbul/backend/handler.go
tkporter: That's what I consider the empty `istanbul.Message` that's sent to be

consensus/istanbul/backend/internal/enodes/val_enode_db.go
```diff
@@ -437,19 +437,25 @@ func (vet *ValidatorEnodeDB) iterateOverAddressEntries(onEntry func(common.Addre
 	return iter.Error()
 }
 
+// ValEnodeEntryInfo contains information for an entry of the val enode table
 type ValEnodeEntryInfo struct {
-	Enode   string `json:"enode"`
-	Version uint   `json:"version"`
+	Enode   string   `json:"enode"`
+	Version uint     `json:"version"`
+	ID      enode.ID `json:"id"`
```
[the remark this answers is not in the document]
tkporter: Just made debugging easier for me when comparing IDs of nodes. I'm not passionate about keeping it

consensus/istanbul/backend/handler.go
tkporter: Do you mean that "VersionedEnodeCertificate" should or shouldn't ever be unsigned?

What do you think of:

`type versionedEnode struct` -> `type versionedEnodeCertificate struct`
`istanbulVersionedEnodeMsg` -> `istanbulVersionedEnodeCertificateMsg`
`istanbulValidatorProofMsg ` -> `istanbulValidatorHandshakeMsg`

consensus/istanbul/backend/announce.go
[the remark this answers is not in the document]
kevjue: My main concern here is that the enode certificate and version number should only be generated when the enodeURL assignment is changed.

As it's currently implemented, those would be updated and generated on every call to `retrieveSelfVersionedEnodeMsg`.  That results in two issues:

1) The certificate will be generated every time `registeredpeer` is called, so there is going to be alot of redundant work done by the done.
2) More importantly, the version number will be updated every time `registeredpeer` is called, and hence there will be an expensive gossiped message every time that function is called.  That will also potentially make all the other validators gossip queryMessage (another *very* expensive networking call).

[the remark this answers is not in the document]
tkporter: `retrieveVersionedEnodeMsg` will only generate the enode certificate & version number if one hasn't been previously generated, which can only happen once upon the first call to `retrieveVersionedEnodeMsg`. Otherwise the previously generated enode certificate & version number is returned

So the only way the enode certificate/version is updated is if:

1. `generateAndGossipAnnounce` generates a new announce message
2. The node is a validator and the enode certificate hasn't been previously generated by `generateAndGossipAnnounce` and the certificate is needed for the handshake or in `RegisterPeer` if the peer is a proxy
3. The proxy receives an enode certificate from the proxied validator as a result of a proxied validator's newly generated announce or the proxied validator's `RegisterPeer`

consensus/istanbul/backend/handler.go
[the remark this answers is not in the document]
nategraf: Looking at this a bit more, I think this makes more sense to me and seems like the best solution.

consensus/istanbul/backend/handler.go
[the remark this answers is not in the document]
nategraf: Looking at the `generateValidatorProofMessage` again after your latest refactor, the logic seems to be slimmed down a bit and I would support moving this functions contents to the call site. I think it'll end up being pretty short, and easier to read.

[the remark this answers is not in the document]
nategraf: It would also remove the need to pass the second return value which amounts to `peer.PurposeIsSet(p2p.ValidatorPurpose)`

consensus/istanbul/backend/handler.go
nategraf: I would recommend leaving the handshake timeout as 5 seconds, and refactor this code a bit to allow for retrieving the cached value without triggering an update (or at least not a synchronous update). It'd be better to use stale data here than timeout the handshake.

consensus/istanbul/backend/handler.go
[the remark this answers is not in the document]
tkporter: I agree, good suggestion! fixed

consensus/istanbul/backend/handler.go
tkporter: Thought about this and I agree, I pushed some code that uses a previously cached value and asynchronously updates

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,175 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	}
+
+	return true, nil
+}
+
+// readValidatorProofMessage reads a validator proof message as a part of the
+// istanbul handshake. Returns if the peer is a validator or if an error occurred.
+func (sb *Backend) readValidatorProofMessage(peer consensus.Peer) (bool, error) {
+	logger := sb.logger.New("func", "readValidatorProofMessage")
+	peerMsg, err := peer.ReadMsg()
```
nategraf: Based on how low level `ReadMsg` is, we should verify the message code and size before decoding, much as the `peer.readStatus` method does. https://github.com/celo-org/celo-blockchain/blob/b7a7de550837eeb8ac1b30aa0f9baf4721c7d3a4/eth/peer.go#L454-L468

consensus/istanbul/backend/handler.go
nategraf: Opening this just to ask about a statement here. 

> but still would not be able to make yourself be a ValidatorPurpose peer

Is that true? It seems that the Upsert below makes the enode a trusted peer. I don't know if this clarification makes a difference to the functionality but I wanted to ask to make sure I understand what is going on here.

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,175 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	if !regAndActiveVals[msg.Address] {
+		logger.Debug("Received a validator proof message from peer with address that is not a registered or active validator", "msg.Address", msg.Address)
+		return false, nil
+	}
+
+	// If the versionedEnode message is too old, we don't count this proof as valid
+	// An error is given if the entry doesn't exist, so we ignore the error.
+	knownVersion, err := sb.valEnodeTable.GetVersionFromAddress(msg.Address)
+	if err == nil && versionedEnode.Version < knownVersion {
+		return false, nil
```
nategraf: I would like to see a `Debug` log here because this should not happen under normal conditions and would be worth noting. It may even be part of an attempted security violation, much like the use of an expired TLS certificate, but I cannot immediately think of a reasonable attack vector as long as this check is in place.

consensus/istanbul/backend/internal/enodes/val_enode_db.go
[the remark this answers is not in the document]
nategraf: +1 to removing it, as it duplicates the data in ID.

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,175 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+			return false, p2p.DiscReadTimeout
+		case isValidator := <-isValidatorCh:
+			return isValidator, nil
+		}
+	}
+}
+
+// shouldSendValidatorProof determines if this node should send a
+// validator proof revealing its address to a peer
+func (sb *Backend) shouldSendValidatorProof(peer consensus.Peer) (bool, error) {
```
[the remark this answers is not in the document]
tkporter: Oops! Deleted this. `peer.PurposeIsSet(p2p.ValidatorPurpose)` does in fact replace this

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,175 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+		msgBytes, err := validatorProofMsg.Payload()
+		if err != nil {
+			errCh <- err
+			return
+		}
+		err = peer.Send(istanbulValidatorProofMsg, msgBytes)
+		if err != nil {
+			errCh <- err
+		}
+		isValidatorCh <- containsProof
```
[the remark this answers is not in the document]
tkporter: If the peer is validatorpurpose it will automatically bypass the max peer check, but changing this to be `peer.PurposeIsSet(p2p.ValidatorPurpose)` is much more clear. Fixed!

consensus/istanbul/backend/handler.go
tkporter: You're right that Upsert makes a node a trusted ValidatorPurpose peer, however the node that's being upserted is the one from the message content (which can't be spoofed as it's signed by the validator who originally created it)  , not the enode of the peer

consensus/istanbul/backend/backend.go
```diff
@@ -848,7 +851,66 @@ func (sb *Backend) retrieveRegisteredAndElectedValidators() (map[common.Address]
...
+	sb.cachedRegisteredElectedValSet = validatorsSet
+	sb.cachedRegisteredElectedValSetTimestamp = time.Now()
+
+	return validatorsSet, nil
+}
 
+// retrieveMaybeStaleRegisteredAndElectedValidators returns the most recently
+// cached set of registered and elected validators and asynchronously updates
+// the cache if it is older than 1 minute. If no set has ever been cached, this
+// function may block for a couple seconds while retrieving the uncached set.
```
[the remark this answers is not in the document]
tkporter: What should happen if this returns nil during the handshake?
If we decide not to wait for the set to be retrieved, the handshake fails and there will be a few dozen seconds before the other node tries to reconnect with us.
If we decide to retrieve the set, then the call to this function has previously started the  asynchronous process of updating the cache and there's a high chance we'll end up doing double-work. If it times out we also fail the handshake

Having a higher chance of correctly accepting the remote peer as a validator feels like it's worth the chance of timing out to me. If you're okay with the chance of double-work I'd be up for changing this to return nil

consensus/istanbul/backend/backend.go
```diff
@@ -848,7 +851,66 @@ func (sb *Backend) retrieveRegisteredAndElectedValidators() (map[common.Address]
...
+func (sb *Backend) retrieveMaybeStaleRegisteredAndElectedValidators() (map[common.Address]bool, error) {
+	sb.cachedRegisteredElectedValSetMu.Lock()
```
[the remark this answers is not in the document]
tkporter: Changed it to a RWMutex that doesn't lock when calling the function that will take a while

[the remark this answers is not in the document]
nategraf: One result of this implementation, is that if a barrage of connection requests come in together when the cache needs to be updated, it will kick off a goroutine to update it for each one, until it is finally updated. It would be best to have machoism to ensure only one task is created to update the cache.

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
nategraf: It's not so much that I disagree with risking a handshake timeout in this case, it's that I dislike the interface and and think this should be a caller decision to retry with the blocking path. I raise this, and similar points, to make it easier for readers of the calling code to understand the flow, and to head off any future confusion where someone might use this function without fully understanding its rough edges and create a bigger issue.

p2p/server.go
```diff
@@ -955,11 +955,11 @@ running:
 				}
 				name := truncateName(c.name)
 				p.log.Debug("Adding p2p peer", "addr", p.RemoteAddr(), "peers", len(peers)+1, "name", name)
-				go srv.runPeer(p)
 				peers[c.node.ID()] = p
 				if p.Inbound() {
 					inboundCount++
 				}
+				go srv.runPeer(p)
```
nategraf: Looks like you a race condition. Nice catch! 🙌

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,152 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	for {
+		select {
+		case err := <-errCh:
+			return false, err
+		case <-timeout.C:
+			return false, p2p.DiscReadTimeout
+		case isValidator := <-isValidatorCh:
+			return isValidator, nil
+		}
+	}
```
nategraf: Just realized this doesn't need to be a loop because you return from all cases.
```suggestion
	select {
	case err := <-errCh:
		return false, err
	case <-timeout.C:
		return false, p2p.DiscReadTimeout
	case isValidator := <-isValidatorCh:
		return isValidator, nil
	}
```

eth/handler.go
```diff
@@ -400,13 +423,10 @@ func (pm *ProtocolManager) handle(p *peer) error {
 // peer. The remote connection is torn down upon returning any error.
 func (pm *ProtocolManager) handleMsg(p *peer) error {
 	// Read the next message from the remote peer, and ensure it's fully consumed
-	msg, err := p.rw.ReadMsg()
+	msg, err := p.ReadMsg()
```
kevjue: I see you created a new peer function `ReadMsg`.  Is there a specific reason for that?

les/client_handler.go
```diff
@@ -99,8 +99,13 @@ func (h *clientHandler) runPeer(version uint, p *p2p.Peer, rw p2p.MsgReadWriter)
 }
 
 func (h *clientHandler) handle(p *peer) error {
-	if h.backend.peers.Len() >= h.backend.config.LightPeers && !p.Peer.Info().Network.Trusted {
-		return p2p.DiscTooManyPeers
+	if !p.Peer.Info().Network.Trusted {
```
kevjue: See my comment regarding adding the incoming peer as a validator peer in the handshake.  If that's done, then this doesn't need to be checked here.

p2p/server.go
```diff
@@ -405,6 +407,19 @@ func (srv *Server) PeerCount() int {
 	return count
 }
 
+// InboundCount returns the number of inbound peers.
+func (srv *Server) InboundCount() int {
```
kevjue: See my comment regarding adding the incoming peer as a validator peer in the handshake. If that's done, then function doesn't need to be created.

consensus/istanbul/backend/handler.go
tkporter: Oops! thanks

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,150 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+
+	// Ensure the node in the versionedEnode matches the peer node
+	if node.ID() != peer.Node().ID() {
+		logger.Warn("Peer provided incorrect node ID in versionedEnode", "versionedEnode enode url", versionedEnode.EnodeURL, "peer enode url", peer.Node().URLv4())
+		return false, errors.New("Incorrect node in versionedEnode")
+	}
+
+	// Check if the peer is within the registered/elected valset
+	regAndActiveVals, err := sb.retrieveCachedRegisteredAndElectedValidators()
```
tkporter: I just realized that this lets a validator who is registered and not active bypass the max peer check. Right now only the active validators are meant to be a part of the overlay network, but this is meant to change to the "announce set"

@kevjue I know you're working on the announce set, does that involve including the entire announce set in the overlay network or just limiting announce messages to the announce set and keeping peer connections amongst those that are active?

nategraf:
I merged in `master` so this branch can be diffed with `trevor/fast-discovery`. This can be undone with `git reset --hard e3abd2634 && git push -f`.

eth/handler.go
```diff
@@ -338,6 +335,32 @@ func (pm *ProtocolManager) handle(p *peer) error {
...
+		}
+		forcePeer = isValidator
+		p.Log().Trace("Peer completed Istanbul handshake", "forcePeer", forcePeer)
+	}
+	// Ignore max peer and max inbound peer check if:
+	//  - this is a trusted or statically dialed peer
+	//  - the peer is from from the proxy server (e.g. peers connected to this node's internal network interface)
+	//  - forcePeer is true
+	if !forcePeer {
+		isStaticOrTrusted := p.Peer.Info().Network.Trusted || p.Peer.Info().Network.Static
```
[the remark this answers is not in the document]
tkporter: Just added! Great idea

I still only make the check inside the `if !forcePeer {` block because there isn't a guarantee that the new peer is trusted just yet (because `AddTrustedPeer` involves sending to a channel that's read by the server event loop goroutine), but now we don't expose the inbound count outside of the p2p server

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
tkporter: Great point-- I added some logic to make sure that only one goroutine is doing the work of `updateCachedValidatorConnSet` at a time. If another goroutine is currently doing the work of updating the cache, the new goroutine will wait for the work to be finished & return any error that arose in the work-goroutine

It does this using `sync.Cond`, here's a playground I made that does something similar: https://play.golang.org/p/MvpS1Zm-pnw

If you have a different solution in mind maybe this should be left to a subsequent PR?

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,172 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
 }
+
+// Handshake allows the initiating peer to identify itself as a validator
+func (sb *Backend) Handshake(peer consensus.Peer) (bool, error) {
+	// Only written to if there was a non-nil error when sending or receiving
+	errCh := make(chan error)
+	isValidatorCh := make(chan bool)
+
+	sendHandshake := func() {
+		var validatorProofMsg *istanbul.Message
```
nategraf: Naming this `validatorProofMsg` is a little misleading because its actaully the `versionedEnode`

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,172 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+		return false, nil
+	}
+
+	var versionedEnode versionedEnode
+	err = rlp.DecodeBytes(msg.Msg, &versionedEnode)
+	if err != nil {
+		return false, err
+	}
+
+	node, err := enode.ParseV4(versionedEnode.EnodeURL)
```
nategraf: Should we use `enode.Parse` instead to support the newer `enr` scheme? https://github.com/celo-org/celo-blockchain/blob/5c9f90e6c0495a20473a65c3cd85c516adcfc08f/p2p/enode/node.go#L65

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,172 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+func (sb *Backend) Handshake(peer consensus.Peer) (bool, error) {
+	// Only written to if there was a non-nil error when sending or receiving
+	errCh := make(chan error)
+	isValidatorCh := make(chan bool)
+
+	sendHandshake := func() {
+		var validatorProofMsg *istanbul.Message
+		var err error
+		peerIsValidator := peer.PurposeIsSet(p2p.ValidatorPurpose)
+		if peerIsValidator {
```
nategraf: I was thinking about this, and it seems to be that non-proxy full-nodes would also create and send their own self-signed enode certificate to a peer if it is marked as a validator, even though the full-node, if operating correctly would not want to be considered a validator by that peer. (In practice, this would not happen under most circumstances because they should not have any ValidatorPurpose connections)
Given this, you might add an `sb.coreStarted` check here as a way to see if this node intends to be seen as a validator.

consensus/istanbul/backend/handler.go
nategraf: +1 to that suggestion, although I was going to suggest just `enodeCertificate` rather than `versionedEnodeCertificate`. (The version is a case of a common security measure found in certificates)

eth/handler.go
```diff
@@ -338,6 +335,30 @@ func (pm *ProtocolManager) handle(p *peer) error {
...
+	// Ignore max peer and max inbound peer check if:
+	//  - this is a trusted or statically dialed peer
+	//  - the peer is from from the proxy server (e.g. peers connected to this node's internal network interface)
+	//  - forcePeer is true
+	if !forcePeer {
+		if err := p.Peer.Server.CheckPeerCounts(p.Peer); err != nil {
+			return err
+		}
+		isStaticOrTrusted := p.Peer.Info().Network.Trusted || p.Peer.Info().Network.Static
+		if !isStaticOrTrusted && pm.peers.Len() >= pm.maxPeers && p.Peer.Server != pm.proxyServer {
```
nategraf: How is this check different than the one done by `CheckPeerCounts`? If this is correct, could you add a comment to highlight that?

consensus/istanbul/backend/announce.go
```diff
@@ -835,18 +800,43 @@ func (sb *Backend) handleVersionedEnodeMsg(peer consensus.Peer, payload []byte)
...
+	// sent a versioned enode for the proxy to use in handshakes
+	if sb.config.Proxy && msg.Address == sb.config.ProxiedValidatorAddress {
+		// There may be a difference in the URLv4 string because of `discport`,
+		// so instead compare the ID
+		selfNode := sb.p2pserver.Self()
+		if parsedNode.ID() != selfNode.ID() {
+			logger.Warn("Received Istanbul Versioned Enode message with an incorrect enode url", "message enode url", versionedEnode.EnodeURL, "self enode url", sb.p2pserver.Self().URLv4())
+			return errors.New("Incorrect enode url")
+		}
+		sb.setVersionedEnodeMsg(&msg)
```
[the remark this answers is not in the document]
tkporter: Nice catch! To avoid the work of having to decode the `istanbul.Message` contents to compare the existing version, I changed the surrounding if condition to make sure the peer is the proxied validator.

consensus/istanbul/backend/handler.go
tkporter: My feeling is that we should wait to do this until we decide what our plans are with the ENR scheme and the announce protocol. Maybe this could be left to the new announce protocol changes? It seems like moving to ENR is a step in the right direction and the new announce protocol would be the time to switch

consensus/istanbul/backend/handler.go
tkporter: Checking `sb.coreStarted` wouldn't work if this node is a proxy that wants to be seen as its proxied validator. If the node isn't using modified code, I don't see a situation where a non-proxy full node would have an incorrect ValidatorPurpose peer

eth/handler.go
```diff
@@ -903,7 +924,7 @@ func (pm *ProtocolManager) FindPeers(targets map[enode.ID]bool, purpose p2p.Purp
 	for _, p := range pm.peers.Peers() {
 		id := p.Node().ID()
 		if targets[id] || (targets == nil) {
-			if purpose == p2p.AnyPurpose || p.Peer.StaticNodePurposes.IsSet(purpose) || p.Peer.TrustedNodePurposes.IsSet(purpose) {
+			if purpose == p2p.AnyPurpose || p.PurposeIsSet(purpose) {
```
kevjue: nit:  Consider moving the `p2p.AnyPurpose` considition into `peer.PurposeIsSet`.

eth/handler.go
```diff
@@ -76,7 +76,8 @@ type ProtocolManager struct {
 
 	txpool     txPool
 	blockchain *core.BlockChain
-	maxPeers   int
+
```
kevjue: nit:  extra white line

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,154 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+			return
+		}
+		err = peer.Send(istanbulValidatorHandshakeMsg, msgBytes)
+		if err != nil {
+			errCh <- err
+		}
+		isValidatorCh <- peerIsValidator
+	}
+	readHandshake := func() {
+		isValidator, err := sb.readValidatorProofMessage(peer)
```
kevjue: Regarding naming conventions, seems like `readValidatorProofMessage` should be renamed to `readIstanbulValidatorHandshakeMsg`.

tkporter: Oops, thanks

consensus/istanbul/backend/handler.go
```diff
@@ -286,3 +301,154 @@ func (sb *Backend) UnregisterPeer(peer consensus.Peer, isProxiedPeer bool) {
...
+	// An error is given if the entry doesn't exist, so we ignore the error.
+	knownVersion, err := sb.valEnodeTable.GetVersionFromAddress(msg.Address)
+	if err == nil && enodeCertificate.Version < knownVersion {
+		logger.Debug("Received a validator handshake message with an old version", "received version", enodeCertificate.Version, "known version", knownVersion)
+		return false, nil
+	}
+
+	// By this point, we know the peer is a validator and we update our val enode table accordingly
+	// Upsert will only use this entry if the version is new
+	err = sb.valEnodeTable.Upsert(map[common.Address]*vet.AddressEntry{msg.Address: {Node: node, Version: enodeCertificate.Version}})
```
kevjue: I thought a lot about this, and I think that if this node is a proxy, then it should not do the `upsert` here.  I think that the proxy should only send the certificate to the proxied validator, which will in turn update its valEnodeTable and then share that with its proxy.

This will  make reasoning of proxies' valEnodeTable updates/inserts much easier (e.g it's always the proxied validator that is driving the updates).  Also, when there are multiple proxies for a validator, doing it this way will prevent any race conditions on proxies valenodetables, since there would be only the proxied validator that is updating it and not both the proxied validator and the remote validator.

If we decide to make this change, this also implies that even though the remote validator sent the certificate to the proxy, the proxy should not return `true`, since there is a small possibility that the wrong proxy received it.

tkporter: Good point! I'm definitely in favor of this to keep the proxied validator in control. I think the issue of a validator connecting to the wrong proxy in the multiproxy setup is definitely a real issue. I'm also in support of changing the val enode share msg to be sent whenever the val enode table is updated to avoid the case that a proxy doesn't receive the entry until the timer ticks (up to a minute)

consensus/istanbul/backend/announce.go
```diff
@@ -687,3 +692,193 @@ func (sb *Backend) checkPeersAnnounceVersions() {
...
+		sb.setEnodeCertificateMsg(&msg)
+		return nil
+	}
+	// TODO: remove this check to allow non-proxy peers to send this message
+	// Issue tracked here: https://github.com/celo-org/celo-blockchain/issues/884
+	if sb.proxyNode == nil || sb.proxyNode.peer == nil || sb.proxyNode.peer.Node().ID() != peer.Node().ID() {
+		logger.Warn("Received Istanbul Enode Certificate message from invalid peer")
+		return errUnauthorizedAnnounceMessage
+	}
+
```
kevjue: wdyt of only executing the code below if this node is a proxied validator (e.g. `sb.config.Proxied && sb.proxyNode != nil && sb.proxyNode.peer != nil && sb.proxyNode.peer.ID() == peer.ID()`)?

That way, this function will only process the message if it's proxy (https://github.com/celo-org/celo-blockchain/pull/873/files#diff-49675a1e6ab38a575ed8c1e9a5f31145R804) or if it's a proxied validator (via the condition that I proposed above in this comment).

For all other cases, it would return an `errUnauthorizedAnnounceMessage` error (and the above condition and code block can be removed).

consensus/istanbul/backend/backend.go
```diff
@@ -780,6 +795,9 @@ func (sb *Backend) addProxy(node, externalNode *enode.Node) error {
 	sb.p2pserver.AddPeer(node, p2p.ProxyPurpose)
 
 	sb.proxyNode = &proxyInfo{node: node, externalNode: externalNode}
+	if err := sb.updateAnnounceVersion(newAnnounceVersion()); err != nil {
```
kevjue: Does this need to be done here?  Seems like proxy is send the enodeCertificate in `addPeer`.

consensus/istanbul/backend/engine.go
```diff
@@ -646,6 +646,10 @@ func (sb *Backend) StartValidating(hasBadBlock func(common.Hash) bool,
 		headBlock := sb.GetCurrentHeadBlock()
 		valset := sb.getValidators(headBlock.Number().Uint64(), headBlock.Hash())
 		sb.RefreshValPeers(valset)
+		// For a proxied validator, this is called in `sb.addProxy`
+		if err := sb.updateAnnounceVersion(newAnnounceVersion()); err != nil {
```
kevjue: Does this need to be done here?  Seems like the proxy is sent the enodeCertificate in `addPeer`.

consensus/istanbul/backend/backend.go
```diff
@@ -825,19 +843,76 @@ func (sb *Backend) ConnectToVals() {
...
+	sb.cachedValidatorConnSetMu.RLock()
+	defer sb.cachedValidatorConnSetMu.RUnlock()
+	return sb.cachedValidatorConnSet, nil
+}
+
+// updateCachedValidatorConnSet updates the cached validator conn set. If another
+// goroutine is simultaneously updating the cached set, this goroutine will wait
+// for the update to be finished to prevent the update work from occurring
+// simultaneously.
+func (sb *Backend) updateCachedValidatorConnSet() (err error) {
```
kevjue: nice function!

consensus/istanbul/backend/announce.go
tkporter: That's what the above if statement checks for-- I'll add `!sb.config.Proxied` to the condition as well to make it clearer

consensus/istanbul/backend/engine.go
tkporter: This call is for a non proxied validator, but the benefit of doing this here for a non-proxied validator is that `sb.enodeCertificateMsg` will be created without having to wait for the first announce msg to be generated/gossiped (iirc that takes about a minute). This allows the validator to prove itself during the handshake if needed during that time (& at startup I see this being pretty important)

consensus/istanbul/backend/backend.go
tkporter: I like having this here so that the announce version is definitely updated whenever there's a change in proxies (I know the RPC isn't well tested right now but better to support it)

But you're right, if `sb.p2pserver.AddPeer(node, p2p.ProxyPurpose)` (which is not synchronous) causes the peer to be registered before `updateAnnounceVersion` is called, the peer would have been sent the old enodeCertificate when it was registered, and then immediately after a new one when `updateAnnounceVersion` is called. I'm going to move `updateAnnounceVersion` to be called before `AddPeer` so that the proxy is guaranteed to only be sent the correct message when it's registered as a peer.

consensus/istanbul/backend/handler.go
tkporter: I thought about this some more, and I think sending a val enode share msg when upserting is probably a bad idea. These messages aren't small by any means and it could flood proxied validator <-> proxy communication if someone decided to repeatedly increase their version and send an enodeCertificate. What do you think about the proxy sending the enodeCertificate to the proxied validator, and then having the proxied validator send it back to the proxy if it wants to? That way the race condition is solved and we don't risk flooding the proxy with a ton of data

consensus/istanbul/backend/announce.go
kevjue: OK.  In that case, could you add some a comment on what cases that condition covers as well as the "fall through" case.  

The main reason for my initial comment is that it took me over 15 minutes to figure out what cases those conditions covered, so I think the comments will be very valuable.

consensus/istanbul/backend/handler.go
nategraf: Oh right 🙈Looks good to me

consensus/istanbul/backend/handler.go
tkporter: For some reason I thought the val enode share msg would cause the proxy to replace its val enode table with the proxied validator's table, but looking at the code it seems to be an `Upsert` that's done, so that eliminates the downsides of a race where:

1. Proxy upserts new entry from an enodeCertificate
2. Proxied validator sends val enode share msg
3. Proxied validator receives forwarded enodeCertificate
4. Proxy now receives val enode share msg

Trying to avoid letting multiproxy implementation details derail this PR too much, but IMO I think in the multiproxy setup the proxied validator should be in charge of which proxy will be used to connect to a new remote validator (and therefore should dictate the val enode table of the proxies).

I think in the single proxy setup it doesn't ultimately matter too much if the proxy upserts or not, but in the multiproxy setup we'll have to give this a deeper think. Are you ok with for the time being having the proxy give the proxied validator the authority to upsert, and then we can think this through more when we come out with multiproxy?

Another benefit to giving the authority to the proxied validator is if a proxy has been removed and a remote validator tries to connect to it. If the decision is left up to the proxied validator, then the remote validator will not incorrectly connect to a removed proxy.

PR #878, opened 2020-02-20
Retrieval of header from freezer bugfix

kevjue:
### Description

This PR if the bugfix for failing to retrieve blocks from the freezer. 

In the headerRLP retrieve function, the code will now retrieve the hash from the decoded header, instead of calculating it.   As mentioned by @mcortesi , the header hash is NOT simply the keccak hash of the RLP encoded header.  It is the hash of the RLP encoded header with the aggregated signature field excluded, since it is not consistent among all the validators.

### Tested

Verified that blocks stored in the freezer can be retrieved.

### Other changes

None

### Related issues

N/A

### Backwards compatibility

Yes


core/rawdb/accessors_chain.go
```diff
@@ -175,13 +175,21 @@ func WriteFastTrieProgress(db ethdb.KeyValueWriter, count uint64) {
...
 	data, _ := db.Ancient(freezerHeaderTable, number)
-	if len(data) > 0 && crypto.Keccak256Hash(data) == hash {
-		return data
+	if len(data) > 0 {
+		// Extra hash comparison is necessary since ancient database only maintains
+		// the canonical data.
+		header := new(types.Header)
+		if err := rlp.Decode(bytes.NewReader(data), header); err != nil {
+			log.Error("Invalid block header RLP", "hash", hash, "err", err)
+			return nil
```
nategraf: It seems unclear to me if this should really be returning nil here. It might make more sense to treat this as if the hash didn't match and continue on with the function, while still logging the error.

kevjue: Ya, good point.  Will make this change.

core/rawdb/accessors_chain.go
```diff
@@ -175,13 +175,19 @@ func WriteFastTrieProgress(db ethdb.KeyValueWriter, count uint64) {
...
 // ReadHeaderRLP retrieves a block header in its raw RLP database encoding.
 func ReadHeaderRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValue {
-	// First try to look up the data in ancient database. Extra hash
-	// comparison is necessary since ancient database only maintains
-	// the canonical data.
+	// First try to look up the data in ancient database.
 	data, _ := db.Ancient(freezerHeaderTable, number)
-	if len(data) > 0 && crypto.Keccak256Hash(data) == hash {
-		return data
+	if len(data) > 0 {
```
asaj: Do we need to make a similar change on line 201?

kevjue: Ya, nice catch.

core/rawdb/accessors_chain.go
```diff
@@ -192,8 +197,13 @@ func ReadHeaderRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValu
 	// but when we reach into leveldb, the data was already moved. That would
 	// result in a not found error.
 	data, _ = db.Ancient(freezerHeaderTable, number)
-	if len(data) > 0 && crypto.Keccak256Hash(data) == hash {
-		return data
+	if len(data) > 0 {
+		header := new(types.Header)
```
mcortesi: maybe extract this into a function... can be one inside the current function, but to extract common logic

mcortesi: something like
```go
if len(data) > 0 && sameHeader(data, hash) {
  return data
}
```

core/rawdb/accessors_chain.go
```diff
@@ -175,13 +174,19 @@ func WriteFastTrieProgress(db ethdb.KeyValueWriter, count uint64) {
...
+	// First try to look up the data in ancient database.
 	data, _ := db.Ancient(freezerHeaderTable, number)
-	if len(data) > 0 && crypto.Keccak256Hash(data) == hash {
-		return data
+	if len(data) > 0 {
+		// Extra hash comparison is necessary since ancient database only maintains
+		// the canonical data.
+		header := new(types.Header)
+		if err := rlp.Decode(bytes.NewReader(data), header); err != nil {
+			log.Error("Invalid block header RLP", "hash", hash, "err", err)
```
mcortesi: would change to something with better context, like: "Error decoding stored block header" and print the block number, no the hash. Since you need the number to query the db

PR #881, opened 2020-02-25
Continue to distribute rewards if a single validator fails

nategraf:
### Description

Currently if distributing rewards to an individual validator fails, reward distribution to other validators stops, without error. This PR changes that to log an error and continue instead.

### Tested

E2e tests

### Related issues

- Related https://github.com/celo-org/celo-labs/issues/306

### Backwards compatibility

Potentially a hard-fork


asaj:
What motivated this change? In some sense, I think it's nice to think of all of the proof-of-stake actions as a single "transaction", such that either everything succeeds or nothing does.

PR #889, opened 2020-02-29
Decrease announce validator set size

kevjue:
### Description

Currently, all registered validators will broadcast their enodeURL to all other registered validators and currently elected validators.

To decrease the exposure of those enodeURLs, this PR contains changes so that the enodeURL is broadcast to the next threshold's elected validators + 10 validators that are just below the election cutoff.

### Tested

* e2e CI tests.
* Ran a testnet with 10 elected validators and 30 registered validators, and verified that the enodeURL is shared with only 20 nodes.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes.

### DIFF
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -173,12 +173,12 @@ func (sb *Backend) announceThread() {
 func (sb *Backend) shouldGenerateAndProcessAnnounce() (bool, error) {
 
 	// Check if this node is in the registered/elected validator set
-	regAndElectedVals, err := sb.retrieveRegisteredAndElectedValidators()
+	validatorConnSet, err := sb.retrieveValidatorConnSet()
 	if err != nil {
 		return false, err
 	}
 
-	return sb.coreStarted && regAndElectedVals[sb.Address()], nil
+	return sb.coreStarted && validatorConnSet[sb.Address()], nil
 }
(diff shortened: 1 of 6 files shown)


consensus/istanbul/backend/announce.go
```diff
@@ -518,13 +518,13 @@ func (sb *Backend) validateAnnounce(announceData *announceData) (bool, error) {
...
-	regAndActiveVals, err := sb.retrieveRegisteredAndElectedValidators()
+	validatorConnSet, err := sb.retrieveValidatorConnSet()
 	if err != nil {
 		return false, err
 	}
 
-	if len(announceData.AnnounceRecords) > 2*len(regAndActiveVals) {
-		logger.Info("Number of announce message encrypted enodes is more than two times the size of the current reg/elected validator set", "num announce enodes", len(announceData.AnnounceRecords), "reg/elected val set size", len(regAndActiveVals))
+	if len(announceData.AnnounceRecords) > 2*len(validatorConnSet) {
+		logger.Info("Number of announce message encrypted enodes is more than two times the size of the current reg/elected validator set", "num announce enodes", len(announceData.AnnounceRecords), "reg/elected val set size", len(validatorConnSet))
```
asaj: Message needs updating, `reg/elected`

PR #890, opened 2020-03-02
Filters gold transfers from tx pool if transfers are frozen

alecps:
### Description

This PR adds checks in core/tx_pool.go and light/txpool.go that filter transactions transferring gold (i.e. with tx.Value() > 0). The checks read from layer 2 contracts to determine if gold is frozen and then to check if either the sender or receiver is whitelisted. To do this, ./freezer and ./transfer_whitelist are added to ./contract_comm.

### Tested

e2e tests added to celo-monorepo in #[3011](https://github.com/celo-org/celo-monorepo/pull/3011)

### Other changes

Some minor formatting changes in core/tx_pool_test.go. One corrects a comment typo and another was performed automatically on my machine when saving. 

### Related issues

- Fixes #[251](https://github.com/celo-org/celo-labs/issues/251)

### Backwards compatibility

Yes. Since we do not filter transactions as part of the state transition function, we do not need to worry about issues with backwards compatibility if validators were to remove these checks and include frozen gold transactions in blocks. This is acceptable because such behavior would be easy to identify and punish.


kevjue:
The nodes should also verify that all the "frozen" rules are enforced in this function:  https://github.com/celo-org/celo-blockchain/blob/alecps/frozenGold/core/state_transition.go#L316

I see that you enforce them in the tx_pool's validate function, but a malicious validator should ignore those enforcements, and compose a block with gold transfers.   The other validators should then reject that block, and the check for bad transactions in that block are done in the function that I mentioned.

light/txpool.go
```diff
@@ -365,6 +367,21 @@ func (pool *TxPool) validateTx(ctx context.Context, tx *types.Transaction) error
 		return core.ErrNonceTooLow
 	}
 
+	// Ensure gold transfers are whitelisted if transfers are frozen.
```
asaj: Looks like there's a bunch of duplicated code here, can we share this some how?

params/protocol_params.go
```diff
@@ -216,7 +218,9 @@ const (
 	MaxGasForGetRegisteredValidators               uint64 = 2000000
 	MaxGasForGetValidator                          uint64 = 100 * 1000
 	MaxGasForGetWhiteList                          uint64 = 20000
+	MaxGasForGetTransferWhitelist                  uint64 = 20000
```
asaj: Consider 1M

asaj:
@kevjue so we actually debated this quite a bit and this approach ultimately seemed like the best path forward, suggest you and @alecps find some time to discuss

light/txpool.go
alecps: Good thought

PR #891, opened 2020-03-03
Added debiting gas for some precompiled contracts

### LINKED ISSUE #565: Transfer precompile should charge gas
### Expected Behavior

Transfer precompile calls `debitRequiredGas`, charges gas (possibly `CallValueTransferGas`?)

### Current Behavior

No gas is debited (after https://github.com/celo-org/celo-blockchain/pull/551/files)

mrsmkl:
### Description

Added `debitRequiredGas` to `TobinTransfer` and some contracts where it was missing (after updating to geth 1.9?)

Two more things that might be useful
* Why do we have `debitRequiredGas` in the first place instead of just debiting the gas in `RunPrecompiledContract` like upstream geth???
* Move the Celo precompiles to be last in the file

### Related issues

- Fixes #565

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -518,7 +518,7 @@ func runBn256Pairing(input []byte, caller common.Address, evm *EVM, gas uint64) 
 type transfer struct{}
 
 func (c *transfer) RequiredGas(input []byte) uint64 {
-	return params.TxGas
+	return params.CallValueTransferGas
 }
 
 func (c *transfer) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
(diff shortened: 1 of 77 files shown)


core/vm/contracts.go
```diff
@@ -506,6 +522,10 @@ func (c *transfer) RequiredGas(input []byte) uint64 {
 }
 
 func (c *transfer) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
+	gas, err := debitRequiredGas(c, input, gas)
```
asaj: I don't think we actually want to debit gas here. Or, we want to return something much smaller than `TxGas` above. Maybe it should be `CallValueTransferGas`? Wdyt?

mrsmkl: ok, didn't check what `TxGas` is.

PR #893, opened 2020-03-05
First pass at announce protocol

tkporter:
### Description

This is a first pass at the implementation of https://github.com/celo-org/CIPs/pull/6. This PR also includes the handshake PR. This hasn’t been tested very much, and I’m sure there are bugs but I wanted to open it up for review as early as possible while I continue testing it.

## Fast Discovery

Fast discovery involves every node keeping a copy of the latest known announce version for each registered or elected validator. Each version is signed by the original author and takes the shape of:

```
type SignedAnnounceVersion struct {
	Address   common.Address
	Version   uint
	Signature []byte
}
```

These are persisted in a levelDB database.

`SignedAnnounceVersion`s are gossiped throughout the network using the message `istanbulSignedAnnounceVersionsMsg` that contains an array of `SignedAnnounceVersion`s. When a node receives this message, it will regossip the entries without any: duplicates, elements with incorrect signatures, entries that were recently gossiped within the last 5 minutes, and any entries with versions that are not new to the node. When two nodes first peer with each other, they exchange their entire signed announce version DB and will always favor any entries with newer versions.

This can be seen using the RPC `Istanbul.signedAnnounceVersionTableInfo`

## Direct announce

“Direct announce” involves a validator sending a `versionedEnode` directly to another validator. This results in the receiving node upserting the content the versioned enode into the val enode table. 

```
type versionedEnode struct {
	EnodeURL string
	Version  uint
}
```

## Query

Whenever a validator becomes aware of an entry (or multiple entries) in the SignedAnnounceVersionDB that either does not have an entry in the val enode table, or if the version in the val enode table is less than the version in the SignedAnnounceVersionDB, a “query” message is sent out. This is the same thing as a regular announce message, except the content of the announce message is only intended for those other remote validators— ie the validator sending the announce only encrypts for those target validators. The enodeURL of the sending validator uses the ECDSA public keys of the remote validators found in the signature of the remote validator’s `SignedAnnounceVersion` to encrypt its enodeURL.

Validators will keep gossiping out these announce messages targeted for the outdated or non existent remote validators using the old rate of gossiping for now, but I’m thinking of changing this to be a flat 5 minutes.

## More info on announce versions

With the single proxy setup, there are two situations where an announce version is updated:

1. A non-proxied validator starts mining in `StartValidating`
2. A proxied validator adds a proxy in `addProxy`

When a proxy is provided at the startup of a proxied validator, `StartValidating` calls `addProxy`, so in both situations a new announce version will be created in `StartValidating`.

Updating an announce version involves calling `updateAnnounceVersion`, which sets the versioned enode msg that will be used for any handshakes, sends it to a proxied validator for using in handshakes (if applicable), sends a direct announce to all registered or active validator peers with the new version, and then gossips out the new `SignedAnnounceVersion`.

Announcing will only start after this first announce version update and the core starts.

### Tested

Ran a testnet, things looked good. I plan to chaos test and test more in depth

### Other changes

n/a

### Backwards compatibility

NOT backwards compatible


consensus/istanbul/backend/announce.go
```diff
@@ -470,28 +478,30 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
+					return err
+				}
+				enodeURL := string(enodeBytes)
+				node, err := enode.ParseV4(enodeURL)
+				if err != nil {
+					logger.Error("Error parsing enodeURL", "enodeUrl", enodeURL)
 					return err
 				}
-
 				if err := sb.valEnodeTable.Upsert(map[common.Address]*vet.AddressEntry{msg.Address: {Node: node, Version: announceData.Version}}); err != nil {
```
tkporter: Need to send a direct announce if the original author node is already a peer

nategraf:
Changing the base of this PR to `trevor/p2p-handshake-1.9`, on the assumption that the former will be merged first, so that I can focus review comments on what is new.

nategraf:
It seems there is some overlap here with the handshake PR, so I haven't reviewed this one yet, but I do have a couple of comments on the (very good) description.
1. I think `SignedAnnounceVersion` could be reduced to the following because the address could be recovered from the signature. Eliminating the address field could get rid of those bytes on the wire. (Or it could be kept in the struct, but removed for the wire format)
```
type SignedAnnounceVersion struct {
	Version   uint
	Signature []byte
}
```
2. We specify the size of the `uint` (e.g. as `uint64` or `uint32`) when used to prevent any inconsistencies between platforms that would be a pain to debug.

nategraf:
One more comment as well. If the struct signed is `{ address, uint }` or even just `{ uint }` like I proposed above, it introduces some risk that a valid signed message could be found from another context that can be decoded into that interface. (i.e. in some other context, the same key signed a message that is exactly the same size). If that's the case, then nasty things could happen, such as an announce message being propagated with an artificially high version number which would prevent the acceptance of accurate announce messages.
If we want to mitigate this, admittedly obscure, risk then we could add a hardcoded salt value which is appended to the message for signature creation and verification, which would prevent values from other contexts from being used.

kevjue:
I'm still going through this PR (will take me a while), and in general, I think things look great!  I'm in the process of adding comments that are mostly around style and variable renaming considerations, and so far don't see any structural issues.

However, one thing I am still a bit uneasy about (and I know that we've already discussed this in your other PR) is the moving of the p2p connection count check from the p2p layer to the eth/les layer.  One alternative proposal I have is to basically extend the p2p server event loop to have an additional checkpoint (e.g. checkpointPostSubProtocolHandshake) that will do the original p2p connection count check.

This proposal will require that the p2p.proposal is modified so that it has a new `Handshake` function that is separate from the 'Run' function.

kevjue:
Regarding names, wdty of the following:

`Announce` is the name of the protocol to publish enodeURL to select nodes.

`versionedEnode` -> `AssignedEnodeURL` This will reflect the fact that the validator has assigned this enodeURL to the remote validator. Will be much more relevant once multi proxy is merged

`announceData` type -> `AssignedEnodeURLRequestsData` This reflects the fact that the message that send this packet data type is so that the destination validators will respond with it's assigned enodeurl.

`announceData.Version` field -> `Timestamp` Since it will be the local timestamp of when this message is sent, as opposed to version of a specific variable

`announceRecord.EncryptedEnodeURL` field -> `assignedEnodeURL` with type `AssignedEnodeURL`

consensus/istanbul/backend/announce.go
```diff
@@ -170,25 +170,31 @@ func (sb *Backend) announceThread() {
...
 
-			updateAnnounceVersionFunc()
-			go sb.generateAndGossipAnnounce(announceVersion)
-
 			// Use this timer to also prune all announce related data structures.
 			if err := sb.pruneAnnounceDataStructures(); err != nil {
 				logger.Warn("Error in pruning announce data structures", "err", err)
 			}
 
-		case <-announceVersionsCheckTicker.C:
```
kevjue: The signedAnnounceVersions will still need to be intermittently checked with its peers.

consensus/istanbul/backend/backend.go
```diff
@@ -867,6 +878,23 @@ func (sb *Backend) retrieveValidatorConnSet() (map[common.Address]bool, error) {
 	return sb.cachedValidatorConnSet, nil
 }
 
+// retrieveCachedValidatorConnSet returns the most recently cached validator conn
```
kevjue: It seems like this function and `retrieveValidatorConnSet` can be merged into one function with an addition boolean parameter such as `updateAsynchronously`.

consensus/istanbul/backend/handler.go
```diff
@@ -409,14 +407,24 @@ func (sb *Backend) readValidatorHandshakeMessage(peer consensus.Peer) (bool, err
 		return false, errors.New("Incorrect node in enodeCertificate")
 	}
 
-	block := sb.currentBlock()
-	valSet := sb.getValidators(block.Number().Uint64(), block.Hash())
-	if !valSet.ContainsByAddress(sb.ValidatorAddress()) {
-		logger.Trace("This validator is not in the validator set")
+	// Check if the peer is within the validator conn set.
+	validatorConnSet := sb.retrieveCachedValidatorConnSet()
```
kevjue: What's the need for the new function `retrieveCachedValidatorConnSet`?  This function is updating the cache if there is a cache miss.

consensus/istanbul/backend/backend.go
```diff
@@ -193,14 +203,26 @@ type Backend struct {
...
+	lastSignedAnnounceVersionsGossipedMu sync.RWMutex
+
+	announceRunning             bool
+	announceMu                  sync.RWMutex
+	announceThreadWg            *sync.WaitGroup
+	announceThreadQuit          chan struct{}
+	generateAndGossipAnnounceCh chan struct{}
 
-	// Map of the received announce message where key is originating address and value is the msg byte array
-	cachedAnnounceMsgs   map[common.Address]*announceMsgCachedEntry
```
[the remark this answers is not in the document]
tkporter: `signedAnnounceVersionTable` fills that role

consensus/istanbul/backend/backend.go
```diff
@@ -278,7 +307,10 @@ func (sb *Backend) Address() common.Address {
 // Close the backend
 func (sb *Backend) Close() error {
 	sb.delegateSignScope.Close()
-	return sb.valEnodeTable.Close()
+	if err := sb.valEnodeTable.Close(); err != nil {
+		return err
+	}
+	return sb.signedAnnounceVersionTable.Close()
```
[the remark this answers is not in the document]
tkporter: Ill change to make it more clear

consensus/istanbul/backend/announce.go
```diff
@@ -343,52 +352,42 @@ func (sb *Backend) generateAndGossipAnnounce() error {
...
+	if err != nil {
+		return nil, err
 	}
 
 	announceData := &announceData{
 		AnnounceRecords: announceRecords,
-		EnodeURLHash:    istanbul.RLPHash(enodeUrl),
-		// Unix() returns a int64, but we need a uint for the golang rlp encoding implmentation. Warning: This timestamp value will be truncated in 2106.
-		Version: uint(time.Now().Unix()),
+		Version:         announceVersion,
```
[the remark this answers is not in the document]
tkporter: Why not the latest announce version for the sending node? The intended recipient will upsert a new entry into the val enode table with the decrypted enode URL and the version provided in the `announceData`. If the version that's sent in the message is the current timestamp it will result in a different version that's upserted in the recipient validator's val enode table from the one that it should know about via the signed announce version exchanging

consensus/istanbul/backend/handler.go
tkporter: This is because of some feedback in the handshake PR from @nategraf: https://github.com/celo-org/celo-blockchain/pull/873#discussion_r386743099

The rationale is that it'd be better to use some potentially stale data during the handshake to avoid risking a potentially long-blocking function. I know we managed to reduce the execution time for elections recently, do you think this is a non-issue now that we are using the validator con set?

consensus/istanbul/backend/handler.go
```diff
@@ -287,8 +285,8 @@ func (sb *Backend) RegisterPeer(peer consensus.Peer, isProxiedPeer bool) {
 		}
 	}
 
-	if peer.Version() >= 65 {
-		sb.sendGetAnnounceVersions(peer)
+	if err := sb.sendAllSignedAnnounceVersions(peer); err != nil {
```
nategraf: With this change to unconditionally send the announce versions, we should probably also drop support for istanbul version 64 in the `eth` handshake.

consensus/istanbul/backend/announce.go
```diff
@@ -114,9 +120,7 @@ func (sb *Backend) announceThread() {
 				// hence more likely that they will be aware that this node is
 				// within that set.
 				time.AfterFunc(1*time.Minute, func() {
```
nategraf: Suggest getting rid of this function call as it will get announced in one minute as part of the event loop anyway. This only serves to have it announce twice quickly.

consensus/istanbul/backend/announce.go
```diff
@@ -114,9 +120,7 @@ func (sb *Backend) announceThread() {
...
-					if err := sb.generateAndGossipAnnounce(announceVersion); err != nil {
-						logger.Error("Error in gossiping announce", "err", err)
-					}
+					sb.startGossipAnnounceTask()
 				})
 
 				if sb.config.AnnounceAggressiveGossipOnEnablement {
```
nategraf: Nit: Comment on line 138 is a bit misleading now that the announcing variable exists

consensus/istanbul/backend/announce.go
```diff
@@ -170,25 +174,40 @@ func (sb *Backend) announceThread() {
...
+			if shouldAnnounce {
+				err := sb.generateAndGossipAnnounce()
+				if err != nil {
+					logger.Warn("Error in generating and gossiping announce", "err", err)
+				}
+			}
 
 		case <-sb.updateAnnounceVersionCh:
 			updateAnnounceVersionFunc()
 			sb.updateAnnounceVersionCompleteCh <- struct{}{}
```
nategraf: Something is wrong about this. I don't think it should send to the channel here.

consensus/istanbul/backend/announce.go
```diff
@@ -246,6 +256,20 @@ func (sb *Backend) pruneAnnounceDataStructures() error {
 		return err
 	}
 
+	sb.lastSignedAnnounceVersionsGossipedMu.Lock()
+	for remoteAddress := range sb.lastSignedAnnounceVersionsGossiped {
+		if !validatorConnSet[remoteAddress] {
+			logger.Trace("Deleting entry from lastSignedAnnounceVersionsGossiped", "address", remoteAddress, "gossip timestamp", sb.lastSignedAnnounceVersionsGossiped[remoteAddress])
+			delete(sb.lastSignedAnnounceVersionsGossiped, remoteAddress)
```
nategraf: Because this is a DDoS protection with a timeout, we should keep the entry for the address in this table until that timeout has expired.

consensus/istanbul/backend/announce.go
```diff
@@ -370,52 +369,42 @@ func (sb *Backend) generateAndGossipAnnounce(version uint) error {
...
-	sb.cachedAnnounceMsgs[announceAddress] = &announceMsgCachedEntry{MsgVersion: version, MsgPayload: payload}
-
 	return sb.Multicast(nil, payload, istanbulAnnounceMsg)
 }
 
-// This function is a helper function for generateAndGossipAnnounce.
-// It will create the latest announce msg for this node with a given version.
-func (sb *Backend) generateAnnounce(version uint) (*istanbul.Message, common.Address, error) {
+// This function is a helper function for generateAndGossipAnnounce that will
+// use the version for this node's address in the signed announce version db
```
nategraf: Nit: Commenting that this is a helper function isn't particularly helpful to the reader. Could you make this more descriptive.

consensus/istanbul/backend/announce.go
```diff
@@ -370,52 +369,42 @@ func (sb *Backend) generateAndGossipAnnounce(version uint) error {
...
+	announceRecords, err := sb.generateAnnounceRecords(enodeURL)
+	if err != nil {
+		logger.Warn("Error generating announce records", "err", err)
+		return nil, err
+	}
+	if len(announceRecords) == 0 {
+		logger.Trace("No announce records were generated, will not generate announce")
+		return nil, nil
+	}
+	announceVersion, err := sb.signedAnnounceVersionTable.GetVersion(sb.Address())
```
nategraf: Is it possible that we could get stuck here if the entry does not exist or the DB gets corrupted? This may need a fallback for if it returns an error.

consensus/istanbul/backend/announce.go
```diff
@@ -428,12 +417,58 @@ func (sb *Backend) generateAnnounce(version uint) (*istanbul.Message, common.Add
...
+		return nil, err
 	}
 
 	logger.Debug("Generated an announce message", "IstanbulMsg", msg.String(), "AnnounceData", announceData.String())
 
-	return msg, msg.Address, nil
+	return msg, nil
+}
+
+// This is a helper function for generateAnnounce.
```
nategraf: Nit: Please use the golang `generateAnnounceRecords returns...` format.

consensus/istanbul/backend/announce.go
```diff
@@ -428,12 +417,58 @@ func (sb *Backend) generateAnnounce(version uint) (*istanbul.Message, common.Add
...
+	var announceRecords []*announceRecord
+	for _, signedAnnVersionEntry := range allSignedAnnVersionEntries {
+		// Don't generate an announce record for ourselves
+		if signedAnnVersionEntry.Address == sb.Address() {
+			continue
+		}
+
+		valEnode := allValEnodes[signedAnnVersionEntry.Address]
+		// If the version in the val enode table is up to date with the version
+		// we are aware of (or is newer in the case the remote validator sent us
```
nategraf: Nit: Could reword the comment in the parenthesis. I am having trouble understanding what you are trying to say.

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
 		logger.Trace("Going to process an announce msg", "announce records", announceData.AnnounceRecords)
```
nategraf: On line 519, consider continuing instead of bailing out here if the function returned an error.

nategraf:
```suggestion
		logger.Trace("Processing and announce message", "announce records", announceData.AnnounceRecords)
```

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
 		for _, announceRecord := range announceData.AnnounceRecords {
 			if announceRecord.DestAddress == sb.Address() {
```
nategraf: Nit: Please change this to `if announceRecord.DestAddress != sb.Address() { continue }` to reduce the indentation of this block.

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
-				// TODO: Decrypt the enodeURL using this validator's validator key after making changes to encrypt it
-				enodeUrl := string(announceRecord.EncryptedEnodeURL)
-				node, err := enode.ParseV4(enodeUrl)
+				enodeBytes, err := sb.decryptFn(accounts.Account{Address: sb.Address()}, announceRecord.EncryptedEnodeURL, nil, nil)
 				if err != nil {
-					logger.Error("Error in parsing enodeURL", "enodeUrl", enodeUrl)
+					sb.logger.Warn("Error decrypting endpoint", "err", err, "announceRecord.EncryptedEnodeURL", announceRecord.EncryptedEnodeURL)
 					return err
```
nategraf: Consider breaking from the loop instead of returning an error

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
 				}
-
-				if err := sb.valEnodeTable.Upsert(map[common.Address]*vet.AddressEntry{msg.Address: {Node: node, Version: announceData.Version}}); err != nil {
-					logger.Warn("Error in upserting a valenode entry", "AnnounceData", announceData.String(), "error", err)
+				enodeURL := string(enodeBytes)
+				node, err := enode.ParseV4(enodeURL)
+				if err != nil {
+					logger.Error("Error parsing enodeURL", "enodeUrl", enodeURL)
```
nategraf: Please make this log level consistent with others

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
 					return err
 				}
-
+				if err := sb.answerAnnounceMsg(msg.Address, node, announceData.Version); err != nil {
+					logger.Warn("Error answering an announce msg", "target node", node.URLv4(), "error", err)
+					return err
+				}
+				// If the announce was only intended for this node, do not regossip
+				if len(announceData.AnnounceRecords) == 1 {
```
nategraf: This could be an attack vector, because the difference in behavior between a validator for whom the announce is intended or unintended is different. To avoid revealing a node as a validator, it must act indistinguishable from one that is not.

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
-	sb.cachedAnnounceMsgsMu.Unlock()
-
 	// Regossip this announce message
 	return sb.regossipAnnounce(msg, payload)
 }
 
+// answerAnnounceMsg will answer a received announce message. If the target
+// node is already a peer of any kind, an enodeCertificate will be sent.
+// Regardless, the target node will be upserted into the val enode table
+// to ensure this node designates the target node as a ValidatorPurpose peer
```
nategraf: Nit: I find the last statement here mildly confusing. "target" and "valid" in this sentance are ambigous. Its also more detail than is relevant here anyway.

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
+// with a valid version.
+func (sb *Backend) answerAnnounceMsg(address common.Address, node *enode.Node, version uint) error {
+	targetID := node.ID()
```
nategraf: Nit: Just use `node.ID()` directly instead of creating this variable

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
+		targetID: true,
+	}
+	// The target could be an existing peer of any purpose.
+	matches := sb.broadcaster.FindPeers(targetIDs, p2p.AnyPurpose)
+	if matches[targetID] != nil {
+		enodeCertificateMsg, err := sb.retrieveEnodeCertificateMsg()
+		if err != nil {
+			return err
+		}
+		if err := sb.sendEnodeCertificateMsg(matches[targetID], enodeCertificateMsg); err != nil {
```
nategraf: If I understand correctly, we could receive the same gossiped announce message from multiple peers. If so, we should make sure we avoid sending that peer duplicate enode certificates

consensus/istanbul/backend/announce.go
```diff
@@ -497,32 +523,66 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
+	// during the istanbul handshake.
+	if err := sb.valEnodeTable.Upsert(map[common.Address]*vet.AddressEntry{address: {Node: node, Version: version}}); err != nil {
+		return err
+	}
+	return nil
+}
+
 // validateAnnounce will do some validation to check the contents of the announce
 // message. This is to force all validators that send an announce message to
 // create as succint message as possible, and prevent any possible network DOS attacks
```
nategraf: Nit: On line 624, note is no longer accurate without the cache of the message. Please adjust this note.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
 	return nil
```
nategraf: On line 632, because this is the same as the new default announce period, and we do not expect most users to be tweaking that parameter, they are running up against race condition with this DDoS protection. Assuming normally distributed network latency and announce processing time, the message will be rejected 50% of the time when the gossip interval equals the DDoS protection limit. DDoS protection should be adjusted down, or announce interval should be adjusted up. Additional though here is that because these messages range in size so much, we might want to impose a more nuanced limit, like 100 entries or 10MB per 5 minutes. Here is a simply implementation of that idea I was able to whip up https://play.golang.org/p/A0Y9IUCLXRP

nategraf: On line 638, unlocking before updating the time allows a race condition where any number of announce messages from the same address may be regossiped as long as they are received at the same time.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
 }
 
-// ===============================================================
-//
-// define the AnnounceVersion messge format, the getAnnounceVersions send and handle function, and the announceVersions handle function
+// signedAnnounceVersion is a signed message from a validator indicating the most
+// recent version of its enode.
+type signedAnnounceVersion struct {
```
nategraf: Seeing as this is the same as `vet.SignedAnnounceVersionEntry`, one should be a type alias of the other. This will also eliminate athe need for the `newSignedAnnounceVersionFromEntry` function because you can just do a direct type cast.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
 
-type announceVersion struct {
-	ValAddress         common.Address
-	AnnounceMsgVersion uint
+func (sav *signedAnnounceVersion) ECDSAPublicKey() (*ecdsa.PublicKey, error) {
```
nategraf: If you keep the `Address` field (see my previous comment on this PR), it may be useful to verify the address matches the recovered public key (returning an error if it does not)

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+	}
 }
 
-// sendGetAnnounceVersions will send a GetAnnounceVersions message to a specific peer to request it's announceVersion set
-func (sb *Backend) sendGetAnnounceVersions(peer consensus.Peer) {
-	sb.logger.Trace("Sending a GetAnnounceVersions message", "func", "sendGetAnnounceVersions", "peer", peer)
-	go peer.Send(istanbulGetAnnounceVersionsMsg, []byte{})
+// String gives a string representation of signedAnnounceVersion
+func (sav *signedAnnounceVersion) String() string {
+	return fmt.Sprintf("{Address: %v, Version: %v, Signature.length: %v}", sav.Address, sav.Version, len(sav.Signature))
```
nategraf: Nit: Consider including the hex encoded signature here.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+	for i, entry := range allEntries {
+		allSignedAnnounceVersions[i] = newSignedAnnounceVersionFromEntry(entry)
+	}
+	return allSignedAnnounceVersions, nil
+}
 
-	announceVersionsBytes, err := rlp.EncodeToBytes(announceVersions)
+// sendAllSignedAnnounceVersions sends all SignedAnnounceVersions this node
+// has to a peer
+func (sb *Backend) sendAllSignedAnnounceVersions(peer consensus.Peer) error {
```
nategraf: Nit
```suggestion
func (sb *Backend) sendAnnounceVersionTable(peer consensus.Peer) error {
```

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+		validAddresses[signedAnnVersion.Address] = true
+		validEntries = append(validEntries, signedAnnVersion.Entry())
+	}
+	return sb.upsertSignedAnnounceVersionEntries(validEntries)
+}
 
-		sb.cachedAnnounceMsgsMu.RLock()
-		if cachedEntry, ok := sb.cachedAnnounceMsgs[announceVersion.ValAddress]; !ok || cachedEntry.MsgVersion < announceVersion.AnnounceMsgVersion {
-			announcesToRequest = append(announcesToRequest, announceVersion.ValAddress)
+func (sb *Backend) upsertSignedAnnounceVersionEntries(entries []*vet.SignedAnnounceVersionEntry) error {
```
nategraf: Nit: I found the contents of this function very surprising given its signature. The actual upsert step takes up 3 lines at the start of the function and the rest is DDoS protection filtering and regossip. Please extract the upsert to the call site and rename this function something more descriptive of its main function, like "regossipAnnounceVersions".

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+	newEntries, err := sb.signedAnnounceVersionTable.Upsert(entries)
+	if err != nil {
+		logger.Warn("Error in upserting entries", "err", err)
+	}
+
+	// If this node is a validator (checked later as a result of this call) and it receives a signed announce
+	// version from a remote validator that is newer than the remote validator's
+	// version in the val enode table, this node did not receive a direct announce
+	// and needs to announce its own enode to the remote validator.
+	sb.startGossipAnnounceTask()
```
nategraf: Given the 5 minute DDoS cooldown, and the fact that this announcement may contain many more enodes than just the ones that changed, I dont think we should trigger this here.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+
+	// Only regossip entries that are new to this node and do not originate
+	// from an address that we have gossiped a signed announce version for
+	// within the last 5 minutes
+	var signedAnnVersionsToRegossip []*signedAnnounceVersion
+	sb.lastSignedAnnounceVersionsGossipedMu.Lock()
+	for _, entry := range newEntries {
+		lastGossipTime, ok := sb.lastSignedAnnounceVersionsGossiped[entry.Address]
+		if !ok || time.Since(lastGossipTime) >= 5*time.Minute {
```
nategraf: This is a key piece of codee and I feel its a little burried in here. We should extract this constant to the top of file, possibly even making it configurable, and make it more obvious that the DDoS protection is implemented here. Additionally, it would be more ideomatic if this was an `if ... { continue }` to decrease the indent level.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
-		sb.cachedAnnounceMsgsMu.RUnlock()
 	}
+	sb.lastSignedAnnounceVersionsGossipedMu.Unlock()
+	if len(signedAnnVersionsToRegossip) > 0 {
+		return sb.gossipSignedAnnounceVersionsMsg(signedAnnVersionsToRegossip)
+	}
+	return nil
+}
+
+func (sb *Backend) updateAnnounceVersion() {
```
nategraf: This function appears to be unused

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+	sb.updateAnnounceVersionCh <- struct{}{}
+	<-sb.updateAnnounceVersionCompleteCh
+}
 
-	if len(announcesToRequest) > 0 {
-		announcesToRequestBytes, err := rlp.EncodeToBytes(announcesToRequest)
+// setAndShareUpdatedAnnounceVersion generates a new enode certificate message and sends
```
nategraf: Nit: I don't think this comment is up to date with the implementation.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+// it to this node's proxy (if applicable).
+// TODO: When a generated announce no longer creates a new version, the version
+// parameter can be removed and instead have the version generated in this function.
+// Tracked here: https://github.com/celo-org/celo-monorepo/issues/2668
+func (sb *Backend) setAndShareUpdatedAnnounceVersion(version uint) error {
+	logger := sb.logger.New("func", "updateAnnounceVersion")
```
nategraf: Nit: Looks like the logger context didn't get updated

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+	// Send new versioned enode msg to all other registered or elected validators
+	validatorConnSet, err := sb.retrieveValidatorConnSet()
+	if err != nil {
+		return err
+	}
+	// Don't send any messages if this node is not registered or elected
+	if !validatorConnSet[sb.Address()] {
```
nategraf: This seems to be the same idea, but slightly different than the `shouldGenerateAndProcessAnnounce` function. Consider using that function or the same implementation here.

consensus/istanbul/backend/announce.go
```diff
@@ -589,130 +649,311 @@ func (sb *Backend) regossipAnnounce(msg *istanbul.Message, payload []byte) error
...
+		return err
 	}
 
-	return nil
+	// Generate and gossip a new signed announce version
+	newSignedAnnVersion, err := sb.generateSignedAnnounceVersion(version)
+	if err != nil {
+		return err
+	}
+	return sb.upsertSignedAnnounceVersionEntries([]*vet.SignedAnnounceVersionEntry{
```
nategraf: `sb.upsertSignedAnnounceVersionEntries` has quite a few side effects and it is unclear if those are intended here.

consensus/istanbul/backend/backend.go
```diff
@@ -110,10 +110,11 @@ func New(config *istanbul.Config, db ethdb.Database) consensus.Istanbul {
 		selfRecentMessages:                 selfRecentMessages,
 		announceThreadWg:                   new(sync.WaitGroup),
 		announceThreadQuit:                 make(chan struct{}),
+		generateAndGossipAnnounceCh:        make(chan struct{}, 1),
 		updateAnnounceVersionCh:            make(chan struct{}),
 		updateAnnounceVersionCompleteCh:    make(chan struct{}),
 		lastAnnounceGossiped:               make(map[common.Address]time.Time),
-		cachedAnnounceMsgs:                 make(map[common.Address]*announceMsgCachedEntry),
+		lastSignedAnnounceVersionsGossiped: make(map[common.Address]time.Time),
```
nategraf: Nit: Consider renaming this to something like "signedAnnounceVersionsRegossipCooldown" to be more descriptive (and the same comment for the other similar variable on line 201)

consensus/istanbul/backend/backend.go
```diff
@@ -299,7 +307,13 @@ func (sb *Backend) Address() common.Address {
 // Close the backend
 func (sb *Backend) Close() error {
 	sb.delegateSignScope.Close()
-	return sb.valEnodeTable.Close()
+	if err := sb.valEnodeTable.Close(); err != nil {
```
nategraf: We should still attempt to close the other table if the first fails. Returning the concatenation of the two error strings would be reasonable if both fail.

consensus/istanbul/backend/backend.go
```diff
@@ -867,6 +881,15 @@ func (sb *Backend) retrieveValidatorConnSet() (map[common.Address]bool, error) {
 	return sb.cachedValidatorConnSet, nil
 }
 
+// retrieveCachedValidatorConnSet returns the most recently cached validator conn
```
nategraf: Please update this comment to be accurate. Also, I noticed that the only way this cache is updated now is a call to `updateCachedValidatorConnSet`. I think this is fine, but I wanted to make sure it is intentional.

consensus/istanbul/backend/announce.go
tkporter: I added some comments to make it more clear, as well as a comment in `updateAnnounceVersion` to show that it must be called from a separate goroutine. Does that help clarify what's happening?

consensus/istanbul/backend/announce.go
tkporter: I recently changed the pruning to happen every 10 minutes, so I think this is a little less of a worry. Given that this isn't very critical would you be okay with treating this as a future optimization in a later PR?

consensus/istanbul/backend/announce.go
tkporter: Good point, changed this to rely on `announceVersion` in the `announceThread`

consensus/istanbul/backend/announce.go
tkporter: Holding off on this for the time being-- Kevin and I were thinking of storing the ecdsa public key in the db because it can be expensive to recover

consensus/istanbul/backend/announce.go
tkporter: It's used in `addProxy` and `StartValidating`

consensus/istanbul/backend/announce.go
```diff
@@ -194,12 +143,28 @@ func (sb *Backend) announceThread() {
...
+				// Regardless, send the announce so that it will at least be
+				// processed by this node's peers. This is helpful when a network
+				// is first starting up.
+				success, err := sb.generateAndGossipAnnounce()
 				if err != nil {
 					logger.Warn("Error in generating and gossiping announce", "err", err)
 				}
+				// If a retry hasn't been scheduled already by a previous announce,
+				// schedule one.
+				if success && announceRetryTimer != nil {
```
nategraf: Directly, I think the second condition is flipped and is intended to be `announceRetryTimer == nil`. More generally, I think this logic is problematic because we should continue to periodically announce until we obtain that peers enode, either through their announcement or a directly shared enode certificate. If we stop either when the announce is multicasted, or when an error occurs in multicasting, we may not reach them and its possible to get stuck in a state where a pair of validators are ignorant of each others enode and not announcing, which would be bad. I do like the intention here, of shutting down this timer while there are no announcements to make. We could accomplish this by checking for the condition that that the generated announce message is empty. If it is empty, then there is no announcement to be made and as long as we restart the timer when there is a new announcement to be made, this will be a good way to increase the responsiveness to changes.

consensus/istanbul/backend/announce.go
```diff
@@ -587,8 +559,16 @@ func (sb *Backend) answerAnnounceMsg(address common.Address, node *enode.Node, v
...
-func (sb *Backend) validateAnnounce(announceData *announceData) (bool, error) {
-	logger := sb.logger.New("func", "validateAnnounce")
+func (sb *Backend) validateAnnounce(msgAddress common.Address, announceData *announceData) (bool, error) {
+	logger := sb.logger.New("func", "validateAnnounce", "msg address", msgAddress)
+
+	// Ensure the version in the announceData is not older than the signed
+	// announce version known by this node
+	knownVersion, err := sb.signedAnnounceVersionTable.GetVersion(msgAddress)
+	if err == nil && knownVersion < announceData.Version {
+		logger.Info("Announce message has version older than known version from signed announce table", "msg version", announceData.Version, "known version", knownVersion)
```
nategraf: Log level should not be info. If we believe this is a problematic case, it should be warning or if it is not, it should be debug.

consensus/istanbul/backend/announce.go
nategraf: Looking at it again, it makes more sense 👍

consensus/istanbul/backend/announce.go
nategraf: This change is as simple as `if time.Since(sb.lastAnnounceGossiped[remoteAddress]) < signedAnnounceVersionGossipCooldownDuration { continue }` so we should esither include this line now or not plan on it. I agree that it's not so important, so if you disagree with my reasoning, feel free to ignore it.

tkporter: nvm I'll just add it

consensus/istanbul/backend/announce.go
tkporter: Just renamed it to `upsertAndGossipSignedAnnounceVersionEntries` and moved `startGossipAnnounceTask` out of there, hope that makes it clearer.

consensus/istanbul/backend/announce.go
tkporter: I think it'd be nice to trigger an announce task the moment we become aware of an inconsistency between the signed announce version table and the validator enode table. Where else would you suggest? Will think about this more

nategraf:
I am seeing an error in RC0 logs that needs some thought. At [consensus/istanbul/backend/backend.go:924](https://github.com/celo-org/celo-blockchain/blob/1dd42191b954d5c2037327c9261ec98a9966cc4e/consensus/istanbul/backend/backend.go#L924), we query the chain for the validators that are winning or almost winning the current election to include in the connection set so that if they get elected, block production will continue gracefully.

One place this is not handled is if not enough validators are electable and the call reverts. In that case only the current validator set will be allowed to connect until the call succeeds. If this call went from failing to successful close to an epoch edge, the newly elected validators may not be in the connection set for long enough, and in the worst case, block production could stall until they validators connect to each other. The case where this is possible is the first election after standing up the network.

One solution to this would be to call `electNValidatorSigners` with the minimum electable as 0. This way, the frontrunners could always connect to the validator overlay network, even if there are not yet enough to form a quorum.

consensus/istanbul/backend/announce.go
tkporter: Great point. I'll make use of the cooldown period. I don't think it's safe to "answer" the announce msg and then always refuse to answer any later requests in the event that the answer is received by the querying validator (or needs an answer again in the future for some reason)

consensus/istanbul/backend/announce.go
```diff
@@ -516,59 +511,61 @@ func (sb *Backend) handleAnnounceMsg(peer consensus.Peer, payload []byte) error
...
+			}
+			enodeURL := string(enodeBytes)
+			node, err := enode.ParseV4(enodeURL)
+			if err != nil {
+				logger.Warn("Error parsing enodeURL", "enodeUrl", enodeURL)
+				return err
+			}
+			sb.lastAnnounceAnsweredMu.Lock()
+			// Don't answer an announce message that's been answered too recently
+			if lastAnswered, ok := sb.lastAnnounceAnswered[msg.Address]; !ok || time.Since(lastAnswered) < announceAnswerCooldownDuration {
```
nategraf: Why is the rate limit applied here instead of in `answerAnnounce`? It seems to be that we should always upsert, even if we don't send out a certificate. I would recommend moving the cooldown logic into `answerAnnounce`.

Its also a little unclear to me why we need a cooldown for this action because the response is to send an `enodeCertificate` (a fairly small message) directly back to the sender of this message, which does not risk flooding the network. If a cooldown is needed, 5 minutes seems extreme, especially because a number of scenarios can result in receiving an announce more frequently. I would suggest removing this cooldown or substantially reducing it (to something to 10 - 30 seconds)

consensus/istanbul/backend/announce.go
```diff
@@ -618,144 +621,216 @@ func (sb *Backend) validateAnnounce(announceData *announceData) (bool, error) {
...
+		logger.Debug("Last announce gossiped not found")
+		sb.lastAnnounceGossipedMu.RUnlock()
+		return nil
+	}
+	scheduledTime := lastGossip.Time.Add(announceGossipCooldownDuration)
+	sb.lastAnnounceGossipedMu.RUnlock()
+
+	duration := time.Until(scheduledTime)
+	// If by this time the cooldown period has ended, just regossip
+	if duration < 0 {
```
nategraf: Nit: Might as well be `<=`
```suggestion
	if duration <= 0 {
```

consensus/istanbul/backend/announce.go
```diff
@@ -842,37 +918,37 @@ func (sb *Backend) handleSignedAnnounceVersionsMsg(peer consensus.Peer, payload
...
 	for _, entry := range newEntries {
 		lastGossipTime, ok := sb.lastSignedAnnounceVersionsGossiped[entry.Address]
-		if !ok || time.Since(lastGossipTime) >= 5*time.Minute {
-			signedAnnVersionsToRegossip = append(signedAnnVersionsToRegossip, &signedAnnounceVersion{
-				Address: entry.Address,
-				Version: entry.Version,
-				Signature: entry.Signature,
-			})
-			sb.lastSignedAnnounceVersionsGossiped[entry.Address] = time.Now()
+		if ok && time.Since(lastGossipTime) >= signedAnnounceVersionGossipCooldownDuration && entry.Address != sb.ValidatorAddress() {
```
nategraf: Should the inequality be inverted?
```suggestion
		if ok && time.Since(lastGossipTime) < signedAnnounceVersionGossipCooldownDuration && entry.Address != sb.ValidatorAddress() {
```

consensus/istanbul/backend/announce.go
```diff
@@ -1067,27 +1149,39 @@ func (sb *Backend) handleEnodeCertificateMsg(peer consensus.Peer, payload []byte
...
+				}
+				return nil
+			}
+		} else {
+			// If this message is not from the proxied validator, send it to the
+			// proxied validator without upserting it in this node. If the validator
+			// decides this proxy should upsert the enodeCertificate, then it
+			// will send it back to this node.
+			if err := sb.sendEnodeCertificateMsg(sb.proxiedPeer, &msg); err != nil {
+				logger.Warn("Error forwarding enodeCertificate to proxied validator", "err", err)
```
nategraf: I would argue this should be raised to an error because the result of a failure to send this message is that this proxy will not consider the peer a validator.
```suggestion
				logger.Error("Error forwarding enodeCertificate to proxied validator", "err", err)
```

consensus/istanbul/backend/announce.go
```diff
@@ -558,161 +619,421 @@ func (sb *Backend) validateAnnounce(announceData *announceData) (bool, error) {
...
 	}
-	sb.lastAnnounceGossipedMu.RUnlock()
 
 	logger.Trace("Regossiping the istanbul announce message", "IstanbulMsg", msg.String())
 	if err := sb.Multicast(nil, payload, istanbulAnnounceMsg); err != nil {
 		return err
 	}
 
-	sb.lastAnnounceGossipedMu.Lock()
-	defer sb.lastAnnounceGossipedMu.Unlock()
```
nategraf: I seems like we still need this lock. Writing to this array from multiple threads can cause a panic.

PR #897, opened 2020-03-09
Update BLS Zexe to latest

gakonst:
We have been recently updating the BLS Library to support latest Zexe, as well as extend its features by incorporating the BLS SNARK functionality for the ultralight sync via the epochs.

A few things were renamed, some directories were moved around, a lot of files were removed, so this PR takes care of these changes.

- Bumps rust to 1.41
- Updates BLS subtree (removes the Zexe subtree dependency altogether)
- Allows consumers to use the `BLS_RS_PATH` if they want to build the chain with their local dev copy of https://github.com/celo-org/bls-zexe

Ideally, I'd like to remove the subtree dependency at `crypto/bls/bls-zexe` altogether, but my understanding is that you prefer subtrees than having to push a `git clone && git checkout $REV` to whoever builds from source


.circleci/config.yml
```diff
@@ -10,19 +10,19 @@ executors:
     working_directory: ~/repos/celo-monorepo/packages/celotool
     environment:
       GO_VERSION: "1.13.7"
-      CELO_MONOREPO_BRANCH_TO_TEST: master
+      CELO_MONOREPO_BRANCH_TO_TEST: kobigurk/fix_hash_to_group
```
asaj: Can we be sure to fix up https://github.com/celo-org/celo-monorepo/pull/2977 before we merge this?

gakonst: Yeah if this CI passes, we'll merge the branch on the monorepo, point CI back to master and merge.

Issue #558, opened 2019-10-29
Validators should select proposer according to an language independent shuffle


nategraf:
### Expected Behavior

Validators should use the order of validators returned by the Election contract to determine the order for round robin proposer selection.

### Current Behavior

The random seed from the Random contract is used to seed a Golang implemented randomness source, which is used to create the permutation that is used.

asaj:
Do we need to do this? What if we kept things as-is and stopped shuffling in `Election.sol`

nategraf:
The origonal thought when opening this issue is that implementing the shuffle in the EVM will be more portable so it can be potentially adjusted in the future without hard fork and is kinder to any future non-Go implementations of the Celo node (as the current implementation uses the Golang specific random number generator). If we decide we don't care as much about ability to change the shuffle through governance, we can replace the Golang random number generator with something that could be reproduced in another language.

nategraf:
It is currently unclear if we will be making this change based on further discussion.

nategraf:
A step that will be simple to do now and well worth it in the long term is to make the shuffle independent of `math/rand`. Currently an update to that library would result in validators on different versions not being interoperable. This is a bad oversight, but should be easy to fix now. https://github.com/celo-org/celo-blockchain/blob/1c42b5fc115f0dfd58b8f72762098f0caf632ac5/consensus/istanbul/validator/selectors.go#L42

PR #898, opened 2020-03-10
Use language-independent random permutation generator for validator set

nategraf:
### Description

Validator ordering is a consensus issue (i.e. unless a quorum validators do not agree who the proposer is, consensus will not be reached) therefore it is important for that ordering to be stable across different version of the software, and potentially different implementations should another Celo client come to exist in the future.

Currently the client uses the Golang implementation of a random number generator to determine validator ordering. This is fragile because if at any point in the future, Golang changes it's random number algorithm, upgrading to that new version of Golang will be a hard-fork change. Additionally, this would be a blocker for new clients trying to implement the protocol.

This PR implements a simple permutation generator that is under our control and could be implemented in any language to prevent the risk of needing a hard-fork at some point in the future.

### Tested

Included tests for basic functionality, and ran manual distribution checks to make sure all permutations are equally likely.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/558

### Backwards compatibility

Validators from a pervious version will not reach consensus with one after this PR.


consensus/istanbul/validator/random/random.go
```diff
@@ -0,0 +1,60 @@
...
+// compress produces a 64-bit random value from a 256-bit random value.
+func compress(value common.Hash) uint64 {
+	var compressed uint64 = 0
+	for i := 0; i < common.HashLength; i += 8 {
+		compressed ^= binary.BigEndian.Uint64(value[i : i+8])
+	}
+	return compressed
+}
+
+// uniform produces a random integer in the range [0, k) from the provided randomness.
```
nategraf:
```suggestion
// uniform produces an integer in the range [0, k) from the provided randomness.
```

consensus/istanbul/validator/random/random.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Create the unshuffled array [0, 1, ... n-1]
+	array := make([]int, n)
+	for i := 0; i < n; i++ {
+		array[i] = i
+	}
+
+	// Shuffle the array using the Fisher-Yates method.
+	for i := 0; i < n-1; i++ {
+		randomness = sha3.Sum256(append(randomness[:], permutationSalt))
```
mstraka100: This might be fine but it's going to be cleaner and more secure to use a stream cipher here like salsa20: https://godoc.org/golang.org/x/crypto/salsa20#XORKeyStream
You can just use the randomness passed into each function as a key along with a nonce that you increment as a counter, with the input being all 0s or some fixed seed phrase

consensus/istanbul/validator/random/random.go
```diff
@@ -0,0 +1,60 @@
...
+func compress(value common.Hash) uint64 {
+	var compressed uint64 = 0
+	for i := 0; i < common.HashLength; i += 8 {
+		compressed ^= binary.BigEndian.Uint64(value[i : i+8])
+	}
+	return compressed
+}
+
+// uniform produces an integer in the range [0, k) from the provided randomness.
+// Based on Algorithm 4 of https://arxiv.org/pdf/1805.10941.pdf
```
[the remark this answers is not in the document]
nategraf: My understanding is that the randomness must be uniform in `[0, 2^L)` (in this case L=256), but `k` shouldn't need to be a power of two and if it is, that's problematic.

consensus/istanbul/validator/random/random.go
nategraf: Switched to using SHAKE 256 in 8aca95d, much nicer!

consensus/istanbul/validator/random/random.go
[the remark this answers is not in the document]
mstraka100: Ah you're right

consensus/istanbul/validator/random/random.go
mstraka100: SHAKE256 is fine here but will be slower than a stream cipher if efficiency is a concern. Interface is nice though

PR #913, opened 2020-03-11
Delete unused headers, ethash, clique, puppeth

alecps:
### Description

### Tested

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

_Brief explanation of why these changes are/are not backwards compatible. Feel free to remove for trivial, backwards compatible changes._


light/lightchain_test.go
```diff
@@ -242,27 +241,6 @@ func TestBrokenHeaderChain(t *testing.T) {
 	}
 }
 
-func makeHeaderChainWithDiff(genesis *types.Block, d []int, seed byte) []*types.Header {
```
kevjue: Should this be removed?

miner/worker.go
```diff
@@ -445,45 +438,6 @@ func (w *worker) mainLoop() {
 			}
 			w.commitNewWork(req.interrupt, req.noempty, req.timestamp)
 
-		case ev := <-w.chainSideCh:
```
kevjue: Instead of removing this whole case, can you still include it, but replace the case's block with an error log for now.   I think we should dig deeper on the chainsidech and see if that can be removed altogether.

miner/worker_test.go
```diff
@@ -171,9 +168,6 @@ func newTestWorkerBackend(t *testing.T, chainConfig *params.ChainConfig, engine
 	if n > 0 {
 		parent = chain.GetBlockByHash(chain.CurrentBlock().ParentHash())
 	}
-	blocks, _ := core.GenerateChain(chainConfig, parent, engine, db, 1, func(i int, gen *core.BlockGen) {
```
kevjue: Should this be removed?

contract_comm/gasprice_minimum/gasprice_minimum.go
```diff
@@ -137,12 +138,14 @@ func GetGasPriceMinimum(currency *common.Address, header *types.Header, state vm
 func UpdateGasPriceMinimum(header *types.Header, state vm.StateDB) (*big.Int, error) {
 	var updatedGasPriceMinimum *big.Int
 
+	gasLimit, _ := blockchain_parameters.GetBlockGasLimit(header, state)
```
nategraf: Ignoring an error is almost always forbidden. I would say it's reasonable to do here, but I'd add a comment.
```suggestion
	// If an error occurs, the default block gas limit will be returned and a log statement will be produced by contract_comm
	gasLimit, _ := blockchain_parameters.GetBlockGasLimit(header, state)
```

core/blockchain_test.go
```diff
@@ -17,7 +17,6 @@
 package core
```
nategraf: I am not quite comfortable with some of the tests that are eliminated here. Verifying that the code works correctly in the case of forks and reorgs is important even though we don't expect it to happen unless something has gone wrong.

miner/worker_test.go
alecps: My interpretation is that `blocks` is only used to pass an uncle block to the struct below. Since I've deleted that field from the struct, go gives me an error because `blocks` is defined and not used. Is there something I might be missing?

params/config.go
```diff
@@ -67,7 +67,10 @@ var (
 		ConstantinopleBlock: big.NewInt(7280000),
 		PetersburgBlock:     big.NewInt(7280000),
 		IstanbulBlock:       big.NewInt(9069000),
-		Ethash:              new(EthashConfig),
+		Istanbul: &IstanbulConfig{
+			Epoch:          30000,
```
nategraf, line 71: Nit: If it doesn't break anything, 17280 would be the more canonical default value.
```suggestion
			Epoch:          17280,
```

params/config.go
```diff
@@ -106,7 +109,10 @@ var (
 		ConstantinopleBlock: big.NewInt(4230000),
 		PetersburgBlock:     big.NewInt(4939394),
 		IstanbulBlock:       big.NewInt(6485846),
-		Ethash:              new(EthashConfig),
+		Istanbul: &IstanbulConfig{
+			Epoch:          30000,
```
nategraf, line 113: Nit: If it doesn't break anything, 17280 would be the more canonical default value.
```suggestion
			Epoch:          17280,
```

signer/fourbyte/4byte.go
```diff
@@ -4246,7 +4246,6 @@ var __4byteJson = []byte(`{
 "0754e744": "icofunding()",
```
nategraf, line 4246: This file should not change. Please check out the version from `master`

signer/fourbyte/4byte.json
```diff
@@ -114081,7 +114081,6 @@
 "c7238af7": "escapeDestination()",
```
nategraf, line 114081: This file should not change. Please check out the version from `master`

tests/solidity/contracts/OpCodes.sol
```diff
@@ -193,12 +193,6 @@ contract OpCodes {
      //number
      assembly {  pop(number())}
 
-     //difficulty
-     assembly {  pop(difficulty())}
-
-     //gaslimit
-     assembly {  pop(gaslimit())}
-
```
nategraf, line 201: Op code change we discussed. Just wanted to tag this one to make sure it doesn't get mnissed.

core/block_validator.go
```diff
@@ -46,26 +46,19 @@ func NewBlockValidator(config *params.ChainConfig, blockchain *BlockChain, engin
...
-		return err
-	}
-	if hash := types.CalcUncleHash(block.Uncles()); hash != header.UncleHash {
-		return fmt.Errorf("uncle root hash mismatch: have %x, want %x", hash, header.UncleHash)
-	}
 	if hash := types.DeriveSha(block.Transactions()); hash != header.TxHash {
 		return fmt.Errorf("transaction root hash mismatch: have %x, want %x", hash, header.TxHash)
 	}
-	if !v.bc.HasBlockAndState(block.ParentHash(), block.NumberU64()-1) {
+	if !v.bc.HasBlockAndState(block.ParentHash(), block.NumberU64()-1) { // NOTE(Alec): This is never entering.
```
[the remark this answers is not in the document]
alecps: This was just a note I left for myself while debugging. I've removed it

PR #915, opened 2020-03-12
Benchmarking precompiles, setting other parameters

### LINKED ISSUE #901: Finalize blockchain parameters
Finalize blockchain parameters for
* Hardforks
* protocol_params.go

mrsmkl:
### Description

Hopefully the precompile gas costs are now good enough. Many of them have basically no cost (calling precompiles costs anyway). The cost of verifying aggregated seals should be checked later when we have a new network running.

### Related issues

- Fixes #901

### DIFF
--- a/params/protocol_params.go
+++ b/params/protocol_params.go
@@ -138,14 +138,13 @@ const (
 	// Celo precompiled contracts
 	// TODO: make this cost variable- https://github.com/celo-org/geth/issues/250
 	FractionMulExpGas uint64 = 1050 // Cost of performing multiplication and exponentiation of fractions to an exponent of up to 10^3.
-	// TODO(kobigurk):  Figure out what the actual gas cost of this contract should be.
-	ProofOfPossessionGas        uint64 = 50000 // Cost of verifying a BLS proof of possession.
-	GetValidatorGas             uint64 = 5000  // Cost of reading a validator's address.
-	GetEpochSizeGas             uint64 = 1000  // Cost of querying the number of blocks in an epoch.
-	GetBlockNumberFromHeaderGas uint64 = 10000 // Cost of decoding a block header.
-	HashHeaderGas               uint64 = 20000 // Cost of hashing a block header.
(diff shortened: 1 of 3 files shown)


params/protocol_params.go
```diff
@@ -136,16 +136,16 @@ const (
...
-	// TODO(kobigurk):  Figure out what the actual gas cost of this contract should be.
-	ProofOfPossessionGas        uint64 = 50000 // Cost of verifying a BLS proof of possession.
-	GetValidatorGas             uint64 = 5000  // Cost of reading a validator's address.
-	GetEpochSizeGas             uint64 = 1000  // Cost of querying the number of blocks in an epoch.
-	GetBlockNumberFromHeaderGas uint64 = 10000 // Cost of decoding a block header.
-	HashHeaderGas               uint64 = 20000 // Cost of hashing a block header.
-	GetParentSealBitmapGas      uint64 = 500   // Cost of reading the parent seal bitmap from the chain.
-	GetVerifiedSealBitmapGas    uint64 = 55000 // Cost of verifying the seal on a given RLP encoded header.
+	FractionMulExpGas           uint64 = 50     // Cost of performing multiplication and exponentiation of fractions to an exponent of up to 10^3.
+	FractionMulExpExtraGas      uint64 = 1050   // Cost for extra bits
```
nategraf: This appears to be unused. If this is not a mistake, go ahead and delete this constant.
```suggestion
```

params/protocol_params.go
```diff
@@ -136,16 +136,16 @@ const (
...
+	ProofOfPossessionGas        uint64 = 350000 // Cost of verifying a BLS proof of possession.
+	GetValidatorGas             uint64 = 10     // Cost of reading a validator's address.
```
nategraf: In the current implementation, this may require a disk read for historical queries. We don't need to be precise here, partly because I anticipate optimizing this function in the future, but we should make this a bit higher.
```suggestion
	GetValidatorGas             uint64 = 1000     // Cost of reading a validator's address.
```

nategraf:
Thanks for getting this done!

PR #924, opened 2020-03-15
bls-zexe: Removes the "plus one" logic from max non signers and updates to latest version

kobigurk:
The epoch encoding previously contained the "aggregated public key". We've since realised that it doesn't have to be included - the SNARK just computed it and outputs it. This saves a nice amount of constraints.

This is NOT backwards compatible.


asaj:
Didn't review the rust code

kobigurk:
> Didn't review the rust code

That's fine, it was reviewed as part of a bls-zexe PR!

PR #925, opened 2020-03-16
Removes the version check for the ledger, it supports the current geth

kobigurk:


accounts/usbwallet/ledger.go
```diff
@@ -163,10 +163,6 @@ func (w *ledgerDriver) SignTx(path accounts.DerivationPath, tx *types.Transactio
 	if w.offline() {
 		return common.Address{}, nil, accounts.ErrWalletClosed
 	}
-	// Ensure the wallet is capable of signing the given transaction
```
asaj: Should we keep this and just set the minimum version number to 1.0.0?

kobigurk: Yeah, sure!

kobigurk: Actually, on second thought, the check is quite useless. How about I leave it as a comment?

PR #927, opened 2020-03-17
BLS SNARK refactor

kobigurk:
SNARK API is implemented and includes a refactor to the BLS library.

This is backwards compatible.


Makefile
```diff
@@ -100,8 +100,9 @@ else
 ifeq ("$(CARGO_LIPO_exists)","")
 	cargo install cargo-lipo
 endif
-	rustup target add aarch64-apple-ios armv7-apple-ios x86_64-apple-ios i386-apple-ios
-	cd $(BLS_RS_PATH) && cargo lipo --release --targets=aarch64-apple-ios,armv7-apple-ios,x86_64-apple-ios,i386-apple-ios
+	rustup toolchain install 1.41.0
```
gakonst: Who is doing the cross-compilation? Might be worth adding a big note in the README that "if you're compiling to IOS you need at most Rust 1.41, for $XYZ reasons"

PR #931, opened 2020-03-19
Set FullHeaderChainAvailable according to sync mode

mrsmkl:
### Description

`FullHeaderChainAvailable` was false by default, this might have caused some erroneous behavior.

### Tested

Only manual testing.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1861,9 +1861,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node) (chain *core.BlockChain, chai
 	var err error
 	chainDb = MakeChainDatabase(ctx, stack)
 	config, _, err := core.SetupGenesisBlock(chainDb, MakeGenesis(ctx))
-	if ctx.GlobalString(SyncModeFlag.Name) == "lightest" {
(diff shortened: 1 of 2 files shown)


cmd/utils/flags.go
```diff
@@ -1861,9 +1861,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node) (chain *core.BlockChain, chai
 	var err error
 	chainDb = MakeChainDatabase(ctx, stack)
 	config, _, err := core.SetupGenesisBlock(chainDb, MakeGenesis(ctx))
-	if ctx.GlobalString(SyncModeFlag.Name) == "lightest" {
-		config.FullHeaderChainAvailable = false
-	}
+	config.FullHeaderChainAvailable = ctx.GlobalString(SyncModeFlag.Name) != "lightest"
 	if err != nil {
 		Fatalf("%v", err)
 	}
```
nategraf: This error check should have been next to where the error is produced.
```suggestion
	if err != nil {
		Fatalf("%v", err)
	}
	config.FullHeaderChainAvailable = ctx.GlobalString(SyncModeFlag.Name) != "lightest"
```

PR #936, opened 2020-03-23
Use metrics registry to prevent unsafe map access

nategraf:
### Description

`contract_comm` uses a map of registered timers to record metrics for EVM calls. Prior this this PR, the map is used in a thread-unsafe way which can result in a panic if two threads make a call system call simultaneously.

When writing this function originally I overlooked the built in `metrics.Registry` which is designed for concurrent access. This PR uses the registry to avoid this issue.

### Tested

e2e tests

### Backwards compatibility

100%


nategraf:
We do not! Thanks, I removed it in 0a37918

PR #938, opened 2020-03-25
Trace transaction's primary evm.Call()

jfoutts-celo:
### Description

Normally `TransitionDb()` makes one `evm.Call()`, but when a transaction `feeCurrency` is set, it makes three. The last `evm.Call()` (which credits back gas fees) overwrites the debugging trace for the transaction's primary `evm.Call()`.

This PR updates `debitGas` and `creditGasFees` to disable debugging for their calls.

### Tested

Tested that traces the transaction's primary EVM call.


jfoutts-celo:
mariano:  5 days ago
qq, do we still get the log events from this transactions?

jfoutts-celo:
> mariano: 5 days ago
> qq, do we still get the log events from this transactions?

Yes.  From logs we can tell how much cUSD was deducted for gas fees, and everyone it went to.

PR #939, opened 2020-03-27
Cancel istanbul gas changes

mrsmkl:
### Description

This removes EIP 1884 gas changes from the Istanbul hard fork.

### DIFF
--- a/core/vm/eips.go
+++ b/core/vm/eips.go
@@ -45,6 +45,8 @@ func EnableEIP(eipNum int, jt *JumpTable) error {
 // - Define SELFBALANCE, with cost GasFastStep (5)
 func enable1884(jt *JumpTable) {
 	// Gas cost changes
+	// Celo does not include these gas changes at genesis.
+	// It is planned to apply them at a later date.
 	// jt[BALANCE].constantGas = params.BalanceGasEIP1884
 	// jt[EXTCODEHASH].constantGas = params.ExtcodeHashGasEIP1884
 	// jt[SLOAD].constantGas = params.SloadGasEIP1884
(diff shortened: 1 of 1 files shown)


core/vm/eips.go
```diff
@@ -46,9 +45,9 @@ func EnableEIP(eipNum int, jt *JumpTable) error {
 // - Define SELFBALANCE, with cost GasFastStep (5)
 func enable1884(jt *JumpTable) {
 	// Gas cost changes
-	jt[BALANCE].constantGas = params.BalanceGasEIP1884
-	jt[EXTCODEHASH].constantGas = params.ExtcodeHashGasEIP1884
-	jt[SLOAD].constantGas = params.SloadGasEIP1884
+	// jt[BALANCE].constantGas = params.BalanceGasEIP1884
+	// jt[EXTCODEHASH].constantGas = params.ExtcodeHashGasEIP1884
```
nategraf: We should include these changes unless they also break things. My understanding is that they don't.

core/vm/eips.go
```diff
@@ -46,9 +45,9 @@ func EnableEIP(eipNum int, jt *JumpTable) error {
...
+	// jt[SLOAD].constantGas = params.SloadGasEIP1884
```
nategraf: Adding a comment for any confused future readers.
```suggestion
	
	// Celo does not include this gas change at genesis, where the rest of EIP1884 is applied.
	// It is planned to be applied at a later date.
	// jt[SLOAD].constantGas = params.SloadGasEIP1884
```

core/vm/eips.go
mrsmkl: Well they might break stuff similarly to how SLOAD does, but probably not in the case we detected (transfer to governance).

nategraf: Good call. I agree being on the safe side of this is probably better. 👌

PR #941, opened 2020-03-27
Clean up loose ends for new Baklava network client

nategraf:
### Description

Getting ready for the new Baklava launch I noticed a number of small but important loose ends that need to be accounted for. This PR includes:

* Change default gateway fee to 0 so that it is default off
* Change gateway fee processing to accept any recipient if the minimum fee is zero
* Change the client version to 0.9.0 to mark the first pre-release Celo version
* Prevent log spam from the reverting `getAddressFor` method before contracts are deployed
* Add a INFO or WARN log message when a static or trusted peer disconnects to catch cases where a proxy or validator drops their connection.
* Add a buffer to Istanbul chain head subscription to prevent possibly blocking the chain
* Refactor ethstats listener to ensure it never blocks subscriptions

### Tested

Existing unit tests and e2e tests as well as manual examination of the logs

### Backwards compatibility

100%

### DIFF
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -190,6 +190,7 @@ func (s *Service) loop() {
 
 	txEventCh := make(chan core.NewTxsEvent, txChanSize)
 	txSub := txpool.SubscribeNewTxsEvent(txEventCh)
+	defer txSub.Unsubscribe()
 
 	istDelegateSignCh := make(chan istanbul.MessageEvent, istDelegateSignChanSize)
 	istDelegateSignSub := s.backend.SubscribeNewDelegateSignEvent(istDelegateSignCh)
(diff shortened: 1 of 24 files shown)


ethstats/ethstats.go
```diff
@@ -233,15 +236,29 @@ func (s *Service) loop() {
...
+
+	go func() {
+		for {
+			select {
 			case delegateSignMsg := <-istDelegateSignCh:
 				var statsPayload StatsPayload
 				err := json.Unmarshal(delegateSignMsg.Payload, &statsPayload)
 				if err != nil {
-					break HandleLoop
+					log.Error("Error in unmarshalling selegate sign payload", "err", err, "payload", delegateSignMsg.Payload)
```
timmoreton: s/selegate/delegate

params/version.go
```diff
@@ -21,10 +21,10 @@ import (
...
 
 const (
-	VersionMajor = 1        // Major version component of the current release
-	VersionMinor = 9        // Minor version component of the current release
-	VersionPatch = 8        // Patch version component of the current release
-	VersionMeta  = "stable" // Version metadata to append to the version string
+	VersionMajor = 0  // Major version component of the current release
+	VersionMinor = 9  // Minor version component of the current release
+	VersionPatch = 0  // Patch version component of the current release
+	VersionMeta  = "" // Version metadata to append to the version string
```
timmoreton: geth alternates this between `unstable` (not a released version) and `stable` (a released build whose commit corresponds to a tag) -- this would be nicer because otherwise I can't necessarily know this isn't *the* 1.9.0 build

ethstats/ethstats.go
```diff
@@ -251,16 +268,15 @@ func (s *Service) loop() {
...
-				// TODO: This is a hacky measure to avoid blocking here if the
-				// channel is not consuming
+				// Avoid blocking here if the channel is not consuming
 				select {
 				case channel <- &statsPayload:
 				default:
 				}
-
+			case <-quitCh:
+				return
```
timmoreton: I think this works fine but the indentation is hard to follow - have you tested that it quits cleanly?

params/version.go
nategraf: Got it, so right now it should be "unstable"

ethstats/ethstats.go
nategraf: I haven't tested this yet (I wanted to get the reviews going), but I can do that!

nategraf: Decided to remove this 👍

PR #946, opened 2020-03-30
Fix nil dereference deploying contract with value

jfoutts-celo:
### Description

Fix a nil dereference preventing contracts from being deployed with value (e.g. to payable constructor).

### Tested

Tested contracts can be deployed with value.

### Backwards compatibility

Yes.


core/tx_pool.go
```diff
@@ -609,11 +609,15 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 
 	// Ensure gold transfers are whitelisted if transfers are frozen.
 	if tx.Value().Sign() > 0 {
-		to := *tx.To()
 		if isFrozen, err := freezer.IsFrozen(params.GoldTokenRegistryId, nil, nil); err != nil {
 			log.Warn("Error determining if transfers are frozen, will proceed as if they are not", "err", err)
 		} else if isFrozen {
 			log.Info("Transfers are frozen")
+			if tx.To() == nil {
```
asaj: If `from` is whitelisted the tx should still go through

PR #952, opened 2020-04-01
Modify eth message BlockBodies to return block hashes

kevjue:
### Description

This PR modifies the eth subprotocal message `BlockBodies` message so that it will return the block hashes along with the block bodies.

This will allow the fetcher's `filterBodies` to correctly filter out bodies.

### Tested

Ran the unit tests.

### Related issues

- Fixes #952 

### Backwards compatibility

Not backwards compatible since the `BlockBodies` message format is changed.


eth/peer.go
```diff
@@ -294,11 +294,6 @@ func (p *peer) SendBlockHeaders(headers []*types.Header) error {
 	return p2p.Send(p.rw, BlockHeadersMsg, headers)
 }
 
-// SendBlockBodies sends a batch of block contents to the remote peer.
```
asaj, line 297: Was this unused?

eth/fetcher/fetcher.go
```diff
@@ -247,7 +248,7 @@ func (f *Fetcher) FilterHeaders(peer string, headers []*types.Header, time time.
 
 // FilterBodies extracts all the block bodies that were explicitly requested by
 // the fetcher, returning those that should be handled differently.
-func (f *Fetcher) FilterBodies(peer string, transactions [][]*types.Transaction, randomness []*types.Randomness, epochSnarkData []*types.EpochSnarkData, time time.Time) ([][]*types.Transaction, []*types.Randomness, []*types.EpochSnarkData) {
+func (f *Fetcher) FilterBodies(peer string, headerHashes []common.Hash, transactions [][]*types.Transaction, randomness []*types.Randomness, epochSnarkData []*types.EpochSnarkData, time time.Time) ([]common.Hash, [][]*types.Transaction, []*types.Randomness, []*types.EpochSnarkData) {
```
asaj, line 251: nit: Consider `blockHashes`

eth/fetcher/fetcher_test.go
```diff
@@ -178,19 +178,21 @@ func (f *fetcherTester) makeBodyFetcher(peer string, blocks map[common.Hash]*typ
 	// Create a function that returns blocks from the closure
 	return func(hashes []common.Hash) error {
 		// Gather the block bodies to return
+		headerHashes := make([]common.Hash, 0, len(hashes))
```
asaj, line 181: nit: `blockHashes` if you take the suggestion above

edit: And everywhere else, ofc

eth/peer.go
kevjue: Ya, It's unused.

tkporter:
These changes shouldn't affect les, right?

Related-- I took a look at the way LES gets blocks, and I think it's not vulnerable to the issue we were seeing, but you may want to confirm as well.

It seems like a block header is requested, and each individual block request has an associated BlockRequest struct that's kept track of using a request ID:
https://github.com/celo-org/celo-blockchain/blob/16e1c70f363451b66712a99540bc03716d93c565/les/odr.go#L126

asaj:
> These changes shouldn't affect les, right?

Light clients will fetch block bodies in some cases. I'm not sure if they use the same mechanism to match bodies to headers

kevjue:
> LGTM pending a fix to the `TestGetBlockBodies64` unit test failure

Running into a tricky deadlock issue in the unit tests.  Not quite sure how it was working before these changes ...

PR #956, opened 2020-04-02
Add methods to EthClient

mcortesi:
Add/Modify the following methods:
 * net_listening
 * ExtendedHeaderByXX methods
 * SendRawTransaction

 Additionally:
  * Expose txdata struct outside the pacakge
  * Add RPCTransaction to the rpc module

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -29,21 +29,21 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
-//go:generate gencodec -type txdata -field-override txdataMarshaling -out gen_tx_json.go
+//go:generate gencodec -type Txdata -field-override TxdataMarshaling -out gen_tx_json.go
 
 var (
 	ErrInvalidSig = errors.New("invalid transaction v, r, s values")
 )
 
 type Transaction struct {
-	data txdata
+	data Txdata
 	// caches
 	hash atomic.Value
 	size atomic.Value
 	from atomic.Value
 }
 
-type txdata struct {
+type Txdata struct {
 	AccountNonce        uint64          `json:"nonce"    gencodec:"required"`
 	Price               *big.Int        `json:"gasPrice" gencodec:"required"`
 	GasLimit            uint64          `json:"gas"      gencodec:"required"`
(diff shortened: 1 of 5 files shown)


core/types/transaction.go
```diff
@@ -29,21 +29,21 @@ import (
...
-	data txdata
+	data Txdata
 	// caches
 	hash atomic.Value
 	size atomic.Value
 	from atomic.Value
 }
 
-type txdata struct {
+type Txdata struct {
```
kevjue, line 46: How come `Txdata` needs to be made public?  I don't see it being used outside of the `types` packages in this PR.

core/types/transaction.go
```diff
@@ -63,7 +63,7 @@ type txdata struct {
 	Hash *common.Hash `json:"hash" rlp:"-"`
 }
 
-type txdataMarshaling struct {
+type TxdataMarshaling struct {
```
kevjue, line 66: How come `TxdataMarshaling` needs to be made public? I don't see it being used outside of the `types` packages in this PR.

core/types/transaction.go
mcortesi: @yorhodes can provide more context, but while inspecting the transaction on the client, he ended up requiring this

core/types/transaction.go
mcortesi: Not needed, `txDataMarshalling` is only use a parameter to `gencoded` to modify marshalling rules... I'll change this

rpc/types.go
```diff
@@ -68,6 +68,26 @@ const (
 	EarliestBlockNumber = BlockNumber(0)
 )
 
+type RPCTransaction struct {
```
yorhodes, line 71: This is exported because we have an external txpool client
I wonder if we should bring those network, txpool, admin, debug clients in
cc @mcortesi

mcortesi: maybe... but not on this PR ;)

ethclient/ethclient.go
```diff
@@ -129,6 +149,43 @@ func (ec *Client) getBlock(ctx context.Context, method string, args ...interface
 	return types.NewBlockWithHeader(head).WithBody(txs, body.Randomness, body.EpochSnarkData), nil
 }
 
+type ExtendedHeader struct {
```
kevjue, line 152: nit:  Consider renaming `ExtendedHeader` to `HeaderAndTxnHashes`.

PR #957, opened 2020-04-02
Fix error when using http2

mcortesi:
### Description

Found an error message like:

```
http2: Transport: cannot retry err [http2: Transport received Server's graceful shutdown GOAWAY] after Request.Body was written; define Request.GetBody to avoid this error
```

So, this PR implements GetBody

### Tested

Discovered, fixed and tested while developing rosetta

### DIFF
--- a/rpc/http.go
+++ b/rpc/http.go
@@ -175,6 +175,13 @@ func (hc *httpConn) doRequest(ctx context.Context, msg interface{}) (io.ReadClos
 	req.Body = ioutil.NopCloser(bytes.NewReader(body))
 	req.ContentLength = int64(len(body))
 
+	// And also add the same thing to `Request.GetBody`, which allows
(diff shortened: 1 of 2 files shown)


rpc/http.go
```diff
@@ -175,6 +175,13 @@ func (hc *httpConn) doRequest(ctx context.Context, msg interface{}) (io.ReadClos
 	req.Body = ioutil.NopCloser(bytes.NewReader(body))
 	req.ContentLength = int64(len(body))
 
+	// And also add the same thing to `Request.GetBody`, which allows
```
kevjue, line 178: nit:  Consider creating a closure that would be used for both `req.Body` and `req.GetBody`.  IMO it would be easier to understand the code.

mcortesi: done!

PR #958, opened 2020-04-02
Add methods to BoundContract & abigen wrappers

mcortesi:
### Description

New methods to parse event Logs, and expose the ABI object

### Tested

In use in rosetta

### DIFF
--- a/accounts/abi/bind/base.go
+++ b/accounts/abi/bind/base.go
@@ -116,6 +116,11 @@ func DeployContract(opts *TransactOpts, abi abi.ABI, bytecode []byte, backend Co
 	return c.address, tx, c, nil
 }
 
+// ABI returns the bound contract parsed ABI
+func (c *BoundContract) ABI() abi.ABI {
+	return c.abi
+}
+
 // Call invokes the (constant) contract method with params as input values and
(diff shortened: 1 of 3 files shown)


accounts/abi/bind/base.go
```diff
@@ -261,6 +266,19 @@ func (c *BoundContract) transact(opts *TransactOpts, contract *common.Address, i
 	return signedTx, nil
 }
 
+// LogEventName will verify is log belongs to the contract and return the eventName for it
```
kevjue, line 269: nit: `verify is log` -> `verify the log`

accounts/abi/bind/template.go
```diff
@@ -351,6 +360,24 @@ var (
 		}
 	{{end}}
 
+	// TryParseLog attemps to parse a log. Returns the parsed log, evenName and wether it was succesfull
```
kevjue, line 363: nit:  typo `wether`

accounts/abi/bind/base.go
mcortesi: thanks! Solved

accounts/abi/bind/template.go
mcortesi: thanks! Solved

PR #961, opened 2020-04-03
Don't aggregate epochSnarkData on non-epoch blocks

kobigurk:
### Description

EpochSnarkData is empty on non-epoch blocks, so no reason to aggregate it.

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -303,7 +303,7 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


consensus/istanbul/core/core.go
```diff
@@ -303,7 +303,7 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
-		aggregatedEpochValidatorSetSeal, err := GetAggregatedEpochValidatorSetSeal(c.current.Commits())
+		aggregatedEpochValidatorSetSeal, err := GetAggregatedEpochValidatorSetSeal(c.current.Proposal().Number().Uint64(), c.config.Epoch, c.current.Commits())
```
asaj, line 306: nit: just use `proposal.Number().Uint64()`

asaj:
FYI #960 will fix the checkout-monorepo step, should be merged shortly

gakonst:
@kobigurk is coverage flaky? Can we merge this?

PR #964, opened 2020-04-03
Improve help for the proof-of-possession command

nategraf:
### Description

Had to open up the code to figure out the arg ordering for the `proof-of-possession` command, os I wrote a bit more description for it while I was in there.

https://github.com/celo-org/celo-blockchain/blob/62a6626e928cfb29eef7a0a1e488dd65d546fe2e/cmd/geth/accountcmd.go#L124-L130

### Other Changes

* Remove an used config parameter on the transaction pool and as a result fix `dumpconfig`


cmd/geth/accountcmd.go
```diff
@@ -122,6 +122,11 @@ Print a short summary of all accounts`,
 				},
 				Description: `
 Print a proof-of-possession signature for the given account.
+
+Creates a signature over the given message, for example an account address,
+with the private key corrosponsding to a given signer address. The singature
+can then be used to prove possession of the signing key, for exmaple to
+authrorize a the signer to act as a validator in the Celo protocol.
```
kevjue, line 129: Some spelling corrections.

```suggestion
Creates a signature over the given message, for example an account address,
with the private key corresponding to a given signer address. The signature
can then be used to prove possession of the signing key, for example to
authorize a the signer to act as a validator in the Celo protocol.
```

PR #969, opened 2020-04-06
Batch BLS Verification in Ultralight mode

gakonst:
### Description

Batch BLS Verification in Ultralight mode. Currently, we check BLS signatures on a per-epoch basis. This PR allows us to get the data from all headers, and batch verify them with a pairing equation which in total costs us n+1 pairings vs 2n.

Depends on #973 

### Tested

Have added a unit test for the function 

### Other changes

- Updates bls-zexe to latest.
- Exposes pubkey aggregation in bls-crypto

### Backwards compatibility

This should be backwards compatible.


consensus/istanbul/backend/engine.go
```diff
@@ -195,8 +195,15 @@ func (sb *Backend) VerifyHeaders(chain consensus.ChainReader, headers []*types.H
 	} else {
 		go func() {
 			err := sb.batchVerifyHeaders(chain, headers)
-			// only 1 result will be returned by the channel
-			results <- err
+			// even though we get 1 return value from the batched verification
```
gakonst, line 198: So I was not too happy with this change, but it seems like there's no good alternative? In the `if` branch we have an array of results due to iterating over the headers, but in the `else` branch, since we're doing _batch_ verification there is just 1 result _for all headers_. I chose to just feed that result in all slots in the channel. Happy to discuss alternatives (not sure if there are any, without significant changes in the interfaces)

gakonst:
The PR should be good for review, the `end-to-end-transfer-test` is flaky.

consensus/istanbul/backend/engine.go
```diff
@@ -175,20 +177,112 @@ func (sb *Backend) verifyCascadingFields(chain consensus.ChainReader, header *ty
...
+
+		// short-circuit if there was no signature found
+		if len(extra.AggregatedSeal.Signature) == 0 {
+			return errEmptyAggregatedSeal
+		}
+
+		// 1. Get the aggregate public key for this epoch
+
+		// 1.1. Get the snapshot
+		snap, err := sb.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, headers[:i])
```
kevjue, line 241: The `snapshot` function assumes that all of the headers passed in the `parents` parameter are already verified.   Does that assumption break anything here?

consensus/istanbul/backend/engine.go
```diff
@@ -285,27 +379,38 @@ func (sb *Backend) verifyAggregatedSeal(headerHash common.Hash, validators istan
...
+	if err != nil {
+		logger.Error("Unable to verify aggregated signature", "err", err)
+		return errInvalidSignature
+	}
+
+	return nil
+}
+
+// Find which public keys signed from the provided validator set
+func hasQuorum(validators istanbul.ValidatorSet, aggregatedSeal types.IstanbulAggregatedSeal) ([]blscrypto.SerializedPublicKey, error) {
```
kevjue, line 399: nit: Perhaps rename this to `getAggregatedSealSigners`?

consensus/istanbul/backend/engine.go
```diff
@@ -175,20 +177,112 @@ func (sb *Backend) verifyCascadingFields(chain consensus.ChainReader, header *ty
...
+		// 1.1. Get the snapshot
+		snap, err := sb.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, headers[:i])
+		if err != nil {
+			return err
+		}
+
+		// 1.2 Get the subset of validator public keys that signed
+		// based on the sig's bitmap
+		asig := extra.AggregatedSeal
+		publicKeys, err := hasQuorum(snap.ValSet, asig)
```
kevjue, line 249: Related to my comment above, is the ValSet here trusted?  Or will the batchVerify function below catch an invalid valset?

consensus/istanbul/backend/engine.go
gakonst: The `batchVerifyHeaders` function is called with _all_ the headers which are provided to `engine.VerifyHeaders`. It then gathers all the signatures and validator sets from each epoch and forwards them to the signature verification algorithm. The verification algorithm _will_ fail if a "fake" validator public key is provided, if the signature is made by a different key.


 
Finally, my understanding is that the `parents` parameter is used as a cache, so if this is still a concern we can simply leave it as `nil`.

consensus/istanbul/backend/engine.go
gakonst: If an invalid validator set is provided, then the signature verification will fail

consensus/istanbul/backend/engine.go
```diff
@@ -175,20 +177,112 @@ func (sb *Backend) verifyCascadingFields(chain consensus.ChainReader, header *ty
...
+					return
+				case results <- err:
+				}
 			}
-		}
-	}()
+		}()
+	} else {
+		go func() {
+			err := sb.batchVerifyHeaders(chain, headers)
```
asaj, line 197: Should we batch verify headers even when `FullHeaderChainAvailable` == true? In those cases for the most part the validator set will be the same for every header in the batch

consensus/istanbul/backend/engine.go
```diff
@@ -175,20 +177,112 @@ func (sb *Backend) verifyCascadingFields(chain consensus.ChainReader, header *ty
...
+				}
+			}
+		}()
+	}
 	return abort, results
 }
 
+// batchVerifyHeaders is called in ultralight mode and performs batch BLS signature
+// verification over all provided headers and their corresponding validator sets
+func (sb *Backend) batchVerifyHeaders(chain consensus.ChainReader, headers []*types.Header) error {
```
asaj, line 214: This seems to have some duplicated code from `verifyHeader`, see my suggestion above for how this might be shared

consensus/istanbul/backend/engine.go
asaj: Hmm yeah I'm not totally sure but I think @kevjue is onto something, nice catch.

When calling `snapshot` on header `H` in epoch N, it uses `parents` to build the validator set snapshot for `N` (and previous epochs, if necessary) by looking at the validator set diffs in `parents`.

A malicious full node could send a batch of headers that includes an invalid header with a validator set diff that changes out the validator set to keys controlled by the attacker. This batch of headers would get rejected, but not before calling `snapshot` on them, which calls `apply`, which would store the attacker's validator set. Then, the same full node could send another batch of headers signed by the attacker, which would get accepted.

Instead, we may need to pull a snapshot for the first header to get the validator set at the beginning of the batch, and then "manually" update this validator set whenever we see epoch blocks as we iterate over the headers.

Finally, if all checks pass and the signatures look good, then we call `snapshot` on the last header in the batch, passing all the parent headers, to update the validator set snapshot to the latest epoch.

consensus/istanbul/backend/engine.go
kobigurk: In this case we can even do better, tbh.

asaj: Cool, it sounds like it would be both more efficient *and* simplify the code, so I suggest we go for it

consensus/istanbul/backend/engine.go
```diff
@@ -175,20 +177,112 @@ func (sb *Backend) verifyCascadingFields(chain consensus.ChainReader, header *ty
...
+		// 1.2 Get the subset of validator public keys that signed
+		// based on the sig's bitmap
+		asig := extra.AggregatedSeal
+		publicKeys, err := getAggregatedSealSigners(snap.ValSet, asig)
+		if err != nil {
+			return err
+		}
+
+		// 1.3. Aggregate them
+		apk, err := blscrypto.AggregatePublicKeys(publicKeys)
```
asaj, line 255: Is this step expensive? Is there anything we can do to take advantage of the fact that the validator set changes slowly?

asaj:
Have we done any benchmarking from which we can infer the impact of this change?

PR #971, opened 2020-04-06
Change proxy and proxied checks

jcortejoso:
### Description

Fixes for celostats integration

### Tested

Tested on a testnet


consensus/istanbul/backend/backend.go
```diff
@@ -273,18 +273,20 @@ type Backend struct {
...
 func (sb *Backend) IsProxiedValidator() bool {
-	return sb.proxyNode != nil && sb.proxyNode.peer != nil
+	return sb.config.Proxied
 }
 
 // SendDelegateSignMsgToProxy sends an istanbulDelegateSign message to a proxy
 // if one exists
 func (sb *Backend) SendDelegateSignMsgToProxy(msg []byte) error {
-	if !sb.IsProxiedValidator() {
+	if !sb.IsProxiedValidator() || sb.proxyNode.peer == nil {
```
timmoreton, line 289: panic if `sb.ProxyNode == nil` here ?

tkporter: I could see this state happening if a validator & proxy disconnection happens right after a validator receives a delegate sign message from the proxy, so I don't think we should purposely cause a validator to crash

timmoreton: what I mean is it would panic if `sb.ProxyNode == nil`

timmoreton: consider ```if !sb.IsProxiedValidator() || sb.proxyNode == nil || sb.proxyNode.peer == nil {```

tkporter: Ahh good catch

Issue #968, opened 2020-04-06
`makeBlock` test deadlock


gakonst:
### Expected Behavior

`makeBlock` should create a new block which gets signed by all the validators in the current engine. This is useful if you want to write multivalidator tests

### Current Behavior

Code hangs if `makeBlock` is called with >1 validators in the current chain

### Steps to Reproduce Behavior

Run: 
`go test -v ./consensus/istanbul/backend/... -run "TestVerifySeal"` in branch https://github.com/celo-org/celo-blockchain/tree/gakonst/engine-test-deadlock, and watch the test never complete

PR #973, opened 2020-04-07
refactor(consensus): create shared test utils and deduplicate code

gakonst:
As title. 

- Improves writing consensus/backend tests around BLS signatures. There was a lot of duplicate code which could be abstracted in nicer ways. 
    - The 2 BLS signing functions have been combined in 1, which takes a flag for choosing hashing mode and accepts extra data. This is functionally equivalent with what we had before, just more flexible and less code.
- Moves shared test utilities in a shared file
- Fixes tests which were misleadingly passing. 
    - The `eventLoop` + `Seal`+ multiple validators patterns was kind of confusing. I've changed `makeBlock` to now work with multiple validators and filling the block's aggregate signature field properly.

This is _not_ a breaking change

Closes https://github.com/celo-org/celo-blockchain/issues/968

(note that this PR is necessary in order to be able to test the changes in https://github.com/celo-org/celo-blockchain/pull/969, since they require generating a bunch of blocks with valid BLS signatures)


gakonst:
FWIW I was told the governance test is still flaky

consensus/istanbul/backend/test_utils.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/crypto"
+	blscrypto "github.com/ethereum/go-ethereum/crypto/bls"
+	"github.com/ethereum/go-ethereum/crypto/ecies"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+// in this test, we can set n to 1, and it means we can process Istanbul and commit a
```
asaj, line 28: Comment no longer relevant?

consensus/istanbul/backend/engine_test.go
```diff
@@ -354,29 +99,37 @@ func TestSealStopChannel(t *testing.T) {
...
+	engine.Commit(otherBlock, types.IstanbulAggregatedSeal{}, types.IstanbulEpochValidatorSetSeal{})
 
-	go func() { engine.Seal(chain, block, results, nil) }()
 	select {
-	case <-results:
-		t.Error("seal should not be completed")
-	case <-time.After(2 * time.Second):
+	case res := <-results:
+		t.Fatal("seal should not be completed", res)
+	case <-time.After(100 * time.Millisecond):
```
asaj, line 132: Why can this be so much faster now?

consensus/istanbul/backend/test_utils.go
gakonst: No it is still needed, you still need to make a `Seal` -> `Commit` call to get blocks signed by multiple validators

consensus/istanbul/backend/engine_test.go
gakonst: This was leftover from the initial instanbul implementation, it could've been smaller anyway

gakonst:
@asaj this seems to be the flaky test that's failing, I've rebased this to latest master

PR #974, opened 2020-04-07
Removed etherbase from login to avoid account unlocks on proxies

H34D:
### Description

It is not longer required to unlock the etherbase account on the proxy to connect to celostats

### Backward compatibility

This change is breaking because the ID field is being removed from the auth message. This was already adapted in celostats and should not make any different now. It can be safely removed. Celostats now uses proof.address to do the same. 

### Tested

Tested with my local testnet

- tested in proxied environments
- tested in non proxied environments


consensus/istanbul/backend/backend.go
```diff
@@ -275,12 +275,12 @@ type Backend struct {
...
 // IsProxy returns if instance has proxy flag
 func (sb *Backend) IsProxy() bool {
-	return sb.config.Proxy
+	return sb.proxiedPeer != nil || sb.config.Proxy
 }
 
 // IsProxiedValidator returns if instance has proxied validator flag
 func (sb *Backend) IsProxiedValidator() bool {
-	return sb.config.Proxied
+	return (sb.proxyNode != nil && sb.proxyNode.peer != nil) || sb.config.Proxied
```
asaj, line 283: Ditto

ethstats/ethstats.go
```diff
@@ -388,15 +376,23 @@ func (s *Service) login(conn *websocket.Conn, sendCh chan *StatsPayload) error {
 		network = fmt.Sprintf("%d", lesInfo.Network)
 		protocol = fmt.Sprintf("les/%d", les.ClientProtocolVersions[0])
 	}
-	if s.eth != nil {
-		etherBase, err = s.eth.Etherbase()
-		if err != nil {
-			return err
+
+	if s.backend.IsProxy() {
+		// Proxy needs a delegate send here to get ACK
```
asaj, line 381: I'm not totally clear on what's happening here

asaj:
Can you complete the PR template? I'm particularly interested on how backwards compatible these changes are

ethstats/ethstats.go
```diff
@@ -567,9 +555,8 @@ type nodeInfo struct {
 
 // authMsg is the authentication infos needed to login to a monitoring server.
 type authMsg struct {
-	ID      string         `json:"id"`
-	Address common.Address `json:"address"`
-	Info    nodeInfo       `json:"info"`
+	ID   string   `json:"id"`
```
asaj, line 558: Is this change backwards compatible?

consensus/istanbul/backend/backend.go
```diff
@@ -275,12 +275,12 @@ type Backend struct {
 
 // IsProxy returns if instance has proxy flag
 func (sb *Backend) IsProxy() bool {
-	return sb.config.Proxy
+	return sb.proxiedPeer != nil || sb.config.Proxy
```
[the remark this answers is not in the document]
H34D: there are scenarios where the proxied peer is not yet connected, or has disconnect from the proxy (during the network rampup, or simple when the validator reboots). During this time the proxy should still identify itself as a proxy i.e. to start the celostats daemon. So either we have a proxied peer connect or we are configured as a proxy makes the node a proxy.

consensus/istanbul/backend/backend.go
H34D: there are scenarios where the proxied peer is not yet connected, or has disconnect from the proxy (during the network rampup, or simple when the validator reboots). During this time the proxy should still identify itself as a proxy i.e. to start the celostats daemon. So either we have a proxied peer connect or we are configured as a proxy makes the node a proxy.

ethstats/ethstats.go
H34D: a proxy is not able to sign therefore it has to forward all the sign operations to the validator and sign it with the validator signer

ethstats/ethstats.go
H34D: yes it is, celostats was using this field since a couple of days ago. I removed it there first now it can be removed here. I don't know any more usages of this field.

ethstats/ethstats.go
H34D: > Can you complete the PR template? I'm particularly interested on how backwards compatible these changes are

will do

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
tkporter: `sb.proxiedPeer != nil` will only be possible if `sb.config.Proxy` is true, so we can keep this as `sb.config.Proxy`. In general I'm hesitant about editing a function like this because it might have unintended consequences

consensus/istanbul/backend/backend.go
tkporter: Same thing, `(sb.proxyNode != nil && sb.proxyNode.peer != nil)` can only be true if `sb.config.Proxied` is true, so this is the same as just returning `sb.config.Proxied`

ethstats/ethstats.go
```diff
@@ -68,7 +68,7 @@ const (
 	// connectionTimeout waits for the websocket connection to be established
 	connectionTimeout = 10
 	// delegateSendTimeout waits for the proxy to sign a message
-	delegateSendTimeout = 5
+	delegateSendTimeout = 10
```
tkporter, line 71: Curious if we've been seeing long response times for the delegate signing?

I thin `delegateSend` in this file isn't really clear what is happening, should this be renamed `delegateSignTimeout`?

ethstats/ethstats.go
tkporter: Would a geth instance still sending `ID, Address, Info` still be able to authenticate with the updated celostats?

ethstats/ethstats.go
tkporter: If a delegate sign message is sent from the proxy to the validator in the `sendStats` call on line 410, then why is the proxy now waiting for delegate signed msg from the proxied validator before the `sendStats` call?

ethstats/ethstats.go
H34D: yes

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
H34D: i rolled back that change

consensus/istanbul/backend/backend.go
H34D: i rolled back that change

ethstats/ethstats.go
H34D: renamed, timeout back to 5 seconds

H34D: renamed

ethstats/ethstats.go
H34D: it will wait for the channel in 383 and 388

H34D:
> Can you complete the PR template? I'm particularly interested on how backwards compatible these changes are

done, i will now do the retest. it took me quite a while to get the previous version running.

kevjue:
> This change is breaking because the ID field is being removed from the auth message. This was already adapted in celostats and should not make any different now. It can be safely removed. Celostats now uses proof.address to do the same.

Just to understand this a bit better, will celostats still support clients that have this change and clients that don't?

H34D:
> Just to understand this a bit better, will celostats still support clients that have this change and clients that don't?

yes, celostats will still support clients that have this change and clients that don't

PR #975, opened 2020-04-07
Adds address confirmation to Ledger

kobigurk:
### Description

This allows users to confirm the address they receive from other commands are an address that exists on the Ledger, confirming the Ledger actually contains the address private key.

The command format is `geth account confirm-address ADDRESS`.

### Tested

Tested that normal transaction sending and proofs-of-possession are not affected and that when running the confirm-address command the address is displayed on the Ledger.

This is backwards compatible.

### DIFF
--- a/accounts/accounts.go
+++ b/accounts/accounts.go
@@ -71,7 +71,7 @@ type Wallet interface {
 	// resources (especially important when working with hardware wallets).
 	Open(passphrase string) error
 
-  // ConfirmAddress shows the address of the given path on the wallet's display.
+	// ConfirmAddress shows the address of the given path on the wallet's display.
 	ConfirmAddress(path DerivationPath) (common.Address, error)
 
 	// Close releases any resources held by an open wallet instance.
--- a/accounts/accounts.go
+++ b/accounts/accounts.go
@@ -71,6 +71,9 @@ type Wallet interface {
 	// resources (especially important when working with hardware wallets).
 	Open(passphrase string) error
 
+  // ConfirmAddress shows the address of the given path on the wallet's display.
+	ConfirmAddress(path DerivationPath) (common.Address, error)
+
 	// Close releases any resources held by an open wallet instance.
 	Close() error
 
--- a/accounts/usbwallet/wallet.go
+++ b/accounts/usbwallet/wallet.go
@@ -56,7 +56,7 @@ type driver interface {
 	// or may not be used by the implementation of a particular wallet instance.
 	Open(device io.ReadWriter, passphrase string) error
 
-  // ConfirmAddress shows the address of the given path on the wallet's display.
+	// ConfirmAddress shows the address of the given path on the wallet's display.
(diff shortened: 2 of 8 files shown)


accounts/accounts.go
```diff
@@ -71,6 +71,9 @@ type Wallet interface {
 	// resources (especially important when working with hardware wallets).
 	Open(passphrase string) error
 
+  // ConfirmAddress shows the address of the given path on the wallet's display.
```
gakonst, line 74: `gofmt`

accounts/usbwallet/wallet.go
```diff
@@ -496,6 +499,29 @@ func (w *wallet) Derive(path accounts.DerivationPath, pin bool) (accounts.Accoun
...
+// ConfirmAddress implements accounts.Wallet, showing the address on the device.
+func (w *wallet) ConfirmAddress(path accounts.DerivationPath) (common.Address, error) {
+	// Try to derive the actual account and update its URL if successful
+	w.stateLock.RLock() // Avoid device disappearing during derivation
+
+	if w.device == nil {
+		w.stateLock.RUnlock()
+		return common.Address{}, accounts.ErrWalletClosed
+	}
+	<-w.commsLock // Avoid concurrent hardware access
```
gakonst, line 511: Is this a standard pattern? I'd expect that there's a `sync.Mutex` here instead of using an empty channel as a lock structure

kobigurk:
I'm not sure that if we wrote this from scratch we'd do the same, but I
follow geth's approach.

cmd/geth/accountcmd.go
```diff
@@ -303,6 +321,55 @@ func accountProofOfPossession(ctx *cli.Context) error {
...
 
+func accountConfirmHardwareAddress(ctx *cli.Context) error {
+	if len(ctx.Args()) != 1 {
+		utils.Fatalf("Please specify the address to confirm on the hardware wallet.")
+	}
+
+	stack, _ := makeConfigNode(ctx)
+	am := stack.AccountManager()
+
+	address := common.HexToAddress(ctx.Args()[0])
```
gastonponti, line 332: Before you asked for `len(ctx.Args()) != 1` so probably it means that the address is not in the first position, but then you are extracting the address from there... Or am I missing something?

kobigurk: If it's different than 1 then it fails and exits :)

gastonponti: yes, but that means that it has two or more params. My question is if the index `0` is still the address, or maybe `1`

kobigurk: Yeah, the context starts from the arguments, it doesn't contain the program name or anything. You can see an example here: https://github.com/celo-org/celo-blockchain/blob/7a9fe442d0b8480a3e7a4f5a42ac288e0ba35cc8/cmd/geth/accountcmd.go#L257

PR #978, opened 2020-04-08
Fix missing reference to libbls_crypto causing linking errors on iOS

jeanregisser:
### Description

Fix missing reference to libbls_crypto causing linking errors on iOS

### Tested

Compilation works again on iOS.

### Backwards compatibility

Yes.


kobigurk:
If this works, let's go for it for now.

A note for the future - linking bls_crypto shouldn't be needed, since epoch_snark contains all the functions from bls_crypto. We need to investigate this more deeply in the future.

PR #979, opened 2020-04-09
Adds ERC20 tokens list to Ledger

kobigurk:
### Description

Ledger supports a dynamic ERC20 mechanism, requiring that the tokens are signed by some authority defined by a public key in the Ledger app.
This PR adds automatic detection from a signed token list and providing these details for the To and FeeCurrency addresses to the Ledger before an ERC20 transaction.

### Tested

Tested against the Ledger and RC1.

### Backwards compatibility

This is backwards compatible.


kobigurk:
@timmoreton @asaj the token list embedded now is just cUSD from alfajores staging. Let's update it with both:

* cUSD and cGLD from all the testnets
* Signatures by the new private key on the SF ledger

accounts/usbwallet/ledger.go
```diff
@@ -325,11 +338,27 @@ func (w *ledgerDriver) ledgerSign(derivationPath []uint32, tx *types.Transaction
...
 		err   error
 	)
+
+	if ledger.IsERC20Transfer(tx.Data()) {
+		err = w.ledgerProvideERC20(*tx.To(), chainID)
+		if err != nil && err != ledger.ErrCouldNotFindToken {
+			return common.Address{}, nil, err
+		}
+
+		if tx.FeeCurrency() != nil {
```
gastonponti, line 353: I don't know if I'm following the case here...
If the feeCurrency is for example cUSD, it means that the transaction will be a cUSD one? It's not possible to just transfer cGOLD paying fees using a stable token?

accounts/usbwallet/ledger/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+	"bytes"
+	"encoding/base64"
+	"encoding/binary"
+	"errors"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const TokensBlob = "AAAAaARjVVNEdl3oFoRYYedaJfyhIrtomLixKCoAAAASAACk7DBFAiEApwQFHNBKXp+V2jq8BMD2y/5AwC9bhPQ2H4hT/vMl/B4CIFalOVtBFGREUKMU/F5vDlJLeQrTn6GQeDertpB2FpMvAAAAZwRjR0xERx7ON1DaI3+TuOM5xTaYm4l4pDgAAAASAACk7DBEAiAtUE03OVEDTuAgMR8CeUiXVB7Uqa+4gQHaZtXsU0tjMQIgTsY5//96mGMOtyoQFgjDrh3kztxlUO90MJOLElo6baQ=" // #nosec
```
gastonponti, line 13: Maybe you could extract it in another file to keep things cleaner. One thing is how to use the data, and the other is the data itself

accounts/usbwallet/ledger/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return nil, ErrCouldNotFindToken
+}
+
+func IsERC20Transfer(data []byte) bool {
+	if len(data) < 4 {
+		return false
+	}
+
+	if !bytes.Equal(data[:4], []byte{0xa9, 0x05, 0x9c, 0xbb}) {
```
gastonponti, line 97: Could you add a comment of what `[]byte{0xa9, 0x05, 0x9c, 0xbb}` suppose to be? In the future someone could struggle with this, haha 😄

accounts/usbwallet/ledger.go
kobigurk: Yeah, it should be possible!

There are two calls to `ledgerProvideERC20` - one for the `To` and one for the `FeeCurrency`. The Ledger app knows to save the last 2 calls of `ledgerProvideERC20` it received, so it should have info about both.

accounts/usbwallet/ledger/tokens.go
kobigurk: Good idea, I'll do that.

accounts/usbwallet/ledger/tokens.go
kobigurk: Yes :D

accounts/usbwallet/ledger.go
gastonponti: Ahhh but does means that you are somehow deleting the `To` field from the transaction? otherwise you'll getting both times the same contractData

PR #980, opened 2020-04-09
Delete gitignore so libs are included in npm publish script

cmcewen:
## Description

The npm package was missing the ios libraries due to the gitignore


jeanregisser:
Nice find! 👍 

This is what caused the original linking errors I got on iOS which made me fix it incorrectly in #978 (I've now open #981 to revert it).

I had cloned `celo-blockchain` directly to replace `@celo/client` in my `node_modules` and didn't experience this problem in this context, which made me think the fix I did was right.

jeanregisser:
@Mergifyio update

PR #981, opened 2020-04-10
Revert "Fix missing reference to libbls_crypto causing linking errors on iOS (#978)"

jeanregisser:
## Description

This reverts commit d2eca3f142b1cac962b8245806d02e382b2690d5.

The problem was the lib missing in the published npm package, now fixed by @cmcewen in #980.

I had cloned `celo-blockchain` directly to replace `@celo/client` in my `node_modules` and didn't experience this problem in this context, which made me think the fix I did was right.

My bad 🤦 

## Tested

Compilation of the wallet is successful.


jeanregisser:
@Mergifyio update

PR #987, opened 2020-04-13
[DO NOT MERGE] Updates bls-zexe to use upstream Zexe for the map to curve

kobigurk:
## Description

Upstream Zexe now has a method for map to curve which is slightly different than ours - the greatest bit location is different. This updates bls-zexe to use it.

This is NOT backwards compatible.


gakonst:
LGTM. THIS IS A BREAKING CHANGE, so let's make sure we all agree on it. As discussed in private, the only thing this changes are the values of hashes. As such, it is very limited in scope, and we have already tested that the behavior is unchanged beyond that scope.

gakonst:
@kobigurk think this can be closed now since #1022 is merged

kobigurk:
Right, thanks!

PR #991, opened 2020-04-13
Adds epoch validator snark bitmap to block body

kobigurk:
## Description

In Kevin's words:

```
As discussed in the protocol meeting, the possible issue that I discovered during the BLS aggregated seal / lightest sync audit, is the small chance that the epoch encoding data aggregated seal's validator bitmask may be incorrect.  I have more details in the thread.

The encoded epoch data's aggregated seal's bitmask appears to be the same one that the aggregated commit seal uses. (edited) 

However, it's possible that when a node syncs a block, that the block header and block body are retrieved from different nodes.

And those node may have originally retrieved a finalized block from two different validators that have a different seal validator bitmap.

So I think a possible fix is for a bitmap to be added for the encoded epoch data's seal in the block body.
```

We've implemented this suggested fix.

## Tested

Unit tests run.

## Backwards compatibility

It is backwards compatible for any non-validators and not for any validator node.


core/types/block.go
```diff
@@ -130,26 +130,29 @@ func (r *Randomness) EncodeRLP(w io.Writer) error {
 }
 
 type EpochSnarkData struct {
+	Bitmap    *big.Int
 	Signature []byte
 }
 
 func (r *EpochSnarkData) Size() common.StorageSize {
-	return common.StorageSize(blscrypto.SIGNATUREBYTES)
+	return common.StorageSize(blscrypto.SIGNATUREBYTES + (r.Bitmap.BitLen() / 8))
```
kevjue, line 138: nit:  Consider adding a comment stating that this function returns an estimated size.

kobigurk: Done!

PR #995, opened 2020-04-15
Adding testnet flags for baklava and alfajores

### LINKED ISSUE #994: Add a testnet specific flags for baklava and alfajores for geth.
**Expected Behavior**

Upstream geth already has command line options to connect and configure to non mainnet ethereum networks.   We should create similar ones for Celo testnets (alfajores and baklava).

This is part of epic:  https://github.com/celo-org/celo-blockchain/issues/1012

**Current Behavior**

The celo blockchain client doesn't have any flags for the celo testnets.

mrsmkl:
## Description

Added flags for running Baklava or Alfajores

## Other changes

Baklava and Alfajores aren't compatible with each another, and also neither is compatible with current master (RC1)

## Tested

Able to sync with both alfajores and baklava.

## Related issues

- Fixes #994

## Backwards compatibility

Using old genesis json configs for baklava and alfajores probably won't work anymore.


mrsmkl:
Looks like baklava and alfajores are incompatible, perhaps these flags will make more sense when we can use just one client?

cmd/geth/chaincmd.go
```diff
@@ -184,6 +184,8 @@ Use "ethereum dump 0" to dump the genesis block.`,
 			utils.TestnetFlag,
 			utils.RinkebyFlag,
```
timmoreton, line 185: is it worth removing these here? (or in a separate PR?)

timmoreton:
looks like a flaky test `TestUnlockFlagWrongPassword`

PR #997, opened 2020-04-18
Begin 1.1.0 release cycle

timmoreton:
## Description

* Update master version to `1.1.0-unstable`
* Branch `rc0` is tracking `1.0.x`

### DIFF
--- a/params/version.go
+++ b/params/version.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	VersionMajor = 1          // Major version component of the current release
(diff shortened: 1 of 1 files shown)


asaj:
> Branch rc0 is tracking 1.0.x
Should this be rc1? Should we have a `releases/1.0` branch instead?

mcortesi:
I guess this arrives quite late.... can i close? @asaj @timmoreton

PR #1022, opened 2020-05-03
Simplifies consuming bls-zexe dependency

kobigurk:
### Description

Rust is not needed anymore to build bls-zexe. https://github.com/celo-org/celo-bls-go is the Go module that is now used. It's versioned and provides pre-built binaries for all the supported platforms.

### Other changes

Updates to latest bls-zexe that uses a compatibility mode to use the current consensus-compatible hash-to-group.

### Tested

Tested all the different supported platforms:

* MacOS
* Linux
* ARM
* Android
* iOS

End-to-end tests check syncing and verified syncing against RC1.


kobigurk:
> Fantastic. Can we ensure that a node can sync with this change before merging it?

End-to-end tests check syncing and verified against RC1 now!

In any case, I'll merge this after properly releasing `celo-bls-go`, when we merge the `bls-zexe` PR.

PR #1025, opened 2020-05-04
Added plumo flag

### LINKED ISSUE #1014: Make --plumo and alias for --lightest
### Expected Behavior

Make --plumo and alias for --lightest

### Current Behavior

There is no --plumo command line option

mrsmkl:
### Description

Just added `plumo` as an alias for `lightest` in sync mode flag (didn't remove `lightest`).
Also changed the directory name.

### Tested

Tested with alfajores using command `./build/bin/geth --alfajores --syncmode plumo`

### Related issues

- Fixes #1014

### Backwards compatibility

No, changes the directory name.

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -214,7 +214,7 @@ func initGenesis(ctx *cli.Context) error {
 	stack := makeFullNode(ctx)
 	defer stack.Close()
 
-	for _, name := range []string{"chaindata", "lightchaindata", "lightestchaindata"} {
+	for _, name := range []string{"chaindata", "lightchaindata", "plumochaindata"} {
 		chaindb, err := stack.OpenDatabase(name, 0, 0, "")
 		if err != nil {
 			utils.Fatalf("Failed to open database: %v", err)
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -216,7 +216,7 @@ var (
 	defaultSyncMode = eth.DefaultConfig.SyncMode
 	SyncModeFlag    = TextMarshalerFlag{
 		Name:  "syncmode",
-		Usage: `Blockchain sync mode ("fast", "full", "light", or "lightest")`,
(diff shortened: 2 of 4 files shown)


kobigurk:
Do we maybe need two names? Plumo for SNARK sync and another for epoch header sync?

kevjue:
> Do we maybe need two names? Plumo for SNARK sync and another for epoch header sync?

Good point.  Are they not "one and the same".  E.g. if the snark is available, the client will download that, but if it isn't it will download the epoch headers?

kobigurk:
> > Do we maybe need two names? Plumo for SNARK sync and another for epoch header sync?
> 
> Good point. Are they not "one and the same". E.g. if the snark is available, the client will download that, but if it isn't it will download the epoch headers?

Since this isn't integrated yet, we've yet to decide at that level. I agree that intuitively it seems that if the snark is available, you want to use that and fallback.

kevjue:
> > > Do we maybe need two names? Plumo for SNARK sync and another for epoch header sync?
> > 
> > 
> > Good point. Are they not "one and the same". E.g. if the snark is available, the client will download that, but if it isn't it will download the epoch headers?
> 
> Since this isn't integrated yet, we've yet to decide at that level. I agree that intuitively it seems that if the snark is available, you want to use that and fallback.

Seems like this is a product question.  E.g. we could have the --plumo flag behave as implemented in the PR, but that runs the risk that it's meaning will change if during the snark integration it's decided to have two separate modes (resulting in a bad UX).

I agree with you @kobigurk that we should hold off on it until that's decided.

@timmoreton @pranaymohan - WDYT?

pranaymohan:
Synced with @kevjue and agree we should hold off on merging until it is decided what we want to expose to the user (eg. give them the option to sync via downloading epoch headers vs SNARK). 

I would be curious if there is any reason someone would run in light client mode and would rather have the raw headers than the SNARK, but choose not to run a full node... If this doesn't feel like a desirable use case, then trying to download SNARK and falling back to epoch headers sounds right.

PR #1031, opened 2020-05-07
Making lightchain compatible with lightest sync mode

### LINKED ISSUE #1011: Existing or new lightchain object SBAT be compatible with lightest sync downloader
### Expected Behavior

The current lightchain object assumes that the lightest sync downloader will download the full header chain.

This is part of this epic:  https://github.com/celo-org/celo-blockchain/issues/1010

### Current Behavior

The existing lightchain object or a new one should be able to work with downloader that only downloads the epoch headers.

lucasege:
### Description
Adding a `HeaderRequest` object to the `odr` to allow a lightchain to request a block header via a block number or a block hash. This allows the `lightest` client to request block information by first requesting headers and then following the usual `light` client flow using the header.

### Other changes

This involves changes in `les.client_handler` to expect these header requests.
This also involves changes to `les/fetcher` to allow for `peer.hasBlock` to optimistically return true if no block number is supplied which allows for requesting a block via just a hash.

### Tested

Added a unit test for this, and tested via a local testnet.

Still TODO:
* Tests to catch bad signatures

### Related issues

- Fixes #1011 

### Backwards compatibility

Yes


les/odr_requests.go
```diff
@@ -125,6 +128,46 @@ func (r *BlockRequest) Validate(db ethdb.Database, msg *Msg) error {
...
 
+//TODO(lucas): Comments
+type HeaderRequest light.HeaderRequest
+
+func (r *HeaderRequest) GetCost(peer *peer) uint64 {
+	return peer.GetRequestCost(GetBlockHeadersMsg, 1)
+}
+
+func (r *HeaderRequest) CanSend(peer *peer) bool {
+	// TODO(lucas): `HasBlock` doesn't work for the hash only case
```
kevjue, line 139: Out of curiosity, since the `HeaderRequest` is is serving `odr_utils.GetHeaderByHash`, how does the request figure out if the peer can serve it?

light/odr.go
```diff
@@ -119,6 +119,23 @@ func (req *BlockRequest) StoreResult(db ethdb.Database) {
...
+
+type HeaderRequest struct {
+	OdrRequest
+	Origin hashOrNumber
+	Header *types.Header
+}
+
+func (req *HeaderRequest) StoreResult(db ethdb.Database) {
+	rawdb.WriteHeader(db, req.Header)
+	rawdb.WriteTd(db, req.Header.Hash(), req.Header.Number.Uint64(), big.NewInt(int64(req.Header.Number.Uint64()+1)))
```
kevjue, line 135: Is there a function somewhere that will calculate the TD for our consensus engine?  Seems that we're hard coding `Number+1` all over the place for TD.

light/odr.go
```diff
@@ -119,6 +119,23 @@ func (req *BlockRequest) StoreResult(db ethdb.Database) {
...
+	Number uint64
+}
+
+type HeaderRequest struct {
+	OdrRequest
+	Origin hashOrNumber
+	Header *types.Header
+}
+
+func (req *HeaderRequest) StoreResult(db ethdb.Database) {
```
kevjue, line 133: Seems like the function `HeaderChain.WriteHeader` does a lot of the same functionality as this function.  How are they different?

les/client_handler.go
```diff
@@ -232,9 +232,27 @@ func (h *clientHandler) handleMsg(p *peer) error {
...
-				log.Debug("Failed to deliver headers", "err", err)
+		} else { // TODO(lucas): this is likely imperfect
+			h.backend.retriever.lock.RLock()
+			headerRequested := h.backend.retriever.sentReqs[resp.ReqID]
+			h.backend.retriever.lock.RUnlock()
+			if headerRequested != nil {
+				contiguousHeaders := true
+				if h.syncMode == downloader.LightestSync {
+					contiguousHeaders = false
+				}
```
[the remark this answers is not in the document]
lucasege: `fullChainAvailable` is not on `headerChain` and cannot easily be accessed at this point. Maybe I should add to headerchain?

light/odr.go
kevjue, line 133: Should this function verify the headers before saving it to the chaindb (https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/engine.go#L104)?

kevjue: Actually, looks like this header verification function should be used: https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/engine.go#L221.

les/odr_requests.go
lucasege: Yeah this was something I wasn't sure how to implement - this change is in `les/fetcher.peerHasBlock` to optimistically return true if the provided blocknumber is 0, otherwise it gets confused with the genesis block (`number` is uint64 so I can't provide a clearer sentinel value here).
This will lead to some false positives, but that seemed ok to me. Would appreciate your input

light/odr.go
lucasege: Not that I know of, usually `WriteTd` uses `Block.TotalDifficulty()` which just does this same thing `Number+1`

light/odr.go
lucasege: Currently this is done in `client_handler` via `lightchain.InsertHeaderChain` which calls `headerchain.insertHeaderChain` which calls `engine.verifyHeaders`
`Request.Validate` doesn't have a hook to the engine object from what I see so I don't know if I can do this here

lucasege: Don't have access to `HeaderChain` here

kevjue: Are the `StoreResult` writes also done here:  https://github.com/celo-org/celo-blockchain/pull/1031/files#diff-3e970814285a79d487cdf2e12822e0bcR244 in `client_handler`?

If so, would it make sense to remove the writes in `StoreResult`?

les/odr_requests.go
kevjue: Ah ok, the change if `fetcher.peerHasBlock` makes much more sense to me now.

I can't really think of any good ideas other than the obvious (e.g. change the `fetcher.peerHasBlock`'s `number` parameter to have type `*int64` and then check if it's nil).  That may be worth doing if there are not that many `fetcher.peerHasBlock` invocations that you need to change.

les/client_handler.go
```diff
@@ -232,9 +232,27 @@ func (h *clientHandler) handleMsg(p *peer) error {
 		p.fcServer.ReceivedReply(resp.ReqID, resp.BV)
 		if h.fetcher.requestedID(resp.ReqID) {
 			h.fetcher.deliverHeaders(p, resp.ReqID, resp.Headers)
-		} else {
-			if err := h.downloader.DeliverHeaders(p.id, resp.Headers); err != nil {
-				log.Debug("Failed to deliver headers", "err", err)
+		} else { // TODO(lucas): this is likely imperfect
+			h.backend.retriever.lock.RLock()
```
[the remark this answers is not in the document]
lucasege: I believe this is possible, seems like the odr code will try sending a request and if it fails it well keep sending to new peers, under the same request ID.
This seems fine to me, since even if a request gets answered twice it will still need to pass validation, and if both responses pass validation then we just update the header in the db twice.
Am I missing something?

les/odr_requests.go
lucasege: Implemented this now, please take a look. Makes a lot of things a bit uglier with `&`s and `*`s but works for my tests.

light/odr.go
lucasege: Yes this looks right, good catch. I'll modify `storeResult`

les/client_handler.go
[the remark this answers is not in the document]
kevjue: That makes sense.

light/odr.go
kevjue: I'm actually now wondering if the write to the header chain and write to the canonical hash store needs to be atomic.  I'd presume not (since I doubt it's done in upstream geth), but I think it would be worth looking into it a bit of what happens if they are inconsistent.  E.g. what happens if one is written and the other isn't (say from a client crash).

light/odr.go
```diff
@@ -119,6 +119,25 @@ func (req *BlockRequest) StoreResult(db ethdb.Database) {
 	rawdb.WriteBodyRLP(db, req.Hash, req.Number, req.Rlp)
 }
 
+type hashOrNumber struct {
```
kevjue, line 122: nit:  Consider renaming to `blockHashOrNumber`.

light/odr.go
lucasege: Yeah this is a good point, this does introduce the potential for inconsistent reads - the impact seems small but not 0.
Wonder if I should look into updating `InsertHeaderChain` or if there's another solution, will keep investigating

kevjue: Seems like there is an expectation that the header is in the chaindb if the canonicalhash is in the db.

https://github.com/celo-org/celo-blockchain/blob/master/light/odr_util.go#L35

Does your changes maintain that constraint?

kevjue:
Regarding testing, IMO I think adding some test cases to the e2e sync test (where mode == lightest) would provide good coverage.

light/odr.go
lucasege: I believe it does maintain this constraint, just not the other direction - if the canonical hash is in the db then the header is. 
The header being in the db does not necessarily mean the hash is in the db (since there is a delay before `storeResult` is called.
This is what I meant by thinking it's ok, but that just means I haven't found an area that makes it not Ok yet

lucasege:
> Regarding testing, IMO I think adding some test cases to the e2e sync test (where mode == lightest) would provide good coverage.

Will do, that's the plan. I'll add comments to this PR and then finish the e2e tests in another PR and send back to you

light/odr.go
kevjue: Sounds good!

kevjue:
> > Regarding testing, IMO I think adding some test cases to the e2e sync test (where mode == lightest) would provide good coverage.
> 
> Will do, that's the plan. I'll add comments to this PR and then finish the e2e tests in another PR and send back to you

One other test case that I just thought of is an adversarial test case where the client downloads a header with an invalid BLS seal.

Unfortunately this will be hard to add in the e2e test suite, but perhaps can be done via the unit tests.

lucasege:
> > > Regarding testing, IMO I think adding some test cases to the e2e sync test (where mode == lightest) would provide good coverage.
> > 
> > 
> > Will do, that's the plan. I'll add comments to this PR and then finish the e2e tests in another PR and send back to you
> 
> One other test case that I just thought of is an adversarial test case where the client downloads a header with an invalid BLS seal, and then rejects it.
> 
> Unfortunately this will be hard to add in the e2e test suite, but perhaps can be done via the unit tests.

Spent some time working on this, and judged that it was taking too much time and touching too many different files to make it a good fit for this PR. I've moved it to `lucasege/istanbul-simulation` and tracked in #1044 .

PR #1033, opened 2020-05-11
Fix/celostats

H34D:
### Description

- Some refactorings
- Changed error handling slightly on SendDelegateSignMsgToProxiedValidator
- Used name from `--ethstats=name@stats.celo.org` only as the name not as id now. So the name can now be empty
- Updated gorilla websockets
- transport latency as number not as string

### Related

https://github.com/celo-org/celo-monorepo/issues/3763


ethstats/ethstats.go
```diff
@@ -545,7 +544,13 @@ func (s *Service) readLoop(conn *websocket.Conn) {
 			log.Info("Ping from websocket server", "msg", msg)
 			// Primus server might want to have a pong or it closes the connection
 			var serverTime = fmt.Sprintf("primus::pong::%d", time.Now().UnixNano()/int64(time.Millisecond))
-			conn.WriteJSON(serverTime)
+			var pong = map[string]interface{}{
```
diminator, line 547: Is this coming from the Primus protocol?

ethstats/ethstats.go
```diff
@@ -621,7 +626,7 @@ func (s *Service) reportLatency(conn *websocket.Conn, sendCh chan *StatsPayload)
 		// Ping timeout, abort
 		return errors.New("ping timed out")
 	}
-	latency := strconv.Itoa(int((time.Since(start) / time.Duration(2)).Nanoseconds() / 1000000))
+	latency := int((time.Since(start) / time.Duration(2)).Nanoseconds() / 1000000)
```
diminator, line 629: Is this what Primus expects?

ethstats/ethstats.go
H34D: not this change is coming from the linter, writeJSON should write a JSON structure. I was never able to trigger this piece of code so it might make no difference.

ethstats/ethstats.go
H34D: no this is what celostats expects, this is the payload of the latency response. The value is converted to number on arrival at celostats to be used as number in latency calculation. To save this conversion it makes sense to transfer it as number from the start

ethstats/ethstats.go
diminator: from https://github.com/primus/primus#protocol
```
primus::pong::<ping> client -> server, The pong is the response to the ping packet. It echoes back the exact value that it received.
```

So the packet that Primus expects should be of the form `primus::pong::<ping>`, which I'm not sure now will be the case. Can you verify?

H34D: if this is the case WriteJSON is not the right method to do this. i will check

H34D: i changed it to write message

consensus/istanbul/backend/backend.go
```diff
@@ -301,13 +301,15 @@ func (sb *Backend) SendDelegateSignMsgToProxy(msg []byte) error {
 
 // SendDelegateSignMsgToProxiedValidator sends an istanbulDelegateSign message to a
 // proxied validator if one exists
-func (sb *Backend) SendDelegateSignMsgToProxiedValidator(msg []byte) error {
+func (sb *Backend) SendDelegateSignMsgToProxiedValidator(msg []byte) {
```
kevjue, line 304: Is there a specific reason that you removed the returned error?

ethstats/ethstats.go
```diff
@@ -97,14 +96,14 @@ type blockChain interface {
...
-	etherBase common.Address
-	node      string // Name of the node to display on the monitoring page
-	host      string // Remote address of the monitoring service
+	server        *p2p.Server              // Peer-to-peer server to retrieve networking infos
+	eth           *eth.Ethereum            // Full Ethereum service if monitoring a full node
+	les           *les.LightEthereum       // Light Ethereum service if monitoring a light node
+	engine        consensus.Engine         // Consensus engine to retrieve variadic block fields
+	backend       *istanbulBackend.Backend // Istanbul consensus backend
+	etherBase     common.Address
+	nodeName      string // Name of the node to display on the monitoring page
```
kevjue, line 105: I appreciate the renamings here.

ethstats/ethstats.go
```diff
@@ -395,10 +394,10 @@ func (s *Service) login(conn *websocket.Conn, sendCh chan *StatsPayload) error {
 		}
 	}
 	auth := &authMsg{
-		ID:      s.node,
+		ID:      s.backend.Address().String(),
```
kevjue, line 397: Is this line run by a proxy?  If so, not that the address returned here will be the proxy's etherbase's address.  Not the proxied validator address.

To get the proxied validator address, you need to use `backend.ValidatorAddress`.

ethstats/ethstats.go
```diff
@@ -599,7 +601,7 @@ func (s *Service) reportLatency(conn *websocket.Conn, sendCh chan *StatsPayload)
 	start := time.Now()
 
 	ping := map[string]interface{}{
-		"id":         s.node,
+		"id":         s.backend.Address().String(),
```
kevjue, line 604: See above comment.

ethstats/ethstats.go
```diff
@@ -621,13 +623,13 @@ func (s *Service) reportLatency(conn *websocket.Conn, sendCh chan *StatsPayload)
...
 	}
-	latency := strconv.Itoa(int((time.Since(start) / time.Duration(2)).Nanoseconds() / 1000000))
+	latency := int((time.Since(start) / time.Duration(2)).Nanoseconds() / 1000000)
 
 	// Send back the measured latency
 	log.Trace("Sending measured latency to ethstats", "latency", latency)
 
 	stats := map[string]interface{}{
-		"id":      s.node,
+		"id":      s.backend.Address().String(),
```
kevjue, line 632: See above comment.

ethstats/ethstats.go
```diff
@@ -756,7 +758,7 @@ func (s *Service) reportBlock(conn *websocket.Conn, block *types.Block) error {
 	log.Trace("Sending new block to ethstats", "number", details.Number, "hash", details.Hash)
 
 	stats := map[string]interface{}{
-		"id":    s.node,
+		"id":    s.backend.Address().String(),
```
kevjue, line 761: See above comment.

ethstats/ethstats.go
H34D: thanks

consensus/istanbul/backend/backend.go
H34D: yes, this method it called async, the error was never handled during async handling. So i decided to rather log the error instead of just hiding it by not handling it at all. Its not good, but its better than it was before.

ethstats/ethstats.go
H34D: The line should not be run by a proxy. The ID here is less important anyway because it will be overridden by the address of the account that was used to sign the auth, stats and block informations. It just can not be empty, otherwise celostats will fail.

ethstats/ethstats.go
H34D: see above answer

ethstats/ethstats.go
H34D: see above answer

ethstats/ethstats.go
H34D: see above answer

asaj:
@H34D does this still need to be merged?

H34D:
> @H34D does this still need to be merged?

yes!

H34D:
@asaj

PR #1035, opened 2020-05-12
Remove start time dependency for announce protocol and proxy

### LINKED ISSUE #947: Connecting to proxy and announce protocol shouldn't depend on istanbul.StartValidating function
### Expected Behavior

Connecting to proxy and announce protocol shouldn't depend on istanbul.StartValidating function to be called.  Specifically, the validator overlay network shouldn't require the validators to also be running the consensus protocol.

This is part of epic:  https://github.com/celo-org/celo-blockchain/issues/1012

### Current Behavior

Connecting to proxy and announce protocol depends on istanbul.StartValidating function to be called.

How can we reproduce this?

### Logs

Are there any logs?

### System information

Run `geth version`

kevjue:
### Description

This PR removes the announce protocol's and establishment of proxy connection's dependency that the machines local time is greater than or equal to the network start time set in the genesis file.

### Other changes

Started effort in splitting up some of the `istanbul.backend`'s bulky files.  Specifically, does the following:

1. moved all of the new eth message codes into `istanbul/protocol.go` and any related utility functions
1. created new file for the gossip caches
1. created new file for the message sender functions
1. Removed the `candidates` property from the istanbul rpc api

### Tested

1. Deployed a testnet with a start timestamp about half an hour in the future.  Then ensured that the validator connections were established beforehand and that the testnet started validating at the start timestamp.
2. Unit tests

### Related issues

- Fixes #947

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -419,7 +419,7 @@ func (sb *Backend) generateAndGossipQueryEnode(version uint, enforceRetryBackoff
 			return err
 		}
 
-		if err := sb.Multicast(nil, payload, istanbulQueryEnodeMsg); err != nil {
+		if err := sb.Gossip(payload, istanbul.QueryEnodeMsg); err != nil {
 			return err
 		}
 
@@ -501,7 +501,7 @@ func (sb *Backend) generateQueryEnodeMsg(version uint, queryEnodeDestAddresses [
 	}
 
 	msg := &istanbul.Message{
-		Code:      istanbulQueryEnodeMsg,
+		Code:      istanbul.QueryEnodeMsg,
 		Msg:       queryEnodeBytes,
 		Address:   sb.Address(),
 		Signature: []byte{},
@@ -544,10 +544,16 @@ func (sb *Backend) generateEncryptedEnodeURLs(enodeURL string, queryEnodeDestAdd
 }
 
 // This function will handle a queryEnode message.
-func (sb *Backend) handleQueryEnodeMsg(peer consensus.Peer, payload []byte) error {
+func (sb *Backend) handleQueryEnodeMsg(addr common.Address, peer consensus.Peer, payload []byte) error {
 	logger := sb.logger.New("func", "handleQueryEnodeMsg")
 
 	msg := new(istanbul.Message)
+
+	// Since this is a gossiped messaged, mark that the peer gossiped it and check to see if this node already gossiped it
+	sb.markPeerGossipCache(addr, payload)
+	if sb.checkSelfGossipCache(payload) {
+		return nil
+	}
 
 	// Decode message
 	err := msg.FromPayload(payload, istanbul.GetSignatureAddress)
@@ -713,7 +719,7 @@ func (sb *Backend) regossipQueryEnode(msg *istanbul.Message, msgTimestamp uint, 
 	}
 
 	logger.Trace("Regossiping the istanbul queryEnode message", "IstanbulMsg", msg.String())
-	if err := sb.Multicast(nil, payload, istanbulQueryEnodeMsg); err != nil {
+	if err := sb.Gossip(payload, istanbul.QueryEnodeMsg); err != nil {
 		return err
 	}
 
@@ -835,7 +841,7 @@ func (sb *Backend) encodeVersionCertificatesMsg(versionCertificates []*versionCe
 		return nil, err
 	}
 	msg := &istanbul.Message{
-		Code: istanbulVersionCertificatesMsg,
+		Code: istanbul.VersionCertificatesMsg,
 		Msg:  payload,
 	}
 	msgPayload, err := msg.Payload()
@@ -853,7 +859,7 @@ func (sb *Backend) gossipVersionCertificatesMsg(versionCertificates []*versionCe
 		logger.Warn("Error encoding version certificate msg", "err", err)
 		return err
 	}
-	return sb.Multicast(nil, payload, istanbulVersionCertificatesMsg)
+	return sb.Gossip(payload, istanbul.VersionCertificatesMsg)
 }
 
 func (sb *Backend) getAllVersionCertificates() ([]*versionCertificate, error) {
@@ -882,12 +888,18 @@ func (sb *Backend) sendVersionCertificateTable(peer consensus.Peer) error {
 		logger.Warn("Error encoding version certificate msg", "err", err)
 		return err
 	}
-	return peer.Send(istanbulVersionCertificatesMsg, payload)
+	return peer.Send(istanbul.VersionCertificatesMsg, payload)
 }
 
-func (sb *Backend) handleVersionCertificatesMsg(peer consensus.Peer, payload []byte) error {
+func (sb *Backend) handleVersionCertificatesMsg(addr common.Address, peer consensus.Peer, payload []byte) error {
 	logger := sb.logger.New("func", "handleVersionCertificatesMsg")
 	logger.Trace("Handling version certificates msg")
+
+	// Since this is a gossiped messaged, mark that the peer gossiped it and check to see if this node already gossiped it
+	sb.markPeerGossipCache(addr, payload)
+	if sb.checkSelfGossipCache(payload) {
+		return nil
+	}
 
 	var msg istanbul.Message
 	if err := msg.FromPayload(payload, nil); err != nil {
@@ -1043,7 +1055,7 @@ func (sb *Backend) setAndShareUpdatedAnnounceVersion(version uint) error {
 		destAddresses[i] = address
 		i++
 	}
-	err = sb.Multicast(destAddresses, payload, istanbulEnodeCertificateMsg)
+	err = sb.Multicast(destAddresses, payload, istanbul.EnodeCertificateMsg, false)
 	if err != nil {
 		return err
 	}
@@ -1139,7 +1151,7 @@ func (sb *Backend) generateEnodeCertificateMsg(version uint) (*istanbul.Message,
 		return nil, err
 	}
 	msg := &istanbul.Message{
-		Code:    istanbulEnodeCertificateMsg,
+		Code:    istanbul.EnodeCertificateMsg,
 		Address: sb.Address(),
 		Msg:     enodeCertificateBytes,
 	}
@@ -1159,7 +1171,7 @@ func (sb *Backend) generateEnodeCertificateMsg(version uint) (*istanbul.Message,
 // will send it back to this node.
 // If the proxied validator sends an enode certificate for itself to this node,
 // this node will set the enode certificate as its own for handshaking.
-func (sb *Backend) handleEnodeCertificateMsg(peer consensus.Peer, payload []byte) error {
+func (sb *Backend) handleEnodeCertificateMsg(_ common.Address, peer consensus.Peer, payload []byte) error {
 	logger := sb.logger.New("func", "handleEnodeCertificateMsg")
 
 	var msg istanbul.Message
@@ -1272,7 +1284,7 @@ func (sb *Backend) sendEnodeCertificateMsg(peer consensus.Peer, msg *istanbul.Me
 		logger.Error("Error getting payload of enode certificate message", "err", err)
 		return err
 	}
-	return peer.Send(istanbulEnodeCertificateMsg, payload)
+	return peer.Send(istanbul.EnodeCertificateMsg, payload)
 }
 
 func (sb *Backend) setEnodeCertificateMsg(msg *istanbul.Message) error {
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -260,7 +260,7 @@ func (sb *Backend) shouldSaveAndPublishValEnodeURLs() (bool, error) {
 		return false, err
 	}
 
-	return sb.coreStarted && validatorConnSet[sb.Address()], nil
+	return validatorConnSet[sb.Address()], nil
 }
 
 // pruneAnnounceDataStructures will remove entries that are not in the validator connection set from all announce related data structures.
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -554,6 +554,7 @@ func (sb *Backend) handleQueryEnodeMsg(addr common.Address, peer consensus.Peer,
 	if sb.checkSelfGossipCache(payload) {
 		return nil
 	}
+	defer sb.markSelfGossipCache(payload)
 
 	// Decode message
 	err := msg.FromPayload(payload, istanbul.GetSignatureAddress)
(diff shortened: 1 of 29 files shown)


kevjue:
@gastonponti - I remember that you requested to review this PR.

Note that the actual code changes for the start time dependency is just a fraction of the total code changes (as much of it is for the `Other Changes`).

For the start time dependency changes, you can view the following files:

https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-37def8a8c5ffccd7af569d23609a15ba
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-5b7bdb45b66a5017bf89b46e6ef7a7f4
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-3a5c8ea459d421751fb70343be214598
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-49675a1e6ab38a575ed8c1e9a5f31145R263
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-6e58dd6edf43343da874fb58a00f6727R285
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-e968de38f75ea48cdf62c94527dbec22
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-ceb9235ce1650af05288eb69654a4d7f
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-a30282077b8ab0dd89737f2efcb67a50
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-cea813cccd5fa85511c11995e7f5c8c6
https://github.com/celo-org/celo-blockchain/pull/1035/files#diff-e42690942282286ede9e53bbe7b0c248R548

consensus/istanbul/backend/handler.go
```diff
@@ -92,39 +73,16 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, peer consensus.Pe
...
+		if msg.Code == istanbul.ConsensusMsg {
 			err := sb.handleConsensusMsg(peer, data)
 			return true, err
-		} else if msg.Code == istanbulFwdMsg {
+		} else if msg.Code == istanbul.FwdMsg {
 			err := sb.handleFwdMsg(peer, data)
 			return true, err
 		} else if announceHandlerFunc, ok := sb.istanbulAnnounceMsgHandlers[msg.Code]; ok { // Note that the valEnodeShare message is handled here as well
-			go announceHandlerFunc(peer, data)
+			go announceHandlerFunc(addr, peer, data)
```
tkporter, line 83: It seems like `addr` is the address derived from the peer's node public key. Is this intentional? It feels like the desired behavior, but want to confirm.

I think it could also be helpful to make this more clear in the code that the address isn't necessarily the signer address

consensus/istanbul/protocol.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2017 The celo Authors
```
tkporter, line 1: This file is great!

consensus/istanbul/protocol.go
```diff
@@ -0,0 +1,59 @@
...
+	EnodeCertificateMsg    = 0x17
+	ValidatorHandshakeMsg  = 0x18
+)
+
+func IsIstanbulMsg(msg p2p.Msg) bool {
+	return msg.Code >= ConsensusMsg && msg.Code <= ValidatorHandshakeMsg
+}
+
+// IsGossipedMsg specifies which messages should be gossiped throughout the network (as opposed to directly sent to a peer).
+func IsGossipedMsg(msgCode uint64) bool {
```
tkporter, line 57: Is this used?

consensus/istanbul/backend/gossip_cache.go
```diff
@@ -0,0 +1,61 @@
...
+package backend
+
+import (
+	lru "github.com/hashicorp/golang-lru"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/consensus/istanbul"
+)
+
+func (sb *Backend) markPeerGossipCache(peerNodeAddr common.Address, payload []byte) {
```
tkporter, line 26: Generally I find the names of the functions in here a little confusing, I had to read the contents to get exactly what was happening

consensus/istanbul/backend/message_senders.go
```diff
@@ -0,0 +1,116 @@
...
+// Multicast will send the eth message (with the message's payload and msgCode field set to the params
+// payload and ethMsgCode respectively) to the nodes with the signing address in the destAddresses param.
+// If this node is proxied and destAddresses is not nil, the message will be wrapped
+// in an istanbul.ForwardMessage to ensure the proxy sends it to the correct
+// destAddresses.
+func (sb *Backend) Multicast(destAddresses []common.Address, payload []byte, ethMsgCode uint64, sendToSelf bool) error {
+	var err error
+
+	if sb.IsProxiedValidator() {
+		if err := sb.sendForwardMsgToProxy(destAddresses, ethMsgCode, payload); err != nil {
```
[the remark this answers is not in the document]
kevjue: No, it's not.  Nice catch.

consensus/istanbul/backend/message_senders.go
```diff
@@ -0,0 +1,116 @@
...
+// Gossip will gossip the eth message to all connected peers
+func (sb *Backend) Gossip(payload []byte, ethMsgCode uint64) error {
+	logger := sb.logger.New("func", "Gossip")
+
+	// Get all connected peers
+	peersToSendMsg := sb.broadcaster.FindPeers(nil, p2p.AnyPurpose)
+
+	// Mark that this node gossiped this message, so that it will ignore it if
+	// one of it's peers sends the message to it.
+	sb.markSelfGossipCache(payload)
```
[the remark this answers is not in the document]
kevjue: Good question.

It's done in both places to mark that a specific node is aware of a gossiped message.  E.g. let's say that the message originated from this node and gossips it out (which marks the cache here).  Then whenever this node gets the same message is should ignore it.

On the other hand, if this node gets a gossip message twice (say from two different neighbors), the first instance it'll mark the cache in the handler, and then ignore the second instance.

consensus/istanbul/protocol.go
kevjue: Nope.

consensus/istanbul/backend/handler.go
kevjue: The main purpose of `addr` is for the gossip cache, as the one to mark message's sent/received from specific peers is keys on `addr`.

That's a good point in making it clear of what address it is, and I'll add comments for that.  This probably should be changed to use `nodeID` instead, but I was keeping what the original istanbul implementation did ....

consensus/istanbul/backend/backend.go
```diff
@@ -992,3 +873,34 @@ func (sb *Backend) retrieveUncachedValidatorConnSet() (map[common.Address]bool,
...
+	}
+
+	// Note that we are not signing message.  The message that is being wrapped is already signed.
+	msg := istanbul.Message{Code: istanbul.FwdMsg, Msg: fwdMsgBytes, Address: sb.Address()}
+	fwdMsgPayload, err := msg.Payload()
+	if err != nil {
+		return err
+	}
+
+	go sb.proxyNode.peer.Send(istanbul.FwdMsg, fwdMsgPayload)
```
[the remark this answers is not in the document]
kevjue: Ya, I totally agree that we should be consistent.  However, do you have strong opinions of doing this now?  My plan is to "phase" some of this refactoring as feature/bug fix PRs are being pushed.  Things on the roadmap are:

1). Create new subpackage for proxy (this will be part of the multiproxy PR).
2). Create new subpackage for announce.
3). Do more splitting up of the backend package and settle on a multithreaded vs singlethreaded messages.

Btw, historically, we have run into problems b/c of the "single threaded"-ness of our core package.  So I think that a significant amount of thought should go into #3.

consensus/istanbul/backend/engine.go
```diff
@@ -668,6 +653,54 @@ func (sb *Backend) StopAnnouncing() error {
...
+
+	if !sb.IsProxiedValidator() {
+		return istanbul.ErrValidatorNotProxied
+	} else {
+		if sb.config.ProxyInternalFacingNode != nil && sb.config.ProxyExternalFacingNode != nil {
+			if err := sb.addProxy(sb.config.ProxyInternalFacingNode, sb.config.ProxyExternalFacingNode); err != nil {
+				sb.logger.Error("Issue in adding proxy on istanbul start", "err", err)
+			}
+		}
+		go sb.sendValEnodesShareMsgs()
```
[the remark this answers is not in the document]
kevjue: Renamed to `valEnodesShareThread`

consensus/istanbul/backend/gossip_cache.go
kevjue: renamed to the following:

`markPeerGossipCache` -> `markMessageProcessedByPeer`
`checkPeerGossipCache` -> `checkIfMessageProcessedByPeer`
`markSelfGossipCache` -> `markMessageProcessedBySelf`
`checkSelfGossipCache` -> `checkIfMessageProcessedBySelf`

consensus/istanbul/backend/message_senders.go
```diff
@@ -79,29 +77,28 @@ func (sb *Backend) Gossip(payload []byte, ethMsgCode uint64) error {
...
 
 	return sb.sendMsg(peersToSendMsg, payload, ethMsgCode)
 }
 
 // sendMsg will send the eth message (with the message's payload and msgCode field set to the params
-// payload and ethMsgCode respectively) to either the nodes with the signing address in the destAddresses param,
-// or to all the connected peers with this node (if gossip parameter set to true).
+// payload and ethMsgCode respectively) to all the nodes destPeers param.
 func (sb *Backend) sendMsg(destPeers map[enode.ID]consensus.Peer, payload []byte, ethMsgCode uint64) error {
 	logger := sb.logger.New("func", "multicast")
```
tkporter, line 103: tiny nit: this func has changed

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
tkporter: No strong opinion about this now and I agree this isn't a straightforward decision, so sounds good

consensus/istanbul/backend/announce.go
```diff
@@ -544,11 +544,18 @@ func (sb *Backend) generateEncryptedEnodeURLs(enodeURL string, queryEnodeDestAdd
...
 	logger := sb.logger.New("func", "handleQueryEnodeMsg")
 
 	msg := new(istanbul.Message)
 
+	// Since this is a gossiped messaged, mark that the peer gossiped it (and presumably processed it) and check to see if this node already processed it
+	sb.markMessageProcessedByPeer(addr, payload)
+	if sb.checkIfMessageProcessedBySelf(payload) {
+		return nil
+	}
+	defer sb.markMessageProcessedBySelf(payload)
```
tkporter, line 557: We might as well just call this here instead of deferring to avoid the chance of this message being gossipped to this node from a different peer, right? Same goes for `handleVersionCertificatesMsg`.

One interesting thing we should think about is if we are calling `markMessageProcessedBySelf` (either deferred or not deferred), is it possible that we could receive a message from someone who just joined the validator conn set but this node has not updated the cached validator conn set and then rejects the message at `if !validatorConnSet[msg.Address] {`?

Also if we want to be extra safe, because the announce handler functions are called in a goroutine we may want to add mutexes to these `markMessage*` functions. As far as I can tell there isn't too bad of a worst case if a race condition happens though

consensus/istanbul/backend/backend.go
```diff
@@ -992,3 +873,34 @@ func (sb *Backend) retrieveUncachedValidatorConnSet() (map[common.Address]bool,
 	logger.Trace("Returning validator conn set", "validatorsSet", validatorsSet)
 	return validatorsSet, nil
 }
+
+func (sb *Backend) sendForwardMsgToProxy(finalDestAddresses []common.Address, ethMsgCode uint64, payload []byte) error {
+	logger := sb.logger.New("func", "sendForwardMsgToProxy")
+	if sb.proxyNode.peer == nil {
```
tkporter, line 879: This is definitely outside the scope of this PR, but I've always wondered if we risk the peer being set and passing a nil check, and then the peer disconnects and becomes nil, and then `sb.proxyNode.peer.Send` panics. Do you have any thoughts on that?

consensus/istanbul/backend/engine.go
```diff
@@ -668,6 +653,54 @@ func (sb *Backend) StopAnnouncing() error {
...
+	defer sb.proxyHandlerMu.Unlock()
+
+	if sb.proxyHandlerRunning {
+		return istanbul.ErrStartedProxyHandler
+	}
+
+	if !sb.IsProxiedValidator() {
+		return istanbul.ErrValidatorNotProxied
+	} else {
+		if sb.config.ProxyInternalFacingNode != nil && sb.config.ProxyExternalFacingNode != nil {
```
tkporter, line 668: nit: because the if case returns, we could move this out of an else to make the code a little clear

tkporter:
Btw I ran the handshake e2e test with this branch and things look good, so as long as there aren't any big structural changes I'm comfortable with merging this before the e2e test is merged into master

consensus/istanbul/backend/announce.go
kevjue: > We might as well just call this here instead of deferring to avoid the chance of this message being gossipped to this node from a different peer, right? Same goes for `handleVersionCertificatesMsg`

I don't feel too strongly about doing a defer vs executing it here.  The reason I added the defer is so that we it's marked "processed" when it actually is processed.


> One interesting thing we should think about is if we are calling `markMessageProcessedBySelf` (either deferred or not deferred), is it possible that we could receive a message from someone who just joined the validator conn set but this node has not updated the cached validator conn set and then rejects the message at `if !validatorConnSet[msg.Address] {`?

Ya, great point.   In general, I see that there is a high level tradeoff between how aggressively we want these gossip messages to be propagated throughout the network and how much safeguards we want to put in to prevent an attacker to flood the network.  E.g. in the `if !validatorConnSet[msg.Address] {` case, that obviously is needed so that a non validator can't flood the network, but to your point, there is a small chance that enough validators are not synced up and the gossip message is not communicated to the whole network.

I think that the only solution to that is to make sure that retry messages are sent once in a while (similar to the retry of round change messages in the consensus code).   And this is done by the announce thread in that it will periodically send out a version message.

consensus/istanbul/backend/backend.go
kevjue: Ya, I'm planning to address this in the multi-proxy PR (which includes a major refactoring of the proxy code).

The plan is for the proxy handler to be the single thread set and unset the `peer` field, as well as send messages to that `peer`.  So that will handle any race conditions.

consensus/istanbul/backend/engine.go
kevjue: Ya.  Agreed.  I made the requested changes.

consensus/istanbul/backend/message_senders.go
kevjue: :+1:

PR #1038, opened 2020-05-13
Add RPC and LES support for setting and querying gateway fees

rayyuan:
### Description

Add support for allowing full node to set Gateway Fee, as well as allow light client to query Full Node Gateway Fees. Full nodes that are light servers also broadcast Gateway Fee or Etherbase Changes to Light Client Peers

### Tested

manual testing. network of 4 nodes ( 1 validator, 2 tx-node, 1 light client). Light client connected to the 2 tx-node light server peers. 

### Backwards compatibility

It is backwards compatible


eth/api_backend.go
```diff
@@ -310,5 +310,6 @@ func (b *EthAPIBackend) GatewayFeeRecipient() common.Address {
 }
 
 func (b *EthAPIBackend) GatewayFee() *big.Int {
-	return b.eth.GatewayFee()
+	//return b.eth.GatewayFee()
```
nategraf, line 313: Please revert this change.

eth/backend.go
```diff
@@ -97,8 +97,8 @@ type Ethereum struct {
 
 	networkID     uint64
 	netRPCService *ethapi.PublicNetAPI
-
-	lock sync.RWMutex // Protects the variadic fields (e.g. gas price and etherbase)
+	NodeID        string       // @rayyuan id of node
```
nategraf, line 100: As we discussed, you can have the client get the node ID from its tracking of the peer rather having t  
o store it here.

eth/backend.go
```diff
@@ -312,14 +313,10 @@ func (s *Ethereum) APIs() []rpc.API {
...
 	}
+
 	// Append any APIs exposed explicitly by the consensus engine
 	apis = append(apis, s.engine.APIs(s.BlockChain())...)
 
-	// Append any APIs exposed explicitly by the les server
-	if s.lesServer != nil {
-		apis = append(apis, s.lesServer.APIs()...)
-	}
-
```
nategraf, line 322: Why were the following few lines removed?

eth/backend.go
```diff
@@ -97,8 +97,8 @@ type Ethereum struct {
 
 	networkID     uint64
 	netRPCService *ethapi.PublicNetAPI
-
-	lock sync.RWMutex // Protects the variadic fields (e.g. gas price and etherbase)
+	NodeID        string       // @rayyuan id of node
+	lock          sync.RWMutex // Protects the variadic fields (e.g. gas price and etherbase)
```
nategraf, line 101: Do you mean "variable" fields?

eth/backend.go
```diff
@@ -563,6 +560,7 @@ func (s *Ethereum) GatewayFeeRecipient() common.Address { return common.Address{
 func (s *Ethereum) GatewayFee() *big.Int                { return common.Big0 }
 func (s *Ethereum) Synced() bool                        { return atomic.LoadUint32(&s.protocolManager.acceptTxs) == 1 }
 func (s *Ethereum) ArchiveMode() bool                   { return s.config.NoPruning }
+func (s *Ethereum) NetRPCService() *ethapi.PublicNetAPI { return s.netRPCService } //@rayyuan, need this to access the publicnetapi
```
nategraf, line 563: Why is this added?

internal/ethapi/api.go
```diff
@@ -1771,6 +1771,11 @@ func (s *PublicNetAPI) Listening() bool {
 	return true // always listening
 }
 
+//return p2p Server
```
nategraf, line 1774: Why is this function added?

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
 }
 
+//GatewayFee returns the current gateway fee of this light server
+func (api *PrivateLightServerAPI) GatewayFee() (gf *big.Int, err error) {
+	return api.server.handler.gatewayFee, nil
+}
+
+//SetGatewayFee allows this light server node to set a gateway fee
+func (api *PrivateLightServerAPI) SetGatewayFee(gf *big.Int) error {
+	if gf.Cmp(big.NewInt(0)) == -1 {
```
nategraf, line 65: `gf.Cmp(common.Big0) < 0` is prefered

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+		return errors.New("Invalid Gateway Fee Value")
```
nategraf, line 66: Error messages would be lower case. Also error messages are usually defined as module-level constants i
n Geth.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+	}
+	api.server.handler.gatewayFee = gf
+	if err := api.BroadcastGatewayFeeUpdate(); err != nil {
```
nategraf, line 69: You should compare with the existing gateway fee and only broadcast if it has changed.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+		return err
+	}
+	return nil
+}
+
+// SetEtherbase sets the etherbase of the miner. Not sure if there should be one here or not because theres already one in the eth api
+func (api *PrivateLightServerAPI) SetEtherbase(etherbase common.Address) common.Address { //could do the casting thing again for test. Also, sending push notif here. Generally assume not changing etherbase. but if it can, need more checks on cache get
```
nategraf, line 76: `SetEtherbase` should return `error` instead of `common.Address`

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+	api.server.handler.etherbase = etherbase
+	if etherbase != common.ZeroAddress {
```
nategraf, line 78: Setting the gateway fee recipient to common.ZeroAddress _should_ result in a broadcast. That would mean
 this peer is now accepting transactions without fee.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+		api.BroadcastGatewayFeeUpdate()
+	}
+	return api.server.handler.etherbase
+}
+
+//get the unique node id of current node
+func (api *PrivateLightServerAPI) NodeID() string {
```
nategraf, line 85: Getting the NodeID should not be part of the les protocol. That information can be retrieved via https:
//github.com/ethereum/go-ethereum/wiki/Management-APIs#admin_nodeinfo

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+	return api.server.handler.nodeID
+}
+
+//This sends messages to light client peers whenever this light server updates gateway fee.
+func (api *PrivateLightServerAPI) BroadcastGatewayFeeUpdate() error {
```
nategraf, line 90: Id say this should at least be renamed to `BroadcastGatewayFee` (it may be used even in absence of an update). Additionally, it should be moved out of the API and into a more core piece of the logic like we discussed.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+	lightClientPeerNodes := api.server.peers.AllLightClientPeers()
+	if len(lightClientPeerNodes) == 0 {
+		return errors.New("No light client peers")
```
nategraf, line 93: I would also be reasonable to not return an error here. In the case that there are no peers, doing nothing is the intended outcome. You could imagine that a node operator may occasionally call this API to make sure that their light clients are up to date, and it would be unhelpful to return an error to them in the cases where they temporary have no peers.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+	}
+
+	currGatewayFee := api.server.handler.gatewayFee
```
nategraf, line 96: There is no need to create variables here. It would be better to simply use the value directly.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+	currEtherbase := api.server.handler.etherbase
+	currNodeID := api.server.handler.nodeID
+	if currGatewayFee.Cmp(big.NewInt(0)) >= 0 && currEtherbase != common.ZeroAddress {
```
nategraf, line 99: Instead of nesting all of the logic in the `if` statement, return early on the negative of this condition and put the rest of the logic below. Go style generally tries to reduce nesting https://golang.org/doc/effective_go.html#if

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+		for _, lightClientPeer := range lightClientPeerNodes {
+			reply := lightClientPeer.ReplyGatewayFee(genReqID(), GatewayFeeResps{GatewayFee: currGatewayFee.Uint64(), Etherbase: currEtherbase, NodeID: currNodeID}) //this works as push notification
```
nategraf, line 101: Please break up this line so that it is easier to read.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+			if reply != nil {
```
nategraf, line 102: Simmilar to the comment above: use `if reply == nil { continue }` instead.

les/api.go
```diff
@@ -52,6 +55,70 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+					}
+				})
+			}
+		}
+	}
+
+	return nil
+}
+
+func (api *PrivateLightServerAPI) Etherbase() (eb common.Address, err error) {
```
nategraf, line 118: We should move away form calling this `Etherbase`. Better would be `GatewayFeeRecipient`.

les/api.go
```diff
@@ -352,3 +419,60 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+	return &LightClientAPI{le}
+}
+
+// PullPeerGatewayFees updates cache by pulling gateway fee peer nodes
+func (api *LightClientAPI) GatewayFeeCache() map[string]*gatewayFeeEtherbase {
+	return api.le.handler.gatewayFeeCache.gatewayFeeMap
+}
+
+// PullPeerGatewayFees updates cache by pulling gateway fee peer nodes
+func (api *LightClientAPI) PullPeerGatewayFees() error {
```
nategraf, line 438: The complimentary word to `Broadcast` should be `Query` or `Request`, although this isnt a big deal and if you feel strongly, leave it.

les/api.go
```diff
@@ -352,3 +419,60 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+	peerNodes := api.le.peers.AllPeers()
+	if len(peerNodes) == 0 {
+		return errors.New("No peers")
```
nategraf, line 441: Having 0 peers probably should not return an error here.

les/api.go
```diff
@@ -352,3 +419,60 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+		cost := peerNode.GetRequestCost(GetGatewayFeeMsg, int(1))
+		err := peerNode.RequestGatewayFee(reqId, cost)
+		if err != nil {
+			continue
+		}
+	}
+	return nil
+}
+
+func (api *LightClientAPI) MinPeerGatewayFee() map[string]*gatewayFeeEtherbase { //get min gwFee as well as corresponding etherbase from cache. Can't return a tuple for some reason
```
nategraf, line 456: As we discussed, I would recommend renaming this to `SuggestGatewayFee` to allow for more generic implementations in the backend. This will be important for users of light clients that want to use a ledger or other signing mechanism outside of Geth

les/api.go
```diff
@@ -352,3 +419,60 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+	gatewayFeeMap := api.GatewayFeeCache()
+	minMap := make(map[string]*gatewayFeeEtherbase)
```
nategraf, line 458: Why is the return value a map? Wont it only have one entry? If that is the case, a `struct` should be better.

les/api.go
```diff
@@ -352,3 +419,60 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+
+	if len(gatewayFeeMap) == 0 {
+		log.Info("No Peers")
```
nategraf, line 461: I don't think an error should be logged here, but returning one to the user makes sense.

les/api.go
```diff
@@ -352,3 +419,60 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+		return minMap
+	} else {
```
nategraf, line 463: There is no need to use `else` here. https://golang.org/doc/effective_go.html#if

les/api_backend.go
```diff
@@ -290,5 +289,5 @@ func (b *LesApiBackend) GatewayFeeRecipient() common.Address {
 
 func (b *LesApiBackend) GatewayFee() *big.Int {
 	// TODO(nategraf): Create a method to fetch the gateway fee values of peers along with the coinbase.
-	return eth.DefaultConfig.GatewayFee
+	return big.NewInt(23)
```
nategraf, line 292: Please clean up this change. I don't recall immediatly what should be here, but it should not be a constant.

les/client_handler.go
```diff
@@ -45,6 +46,47 @@ type clientHandler struct {
 	closeCh  chan struct{}
 	wg       sync.WaitGroup // WaitGroup used to track all connected peers.
 	syncDone func()         // Test hooks when syncing is done.
+
+	//@rayyuan
+	gatewayFeeCache *gatewayFeeCache
+}
+
+type gatewayFeeEtherbase struct {
```
nategraf, line 54: More information will likely be added to this type, especially because the gateway fee will need to be 
split into different parts for each supported fee currency in the future. As such a more generic name would be better. Something like `GatewayFeeInformation`

les/client_handler.go
```diff
@@ -45,6 +46,47 @@ type clientHandler struct {
...
+	GatewayFee *big.Int
+	Etherbase  common.Address
+}
+
+func newGatewayFeeEtherbase(gatewayFee *big.Int, etherbase common.Address) *gatewayFeeEtherbase {
```
nategraf, line 59: Because this `struct` has no private fields, there is not need to have this function. Simply create the
 object inline.

les/client_handler.go
```diff
@@ -45,6 +46,47 @@ type clientHandler struct {
...
+	ret := &gatewayFeeEtherbase{
+		GatewayFee: gatewayFee,
+		Etherbase:  etherbase,
+	}
+	return ret
+}
+
+type gatewayFeeCache struct {
+	mux           *sync.RWMutex
```
nategraf, line 68: The word "mux" means something specific and different. Calling this "mutex" or "lock" is better.

les/client_handler.go
```diff
@@ -45,6 +46,47 @@ type clientHandler struct {
...
+
+func newGatewayFeeCache() *gatewayFeeCache {
+	cache := &gatewayFeeCache{
+		mux:           new(sync.RWMutex),
+		gatewayFeeMap: make(map[string]*gatewayFeeEtherbase),
+	}
+	return cache
+}
+
+func (c *gatewayFeeCache) update(nodeID string, val *gatewayFeeEtherbase) error {
```
nategraf, line 80: I see that updates are done under a lock, but that reads are accessed directly. It is unsafe to access 
the values without a lock, so functions will need to be added for that as well.

les/client_handler.go
```diff
@@ -176,6 +220,7 @@ func (h *clientHandler) handle(p *peer) error {
 // peer. The remote connection is torn down upon returning any error.
 func (h *clientHandler) handleMsg(p *peer) error {
 	// Read the next message from the remote peer, and ensure it's fully consumed
+	p.Log().Info("CLIENT HandleMsg: Ray")
```
nategraf, line 223: Please make sure to clean this up

les/client_handler.go
```diff
@@ -353,6 +398,25 @@ func (h *clientHandler) handleMsg(p *peer) error {
...
+			Data      GatewayFeeResps
+		}
+
+		if err := msg.Decode(&resp); err != nil {
+			return errResp(ErrDecode, "msg %v: %v", msg, err)
+		}
+
+		p.fcServer.ReceivedReply(resp.ReqID, resp.BV)
+
+		currGatewayFee := big.NewInt(int64(resp.Data.GatewayFee))
```
nategraf, line 414: Always handle gateway fee as a bit.Int or an appropriate wire format (e.g. hexutil.Big) casting to and
 from `int64` is unsafe.

les/client_handler.go
```diff
@@ -353,6 +398,25 @@ func (h *clientHandler) handleMsg(p *peer) error {
...
+		currEtherbase := resp.Data.Etherbase
+		nodeID := resp.Data.NodeID
```
nategraf, line 416: As we discussed, it is better to get this information form the peer object (`p`) here.

les/costtracker.go
```diff
@@ -45,6 +45,7 @@ var (
 		SendTxV2Msg:            {0, 450000},
 		GetTxStatusMsg:         {0, 250000},
 		GetEtherbaseMsg:        {10000, 1},
+		GetGatewayFeeMsg:       {10000, 1},
```
nategraf, line 48: A quick look at the definition of this struct makes me think we should change this. It seems much cheap
er than the other overations here, and should have `0` base cost because it costs nothing to initialy get this info
. I would suggest something like `{0, 10000}`.

les/protocol.go
```diff
@@ -43,7 +43,7 @@ var (
 )
 
 // Number of implemented message corresponding to different protocol versions.
-var ProtocolLengths = map[uint]uint64{lpv2: 24, lpv3: 26}
+var ProtocolLengths = map[uint]uint64{lpv2: 24, lpv3: 28}
```
nategraf, line 46: We may need to define `lpv4` to maintain backward compatibility but I am not quite sure.

les/server_handler.go
```diff
@@ -55,6 +55,7 @@ const (
 	MaxTxSend                = 64  // Amount of transactions to be send per request
 	MaxTxStatus              = 256 // Amount of transactions to queried per request
 	MaxEtherbase             = 1
+	MaxGatewayFee            = 1000000 //harcoded temp limit for now @rayyuan
```
nategraf, line 58: This should probably be 1 simply because there is no reason for it to be any higher. (Why would someone
 request multiple gateway fee responses?)

les/server_handler.go
```diff
@@ -179,6 +184,7 @@ func (h *serverHandler) handle(p *peer) error {
 // peer. The remote connection is torn down upon returning any error.
 func (h *serverHandler) handleMsg(p *peer, wg *sync.WaitGroup) error {
 	// Read the next message from the remote peer, and ensure it's fully consumed
+	p.Log().Info("Arrived in HandleMsg: Ray")
```
nategraf, line 187: Please make sure to clean this up.

les/server_handler.go
```diff
@@ -865,9 +870,29 @@ func (h *serverHandler) handleMsg(p *peer, wg *sync.WaitGroup) error {
...
+	case GetGatewayFeeMsg:
+		p.Log().Trace("Received gatewayFee request")
+		var req struct {
+			ReqID uint64
+		}
+		if err := msg.Decode(&req); err != nil {
+			return errResp(ErrDecode, "msg %v: %v", msg, err)
+		}
+
+		if h == nil || h.gatewayFee == nil {
```
nategraf, line 882: It is assumed that `h` is non-nil here, or if it is a very bad error has occured. You do not need to c
heck that `h` is `nil`. If `gatewayFee` is `nil`, a response should still be sent. The response should indicate tha
t ethe gateway fee is `0`

les/server_handler.go
```diff
@@ -865,9 +870,29 @@ func (h *serverHandler) handleMsg(p *peer, wg *sync.WaitGroup) error {
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				reply := p.ReplyGatewayFee(req.ReqID, GatewayFeeResps{GatewayFee: h.gatewayFee.Uint64(), Etherbase: h.etherbase, NodeID: h.nodeID})
+				sendResponse(req.ReqID, 1, reply, 10)
+			}()
+		}
 
 	default:
-		p.Log().Trace("Received invalid message", "code", msg.Code)
```
nategraf, line 870: Why was the log message on this line removed?

eth/backend.go
rayyuan: I removed this part because I think it's a duplicate caused by a merge or something. The les APIs are already appended in line 311

eth/backend.go
rayyuan: I didn't add the sync.RWMutex. It was a field from before

les/client_handler.go
rayyuan: for reads, would it be okay to just do h.gatewayFeeCache.mutex.Lock(), or should I have another function for it?

nategraf: You can use the same lock method. In reads are expected to be **much** more common than writes you could consider using a `RWMutex` (https://golang.org/pkg/sync/#RWMutex) instead, but I would not worry about that too much unless performance is shown to be an issue in production.

les/protocol.go
```diff
@@ -77,6 +77,9 @@ const (
 	// Protocol messages introduced in LPV3
 	StopMsg   = 0x18
 	ResumeMsg = 0x19
+	//@rayyuan gateway fee msgs
```
nategraf, line 80: Make sure you clean these kinds of things up
```suggestion
```

internal/web3ext/web3ext.go
```diff
@@ -822,6 +838,22 @@ web3._extend({
...
 			getter: 'les_serverInfo'
 		}),
+		new web3._extend.Property({
+			name: 'gatewayFee',
+			getter: 'les_gatewayFee'
+		}),
+		new web3._extend.Property({
+			name: 'suggestGatewayFee',
+			getter: 'les_suggestGatewayFee'
+		}),
```
nategraf, line 848: I would recommend moving this from properties to methods. Because the suggestion could conceivable be computationally non-trivial or return a non-deterministic result, it would be better to use the semantics of calling a method rather than treating it as a property access.

les/api.go
```diff
@@ -52,6 +55,40 @@ func NewPrivateLightServerAPI(server *LesServer) *PrivateLightServerAPI {
...
+		api.server.handler.gatewayFee = gf
+		if err := api.server.BroadcastGatewayFeeInfo(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// SetEtherbase sets the etherbase of the full node
+func (api *PrivateLightServerAPI) SetEtherbase(etherbase common.Address) error {
```
nategraf, line 78: `SetEtherbase` should be changed to `SetGatewayFeeRecipient` to match the method below and to disambiguate with https://github.com/ethereum/go-ethereum/wiki/Management-APIs#miner_setetherbase

les/api.go
```diff
@@ -352,3 +389,38 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+type LightClientAPI struct {
+	le *LightEthereum
+}
+
+func NewLightClientAPI(le *LightEthereum) *LightClientAPI {
+	return &LightClientAPI{le}
+}
+
+func (api *LightClientAPI) GatewayFeeCache() map[string]*GatewayFeeInformation {
+	return api.le.handler.gatewayFeeCache.gatewayFeeMap
```
nategraf, line 403: As we discussed,  the gateway map is guarded by a mutex to prevent concurrent access (and the `panic` it causes). Returning the map from this method throws that protection out the window and allows the caller, in this case the RPC server code, to access to the map freely. If a write operation were to be made to the map while it is being accessed to produce the JSON serialized form, a panic would occur. If you want to return this map, it needs to be copied before being released.

les/api.go
```diff
@@ -352,3 +389,38 @@ func (api *PrivateLightAPI) GetCheckpointContractAddress() (string, error) {
...
+}
+
+// RequestPeerGatewayFees updates cache by pulling gateway fee peer nodes
+func (api *LightClientAPI) RequestPeerGatewayFees() error {
+	peerNodes := api.le.peers.AllPeers()
+	for _, peerNode := range peerNodes {
+		cost := peerNode.GetRequestCost(GetGatewayFeeMsg, int(1))
+		err := peerNode.RequestGatewayFee(genReqID(), cost)
+		if err != nil {
+			continue
```
nategraf, line 413: This error should not be ignored. it should at least be logged to the Geth console. Also, as written this `continue` can be omitted because this is the last statement of the loop.

les/client_handler.go
```diff
@@ -373,6 +418,31 @@ func (h *clientHandler) removePeer(id string) {
 	h.backend.peers.Unregister(id)
 }
 
+func (h *clientHandler) MinPeerGatewayFee() (map[string]*GatewayFeeInformation, error) {
```
nategraf, line 421: Based on the way nature of this method, I would change it to be a method of the cache instead of the `clientHandler`.

les/odr_requests.go
```diff
@@ -307,6 +308,11 @@ type HelperTrieResps struct { // describes all responses, not just a single one
 	AuxData [][]byte
 }
 
+type GatewayFeeInformation struct {
```
nategraf, line 311: Why is this `struct` defined here? Because it is not used in this file, it seems like there is probably a better place for it.

les/peer.go
```diff
@@ -976,6 +992,30 @@ func (ps *peerSet) AllPeers() []*peer {
 	return list
 }
 
+//Ray: Get all peers that only support les and have no other caps.
```
nategraf, line 995: I just found a better way to make this check. A peer is a client if `p.fcServer != nil` https://github.com/celo-org/celo-blockchain/blob/7d6b3c511560f66e94ae4d3ae2ea903975bde569/les/peer.go#L115

internal/web3ext/web3ext.go
```diff
@@ -838,14 +842,6 @@ web3._extend({
 			name: 'serverInfo',
 			getter: 'les_serverInfo'
 		}),
-		new web3._extend.Property({
-			name: 'gatewayFee',
-			getter: 'les_gatewayFee'
-		}),
```
nategraf: Was removing this an accident?

les/api.go
```diff
@@ -74,8 +74,8 @@ func (api *PrivateLightServerAPI) SetGatewayFee(gf *big.Int) error {
 	return nil
 }
 
-// SetEtherbase sets the etherbase of the full node
-func (api *PrivateLightServerAPI) SetEtherbase(etherbase common.Address) error {
+// SetGatewayFeeRecipient sets the etherbase of the Gateway Fee Recipient
```
nategraf, line 77:
```suggestion
// SetGatewayFeeRecipient sets the etherbase of the gateway fee recipient
```

les/api.go
```diff
@@ -410,15 +410,15 @@ func (api *LightClientAPI) RequestPeerGatewayFees() error {
...
+			return err
 		}
 	}
 	return nil
 }
 
 // SuggestGatewayFee suggests the best light server to choose based on different factors. Currently only minPeerGatewayFee.
-func (api *LightClientAPI) SuggestGatewayFee() (map[string]*GatewayFeeInformation, error) {
-	bestGatewayFeeInfo, err := api.le.handler.MinPeerGatewayFee()
+func (api *LightClientAPI) SuggestGatewayFee() (*FullGatewayFeeInformation, error) {
```
nategraf, line 420: You don't need to return the node ID from this method. Instead it should just return `GatewayFeeInformation` and  `FullGatewayFeeInformation` can be removed.

les/odr_requests.go
```diff
@@ -308,10 +307,10 @@ type HelperTrieResps struct { // describes all responses, not just a single one
...
 }
 
-type GatewayFeeInformation struct {
-	GatewayFee *big.Int
-	Etherbase  common.Address
-}
+// type GatewayFeeInformation struct {
+// 	GatewayFee *big.Int
+// 	Etherbase  common.Address
+// }
```
nategraf, line 313:
```suggestion
```
Please remove this instead of just commenting it out.

internal/web3ext/web3ext.go
rayyuan: im not sure when i added that but there is no gatewayFee method in the les api

les/api.go
rayyuan: okay I'll remove that struct

Issue #1005, opened 2020-04-23
BLS signing delays longer than expected


mstraka100:
### Expected Behavior

When running a 100-validator network with a BLS signing delay, blocktimes will not increase significantly more than the length of this delay

### Current Behavior

When running a 100-validator network with a 2-second BLS signing delay and 5s targeted blocktimes, every 20-30 blocks will show a blocktime of ~1 minute. It is possible but rare for blocks to take between 12 and 50 seconds. Occasionally, a block `n+1` taking ~12 seconds will occur immediately after a block `n` taking ~60 seconds.

This behavior does not occur consistently, as the network will over some periods of time have no delays while achieving 5s blocktimes. This is always the case with 10 validators and a 2-second signing delay.

Similar behavior has been observed with:

- an otherwise identical 100-validator testnet with a 0.7s signing delay
- an otherwise identical 100-validator testnet with a 2s signing delay, which each validator at startup has a 50% chance of being assigned. This is to simulate 50% of validators using a ledger wallet for BLS signatures

Both of the above networks result in less frequent and less severe delays on average, but not always.

### Steps to Reproduce Behavior

Run a testnet with 100 validators, blocktime 5 seconds, and timeout of 3000ms, on this commit: https://github.com/celo-org/celo-blockchain/commit/ee29b2ef583a6a1c581050204e8bc6a71ede2db0

### Logs

This issue appears to occur at the same time as the error described here: https://github.com/celo-org/celo-blockchain/issues/1004

### System information

```Celo
Version: 0.10.0-unstable
Git Commit: ee29b2ef583a6a1c581050204e8bc6a71ede2db0
Architecture: amd64
Protocol Versions: [65 64]
Go Version: go1.13.6
Operating System: darwin
GOPATH=
GOROOT=/usr/local/go```

nategraf:
@trianglesphere do you think is related to the round change behavior you've been reseaching?

trianglesphere:
Depending on the load, this could cause a cascading failure through the first couple rounds, but I suspect that is also has something to do with the `inconsistent PREPARE` that is seen when this occurs. I've actually seen that message on baklava once or twice, but wasn't able to track down which node it came from. I think that issue is more likely to result in timeouts and round changes.

trianglesphere:
I believe that is a combination of non-exponential early round changes (to be fixed) and the inconsistent prepare bug (fixed)

PR #1039, opened 2020-05-14
[WIP] BLS separate thread

### LINKED ISSUE #1020: Make BLS signing done in separate thread
### Expected Behavior

The BLS signing should be done in separate thread.

### Current Behavior

Currently, the BLS signing is done within the single threaded istanbul core engine.  Since it's expected that BLS signing can take about 1 second, the signing essentially freezes the core engine, causing round changes in the subsequent sequence.  See https://github.com/celo-org/celo-blockchain/issues/1005 for more details.

See https://github.com/celo-org/celo-blockchain/issues/1005

mrsmkl:
### Description

Just added a new thread so that committing won't block if signing takes too long.
If there are several queued requests, the older ones are ignored and only the latest processed. I think this behavior might cause the validator to miss parent seal signatures but maybe there are no other effects.

### Tested

Need to do testing as described here https://github.com/celo-org/celo-blockchain/issues/1005

### Related issues

- Fixes #1020 

### Backwards compatibility

Yes


mcortesi:
old

PR #1041, opened 2020-05-18
Disables fast sync

kobigurk:
### Description

It seems a few people have stumbled upon BAD BLOCK situations when using fast sync. I've verified that running a fast synced node starts failing on an epoch block on RC1. I think it's best we disable it until we investigate it further.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -200,7 +200,7 @@ var (
 	defaultSyncMode = eth.DefaultConfig.SyncMode
 	SyncModeFlag    = TextMarshalerFlag{
 		Name:  "syncmode",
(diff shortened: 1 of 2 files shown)


gakonst:
This is unrelated to https://github.com/celo-org/celo-blockchain/pull/1022, it has been around for a while. Can we bisect the commit history and find when this first started happening?

kobigurk:
I suspect this is here for a long time... To clarify further, this happens
even on the RC1 docket imagem
The problem is that currently it takes a long time to reproduce - you need
a fast sync node that is already synced, and them stumble upon an epoch
block. I'm guessing we can probably reproduce it with a test chain.

nategraf:
It looks like https://github.com/celo-org/celo-blockchain/pull/1081 has fixed fast sync, so this PR is no longer needed.

PR #1045, opened 2020-05-21
Allow sending transactions to all suitable LES server peers

nategraf:
### Description

Currently when a transaction is sent by an LES client to a server for relay to the network, the client will pick a single node and send the transaction. If the node rejects the transaction, or the request otherwise fails, it will not be sent to any other node and the client operator is not informed.

This PR adjusts the LES transaction relay to send the transaction to any peer that is expected to transmit it to the network, and if that client rejects the transaction or does not respond within 500 ms, it will retry the request with a new peer according to the `les/retreive.go` logic.

### Other changes

* Moved gatewayFee field from tx relay to peer, where it can stay in the future.
* Changed the return type of `les/txRelay.CanRelayTransaction` to `bool`
* Changed the return like of `les/peer.Etherbase` to `common.Address, bool` to encode directly whether or not the etherbase is know. Same for `les/peer.GatewayFee`
* Revert change of `lpv3` protocol length, which could have caused issues with existing nodes
* Add `lpv4` entry as a WIP new protocol version which will support gateway fee queries
* Have CI skip building `docs` in monorepo

### Tested

E2e transfer test and a new unit test for the new function in `les/peer.go`

### Related issues

- Related https://github.com/celo-org/celo-monorepo/issues/3654
  - This PR will cut down on transaction send timeouts due caused by losing connection to a node, or that node otherwise failing.

### Backwards compatibility

100%


les/peer.go
```diff
@@ -780,6 +794,34 @@ func (p *peer) Handshake(td *big.Int, head common.Hash, headNum uint64, genesis
...
+	// Retrieve the gateway fee information known for this peer.
+	gatewayFee := p.GatewayFee()
+	var etherbase common.Address
+	if ptr := p.Etherbase(); ptr != nil {
+		etherbase = *ptr
+	}
+
+	// Check that the transaction meets the peer's gateway fee requirements.
+	// Treat nil (i.e. unknown) peer gateway fee information as possibly free relay.
+	if etherbase != (common.Address{}) && gatewayFee != nil && gatewayFee.Cmp(common.Big0) > 0 {
```
mcortesi, line 814: I remember there was a tricky interpretation on `etherbase`, something like:

```
etherbase == nil => not initialized yet (waiting from msg from the server)
etherbase == ZeroAddr => accept any tx
etherbase == someAddr => only accept tx with someAddr as feeRecipient
```

it's very subtle, and easy to miss... if possible i would make that clear now that we are doing more things with the etherbase.

les/txrelay.go
```diff
@@ -68,93 +55,92 @@ func (self *lesTxRelay) Stop() {
...
 				return func() { peer.SendTxs(reqID, cost, enc) }
 			},
 		}
-		go self.retriever.retrieve(context.Background(), reqID, rq, func(p distPeer, msg *Msg) error { return nil }, self.stop)
+
+		// Check the response to see if the transaction was successfully added to the peer pool or mined.
+		// If an error is returned, the retreiver will retry with any remaining suitable peers.
+		checkTxStatus := func(p distPeer, msg *Msg) error {
+			if msg.MsgType != MsgTxStatus {
+				return errors.New("received unexpected messgae code")
```
mcortesi, line 125: type `message`

les/txrelay.go
```diff
@@ -68,93 +55,92 @@ func (self *lesTxRelay) Stop() {
...
+			status := statuses[0]
+			if status.Error != "" {
+				return errors.New(status.Error)
+			}
+			if status.Status == core.TxStatusUnknown {
+				return errors.New("transaction status unknown")
+			}
+			return nil
+		}
+		go self.retriever.retrieve(context.Background(), reqID, rq, checkTxStatus, self.stop)
```
mcortesi, line 143: `context.Background()` means we won't apply any kind of timeout logic to the request, are we sure about that?

les/txrelay.go
```diff
@@ -68,93 +55,92 @@ func (self *lesTxRelay) Stop() {
...
 
+		// Send a single transaction per request to avoid failure coupling and
+		// because the expected base cost of a SendTxV2 request is 0, so it
+		// cost no extra to send multiple requests with one transaction each.
+		list := types.Transactions{tx}
+		enc, _ := rlp.EncodeToBytes(list)
+
+		// Assemble the request object with callbacks for the distributor.
 		reqID := genReqID()
 		rq := &distReq{
```
mcortesi, line 106: iiuc, what's happening here is that we will relay the tx to every peer that accepts it? or the the first one that accepts it?
If it's the second case, that means we will wait for the response on every peer before trying with the second one..

What's the case here?

les/txrelay.go
nategraf: I've found this part of the code to be a bit of a rats nest as I have been diving into it. The original authors chose to pass through a context variable, but use a different mechanism for cancelations and timeouts. I looked at changing it with the PR, but decided it was too much to change for this piece of work, but I am still workin in this part of the code, so I can still change it in a following PR.

les/peer.go
nategraf: Here is a bit of a simplification 148acf3a75b44f241ce06aa157c779a1821612a8. In that commit I've changes it such that the `Etherbase` and `GatewayFee` methods return an `ok` boolean to indicate whether the values are known.

Issue #820, opened 2020-01-22
Move `GetValidators` and `EpochSize` to Istanbul consensus specific interface


kevjue:
### Expected Behavior

The methods `GetValidators` and `EpochSize` should be in the Istanbul consensus specific interface:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/consensus.go#L168

### Current Behavior

They are currently in the "algorithm agnostic" consensus interface:  https://github.com/celo-org/celo-blockchain/blob/master/consensus/consensus.go#L56

### Steps to Reproduce Behavior

N/A

### Logs

N/A 

### System information

N/A

ocotopdf:
Hey @kevjue

Looks like  the EVM `Context` for `contracts.go` is using the `Engine` interface.

Changing this context to Istanbul necessitates a refactor wherever the EVM context is used. Is that expected?

Thanks,
Erik

kevjue:
Closing this issue, as it's not relevant anymore after this PR:  https://github.com/celo-org/celo-blockchain/pull/913.

PR #1047, opened 2020-05-25
vm, consensus move EpochSize, GetValidators

ocotopdf:
### Description

Moving `EpochSize()` and `GetValidators` to `Istanbul` interface. for #820 

@kevjue am I on the right track here?

### Tested

Ran `make lint`
Ran `make test`

Got `panic: runtime error: invalid memory address or nil pointer dereference [recovered]`
but unsure how this relates to my specific changes.

Unsure how to perform `End to end sync and transfer tests` locally. Is there documentation regarding this?

### DIFF
--- a/cmd/geth/retesteth.go
+++ b/cmd/geth/retesteth.go
@@ -189,6 +189,7 @@ type SRItem struct {
 
 type NoRewardEngine struct {
 	inner     consensus.Engine
+	istanbul  consensus.Istanbul
 	rewardsOn bool
 }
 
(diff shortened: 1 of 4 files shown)


kevjue:
Hey @ocotopdf - Thanks for creating this PR and really sorry for the late reply!

We actually decided to remove all of the consensus engines that are not relevant to us, such as ethash in this PR (https://github.com/celo-org/celo-blockchain/pull/913).

So the validator set and epoch size is now "first class" concepts for consensus in our client, and it doesn't really make sense to move those functions from the `consensus` interface.  So I'm going to close this PR and the issue that I initially created.

However, the cLabs team definitely appreciates your effort here.  Let us know if you would still like to contribute to the Celo client.  We are currently going through an effort to figure out good "first time" issues.

PR #1048, opened 2020-05-27
Update go.mod

malikankit:
Fix incorrect module name for celo-blockchain

### Description

I was wondering if the module name in go.mod is an intentional hack. So I asked on Discord  and was told that it's probably an oversight. Hence, submitting a PR. 

```
@AM This looks like an oversight since the code is itself a fork. You could open an issue on gh to fix it. or a Pull request 

https://discord.com/channels/600834479145353243/600840423958904842/712160929504297020
```

### Other changes

None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/ethereum/go-ethereum
+module github.com/celo-org/celo-blockchain
 
 go 1.13
 


bowd:
@malikankit thanks for your PR

Currently the celo client as a go module still has to be named `github.com/ethereum/go-ethereum` even if that's not really where it is. This is because of how dependent packages are included (otherwise we would need to rename ethereum/goethereum to celo-org/celo-blockchain everywhere in the project), and the fact that we still merge changes from `geth` (if we did rename everywhere that would be a pain). And that's just scratching the surface. We're still working out exactly how we want to handle the celo client being a fork of geth but we're not ready to change the module name right now, and it isn't a trivial change.

That being said, if you're interested in this I'd urge you to think about the broader implication of changing the module name (import path of packages, build scripts, and how to handle merging changes in geth post module name change) and start a conversation around a potential solution.

PR #1050, opened 2020-05-29
Add eth_gasPriceInCurrency to rpc for gas price in different fee currencies

yorhodes:
### Description

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/accounts/abi/bind/backend.go
+++ b/accounts/abi/bind/backend.go
@@ -74,6 +74,9 @@ type ContractTransactor interface {
 	// SuggestGasPrice retrieves the currently suggested gas price to allow a timely
 	// execution of a transaction.
 	SuggestGasPrice(ctx context.Context) (*big.Int, error)
+	// SuggestGasPriceFromCurrency retrieves the currently suggested gas price to allow
+  // a timely execution of a transaction for an specific currency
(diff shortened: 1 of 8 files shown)


accounts/abi/bind/backend.go
```diff
@@ -74,6 +74,9 @@ type ContractTransactor interface {
 	// SuggestGasPrice retrieves the currently suggested gas price to allow a timely
 	// execution of a transaction.
 	SuggestGasPrice(ctx context.Context) (*big.Int, error)
+	// SuggestGasPriceInCurrency retrieves the currently suggested gas price to allow
+	// a timely execution of a transaction for an specific currency
+	SuggestGasPriceInCurrency(ctx context.Context, account common.Address) (*big.Int, error)
```
nategraf, line 79:
```suggestion
	SuggestGasPriceInCurrency(ctx context.Context, currency *common.Address) (*big.Int, error)
```
Currency should be of type `*common.Address`. `nil` is the value in transactions that are paid in cGLD, and the zero address is an invalid currency value. Is it not possible to use that type in RPC?

yorhodes:
closing in favor of https://github.com/celo-org/celo-blockchain/pull/1070

PR #1069, opened 2020-06-08
Improve handling of invalid TobinTax return values

asaj:
### Description

Adds an additional layer of safety in celo-blockchain to handle when the returned tobin tax is undefined or greater than one. In these cases a normal transfer with zero tobin tax will be performed.

In the past, an undefined tobin tax would cause a panic. 

### Other changes

None.
### Tested

Not testede
### Related issues

None
### Backwards compatibility

Yes, as long as there have been no transactions for which a Tobin Tax was > 1 or had denominator == 0.


core/vm/evm.go
```diff
@@ -536,26 +537,43 @@ func (evm *EVM) TobinTransfer(db StateDB, sender, recipient common.Address, gas
...
+				log.Trace("TobinTransfer: Error calling getOrComputeTobinTaxFunctionSelector", "error", err)
+			} else {
+				numerator, denominator, err := parseTobinTax(ret)
+				if err != nil {
+					// The Tobin Tax should always be parsable.
+					log.Error("TobinTransfer: Error calling getOrComputeTobinTaxFunctionSelector", "error", err, "ret", ret)
+				} else {
+					tobinTax := new(big.Int).Div(new(big.Int).Mul(numerator, amount), denominator)
+					evm.Context.Transfer(db, sender, recipient, new(big.Int).Sub(amount, tobinTax))
+					evm.Context.Transfer(db, sender, *reserveAddress, tobinTax)
```
shellygr, line 574: Is it possible that only 1 of the transfers goes through?

asaj: It shouldn't be possible as `Transfer` does not return anything

core/vm/evm.go
```diff
@@ -536,26 +537,43 @@ func (evm *EVM) TobinTransfer(db StateDB, sender, recipient common.Address, gas
...
+			parseTobinTax := func(ret []byte) (*big.Int, *big.Int, error) {
+				// Expected size of ret is 64 bytes because getOrComputeTobinTax() returns two uint256 values,
+				// each of which is equivalent to 32 bytes
+				if binary.Size(ret) != 64 {
+					return nil, nil, errors.New("Length of tobin tax not equal to 64 bytes")
+				}
 				numerator := new(big.Int).SetBytes(ret[0:32])
 				denominator := new(big.Int).SetBytes(ret[32:64])
-				tobinTax := new(big.Int).Div(new(big.Int).Mul(numerator, amount), denominator)
+				if big.NewInt(0).Cmp(denominator) == 0 {
```
mcortesi, line 554: nit: you can use `common.Big0` instead of `big.NewInt(0)`

core/vm/evm.go
```diff
@@ -536,26 +537,43 @@ func (evm *EVM) TobinTransfer(db StateDB, sender, recipient common.Address, gas
 
 	if amount.Cmp(big.NewInt(0)) != 0 {
 		reserveAddress, err := GetRegisteredAddressWithEvm(params.ReserveRegistryId, evm)
-		if err != nil && err != errors.ErrSmartContractNotDeployed && err != errors.ErrRegistryContractNotDeployed {
+		if err != nil && err != contract_errors.ErrSmartContractNotDeployed && err != contract_errors.ErrRegistryContractNotDeployed {
 			log.Trace("TobinTransfer: Error fetching Reserve address", "error", err)
 		}
 
 		if err == nil {
```
mcortesi, line 544: nit: readability

Instead of 3 nested `ifs` you can do:

```go
if err != nil {
  return
}
//...continue
```

core/vm/evm.go
```diff
@@ -536,26 +537,43 @@ func (evm *EVM) TobinTransfer(db StateDB, sender, recipient common.Address, gas
 
 	if amount.Cmp(big.NewInt(0)) != 0 {
 		reserveAddress, err := GetRegisteredAddressWithEvm(params.ReserveRegistryId, evm)
```
mcortesi, line 539: lacking a bit of context here, but can we have a func in `contract_comm` instead of inlining `getOrComputeTobinTax()` here

core/vm/evm.go
```diff
@@ -536,26 +537,43 @@ func (evm *EVM) TobinTransfer(db StateDB, sender, recipient common.Address, gas
 
 	if amount.Cmp(big.NewInt(0)) != 0 {
```
mcortesi, line 538: I suggest we change this logic to something like:

```go
num, den, err := getTobinTax()

if err == nil || num > 0 {
  tobinTax := new(big.Int).Div(new(big.Int).Mul(numerator, amount), denominator)
	evm.Context.Transfer(db, sender, recipient, new(big.Int).Sub(amount, tobinTax))
	evm.Context.Transfer(db, sender, *reserveAddress, tobinTax)  
} else {
  evm.Context.Transfer(db, sender, recipient, amount)
}
```

core/vm/evm.go
asaj: I'm actually not sure I can, since we don't want to exit the function if there's an error. Instead, we want to complete the transfer *without* charging the tobin tax.

core/vm/evm.go
asaj: I don't think so due to circular dependencies (`contract_comm` depends on `core/vm`)

core/vm/evm.go
asaj: Broke `getTobinTax()` out into its own function which I think will improve readability, wdyt?

core/vm/evm.go
```diff
@@ -526,6 +525,33 @@ func (evm *EVM) Create2(caller ContractRef, code []byte, gas uint64, endowment *
 // ChainConfig returns the environment's chain configuration
 func (evm *EVM) ChainConfig() *params.ChainConfig { return evm.chainConfig }
 
+func (evm *EVM) getTobinTax(sender common.Address) (*big.Int, *big.Int, *common.Address, error) {
```
hbandura, line 528: This doesn't seem to be using any private EVM methods or properties, I'd avoid adding it as a responsibility to the EVM. Maybe extract `GetTobinTax(*EVM, common.Address)` or `interface TobinTaxGetter { Get(*EVM, common.Address) }`. This probably also makes it unit-testable.

asaj: Good point, added `getTobinTax(*EVM, common.Address)`

core/vm/evm.go
```diff
@@ -526,6 +525,33 @@ func (evm *EVM) Create2(caller ContractRef, code []byte, gas uint64, endowment *
 // ChainConfig returns the environment's chain configuration
 func (evm *EVM) ChainConfig() *params.ChainConfig { return evm.chainConfig }
 
+func getTobinTax(evm *EVM, sender common.Address) (*big.Int, *big.Int, *common.Address, error) {
```
hbandura, line 528: Is this file from upstream geth? If so, do you think we could take advantage of this being an independent function, and move it to a different file, to make future merges with upstream easier (lessening the amount of differences to check)

asaj: Are we likely to see merge conflicts here? All functions from here on down are additions with respect to upstream, I'm not sure if breaking this out into a separate file will move the needle

PR #1075, opened 2020-06-15
Add geth client Docs

### LINKED ISSUE #1072: Copy docs from geth (gh-pages branch)
### Summary

As part of our [Release Process Documentation](https://github.com/celo-org/celo-monorepo/pull/4045) we decided to surface geth docs in the `celo-blockchain` repo. Geth hosts the docs on github and therefore they are only found in the `gh-pages` branch, and not on `master`. We're thinking of moving them, as is, to the `master` branch.

### Issues with the approach

Making this structural change to the repo will increase complexity when merging upstream geth. Especially if upstream geth makes changes to the docs `gh-pages`, and we have also made changes to our `docs` on master, these will no longer share a common git history that's mergeable (unless we do some fancy git-work, which I'm not totally sure is possible).

Secondly, the docs are constructed as a `jekyll` static site that builds to what you see [here](https://geth.ethereum.org/) and they are constructed to best serve that purpose. Which means that folders don't have a `README.md` which would render in github and that the various links in the markdown files don't work, they are designed to work only when the site is built.

If we include instruction on how to build the static site (which requires installing ruby and a whole bunch of gems) I think it makes more sense to host it ourselves anyway.

Then there's also the question of how much of the content we need to massage. Currently it includes a release page with links to geth binaries and things like that.

### Alternative approach

I would suggest we keep the `gh-pages` way of doing things but we spend some time editing the static site to include branding and remove things like links to geth binaries. We then link to this page when needed.

We should update the docs to reflect differences between `geth` and the `celo-blockchain` client and make sure to keep it up to date when merging upstream geth.

### Initial approach

If we really want to surface these docs on the `master` branch I would:

- [ ] Copy over the docs folder to the `master` branch
- [ ] Edit slightly to remove the release page and introduce some branding
- [ ] Include instruction on how to build and view the docs

bowd:
### Description

This PR aims to include and better surface the client docs in our repo. In order to do this I've:

- [x] Copied over the `geth` docs from the [`gh-pages` branch](https://github.com/ethereum/go-ethereum/tree/gh-pages) into the `/docs` folder on our `master` branch.
- [x] Made some copy and asset changes to make the docs refer to `celo-blockchain`
- [x] Included instruction on how to build the docs in `docs/README.md`
- [x] Removed some sections from the docs which didn't make sense in our context (like the geth binary release page)

> An example of the built docs can be found on the [github page of my fork].(https://bowd.github.io/celo-blockchain/)

### Notes 

Moving the docs to the master branch increases visibility and makes it easier for PRs to also included relevant changes to the docs, rather than having docs on the `gh-pages` branch which would have required an additional PR. With that said changing the structure of the repo will incur additional cost when merging changes from `geth`.  

#### 1. Deploying the docs

Because the docs are formatted as a Jekyll static site, they are best consumed built vs the source that would be visible when browsing github. Thus we should consider deploying them using github pages, which we can luckily still take advantage:

- [ ] A repo admin will need to configure the Github Pages section and choose `master branch /docs folder` as the source
- [ ] Decide on, and set a CNAME that we can then use like `client-docs.celo.org` and start linking to it in various places

#### 2. Updating content

It doesn't really make sense to have these docs unless we actually update them to reflect `celo-blockchain` specific information. Otherwise we might as well just link to the geth docs.

Therefore it's important to have a clear plan to review the full content of the docs and ensure that they are accurate for `celo-blockchain` and also extend them where relevant.

### Related issues

- Fixes #1072


nategraf:
Overall, looks god to me. I'm glad we are getting started on this, although the docs will obviosuly need some work. I made some initial comments on the README and index.html pages

nategraf:
@asaj Is an admin for this repository. Would you be able to set up the GitHub pages?

bowd:
@nategraf good catches :+1: I incorporated all of the suggestions. We'll sync on this during the bubble meeting today and I'll book some time with Asa to make sure we can push this across the finish line.

Issue #185, opened 2019-05-09
Geth SBAT protect against long range attacks


asaj:
### Expected Behavior

Reorgs are limited

### Current Behavior

No limit on reorgs

PR #1076, opened 2020-06-15
Adding checkpoints

mrsmkl:
### Description

Just testing how to set the checkpoints in geth. This should help against long range attacks. If wanted, it should be quite easy to make checkpoints using governance, there is already some support for reading them from smart contracts.

### Related issues

- Related to #185


mcortesi:
too old

PR #1081, opened 2020-06-19
Computing uptimes if they are not up to date

### LINKED ISSUE #1059: Fix fast sync
### Expected Behavior

Fast sync should not encounter a bad block error.

### Current Behavior

If a blockchain is longer than an epoch long, then when a node syncs from the genesis block, it will encounter a BAD BLOCK error when syncing an epoch block.

Running fast sync from the genesis block on any network with chain length > an epoch length should work.

### Logs

Here's a slack thread that includes logs:  https://celo-org.slack.com/archives/C9RSRC8KS/p1589742764366400

### System information

Run `geth version`

mrsmkl:
### Description

When running fast sync, the uptimes aren't updated (except for some number of blocks at the end of fast sync)

### Tested

Fast sync with baklava worked, next going to sync with mainnet.

### Related issues

- Fixes #1059

### Backwards compatibility

Yes.


consensus/istanbul/backend/engine.go
```diff
@@ -777,6 +777,10 @@ func (sb *Backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 		}
 
 		genesis := chain.GetHeaderByNumber(0)
+		if genesis == nil {
```
kevjue, line 780: When does this happen?  Seems like a critical issue if it does happen, and something we should panic on.

mrsmkl: It might happen when geth is exiting (using interrupt). Don't know what's the best way to have clean exit.

core/blockchain.go
```diff
@@ -1345,6 +1379,10 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		if !istanbul.IsFirstBlockOfEpoch(block.NumberU64(), bc.chainConfig.Istanbul.Epoch) {
 			epochNum := istanbul.GetEpochNumber(block.NumberU64(), bc.chainConfig.Istanbul.Epoch)
 
+			if err := bc.checkUptimes(state, epochNum); err != nil {
```
kevjue, line 1382: I see that you are updating the updates here.  WDYT of updating it when it processes the header instead?  AFAIK, uptime only needs the parent aggregated seal field.  And if the logic is moved over there, then it should work for both full and fast sync.

consensus/istanbul/backend/engine.go
kevjue: OK.  Even then, that seems odd.  Since the genesis block should of been written to disk when `geth init` is done.

core/blockchain.go
mrsmkl: I can check. But it might slow down the fast sync a bit.

kevjue: Ya, you make a good point.

IMO, if we want to continue on with the "spirit" of fast sync, then this node should download the uptime values from other nodes, and those values are "committed" onto the header.  But seems like that's going to be big changes.

Also, if the uptimes are computed here for fast sync, is there a chance that they could still be incorrect when this node reaches the head node and transitions into full sync?  E.g. when the node does reach that point, are we guaranteed to have downloaded all the historical blocks?

.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           command: |
             mkdir -p /tmp/test-results
             trap "go-junit-report < /tmp/test-results/go-test.out > /tmp/test-results/go-test-report.xml" EXIT
-            go run build/ci.go test -v | tee /tmp/test-results/go-test.out
+            ( go run build/ci.go test -v || go run build/ci.go test -v ) | tee /tmp/test-results/go-test.out
```
mcortesi, line 48: i would avoid doing this. Let's fix the flaky tests instead.

consensus/istanbul/backend/engine.go
mcortesi: agree with @kevjue. 

This should be a check we do on geth startup, in another context

core/blockchain.go
```diff
@@ -1326,6 +1326,51 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
 	return bc.writeBlockWithState(block, receipts, state)
 }
 
+func (bc *BlockChain) ReadAccumulatedEpochUptime(state *state.StateDB, epoch uint64) (*istanbul.Uptime, error) {
```
mcortesi, line 1329: up to this point the `Blockchain` type didn't know *anything* about uptimes. shouldn't this live somewhere else? I don't see how `uptimes` is a topic related to the `Blockchain` object. 

What do yo think @kevjue?

consensus/istanbul/backend/engine.go
mrsmkl: It's just messed up at some point when it's exiting (probably the DB has shutdown), clearly everything has been ok at startup.

core/blockchain.go
mrsmkl: This will be called only after it has transitioned to full sync, and all the previous blocks should be available.

core/blockchain.go
mrsmkl: Going to remove this function, I didn't use it at the end. They should probably be managed in the same place as the valset snapshot, which is currently at istanbul backend. But the relation between istanbul backend and chain objects is confusing, in some methods it's passed as argument, and others use the `chain` field.

core/blockchain.go
gakonst: I believe that the way the header verification logic is written inside `istanbul/`, it doesn't have access to the state, so we'd need to break some abstraction in order to make this work.

Also, the uptime calculation on non-epoch blocks, so you'd be adding extra information on the header that `lightest` clients do not need.

For the above two reasons, I do not think we should move this logic to the header verification functions

PR #1083, opened 2020-06-25
Add newView from core to worker when it's new proposer

YoduYodu:
### Description

Right now, every validator will create a block proposal for height N, even if it’s extremely unlikely they will need to proposal block N
To optimize, the worker will subscribe to istanbul core engine, which will send an event when the running validator is the new proposer of the round.

### Tested

Not yet.

### Backwards compatibility

Yes


miner/worker.go
```diff
@@ -375,6 +379,8 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
 		case head := <-w.chainHeadCh:
 			headNumber := head.Block.NumberU64()
 			clearPending(headNumber)
```
kevjue, line 381: Do you also need to add a `clearPending` line to your newView handler?  If it does, then I think you to reconsider emitting the NewView event to the worker's channel on whenever there is a new view (as opposed to only if this node is the proposer).

miner/worker.go
```diff
@@ -214,6 +217,7 @@ func newWorker(config *Config, chainConfig *params.ChainConfig, engine consensus
 	// Subscribe events for blockchain
 	worker.chainHeadSub = eth.BlockChain().SubscribeChainHeadEvent(worker.chainHeadCh)
```
kevjue, line 218: Can the `chainHeadSub` be removed?

consensus/istanbul/core/events.go
```diff
@@ -30,3 +30,8 @@ type resendRoundChangeEvent struct {
 type timeoutAndMoveToNextRoundEvent struct {
 	view *istanbul.View
 }
+
+// NewViewEvent is posted when a new view has been created
+type NewViewEvent struct {
+	View *istanbul.View
```
kevjue, line 36: You may need to add a field that contains the new proposer to this event.  Especially if you need to emit this event to the worker for all new views.

miner/worker.go
YoduYodu: Why would we emit the NewView event every time? Is it because we need to let all validators be aware of the new view/round, thus they can clearPending()?

miner/worker.go
YoduYodu: Yes, since its role now has been replaced/optimized by `newViewSub`

miner/worker.go
kevjue: So i'm thinking of the case that a node has become a validator for a view and then sends out a proposal.  And then a new view happens.  What will need to happen to that in-process proposal?  I suspect that `clearPending` will need to be called, but I'm not 100% sure.  Can you look into that.

YoduYodu: Yes, `clearPending` will need to be called here

PR #1085, opened 2020-06-29
Setup binary release CI pipeline

### LINKED ISSUE #1073: Setup pipeline that builds release artefacts
### Summary

As part of our [Release Process](https://github.com/celo-org/celo-monorepo/pull/4045) we need to setup a pipeline that automatically outputs release artefacts based on the code in our `release/<version>` and `master` branches.

### Details

The current proposal is that we do this in Google Cloud Build, in a project managed by cLabs with binaries stored in Cloud Storage and Docker images published to Container Registry.

> Potential tension: Maybe we should think about using OSS tools for our CI pipeline and have that all out in the open. It looks like there are various ways of automatic this with travis/circle/github actions that also sets up the actual release on github with the binaries.

#### Binaries

Binaries should be created for the following systems:

- Linux
- Darwin
- Windows

And the following platforms:

- 386
- amd64

Resulting in 6 binary artefacts of the form:

`celo-blockchain-vX.Y.Z-stable-{system}-{plaftorm}.{exe?}.tar.gz`

Optional/future tasks:

- Extend `celotooljs` functionality to allow signing release binaries and uploading the signatures for ease of use.

#### Docker Images

Docker images will be built and pushed to our registry and tags will be updated.

Optional/future tasks:

- Extend `celotooljs` functionality to allow signing of docker releases via the tar export mechanism (`docker save us.gcr.io/celo-testnet/celo-node:vX.Y.Z`)

bowd:
### Description

This PR includes the necessary bits to setup a CI pipeline that cross-compiles all geth tools into release binaries for the platforms we want/support. Currently these are:

- ✅Linux 32bit/64bit
- ✅Darwin 32bit/64bit (~pending on #1082~ merged!)
- ✅Windows 64bit (~pending on celo-bls-go v0.1.6 release~ ~pending on #1089~ merged!)

Elements of the PR:

- `Dockerfile.binaries` for the container where cross-compilation occurs. It's based on a [fork of xgo](https://github.com/techknowlogick/xgo), and includes some back-ported mingw packages for the windows build.

- Changes to the `geth` build scripts:
   - Introduce a new CI environment in `internal/build/env.go` which which can be used for Google Cloud Build.
   - Create a new ci task in `build/ci.go` that bundles the binaries into well named release archive

- `cloudbuild-binaries.yaml` which defines the CI pipeline. It should be used in conjunction with a trigger on `release/.*` and `master` branches and the trigger should have to variables:
  - `_BUCKET` with the name of the google cloud storage bucket where the artefacts will be stored
  - `_BUILD_TARGETS` comma-separated string of build targets. e.g. `linux/amd64,linux/386`

#### Shortcomings 

- 🔴 The geth `build.Environment` struct requires the commit timestamp, which is then passed to `main.gitDate` in the binaries that it builds. Usually this is extracted from the git but in Cloud Build in the CI steps the git data is stripped and all we have is what's passed through env vars. I have substituted the commit timestamp with the build timestamp instead.
- 🟡 The `xgo` image is really large (3gb) and adds about ~4minutes to the build time

#### Release artefacts

The pipeline outputs release artefacts into `<bucket>/<branch>/<file>`. For example here's the output of a test build:

Screenshot 2020-06-29 at 11 41 00

A few things to notice here:

- The folder is `<bucket>/release/1.1` yet the version is `1.0.0-unstable` this is because there isn't any enforced relationship between the branch name and the version sourced from `params/version.go` which is the authoritative source. So in this case I just created the dummy `release/1.1` branch on my fork in order to play around. Because we're using the branch name in the path all "nightly" version will be stored in the `<bucket>/master` folder, if we setup the trigger on `master`.

- There are 2 archives per platform, one containing only `geth` and the other containing all binary tools located inside `cmd`, just like geth releases. ⚠️ The archives also contain the "COPYING" file, we need to decide if we want to make changes to that ⚠️
 

#### Post-merge TODOs:

- [x] Create a bucket and triggers in `celo-testnet` to start running the pipeline
- [x] Enable more build targets as they are unblocked 
 
### Tested

I have currently tested the CI pipeline in a personal google cloud project and tested the linux binaries in docker.

### Related issues

- Fixes #1073 

### Backwards compatibility

Changes are only related to tooling so no problem with compatibility.

### DIFF
--- a/Dockerfile.binaries
+++ b/Dockerfile.binaries
@@ -1,0 +1,37 @@
+# How to test changes to this file
+# docker build -f Dockerfile -t gcr.io/celo-testnet/geth:$USER .
+# To locally run that image
+# docker rm geth_container ; docker run --name geth_container gcr.io/celo-testnet/geth:$USER
+# and connect to it with
+# docker exec -t -i geth_container /bin/sh
+#
+# Once you are satisfied, build the image using
+# export COMMIT_SHA=$(git rev-parse HEAD)
+# docker build -f Dockerfile --build-arg COMMIT_SHA=$COMMIT_SHA -t gcr.io/celo-testnet/geth:$COMMIT_SHA .
+#
+# push the image to the cloud
+# docker push gcr.io/celo-testnet/geth:$COMMIT_SHA
+#
+# To use this image for testing, modify GETH_NODE_DOCKER_IMAGE_TAG in celo-monorepo/.env file
+
+# Build Geth binaries in the xgo builder container
+FROM techknowlogick/xgo as xgobuilder
+
+# We need a newer version of mingw, backported to Bionic
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt install -y  --no-install-recommends software-properties-common apt-utils
+RUN add-apt-repository -y ppa:mati865/mingw-w64
+RUN apt update && apt -y upgrade
+
+ADD . /go/src/github.com/ethereum/go-ethereum
+WORKDIR /go/src/github.com/ethereum/go-ethereum
+
+ENV GO=1.13
+ARG BUILD_TARGETS
+ARG COMMIT_SHA
+ARG ALLTOOLS
+ENV ALLTOOLS_COMMAND_SEGMENT=""
+RUN if [ "$ALLTOOLS" = "true" ]; then export ALLTOOLS_COMMAND_SEGMENT="--alltools"; fi
+
+RUN echo $COMMIT_SHA > /version.txt
+RUN go run build/ci.go xgo $ALLTOOLS_COMMAND_SEGMENT -- --go=$GO --targets=$BUILD_TARGETS -v ./cmd/geth
--- a/Dockerfile.binaries
+++ b/Dockerfile.binaries
@@ -1,21 +1,50 @@
-# How to test changes to this file
-# docker build -f Dockerfile -t gcr.io/celo-testnet/geth:$USER .
-# To locally run that image
-# docker rm geth_container ; docker run --name geth_container gcr.io/celo-testnet/geth:$USER
-# and connect to it with
-# docker exec -t -i geth_container /bin/sh
+# Purpose:
+# -------
+# This docker image is used by the ./cloudbuild-binaries.yaml CI flow to
+# cross-compile geth for different platforms.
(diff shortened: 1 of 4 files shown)


Dockerfile.binaries
```diff
@@ -0,0 +1,56 @@
+# Purpose:
+# -------
+# This docker image is used by the ./cloudbuild-binaries.yaml CI flow to
+# cross-compile geth for different platforms.
+# At the time of writing this is only linux-{386/amd64} but there are two
+# pending update that will add darwin-{386,amd64} and windows-amd64 support.
```
nategraf, line 6:
```suggestion
# At the time of writing linux-{386/amd64}, darwin-{386,amd64} and windows-amd64 are supported.
```

Dockerfile.binaries
```diff
@@ -0,0 +1,56 @@
...
+# Running this image depends on a series of environment variables:
+#   BUILD_TARGETS       Comma separated list of platforms to build for,
+#                       passed as an arg to xgo. eg: "linux/386,linux/amd64"
+#   TAG_NAME            Name of the tag associated with the current commit
+#   BRANCH_NAME         Name of the branch
+#   REPO_NAME           Name of the repo
+#   COMMIT_SHA          Full SHA of the commit
+#   COMMIT_TIMESTAMP    Commit timestamp
+#                       TODO: currently this is not accurately passed see
+#                             discussion in PR celo-blockchain#<number>
```
nategraf, line 22:
```suggestion
#                             discussion in PR celo-blockchain#1085
```

Dockerfile.binaries
```diff
@@ -0,0 +1,56 @@
...
+#    go run build/ci.go xgo-archive -targets=$_BUILD_TARGETS -in /build -out /archives
+#
+# This will result in build archives stored in ./build/archives
+# In the CI flow these are then uploaded to cloud storage as artefacts.
+
+# Build Geth binaries in the xgo builder container
+FROM techknowlogick/xgo:go-1.13.x
+# techknowlogic/xgo is a fork of karalabe/xgo updated to ubunut-18, it is more maintained
+# by the community and allows us to backport mingw in order to build for windows
+# See discussion in PR celo-blockchain#<number> about downsides of this image
```
nategraf, line 47:
```suggestion
# See discussion in PR celo-blockchain#1085 about downsides of this image
```

PR #1091, opened 2020-06-30
Adds metrics for missed rounds as proposer & downtime events

### LINKED ISSUE #1013: Include metrics for missed rounds as proposer and 12+ block downtime events
### Expected Behavior

Include metrics for missed rounds as proposer and 12+ block downtime events 

This is part of epic: https://github.com/celo-org/celo-blockchain/issues/1012 

### Current Behavior

Those metrics are not being emitted.

trianglesphere:
### Description

Two sets of metrics are added.

1. Adds meter for the number blocks that could have been proposed by the validator but were not. This is not the same as being down for a block.
2. Adds a gauge for the number of blocks that were not signed by the validator in a row when the validator is elected.
3. Adds a meter for blocks that will be counted as downtime events & also reports an error the the logs.

The gauge for missed blocks is reset at each epoch to match how downtime is calculated.

The meter in 1 is complex because in a single sequence (block number), if there are multiple rounds, the proposer is rotated. In the first round, the proposer can propose any block, but in later rounds, if the proposer receives a prepared certificate during round change messages, it must re-propose that block. Thus from the finalized block, it is hard to tell if the proposer did not propose, proposed their own block that timed out, or were required to proposed a previous block. The meter 'blocksMissedRoundsAsProposerMeter' includes all three cases

### Other changes

TODO: Update docs in monorepo.

### Tested

Passes lint, but no specific tests.

### Related issues

- Fixes #1013 

### Backwards compatibility

Are.


consensus/istanbul/backend/handler.go
```diff
@@ -266,10 +267,28 @@ func (sb *Backend) UpdateMetricsForParentOfBlock(child *types.Block) {
...
+	}
+
+	// Clear downtime counter on end of epoch.
+	if istanbul.IsLastBlockOfEpoch(number-1, sb.config.Epoch) {
+		sb.blocksElectedButNotSignedGauge.Update(0)
+	}
+
+	// missed round as proposer
+	if missedRounds > 0 && parentHeader.Coinbase == sb.ValidatorAddress() {
+		sb.blocksMissedRoundsAsProposerMeter.Mark(missedRounds)
```
kevjue, line 290: Is this metric an accumulator for the number of missed rounds when this validator is the proposer?

IIUC, right now it's the total number of missed rounds (regardless of which validator is the proposer) for sequences right after this validator proposed a finalized block.

consensus/istanbul/backend/handler.go
```diff
@@ -266,10 +267,28 @@ func (sb *Backend) UpdateMetricsForParentOfBlock(child *types.Block) {
...
 		sb.blocksElectedAndSignedMeter.Mark(1)
+		sb.blocksElectedButNotSignedGauge.Update(0)
 	} else {
 		sb.blocksElectedButNotSignedMeter.Mark(1)
+		sb.blocksElectedButNotSignedGauge.Update(sb.blocksElectedButNotSignedGauge.Value() + 1)
 		sb.logger.Warn("Elected but didn't sign block", "number", number-1, "address", sb.ValidatorAddress())
 	}
+
+	// Report downtime events
+	if sb.blocksElectedButNotSignedGauge.Value() >= 12 {
```
kevjue, line 278: nit:  What do you think of replacing `12` with a named constant (perhaps that is already defined somewhere)?

consensus/istanbul/backend/handler.go
trianglesphere: It's supposed to be the number of missed round when the validator is the proposer.

My understanding is that missedRounds while coming from the child block, refers back to the parent block, and the parent coinbase is the proposer for that block.

kevjue: Ah, just now seeing that the `missedRound` is from the parent aggregated seal.

Also, the proposer should be based on round, not block number.  E.g. for a specific block number, the proposer will be rotated as rounds increase for that block: https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/validator/selectors.go#L54

trianglesphere: Ok, might need a rename, something like `potentialMissedProposalOpportunity`, but this still remains tricky with how proposer rotations are done.

PR #1096, opened 2020-07-02
Adds epochSnarkData to getBlock API

kobigurk:
### Description

While the data exists in the block body, the API did not expose it.


internal/ethapi/api.go
```diff
@@ -1047,6 +1047,15 @@ func RPCMarshalBlock(block *types.Block, inclTx bool, fullTx bool) (map[string]i
 		"revealed":  hexutil.Bytes(block.Randomness().Revealed.Bytes()),
 		"committed": hexutil.Bytes(block.Randomness().Committed.Bytes()),
 	}
+	epochSnarkData := block.EpochSnarkData()
+	if !epochSnarkData.IsEmpty() {
```
mcortesi, line 1051: i know we use the EmptySnarkData, but it might be good to also check for `nil` just as a future proof measure

kobigurk: Good idea. I'll add that.

Issue #636, opened 2019-11-22
Resubmit block proposal requests for istanbul


asaj:
### Expected Behavior

Within a sequence, the worker can send new requests to the istanbul consensus engine.  This will allow the new proposer to consider recent transactions that it synced.
 
### Current Behavior

Block request created at the beginning of a sequence, never changes until the next sequence (or mining stopped/restarted)

nategraf:
@asaj @timmoreton can we get some more context on this? Is this something that needs to be solved ASAP?

kevjue:
Going through old tickets, and wanted to mention that this should be thought through carefully.  E.g. there is a big safety reason that proposers in new rounds chose the highest prepare certificate's proposal, since it may already been committed by another validator.

PR #1102, opened 2020-07-09
[WIP] Only proposer should send proposal

YoduYodu:
### Description

Right now, every validator will start proposing a new block N after last block N - 1 was inserted into chain. This process involves heavy computation, e.g. transactions & generating random beacon. But only the proposer's processing makes sense.

Instead of kicking off the proposal once the block is inserted, the validator(`worker.go`) can get an event from `consensus engine` when a new proposer of the round has been selected. Thus the validator will be aware that it's the proposer and only so will it proposer the new block, saving the non-proposers computation resources.

### Related Issues

https://github.com/celo-org/celo-blockchain/issues/636

### Tested

Yes


consensus/consensus.go
```diff
@@ -20,6 +20,8 @@ package consensus
 import (
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/event"
```
kevjue, line 23: Looks like you need to run `go fmt` on this file.

consensus/consensustest/mockprotocol.go
```diff
@@ -25,6 +25,9 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/ethereum/go-ethereum/consensus/istanbul"
```
kevjue, line 28: Please run 'go fmt' on this file.

consensus/istanbul/core/events.go
```diff
@@ -27,6 +27,7 @@ type backlogEvent struct {
 type resendRoundChangeEvent struct {
 	view *istanbul.View
 }
+
```
kevjue, line 30: Please remove this new line.

go.sum
```diff
@@ -229,4 +229,4 @@ gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.7 h1:VUgggvou5XRW9mHwD/yXxIYSMtY0zoKQf/v226p2nyo=
 gopkg.in/yaml.v2 v2.2.7/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gotest.tools v2.2.0+incompatible h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
-gotest.tools v2.2.0+incompatible/go.mod h1:DsYFclhRJ6vuDpmuTbkuFWG+y2sxOXAzmJt81HFBacw=
+gotest.tools v2.2.0+incompatible/go.mod h1:DsYFclhRJ6vuDpmuTbkuFWG+y2sxOXAzmJt81HFBacw=
```
kevjue, line 232: Please remove any whitespace that is causing this diff.

miner/worker.go
```diff
@@ -378,13 +387,20 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
 			clearPending(w.chain.CurrentBlock().NumberU64())
 			timestamp = time.Now().Unix()
-			commit(false, commitInterruptNewHead)
+			commit(false, commitInterruptNewHead, true)
 
-		case head := <-w.chainHeadCh:
-			headNumber := head.Block.NumberU64()
+		case <-w.chainHeadCh:
+			if h, ok := w.engine.(consensus.Handler); ok {
+				h.NewWork()
```
kevjue, line 394: I think `NewWork` should be renamed to `NewChainHead`.

miner/worker.go
```diff
@@ -877,6 +893,15 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
 
 	w.updateSnapshot()
 
+	// Short circuit if this validator is not the proposer of the round/view
+	if !isProposer {
+		log.Info("Tong: Is not a proposer, committing empty block")
+		istanbulEmptyBlockCommit()
```
kevjue, line 899: Instead of an empty block submitted to seal, it should be a `nil` block.

miner/worker.go
YoduYodu: There is an existing NewChainHead in `consensus.go`: https://github.com/celo-org/celo-blockchain/blob/83cb6f6b8d5777059f68824fb5d3ffc59c579721/consensus/consensus.go#L141

consensus/istanbul/core/core.go
```diff
@@ -466,11 +464,12 @@ func (c *core) startNewRound(round *big.Int) error {
 	c.processPendingRequests()
 	c.backlog.updateState(c.current.View(), c.current.State())
 
-	if roundChange && c.isProposer() && request != nil {
-		c.sendPreprepare(request, roundChangeCertificate)
-	}
 	c.resetRoundChangeTimer()
 
+	newViewEvent := istanbul.NewViewEvent{NewView: newView, IsProposer: c.isProposer()}
```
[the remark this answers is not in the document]
kevjue: This will mean that a `newViewEvent` will need to be sent from the istanbul engine to the worker when the engine is started.

eth/api_backend.go
```diff
@@ -59,13 +59,8 @@ func (b *EthAPIBackend) SetHead(number uint64) {
...
 
 func (b *EthAPIBackend) HeaderByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Header, error) {
-	// Pending block is only known by the miner
-	if number == rpc.PendingBlockNumber {
-		block := b.eth.miner.PendingBlock()
-		return block.Header(), nil
-	}
 	// Otherwise resolve and return the block
-	if number == rpc.LatestBlockNumber {
+	if number == rpc.LatestBlockNumber || number == rpc.PendingBlockNumber {
```
kevjue, line 63: As described in the discussion in slack, we should use the `pending block` if the node is a full node and the `latest block` if it is a validator.

eth/api_backend.go
```diff
@@ -93,13 +88,8 @@ func (b *EthAPIBackend) HeaderByHash(ctx context.Context, hash common.Hash) (*ty
...
 
 func (b *EthAPIBackend) BlockByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Block, error) {
-	// Pending block is only known by the miner
-	if number == rpc.PendingBlockNumber {
-		block := b.eth.miner.PendingBlock()
-		return block, nil
-	}
 	// Otherwise resolve and return the block
-	if number == rpc.LatestBlockNumber {
+	if number == rpc.LatestBlockNumber || number == rpc.PendingBlockNumber {
```
kevjue, line 92: As described in the discussion in slack, we should use the pending block if the node is a full node and the latest block if it is a validator.

eth/api_backend.go
```diff
@@ -131,13 +121,14 @@ func (b *EthAPIBackend) BlockByNumberOrHash(ctx context.Context, blockNrOrHash r
 }
 
 func (b *EthAPIBackend) StateAndHeaderByNumber(ctx context.Context, number rpc.BlockNumber) (*state.StateDB, *types.Header, error) {
-	// Pending state is only known by the miner
+	var header *types.Header
```
kevjue, line 124: Same comment as above.

eth/api_tracer.go
```diff
@@ -107,15 +107,15 @@ func (api *PrivateDebugAPI) TraceChain(ctx context.Context, start, end rpc.Block
 
 	switch start {
 	case rpc.PendingBlockNumber:
-		from = api.eth.miner.PendingBlock()
+		fallthrough
```
kevjue, line 110: Same comment as above.

miner/worker.go
kevjue: This comment is irrelevant now, given this comment:  https://github.com/celo-org/celo-blockchain/pull/1102#discussion_r454040327

miner/worker.go
kevjue: Good catch.  Looks like this function https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/handler.go#L276 will need to be modified so that the final committed event is sent to the core consensus engine and the `NewWork` function can be removed altogether.

consensus/istanbul/core/core.go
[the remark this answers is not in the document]
YoduYodu: I'm assuming you meant `when is engine is proposer`?

consensus/istanbul/backend/handler.go
```diff
@@ -273,10 +257,20 @@ func (sb *Backend) UpdateMetricsForParentOfBlock(child *types.Block) {
...
 
+	sb.logger.Debug("NewWork called, acquiring core lock", "func", "NewWork")
+	sb.coreMu.RLock()
+	defer sb.coreMu.RUnlock()
+	if !sb.coreStarted {
+		return istanbul.ErrStoppedEngine
+	}
+
+	sb.logger.Debug("Posting FinalCommittedEvent", "func", "NewWork")
+	go sb.istanbulEventMux.Post(istanbul.FinalCommittedEvent{})
```
kevjue, line 272:
```suggestion
	        sb.logger.Debug("Posting FinalCommittedEvent", "func", "NewWork")
	        go sb.istanbulEventMux.Post(istanbul.FinalCommittedEvent{})
	}	
```

consensus/istanbul/backend/handler.go
```diff
@@ -302,6 +296,8 @@ func (sb *Backend) NewChainHead(newBlock *types.Block) {
 	}
 
 	sb.logger.Trace("End NewChainHead", "number", newBlock.Number().Uint64())
+
+	return nil
```
kevjue, line 300:
```suggestion
```

consensus/istanbul/backend/handler.go
```diff
@@ -273,10 +257,20 @@ func (sb *Backend) UpdateMetricsForParentOfBlock(child *types.Block) {
 }
 
 // Actions triggered by a new block being added to the chain.
-func (sb *Backend) NewChainHead(newBlock *types.Block) {
+func (sb *Backend) NewChainHead(newBlock *types.Block) error {
```
kevjue, line 260:
```suggestion
func (sb *Backend) NewChainHead(newBlock *types.Block) {
```

miner/worker.go
```diff
@@ -381,7 +390,13 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
 			commit(false, commitInterruptNewHead)
 
 		case head := <-w.chainHeadCh:
-			headNumber := head.Block.NumberU64()
+			if h, ok := w.engine.(consensus.Handler); ok {
```
kevjue, line 393: You don't need to listen to the `chainHeadCh` anymore.  That `case` can be removed.

eth/api_backend.go
```diff
@@ -59,13 +59,11 @@ func (b *EthAPIBackend) SetHead(number uint64) {
...
-		block := b.eth.miner.PendingBlock()
-		return block.Header(), nil
-	}
-	// Otherwise resolve and return the block
-	if number == rpc.LatestBlockNumber {
+	// Use latest block for validators, otherwise use pending block for full nodes
+	if number == rpc.LatestBlockNumber || number == rpc.PendingBlockNumber {
+		if !b.eth.IsMining() {
+			return b.eth.miner.PendingBlock().Header(), nil
+		}
```
kevjue, line 66:
```suggestion
	// If this is a validator, return the latest block if pending block is requested.  Non proposing validators for IBFT don't have a pending block.  TODO(kevjue) modify the miner so that the pending txns will build up the pending block (similar to what full nodes do).
	if number == rpc.PendingBlockNumber {
		if !b.eth.IsMining() {   // This is a full node
			return b.eth.miner.PendingBlock().Header(), nil
		} else {
		        return b.eth.blockchain.CurrentBlock().Header(), nil
		}
	}
	
	// Otherwise resolve and return the block
	if number == rpc.LatestBlockNumber {
```

PR #1103, opened 2020-07-09
Use correct address to select BLS key when ECDSA key is different

nategraf:
### Description

Currently, when the ECDSA and BLS keys are set differently (e.g. through the `--etherbase` and `--blsbase` flags, the ECDSA address is used to issue signing requests to the wallet, which results in an unknown account error. This PR adds the BLS address as a field to the engine and uses it for signing requests.

### Tested

Currently untested

### Backwards compatibility

100%


nategraf:
Need to add a test to this PR before merging

PR #1115, opened 2020-07-18
Merge 1.9.10 upstream

mcortesi:
### Description

* Merge upstream 1.9.1 - https://github.com/ethereum/go-ethereum/releases/tag/v1.9.1
* Merge upstream 1.9.10 - https://github.com/ethereum/go-ethereum/releases/tag/v1.9.10

### Tested

* ci & unit tests

### Backwards compatibility

Yes


miner/worker.go
```diff
@@ -1010,11 +1013,3 @@ func (w *worker) isIstanbulEngine() bool {
 	_, isIstanbul := w.engine.(consensus.Istanbul)
 	return isIstanbul
 }
-
```
kevjue, line 1013: How come this function is removed?

kevjue:
This upstream PR https://github.com/ethereum/go-ethereum/pull/20453 (which is actually a fix for a bug that @asaj submitted to upstream geth) should make this celo-blockchain PR unnecessary and revert-able:  https://github.com/celo-org/celo-blockchain/pull/333

miner/worker.go
mcortesi: It was added on 1.9.9 for testing purposes... but then when adapting the tests to use our fake consensus engine it was not needed anymore, so the linter complained about not being used

PR #1116, opened 2020-07-21
Integrate Plumo proofs into the lightest sync protocol

lucasege:
### Description
WIP

Tentative plan:
- [x] Implement `AddProof` RPC command for full nodes
- [x] Gossip added proofs to peers
- [x] Add new general les message: "I have val set at epoch X", server responds with correct combination of headers/proofs as available
- [x] Verify proofs within client
- [x] Update light client state based on proof (validator set)
- [ ] Implement lightest client lookbacks - will need the validator set for any epoch we want to query

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


hbandura:
@mcortesi should I close this?

Issue #1105, opened 2020-07-14
Preliminary work for blockchain scalability


kevjue:
* Polish load test
* measure current TPS
* block proposal
* verification optimizations

PR #1123, opened 2020-07-28
Add block processing cache to avoid redundancy

YoduYodu:
### Description

Currently, block processing is executed two times after the validator sent the proposal:
1. Before broadcasting to peers during `preprepare`
2. After Finalizing it during `fetcher.enqueue`.

Instead, we can cache the block processing result in step 1, and use it in step 2, thus avoiding redundant block processing.

----------------------------------
To-do's added at Dec 3, 2020
- [ ] Proposer doesn't need to verify proposal.

- [ ] A Geth scope cache handling verification during consensus, AND downloaded proposal from peers

----------------------------------

### Other changes

Instead of non proposers sending the block to fetcher for insertion, we can handle it in `worker.go` altogether, same process as proposers.

### Related issues

#1105 

### Backwards compatibility

This branch is on upstream 1.9.10 merge.


miner/worker.go
```diff
@@ -551,64 +551,75 @@ func (w *worker) taskLoop() {
...
-				*receipts[i] = *receipt
-				// Update the block hash in all logs since it is now available and not when the
-				// receipt/log of individual transactions were created.
-				for _, log := range receipt.Logs {
-					log.BlockHash = hash
-					// Handle block finalization receipt
-					if (log.TxHash == common.Hash{}) {
-						log.TxHash = hash
+
+			if result.IsProposer {
```
trianglesphere, line 560: If my understanding of the code is correct, this means that the proposer must deep copy the block submitted through the seal function, but other validators sending in the result through the commit function don't have to deep copy the result. Could you add a comment briefly explaining why this is?

consensus/istanbul/backend/backend.go
```diff
@@ -430,21 +432,46 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
...
-	//   to commit channel, which is being watched inside the engine.Seal() function.
-	// - otherwise, we try to insert the block.
-	// -- if success, the ChainHeadEvent event will be broadcasted, try to build
-	//    the next block and the previous Seal() will be stopped.
-	// -- otherwise, a error will be returned and a round change event will be fired.
-	if sb.proposedBlockHash == block.Hash() {
-		// feed block hash to Seal() and wait the Seal() result
-		sb.commitCh <- block
+
+	sealHash := sb.SealHash(block.Header())
```
kevjue, line 436: I think the `sealhash` calculated here will be different than when it's calculated when the cache entry is saved.

consensus/istanbul/backend/backend.go
```diff
@@ -527,6 +554,17 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
 		return 0, err
 	}
 
+	// Cache the BlockProcessResult
```
kevjue, line 557: Instead of clearing the cache here, I think it's better to clear it within `NewChainHead` (basically when a new block has been added to the blockchain).

That way, the `pendingBlockProcessResults ` map will only need to cache results for a single block number.   Note that there still can be multiple processed results in a block number, since there can be different proposals for the same sequence.

consensus/istanbul/backend/backend.go
YoduYodu: `sealHash` will only calculate the hash based on block header without taking `seal` into account. 
And epoch snark data is store outside of header.
So sealHash makes sense here.

consensus/istanbul/backend/test_utils.go
```diff
@@ -144,11 +145,19 @@ func makeBlock(keys []*ecdsa.PrivateKey, chain *core.BlockChain, engine *Backend
...
 	aggregatedSeal := signBlock(keys, block)
-	err := engine.Commit(block, aggregatedSeal, types.IstanbulEpochValidatorSetSeal{})
+	aggregatedEpochValidatorSetSeal := types.IstanbulEpochValidatorSetSeal{}
+	err := engine.Commit(block, aggregatedSeal, aggregatedEpochValidatorSetSeal)
 	if err != nil {
 		return nil, err
 	}
-	block = <-results
+
+	h := block.Header()
```
kevjue, line 154: How come this new code block needs to be added here?

YoduYodu: Previously `engine.Commit` will search newly added `BlockProcessResult` in the cache to return the block. But since the default setting for test_backend is non-proposer. The block is never saved in the `engine.Seal`.

In my new commit, this has been fixed.

consensus/istanbul/backend/backend.go
YoduYodu: In the new commits, since we refactor the `BlockProcessResults` from `istanbul/backend` to proposalVerificationStatus in `istanbul/core`, and it's cleared at every new sequence. Then moving it within `NewChainHead` is no longer needed.

miner/worker.go
```diff
@@ -551,64 +528,26 @@ func (w *worker) taskLoop() {
...
 	for {
 		select {
-		case block := <-w.resultCh:
-			// Short circuit when receiving empty result.
-			if block == nil {
-				continue
-			}
-			// Short circuit when receiving duplicate result caused by resubmitting.
-			if w.chain.HasBlock(block.Hash(), block.NumberU64()) {
+		case result := <-w.resultCh:
```
kevjue, line 531: Let's try to keep the code change as small as possible within this file.  So it's easier to merge upstream.

YoduYodu: Moved deep-copy part back to this file, `pendingTasks` is removed.

consensus/istanbul/backend/backend.go
```diff
@@ -508,34 +491,35 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (time.Duration, error) {
...
 	if istanbul.IsLastBlockOfEpoch(block.Header().Number.Uint64(), sb.config.Epoch) {
 		if err := sb.verifyValSetDiff(proposal, block, state); err != nil {
 			sb.logger.Error("verify - Error in verifying the val set diff", "err", err)
-			return 0, err
+			return 0, nil, err
 		}
 	}
 
-	return 0, err
+	result := &istanbul.BlockConsensusAndProcessResult{SealedBlock: block, Receipts: receipts, Logs: allLogs, State: state}
```
kevjue, line 521: I don't think you need to use the full `BlockConsensusAndProcessResult` here.  Specifically, the field `SealedBlock` doesn't make sense.

Consider creating a new type called `BlockProcessResult` with the `Receipts`, `Logs`, and `State` fields and the `BlockConsensusAndProcessResult` type uses that new type.

consensus/istanbul/core/core.go
```diff
@@ -310,7 +310,13 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
-		if err := c.backend.Commit(proposal, aggregatedSeal, aggregatedEpochValidatorSetSeal); err != nil {
+
+		result, isCached := c.current.GetBlockConsensusAndProcessResult(proposal.Hash())
+		if !isCached {
+			log.Error("BlockConsensusAndProcessResult not found in cache", "number", proposal.Number(), "hash", proposal.Hash())
+			return nil
```
kevjue, line 317: This should return a non nil error.  If the cached entry is not there, then there is a bug in the code.   Please create a new error in `core/errors.go` for that.

consensus/istanbul/core/roundstate.go
```diff
@@ -49,6 +49,7 @@ type RoundState interface {
 	AddParentCommit(msg *istanbul.Message) error
 	SetPendingRequest(pendingRequest *istanbul.Request) error
 	SetProposalVerificationStatus(proposalHash common.Hash, verificationStatus error)
+	SetBlockConsensusAndProcessResult(sealHash common.Hash, blockProcessResult *istanbul.BlockConsensusAndProcessResult)
```
kevjue, line 52: After creating a new type for the `BlockProcessResult`, this function name should be renamed accordingly.

consensus/istanbul/core/roundstate.go
```diff
@@ -70,6 +71,7 @@ type RoundState interface {
 	View() *istanbul.View
 	PreparedCertificate() istanbul.PreparedCertificate
 	GetProposalVerificationStatus(proposalHash common.Hash) (verificationStatus error, isCached bool)
+	GetBlockConsensusAndProcessResult(sealHash common.Hash) (result *istanbul.BlockConsensusAndProcessResult, isCached bool)
```
kevjue, line 74: ditto.

consensus/istanbul/types.go
```diff
@@ -26,25 +26,6 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
-// UptimeEntry contains the uptime score of a validator during an epoch as well as the
```
kevjue, line 29: Did you move these types because of some circular imports?  If so, what was that circular import?

I actually think that the `Uptime` related types still makes sense to be somewhere in the `istanbul` directory.

YoduYodu: Yes.
My changes will make `istanbul` depend on `core/state`, because the `BlockProcessResult` has a statedb in `core/state`.
But there is a test in `core/state` depends on `istanbul`

miner/worker.go
```diff
@@ -551,32 +529,27 @@ func (w *worker) taskLoop() {
...
 			}
+
 			var (
 				sealhash = w.engine.SealHash(block.Header())
 				hash     = block.Hash()
 			)
-			w.pendingMu.RLock()
-			task, exist := w.pendingTasks[sealhash]
-			w.pendingMu.RUnlock()
-			if !exist {
```
kevjue, line 570: Please add a check (along with the error message) that is similar to this, where your check will make sure the `result.State` and `result.Receipts` are not nil.   If they are nil, then there is something very wrong (most likely a bug in the code).

miner/worker.go
```diff
@@ -551,32 +529,27 @@ func (w *worker) taskLoop() {
...
+			// Short circuit when receiving empty result
+			if result == nil {
 				continue
 			}
+			block := result.SealedBlock
 			// Short circuit when receiving duplicate result caused by resubmitting.
 			if w.chain.HasBlock(block.Hash(), block.NumberU64()) {
 				continue
 			}
+
```
kevjue, line 542: Remove extra line?

miner/worker.go
```diff
@@ -596,13 +569,14 @@ func (w *worker) resultLoop() {
...
-				continue
+			if result.State != nil {
+				_, err := w.chain.WriteBlockWithState(block, receipts, logs, result.State, true)
+				if err != nil {
+					log.Error("Failed writing block to chain", "err", err)
+					continue
+				}
+				log.Info("Successfully sealed new block", "number", block.Number(), "sealhash", sealhash, "hash", hash)
 			}
-			log.Info("Successfully sealed new block", "number", block.Number(), "sealhash", sealhash, "hash", hash,
```
kevjue, line 604: it would be nice to not remove the `elapsed` time.  That is actually very interesting information (e.g. it will tell the user how long it took for consensus and execution of the block).  Please try to see if there is a way to preserve that output.

consensus/istanbul/validator.go
```diff
@@ -151,3 +153,13 @@ func SeparateValidatorDataIntoIstanbulExtra(validators []ValidatorData) ([]commo
 
 	return addrs, pubKeys
 }
+
+// ----------------------------------------------------------------------------
+
+// BlockProcessResult caches block process result.
+type BlockConsensusAndProcessResult struct {
```
kevjue, line 160: This should go into the file `istanbul/types.go`

consensus/istanbul/types.go
kevjue: Got it.  I actually think that all the uptime related functions and tests should be moved to the `istanbul` package.  But what you did is fine for now.  There are future plans to redo the uptime related code.

miner/worker.go
YoduYodu: `result.Receipts` could be `nil` if a block has no txns.
`result.State` cannot be nil, will add a nil check for `result.State`

consensus/istanbul/backend/test_utils.go
```diff
@@ -134,28 +134,31 @@ func makeBlock(keys []*ecdsa.PrivateKey, chain *core.BlockChain, engine *Backend
...
 		err := engine.Seal(chain, block, results, nil)
 		if err != nil {
 			panic(err)
 		}
 	}()
 
-	// create the sig and call Commit so that the result is pushed to the channel
+	// setup the BlockConsensusAndProcessResult cache
+	cachedResult := &istanbul.BlockProcessResult{}
+	engine.core.CurrentRoundState().SetBlockProcessResult(block.Hash(), cachedResult)
```
kevjue, line 147: Out of curiosity, why do you need to set the cached process result here?  Whereas in the other test cases when `Commit(` is called, you don't set the cached process result.

consensus/istanbul/core/roundstate.go
```diff
@@ -92,12 +94,15 @@ type roundStateImpl struct {
...
 	preparedCertificate istanbul.PreparedCertificate
 
-	// Verification status for proposals seen in this view
+	// Verification status for proposals seen in this sequence
 	// Note that this field will not get RLP enoded and persisted, since it contains an error type,
 	// which doesn't have a native RLP encoding.  Also, this is a cache, so it's not necessary for it
 	// to be persisted.
 	proposalVerificationStatus map[common.Hash]error
 
+	// Cache for BlockConsensusAndProcessResult in this sequence.
```
kevjue, line 103:
```suggestion
	// Cache for BlockProcessResult in this sequence.
```

miner/worker.go
```diff
@@ -551,32 +529,33 @@ func (w *worker) taskLoop() {
...
 			}
+			block := result.SealedBlock
 			// Short circuit when receiving duplicate result caused by resubmitting.
 			if w.chain.HasBlock(block.Hash(), block.NumberU64()) {
 				continue
 			}
 			var (
 				sealhash = w.engine.SealHash(block.Header())
 				hash     = block.Hash()
+				r        = result.BlockProcessResult
```
kevjue, line 545: nit:  Can this variable be renamed to something more informative.  e.g. `proccessResult`.

miner/worker.go
```diff
@@ -551,32 +529,33 @@ func (w *worker) taskLoop() {
...
-			task, exist := w.pendingTasks[sealhash]
-			w.pendingMu.RUnlock()
-			if !exist {
-				log.Error("Block found but no relative pending task", "number", block.Number(), "sealhash", sealhash, "hash", hash)
+
+			if r == nil || r.State == nil {
+				log.Error("BlockProcessResult cannot be nil", "number", block.Number(), "sealhash", sealhash, "hash", hash)
 				continue
 			}
+
```
kevjue, line 552: remove?

kevjue:
This PR is looking good to me.  As discussed, please run a load test to see the performance gains.

consensus/istanbul/backend/test_utils.go
YoduYodu: It's redundant.

mcortesi:
Closed since it's superseeded by new design/impl of block processing cache

jon-chuang:
Hi @mcortesi , could you point me to this new design/PR/branch please?

PR #1127, opened 2020-08-04
Add peer connection metrics by peer purpose

mcortesi:
### Description

Add number of active (connected) peers metrics discriminated by purpose (validator & proxy).

With this, we should be able to monitor how many validator or proxy connection a node currently has.

### Other changes

Refactors PurposeFlag use:

* Make it immutable, thus easier to track
* Use `Add` , `Remove` vs `Set(bool)`
* Peer now has a single purpose which is the combination of
trustedPurposes & staticPurposes

### Tested

CI, manual

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -163,6 +163,10 @@ func (p *Peer) RemovePurpose(purpose PurposeFlag) {
 	p.purpose = p.purpose.Remove(purpose)
 }
 
+func (p *Peer) HasPurpose(purpose PurposeFlag) bool {
+	return p.purpose.IsSet(purpose)
+}
+
 // ID returns the node's public key.
 func (p *Peer) ID() enode.ID {
 	return p.rw.node.ID()
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -117,8 +117,8 @@ type Peer struct {
 	// events receives message send / receive events if set
 	events *event.Feed
 
-	purposeMu sync.Mutex
-	purpose   PurposeFlag
+	purposesMu sync.Mutex
+	purposes   PurposeFlag
 
 	Server *Server
 }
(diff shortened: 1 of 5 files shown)


p2p/peer.go
```diff
@@ -117,8 +117,8 @@ type Peer struct {
 	// events receives message send / receive events if set
 	events *event.Feed
 
-	StaticNodePurposes  *PurposeFlag
-	TrustedNodePurposes *PurposeFlag
+	purposeMu sync.Mutex
+	purpose   PurposeFlag
```
tkporter, line 121: any reason for not keeping this plural here and in some other places?

p2p/peer.go
```diff
@@ -462,14 +514,14 @@ func (rw *protoRW) ReadMsg() (Msg, error) {
...
-	StaticNodePurposes  string   `json:"staticNodeInfo"`  // Purposes for the static node
-	TrustedNodePurposes string   `json:"trustedNodeInfo"` // Purposes for the trusted node
-	Network             struct {
+	ENR      string   `json:"enr,omitempty"` // Ethereum Node Record
+	Enode    string   `json:"enode"`         // Node URL
+	ID       string   `json:"id"`            // Unique node identifier
+	Name     string   `json:"name"`          // Name of the node, including client type, version, OS, custom data
+	Caps     []string `json:"caps"`          // Protocols advertised by this peer
+	Purposes string   `json:"nodeInfo"`      // Purposes for the peer
+	// TrustedNodePurposes string   `json:"trustedNodeInfo"` // Purposes for the trusted node
```
tkporter, line 523: can remove this

p2p/peer.go
```diff
@@ -462,14 +514,14 @@ func (rw *protoRW) ReadMsg() (Msg, error) {
...
-	Caps                []string `json:"caps"`            // Protocols advertised by this peer
-	StaticNodePurposes  string   `json:"staticNodeInfo"`  // Purposes for the static node
-	TrustedNodePurposes string   `json:"trustedNodeInfo"` // Purposes for the trusted node
-	Network             struct {
+	ENR      string   `json:"enr,omitempty"` // Ethereum Node Record
+	Enode    string   `json:"enode"`         // Node URL
+	ID       string   `json:"id"`            // Unique node identifier
+	Name     string   `json:"name"`          // Name of the node, including client type, version, OS, custom data
+	Caps     []string `json:"caps"`          // Protocols advertised by this peer
+	Purposes string   `json:"nodeInfo"`      // Purposes for the peer
```
tkporter, line 522: json name should change

p2p/peer.go
mcortesi: Yeah, i was hesitant about names. The type is singular `PurposeFlag` but in reality "holds" many flags....
then the singular flags like `ValidatorPurpose` are of type `PurposeFlag` (singular) which looks ok.

I'll change the name here for now

PR #1129, opened 2020-08-06
Adds new metrics for discovery

mcortesi:
### Description

Adds new metrics to the existing ones; with the objective of better understanding discovery. 

```
# Total number of discovered nodes by the les client
les_client_discovered
# Total number of inbound bytes for discovery v4
discover_ingress
# Total number of outbound bytes for discovery v4
discover_egress
# Total number of successful connections with valid handshakes (inbound)
p2p_serves_handshakes
# Total number of successful connections with valid handshakes (outbound)
p2p_dials_handshakes
```

### Other changes

No

### Tested

CI

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes

### DIFF
--- a/les/metrics.go
+++ b/les/metrics.go
@@ -112,6 +112,8 @@ var (
 
 	requestRTT       = metrics.NewRegisteredTimer("les/client/req/rtt", nil)
 	requestSendDelay = metrics.NewRegisteredTimer("les/client/req/sendDelay", nil)
+
+	clientDiscoveredNodesCounter = metrics.NewRegisteredCounter("les/client/discovered", nil) // Counter for discovered nodes
 )
 
 // meteredMsgReadWriter is a wrapper around a p2p.MsgReadWriter, capable of
--- a/les/serverpool.go
+++ b/les/serverpool.go
@@ -198,6 +198,7 @@ func (pool *serverPool) discoverNodes() {
 		if err != nil {
 			continue
 		}
+		clientDiscoveredNodesCounter.Inc(1)
 		pool.discNodes <- enode.NewV4(pubkey, n.IP, int(n.TCP), int(n.UDP))
 	}
 }
--- a/p2p/discover/metrics.go
+++ b/p2p/discover/metrics.go
@@ -1,0 +1,24 @@
+// Copyright 2018 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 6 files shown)


p2p/metrics.go
```diff
@@ -94,6 +95,7 @@ type meteredConn struct {
 	connected time.Time    // Connection time of the peer
 	addr      *net.TCPAddr // TCP address of the peer
 	peer      *Peer        // Peer instance
+	ingress   bool         // Indicates wether connection inbound or outbound
```
tkporter, line 98: teeny nit: `wether` -> `whether`

p2p/discover/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package discover
+
+import "github.com/ethereum/go-ethereum/metrics"
+
+var (
+	ingressTrafficMeter = metrics.NewRegisteredMeter("discover/ingress", nil)
+	egressTrafficMeter  = metrics.NewRegisteredMeter("discover/egress", nil)
```
nategraf, line 23: What are these measuring? Bytes?

les/serverpool.go
```diff
@@ -198,6 +198,7 @@ func (pool *serverPool) discoverNodes() {
 		if err != nil {
 			continue
 		}
+		clientDiscoveredNodesCounter.Inc(1)
```
nategraf, line 201: I mentioned when we were chatting that it would be good to know the number of _new_ discovered nodes in addition to this, which is every node returned by discovery. It looks like [findOrNewNode](https://github.com/celo-org/celo-blockchain/pull/1129/files#diff-507764eec3e28419d12153b0d1fee8cdR442) provides the logic that determines if a node is new, and this might be worth hooking in to.

les/metrics.go
```diff
@@ -112,6 +112,8 @@ var (
 
 	requestRTT       = metrics.NewRegisteredTimer("les/client/req/rtt", nil)
 	requestSendDelay = metrics.NewRegisteredTimer("les/client/req/sendDelay", nil)
+
+	clientDiscoveredNodesCounter = metrics.NewRegisteredCounter("les/client/discovered", nil) // Counter for discovered nodes
```
nategraf, line 116: It seems like a meter might be better than a count here, because if I understand correctly, counters do not report rates over a given time period, but meters do.

p2p/discover/metrics.go
mcortesi: Yes, bytes. The idea always to measure in standard units and then convert that to whatever you need on dispaly (Grafana)

les/metrics.go
mcortesi: When using something like prometheus, you report only the counter. (the simplest thing you can do) and leave those problems to the monitoring engine.

In PromQL you would do: `rate(les_client_discovered[1m])`

p2p/metrics.go
mcortesi: thanks

les/serverpool.go
mcortesi: we chatted offline a decided to better analyze serverpool before adding more metrics

PR #1130, opened 2020-08-06
Adds "get validator public keys" endpoint

kobigurk:
### Description

The Plumo prover needs a list of the BLS public keys for epochs in order to create the witness for the proof.

### DIFF
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -27,9 +27,9 @@ import (
 	"github.com/ethereum/go-ethereum/consensus/istanbul/core"
 	"github.com/ethereum/go-ethereum/consensus/istanbul/validator"
 	"github.com/ethereum/go-ethereum/core/types"
+	blscrypto "github.com/ethereum/go-ethereum/crypto/bls"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/rpc"
-	blscrypto "github.com/ethereum/go-ethereum/crypto/bls"
 )
 
 // API is a user facing RPC API to dump Istanbul state
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -90,8 +90,8 @@ func (api *API) GetValidators(number *rpc.BlockNumber) ([]common.Address, error)
 	return istanbul.MapValidatorsToAddresses(validators), nil
 }
 
-// GetValidatorsPublicKeys retrieves the list of validators public keys that must sign a given block.
-func (api *API) GetValidatorsPublicKeys(number *rpc.BlockNumber) ([]blscrypto.SerializedPublicKey, error) {
+// GetValidatorsBLSPublicKeys retrieves the list of validators BLS public keys that must sign a given block.
+func (api *API) GetValidatorsBLSPublicKeys(number *rpc.BlockNumber) ([]blscrypto.SerializedPublicKey, error) {
 	header, err := api.getParentHeaderByNumber(number)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


consensus/istanbul/backend/api.go
```diff
@@ -89,6 +90,16 @@ func (api *API) GetValidators(number *rpc.BlockNumber) ([]common.Address, error)
 	return istanbul.MapValidatorsToAddresses(validators), nil
 }
 
+// GetValidatorsPublicKeys retrieves the list of validators public keys that must sign a given block.
+func (api *API) GetValidatorsPublicKeys(number *rpc.BlockNumber) ([]blscrypto.SerializedPublicKey, error) {
```
mcortesi, line 94: how is `blscrypto.SerializedPublicKey` encoded to json? does it have `Marshall/Unmarshall` methods already defined?

consensus/istanbul/backend/api.go
```diff
@@ -89,6 +90,16 @@ func (api *API) GetValidators(number *rpc.BlockNumber) ([]common.Address, error)
...
+	header, err := api.getParentHeaderByNumber(number)
+	if err != nil {
+		return nil, err
+	}
+	validators := api.istanbul.GetValidators(header.Number, header.Hash())
```
mcortesi, line 99: Is this an expensive operation? In particular if asked for an old epoch. My "angle" here is.. would this provide a simple way to DoS the node?

internal/web3ext/web3ext.go
```diff
@@ -733,6 +733,12 @@ web3._extend({
 			params: 1,
 			inputFormatter: [null]
 		}),
+		new web3._extend.Method({
+			name: 'getValidatorsPublicKeys',
```
kevjue, line 737: nit:  Perhaps rename to `getValidatorBLSPublicSigningKeys`, to differentiate from the `ECDSA consensus signing keys`.

consensus/istanbul/backend/api.go
kobigurk: We already have a `GetValidators` operation that gets you the addresses, this shouldn't be more expensive than that.

consensus/istanbul/backend/api.go
kobigurk: Hmm, good point. I've seen it was passed as byte arrays between Rust and Go, and in JSON it's serialized as an array of integers.

internal/web3ext/web3ext.go
kobigurk: Good idea.

consensus/istanbul/backend/api.go
mcortesi: Normally we use hex strings. you can use the `hexutil.Bytes` check https://github.com/celo-org/celo-blockchain/blob/d42b997916c74cc17a623e341618cd8190b27d9a/internal/ethapi/api.go#L691

consensus/istanbul/backend/api.go
mcortesi: fair enough :P

consensus/istanbul/backend/api.go
kobigurk: Sounds good :)
How does it happen automatically for `common.Address`? I see methods returning it directly.

mcortesi: Because of...
```
// MarshalText returns the hex representation of a.
func (a Address) MarshalText() ([]byte, error) {
	return hexutil.Bytes(a[:]).MarshalText()
}

// UnmarshalText parses a hash in hex syntax.
func (a *Address) UnmarshalText(input []byte) error {
	return hexutil.UnmarshalFixedText("Address", input, a[:])
}

// UnmarshalJSON parses a hash in hex syntax.
func (a *Address) UnmarshalJSON(input []byte) error {
	return hexutil.UnmarshalFixedJSON(addressT, input, a[:])
}

```

mcortesi: Same happens for Hash, definitely the way to do it.

Check https://github.com/celo-org/celo-blockchain/blob/0ec5ab417559ac3399a73f76219fce043dfc0ef9/common/types.go#L253

kobigurk: Thanks! Done :)

PR #1132, opened 2020-08-07
Mobile discovery improvements

tkporter:
### Description

When a mobile app using mobile geth is closed entirely, it does not gracefully shut down. As a result, the server pool entries were not being properly persisted & loaded upon startup.

This PR:
* Adds the `les_serverPoolEntries` RPC to a new private LES client API to easily view server pool entries
* Saves server pool entries when there is an update to the list of entries that have at least attempted to have been made peers. This is fairly infrequent (there are max 5 entries)
* Allows the creation of an HTTP RPC server on mobile for debugging

### Other changes

Looks like some go.mod/go.sum changes were made automatically, not sure if these are undesirable

I'm still testing this more but wanted to put the PR out

### Tested

Ran on apps & peered more quickly when properly saving server pool entries

### Backwards compatibility

Backward compatible


les/client.go
```diff
@@ -234,6 +234,11 @@ func (s *LightEthereum) APIs() []rpc.API {
 			Version:   "1.0",
 			Service:   NewLightClientAPI(s),
 			Public:    true,
+		}, {
+			Namespace: "les",
+			Version:   "1.0",
+			Service:   NewPrivateLightClientAPI(s),
```
nategraf, line 240: Instead of making a private light client API seperate from a "public" version, Id recommend merging the new function into `LightClientAPI` and making that API private. As far are I can tell, there is no reason to have a "public" light client API.

les/serverpool.go
```diff
@@ -657,6 +656,28 @@ func (pool *serverPool) checkDialTimeout(entry *poolEntry) {
 	pool.setRetryDial(entry)
 }
 
+func (pool *serverPool) Info() []*poolEntryInfo {
+	stateStringMap := make(map[int]string)
```
nategraf, line 660: Could this be extracted as a mapping constructed as a map literal at the package scope instead of doing it inline here?

PR #1136, opened 2020-08-13
CIP-21 Governable "lookbackWindow" parameter

andres-dg:
## Description

* Implementation of [CIP-21](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0021.md).
* Monrepo Counterpart  https://github.com/celo-org/celo-monorepo/pull/4747

Modifies celo-blockchain to obtain `lookbackWindow` parameter from the `BlockchainParameters` smart contract. It assumes the smart contract will return the same value during an epoch.

`lookbackWindow` is used on 3 places:

  1. For metrics; where the parent block signatures are analyzed.
  2. For every block on `Blockchain` we obtain the `loobackWindow` asses the "upness" of each validator on that block
  3. On the last block of the epoch, to compute the uptime score
  

## Tested

Added unit tests.

## Backwards compatibility

**Hardfork Change**. Part of `Donut` Hardfork

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -428,6 +428,7 @@ func (sb *Backend) EpochSize() uint64 {
 }
 
 // LookbackWindow returns the size of the lookback window for calculating uptime (in blocks)
+// Value is constant during an epoch
 func (sb *Backend) LookbackWindow(header *types.Header, state *state.StateDB) (uint64, error) {
 	return uptime.LookbackWindow(sb.chain.Config(), sb.config, header, state)
 }
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -431,10 +431,15 @@ func (sb *Backend) EpochSize() uint64 {
 // LookbackWindow returns the size of the lookback window for calculating uptime (in blocks)
 // Value is constant during an epoch
 func (sb *Backend) LookbackWindow(header *types.Header, state *state.StateDB) uint64 {
+	// Check if donut was already active at the beginning of the epoch
+	// as we want to activate the change at epoch change
+	firstBlockOfEpoch := istanbul.MustGetEpochFirstBlockGivenBlockNumber(header.Number.Uint64(), sb.config.Epoch)
+	isDonutActivated := sb.chain.Config().IsDonut(new(big.Int).SetUint64(firstBlockOfEpoch))
+
 	return uptime.ComputeLookbackWindow(
 		sb.config.Epoch,
 		sb.config.DefaultLookbackWindow,
-		sb.chain.Config().IsDonut(header.Number),
+		isDonutActivated,
 		func() (uint64, error) { return blockchain_parameters.GetLookbackWindow(header, state) },
 	)
 }
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -29,6 +29,7 @@ import (
 	istanbulCore "github.com/ethereum/go-ethereum/consensus/istanbul/core"
 	"github.com/ethereum/go-ethereum/consensus/istanbul/uptime"
 	"github.com/ethereum/go-ethereum/consensus/istanbul/validator"
+	"github.com/ethereum/go-ethereum/contract_comm/blockchain_parameters"
 	gpm "github.com/ethereum/go-ethereum/contract_comm/gasprice_minimum"
 	ethCore "github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/state"
@@ -429,8 +430,13 @@ func (sb *Backend) EpochSize() uint64 {
 
 // LookbackWindow returns the size of the lookback window for calculating uptime (in blocks)
 // Value is constant during an epoch
-func (sb *Backend) LookbackWindow(header *types.Header, state *state.StateDB) (uint64, error) {
-	return uptime.LookbackWindow(sb.chain.Config(), sb.config, header, state)
+func (sb *Backend) LookbackWindow(header *types.Header, state *state.StateDB) uint64 {
+	return uptime.ComputeLookbackWindow(
+		sb.config.Epoch,
+		sb.config.DefaultLookbackWindow,
+		sb.chain.Config().IsDonut(header.Number),
+		func() (uint64, error) { return blockchain_parameters.GetLookbackWindow(header, state) },
+	)
 }
 
 // Finalize runs any post-transaction state modifications (e.g. block rewards)
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
 	istanbulCore "github.com/ethereum/go-ethereum/consensus/istanbul/core"
 	"github.com/ethereum/go-ethereum/consensus/istanbul/validator"
+	"github.com/ethereum/go-ethereum/contract_comm/blockchain_parameters"
 	gpm "github.com/ethereum/go-ethereum/contract_comm/gasprice_minimum"
 	ethCore "github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/state"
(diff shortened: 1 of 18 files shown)


consensus/istanbul/backend/engine.go
```diff
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/ethereum/go-ethereum/contract_comm/blockchain_parameters"
```
andres-dg, line 23: This import might be wrong.

mcortesi (approved):

consensus/istanbul/uptime/config_test.go
```diff
@@ -0,0 +1,74 @@
...
+		isDonut               bool
+		getLookbackWindow     func() (uint64, error)
+	}
+	tests := []struct {
+		name string
+		args args
+		want uint64
+	}{
+		{"returns default if Donut is not active", args{100, 20, false, constant(24)}, 20},
+		{"returns safe minimun if configured is below", args{100, 20, true, constant(2)}, 3},
```
oneeman, line 62: typo in minimum

consensus/istanbul/uptime/config_test.go
```diff
@@ -0,0 +1,74 @@
...
+		{"returns safe maximum if configured is above", args{1000, 20, true, constant(800)}, 720},
+		{"returns epochSize-2 if configured is above", args{100, 20, true, constant(99)}, 98},
+	}
```
oneeman, line 65: There should add a case covering the normal donut case.

consensus/istanbul/uptime/config_test.go
```diff
@@ -0,0 +1,74 @@
...
+	tests := []struct {
+		name      string
+		args      args
+		wantStart uint64
+		wantEnd   uint64
+	}{
+		{"monitoringWindow on first epoch", args{1, 10, 2}, 2, 8},
+		{"monitoringWindow on second epoch", args{2, 10, 2}, 12, 18},
+		{"lookback window too big", args{1, 10, 10}, 10, 8},
+		// TODO: Add test cases.
```
oneeman, line 30: remove this TODO (after adding any cases you think we should add)

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -67,21 +67,36 @@ const (
...
+	"outputs": [
+		{
+		"name": "",
+		"type": "uint256"
+		}
+	],
+	"payable": false,
+	"stateMutability": "view",
+	"type": "function"
+	}
```
oneeman, line 99: The whitespace here is messed up (lines 87-98 should be indented relative to the braces, and 92-93 indented relative to their braces).   Though it was already messed up to some degree before this PR (consistenting of a mix of 2-space indents and tab indents).

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -67,21 +67,36 @@ const (
...
 		"inputs": [],
-		"name": "intrinsicGasForAlternativeFeeCurrency",
+		"name": "getUptimeLookbackWindow",
 		"outputs": [
-		  {
-			"name": "",
-			"type": "uint256"
-		  }
+			{
+				"internalType": "uint256",
```
oneeman, line 77: I don't know much about ABI.   Is this internalType necessary?  Does it make a difference?  Should we have it elsewhere as well?  Is it generated by another tool in which internalType was added in a newer version?

params/config.go
```diff
@@ -111,6 +113,8 @@ var (
 		Istanbul: &IstanbulConfig{
 			Epoch:          17280,
 			ProposerPolicy: 2,
+			BlockPeriod:    5,
+			RequestTimeout: 10000,
```
oneeman, line 117: change to 3000 as we discussed in #1288

consensus/istanbul/config.go
```diff
@@ -81,7 +98,28 @@ var DefaultConfig = &Config{
 	AnnounceAdditionalValidatorsToGossip:           10,
 }
 
-type ProxyConfig struct {
-	InternalNode *enode.Node `toml:",omitempty"` // The internal facing node of the proxy that this proxied validator will peer with
-	ExternalNode *enode.Node `toml:",omitempty"` // The external facing node of the proxy that the proxied validator will broadcast via the announce message
+//ApplyChainConfigToConfig applies chainConfig values to Config
+func ApplyChainConfigToConfig(chainConfig *params.ChainConfig, config *Config) error {
```
oneeman, line 102:
```suggestion
//ApplyParamsChainConfigToConfig applies the istanbul config values from params.chainConfig to the istanbul.Config config
func ApplyParamsChainConfigToConfig(chainConfig *params.ChainConfig, config *Config) error {
```

The word "config" is in there like 10,000 times, but I think given two different config types are involved it's better to call it out in the commend and function name.  Would need to be updated where the function is called, as well.

consensus/istanbul/backend/engine.go
```diff
@@ -416,19 +418,25 @@ func (sb *Backend) UpdateValSetDiff(chain consensus.ChainReader, header *types.H
...
-// Returns the size of the lookback window for calculating uptime (in blocks)
-func (sb *Backend) LookbackWindow() uint64 {
-	return sb.config.LookbackWindow
+// LookbackWindow returns the size of the lookback window for calculating uptime (in blocks)
+// Value is constant during an epoch
+func (sb *Backend) LookbackWindow(header *types.Header, state *state.StateDB) uint64 {
+	return uptime.ComputeLookbackWindow(
+		sb.config.Epoch,
+		sb.config.DefaultLookbackWindow,
+		sb.chain.Config().IsDonut(header.Number),
```
oneeman, line 437: I kind of think this should use `IsDonut(firstBlockOfEpoch)`.  Otherwise, when Donut activates, we could end up with different values within that epoch in a scenario where the value in the contract has by then already been modified through a governance proposal (or initialized to an incorrect value).  That is extremely unlikely to happen, but it still seems more correct to use the beginning of the epoch for the Donut check so that it only takes effect at the beginning of the epoch after activation (or the same epoch if activation is on the epoch's first bock).

Of course, there is a caveat that the CIP says "Starting at HARDFORK_BLOCK, block creation and validation MUST read the lookback window from chain state by calling the BlockchainParameters contract, rather than reading chain paramaters."  But the CIP is actually out of date already, since it doesn't take into account the new contract design where `uptimeLookbackWindow` is a struct rather than `uint256`.

consensus/istanbul/uptime/config_test.go
prestwich, line 62: Should the min/max expected output here be literals, or references to the chain parameters?

consensus/istanbul/config.go
```diff
@@ -17,10 +17,19 @@
...
 import (
+	"fmt"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+const (
+	//MinEpochSize represents minimun block that can conform an epoch
```
oneeman, line 28: There is a typo in "minimum", but also "conform" isn't right here.  I'd just simplify it as follows:

```suggestion
	//MinEpochSize represents the minimum permissible epoch size
```

consensus/istanbul/config.go
```diff
@@ -37,7 +47,7 @@ type Config struct {
 	BlockPeriod                 uint64         `toml:",omitempty"` // Default minimum difference between two consecutive block's timestamps in second
 	ProposerPolicy              ProposerPolicy `toml:",omitempty"` // The policy for proposer selection
 	Epoch                       uint64         `toml:",omitempty"` // The number of blocks after which to checkpoint and reset the pending votes
-	LookbackWindow              uint64         `toml:",omitempty"` // The window of blocks in which a validator is forgived from voting
+	DefaultLookbackWindow       uint64         `toml:",omitempty"` // The default value for the window of blocks in which a validator is forgived from voting
```
oneeman, line 50:
```suggestion
	DefaultLookbackWindow       uint64         `toml:",omitempty"` // The default value for how many blocks in a row a validator must miss to be considered "down"
```

oneeman (commented):
Looks good.  I added a bunch of nitpicks, and a couple of small substantial comments (iirc: one about the Baklava request timeout and one about the `IsDonut()` call.

consensus/istanbul/uptime/config_test.go
mcortesi: done

contract_comm/blockchain_parameters/blockchain_parameters.go
mcortesi: we normally copy and paste the output of truffle compile. For that i wouldn't change it

consensus/istanbul/backend/engine.go
mcortesi: good idea.

oneeman (approved):

prestwich (approved):

consensus/istanbul/uptime/config.go
```diff
@@ -24,32 +24,38 @@ const (
...
+	var returnValue uint64
 
 	if epochSize <= istanbul.MinEpochSize {
 		panic("Invalid epoch value")
 	}
 
-	value, err := getLookbackWindow()
-	if err != nil {
-		value = MinSafeLookbackWindow
+	if !isDonut {
```
oneeman: I usually prefer to flip the then/else to have `if isDonut {` instead, except when there is a reason not to.  Having the shorter block first is nice, but not necessarily worth it.

consensus/istanbul/uptime/config.go
```diff
@@ -24,32 +24,38 @@ const (
 // ComputeLookbackWindow computes the lookbackWindow based on different required parameters.
 // getLookbackWindow represents the way to obtain lookbackWindow from the smart contract
 func ComputeLookbackWindow(epochSize uint64, defaultLookbackWindow uint64, isDonut bool, getLookbackWindow func() (uint64, error)) uint64 {
-	if !isDonut {
-		return defaultLookbackWindow
-	}
+	var returnValue uint64
```
oneeman: Why `returnValue` rather than `value`?  Also, we could put in the signature `(value uint64)`

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -189,3 +204,26 @@ func GetBlockGasLimit(header *types.Header, state vm.StateDB) (uint64, error) {
...
+	)
+	if err != nil {
+		if err == errors.ErrRegistryContractNotDeployed {
+			log.Debug("Error obtaining lookback window", "err", err, "contract", hexutil.Encode(params.BlockchainParametersRegistryId[:]))
+		} else {
+			log.Warn("Error obtaining lookback window", "err", err, "contract", hexutil.Encode(params.BlockchainParametersRegistryId[:]))
+		}
+		return 0, err
+	}
+	return lookbackWindow.Uint64(), nil
```
prestwich, line 228: this will have the side effect of truncating any 256-bit value to a go uint64?

prestwich: basically I'm double-checking that an invalid contract state won't cause a panic or something here

PR #1140, opened 2020-08-18
Multivalidator support

trianglesphere:
### Description

This allows proxies to support multiple (up to 10 based on code defaults) validators at once.
Validators are configured with the enode of the proxy and automatically reach out, so no configuration changes need to be done with the proxy.
Note that the this does not support multiple validator signing addresses.

### Other changes

N/A

### Tested

This will be tested as part of my testing on hot swapping validators with a start/stop API.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes


consensus/istanbul/proxy/proxy_engine.go
```diff
@@ -66,11 +67,10 @@ type proxyEngine struct {
 	logger  log.Logger
 	backend BackendForProxyEngine
 
-	// Proxy's validator
-	// Right now, we assume that there is at most one proxied peer for a proxy
-	// Proxy's validator
-	proxiedValidator   consensus.Peer
-	proxiedValidatorMu sync.RWMutex
+	// Proxied Validators set and count of nodekey addresses
```
YoduYodu, line 70: comment fix: "count of nodekey addresses"

trianglesphere: Nice catch

mcortesi (approved):
Looks good!

consensus/istanbul/proxy/proxied_validator_engine.go
```diff
@@ -598,17 +598,26 @@ func (pv *proxiedValidatorEngine) getValidatorConnSetDiff(validators []common.Ad
...
+func (pv *proxiedValidatorEngine) IsProxyPeer(peerID enode.ID) (bool, error) {
+	proxy, err := pv.getProxy(peerID)
 	if err != nil {
 		return false, err
 	}
 
+	return proxy != nil, nil
+}
+
+func (pv *proxiedValidatorEngine) getProxy(peerID enode.ID) (*Proxy, error) {
```
kevjue, line 610: nit: perhaps this should be generalized to `getProxies`?

PR #1141, opened 2020-08-19
Hotswap Validators

trianglesphere:
### Description

This PR add "replica" validators which can start mining at a given block. Primary validators can be instructed to stop at a given block through the RPC API.

Validators can be started as replicas with the `--istanbul.replica` flag. They should still be provided `--mine` because they start "mining" without participating in consensus. This is done to allow the node to participate in a limited form of announce to get the enode urls of the other validators.

To switch validators
-  spin up a replica
- Once synced & it has the enode urls for the rest of the validators, pick a block in the future to switch on.
- Run `istanbul.stopAtBlock(BLOCK_NUM)` on the primary
- Run `istanbul.startAtBlock(BLOCK_NUM)` on the replica
- You can verify the switch block in `istanbul.replicaState`. Note the switch block number should be the same for the primary and replica.
- spin down the old primary

Added RPC Calls
* `istanbul.start()` Immediately make a replica the primary & participate in consensus.
* `istanbul.stop()` Immediately make a primary a replica & stop participating consensus.
* `istanbul.startAtBlock(blockNum)` The node will start validating on the given block # 
* `istanbul.stopAtBlock(blockNum)` The node will stop validating on the block number. (Will not validate the stop block)
* `istanbul.validating` If this node is actively validating
* `istanbul.replicaState ` Information on the block that this node will switch at and if this is a primary or replica

**To Do for testing**:
- Add test for killing validator and restarting it to check that it persists replica state.

### Other changes

Did rework some of announce & the backend message handler for clarity.

### Tested

Tested manually the following flows:
- Start network, spin up secondary, switch to secondary, switch back to primary.
- Start network, spin up secondary, schedule start & stop, cancel start & stop, switch to secondary
- Start network, spin up secondary, use start/stop API to switch to and from primary (will skip blocks b/c manual)

Unit tests - for `replicaState` and `checkMessage`
e2e tests - done, see  https://github.com/celo-org/celo-monorepo/pull/5029

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes


eth/backend.go
```diff
@@ -251,6 +250,22 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
 				}
 			}
 		}()
+
+		chainEventCh := make(chan core.ChainEvent, 10)
```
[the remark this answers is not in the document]
trianglesphere: Moved to `backend.SetChain`. This does rely on a type cast, but otherwise works.

consensus/istanbul/backend/backend.go
```diff
@@ -303,14 +303,24 @@ func (sb *Backend) IsProxiedValidator() bool {
...
 
 // IsValidator return if instance is a validator (either proxied or standalone)
 func (sb *Backend) IsValidator() bool {
 	return sb.config.Validator
 }
 
+//  IsElectedValidator returns true if instance is an elected validator
+func (sb *Backend) IsElectedValidator() bool {
+	// Check if this node is in the validator connection set
+	validatorConnSet, err := sb.RetrieveValidatorConnSet()
```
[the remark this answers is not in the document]
oneeman: I think this is a naming issue.  I haven't checked all the places this is used, but in the ones that I've checked it replaces a function that was called `shouldSaveAndPublishValEnodeURLs()` and which is implemented this way and intended to include the buffer nodes: https://github.com/celo-org/celo-blockchain/blob/96f69040267fa4c6aec666766660e5dc1e1c7ce3/consensus/istanbul/backend/announce.go#L255-L264

Here is an example of a place it's used that makes this explicit in a comment: https://github.com/celo-org/celo-blockchain/blob/96f69040267fa4c6aec666766660e5dc1e1c7ce3/consensus/istanbul/backend/announce.go#L594-L600

consensus/istanbul/core/replicastate.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func newReplicaState(isReplica bool) ReplicaState {
+	return &replicaStateImpl{
+		isReplica: isReplica,
+		mu:        new(sync.RWMutex),
+	}
+}
+
+func (rs *replicaStateImpl) SetStartValidatingBlock(blockNumber *big.Int) error {
```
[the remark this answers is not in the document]
trianglesphere: Stuff moved, but still need to implement

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
trianglesphere: Yea, I only see it being used in announce which does have a different notion. I didn't love `shouldSaveAndPublishValEnodURLs`, maybe `shouldParticipateInAnnounce` (??, still long, but hopefully conveys the message).

trianglesphere:
Okay, I've done some relatively large changes, but I'm not still fully down with the refactor.
- Split `generic_db.go` out of the `enodes` package. This lets me reuse a simple leveldb interface for the replica state db. This could be split out into it's own PR.
- Rework `backend.HandleMsg` again. This could be split up into it's own PR such that the replica code could be a smaller diff.
- Put `replica state` in `backend` and switch the core on and off based on if it's validating or not. I still need to do more work around removing old traces of hot swap from the core code.
- There's some things in the proxy code here that I'd like to cherry pick onto `simple_mv` (If this PR doesn't get into the multiproxy release).
It is in a place where e2e tests are working, but still in the middle of the refactor.

consensus/istanbul/backend/message_senders.go
```diff
@@ -110,12 +110,16 @@ func (sb *Backend) Gossip(payload []byte, ethMsgCode uint64) error {
 
 // sendMsg will asynchronously send the the Celo messages to all the peers in the destPeers param.
 func (sb *Backend) asyncMulticast(destPeers map[enode.ID]consensus.Peer, payload []byte, ethMsgCode uint64) {
-	logger := sb.logger.New("func", "AsyncMulticastCeloMsg")
+	logger := sb.logger.New("func", "AsyncMulticastCeloMsg", "msgCode", ethMsgCode)
+	if ethMsgCode == istanbul.ConsensusMsg && !sb.IsProxy() && !sb.IsValidating() {
```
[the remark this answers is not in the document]
trianglesphere: Removed

consensus/istanbul/core/request.go
```diff
@@ -42,7 +42,8 @@ func (c *core) handleRequest(request *istanbul.Request) error {
 	}
 
 	// Must go through startNewRound to send proposals for round > 0 to ensure a round change certificate is generated.
-	if c.current.State() == StateAcceptRequest && c.current.Round().Cmp(common.Big0) == 0 {
+	// Also ignore requests when not validating.
+	if c.backend.IsValidating() && c.current.State() == StateAcceptRequest && c.current.Round().Cmp(common.Big0) == 0 {
```
mcortesi, line 46: can we make it that if it's a replica "request" won't even get here? They are redirected somewhere else, in a higher layer (the one that handles switching from replica core impl or primary core impl)

consensus/istanbul/core/preprepare.go
```diff
@@ -27,8 +27,8 @@ import (
 func (c *core) sendPreprepare(request *istanbul.Request, roundChangeCertificate istanbul.RoundChangeCertificate) {
 	logger := c.newLogger("func", "sendPreprepare")
 
-	// If I'm the proposer and I have the same sequence with the proposal
-	if c.current.Sequence().Cmp(request.Proposal.Number()) == 0 && c.isProposer() {
+	// If I'm validating, the proposer, and I have the same sequence with the proposal
+	if c.current.Sequence().Cmp(request.Proposal.Number()) == 0 && c.isProposer() && c.backend.IsValidating() {
```
mcortesi, line 31: why the `isValidating()`? shouldn't core be "stop" if we are not validating?

consensus/istanbul/core/core.go
```diff
@@ -92,6 +92,14 @@ type CoreBackend interface {
 
 	// ParentBlockValidators returns the validator set of the given proposal's parent block
 	ParentBlockValidators(proposal istanbul.Proposal) istanbul.ValidatorSet
+
+	// IsValidating return true if this instance is validating
+	IsValidating() bool
+
+	// IsElectedValidator returns true if instance is an elected validator
+	IsElectedValidator() bool
```
mcortesi, line 100: no used i believe

consensus/istanbul/core/core.go
```diff
@@ -92,6 +92,14 @@ type CoreBackend interface {
...
+
+	UpdateReplicaState() bool
```
mcortesi, line 102: same

consensus/istanbul/backend/engine.go
```diff
@@ -560,6 +561,41 @@ func (sb *Backend) SetChain(chain consensus.ChainReader, currentBlock func() *ty
 	sb.chain = chain
 	sb.currentBlock = currentBlock
 	sb.stateAt = stateAt
+
+	if bc, ok := chain.(*ethCore.BlockChain); ok {
+		// Start listeners
+		chainHeadCh := make(chan ethCore.ChainHeadEvent, 10)
```
mcortesi, line 567: We can encapsulate both goroutines (maybe even have just one) behind a function named so that it conveys the nature of both routines... in fact, having that we don't need to have a `UpdateReplicaState` method, since it would an internal call here. and maybe the routine is `replicaStatusListener()` or something..

same with `newChainHead` do we need a method for that? i believe it was created just because it was a public interface.

consensus/istanbul/backend/engine.go
```diff
@@ -695,6 +737,21 @@ func (sb *Backend) StopProxiedValidatorEngine() error {
 	return nil
 }
 
+// MakeReplica clears the start/stop state & stops this node from participating in consensus
+func (sb *Backend) MakeReplica() {
+	sb.logger.Info("MakeReplica")
+	sb.StopValidating()
```
mcortesi, line 743: StopValidating can return an err

consensus/istanbul/backend/engine.go
```diff
@@ -695,6 +737,21 @@ func (sb *Backend) StopProxiedValidatorEngine() error {
...
+	sb.replicaState.MakeReplica()
+}
+
+// MakePrimary clears the start/stop state & makes this node participate in consensus
+func (sb *Backend) MakePrimary() {
+	sb.logger.Info("MakePrimary")
+	// start core
+	sb.StartValidating(sb.hasBadBlock, sb.processBlock, sb.validateState)
```
mcortesi, line 751: same here, thsi can fail.

consensus/istanbul/backend/handler.go
```diff
@@ -43,70 +43,120 @@ const (
...
-			}
-		} else if sb.IsValidating() {
-			if msg.Code == istanbul.ConsensusMsg {
-				go sb.istanbulEventMux.Post(istanbul.MessageEvent{
-					Payload: data,
-				})
-				return true, nil
-			} else if msg.Code == istanbul.EnodeCertificateMsg {
-				go sb.handleEnodeCertificateMsg(peer, data)
+	} else if sb.IsValidating() {
```
mcortesi, line 96: maybe rename to `IsPrimaryValidator()`

consensus/istanbul/backend/handler.go
```diff
@@ -253,9 +303,9 @@ func (sb *Backend) UpdateMetricsForParentOfBlock(child *types.Block) {
 }
 
 // Actions triggered by a new block being added to the chain.
-func (sb *Backend) NewChainHead(newBlock *types.Block) {
+func (sb *Backend) newChainHead(newBlock *types.Block) {
```
mcortesi, line 306: From our conversation before, it looks that his method in not actually called for every chainHead, right? Or at least, not realtime.

We should review the contents of the method, probably we were assuming otherwise

consensus/istanbul/core/core.go
trianglesphere: `UpdateReplicaState` does get used in `startNewRound`. That goal of that is to make sure that the core checks that it's the primary for the sequence prior to starting that sequence.

consensus/istanbul/backend/engine.go
trianglesphere: Updated. See `consensus/istanbul/backend/engine.go:572` I changed `NewChainHead` -> `newChainHead` because I assumed we added it. It could be reverted back to being a public interface if it's supposed to be part of that interface.

consensus/istanbul/core/request.go
trianglesphere: This is a leftover. Because the core is properly stopped, the event handler should be stopped which in turn won't pass requests to `handleRequest`

consensus/istanbul/core/preprepare.go
trianglesphere: Again, old check.

consensus/istanbul/backend/handler.go
trianglesphere: It's mainly metrics. On epoch transitions it does call refresh val peers for announce, but I think that gets called other places as well.

mcortesi (commented):
will continuew with the review but i wanted to send an update first

consensus/istanbul/backend/backend.go
```diff
@@ -939,3 +957,24 @@ func (sb *Backend) VerifyValidatorConnectionSetSignature(data []byte, sig []byte
 		return istanbul.CheckValidatorSignature(validator.NewSet(validators), data, sig)
 	}
 }
+
+func (sb *Backend) UpdateReplicaState() bool {
```
mcortesi, line 961: It seems to me that there's this "state machine" where we have several inputs:
```
-> newBlockHeader
-> makePrimary/Replica
-> start/stopXXX
-> configureStop/StartBlock
```

and those action might generate state transtitions, which trigger start or stop of subservices.

Now, today, that logic is dispersed in a few methods: `UpdateReplicatState` `Start/StopValidating` `MakeReplica/Primary`. I wonder if it would be better to have the control logic isolated on its own entity, and assign some trigger function of "start/stopXXX" when you enter/exit states.

That way, control logic would be clearer.

consensus/istanbul/backend/engine.go
```diff
@@ -583,6 +628,12 @@ func (sb *Backend) StartValidating(hasBadBlock func(common.Hash) bool,
 	sb.processBlock = processBlock
 	sb.validateState = validateState
 
+	// Don't start core if we're a replica
+	blockNum := new(big.Int).Add(common.Big1, sb.currentBlock().Number())
```
mcortesi, line 632: this function serves 2 purposes:

- enable (configure) core subsystems: Means that it should run, whenever we are primary
- start subsystem: run it

Going back to my idea of state machine, it would be nice to separate both concerns. Also, rn updateReplicaState ends up calling StartValidator for the `start` part, but not to configure it.

mcortesi: Also, what about `Stop` (disable). For example if we have configured `startValidatingAt` but previously we call `stop`?

It's not really clear if this are orthogonal things (enable, versus primary/replica) or the same thing.

consensus/istanbul/backend/backend.go
trianglesphere: There definitely is. I'd put the inputs as follows:
```
-> block number (from bc.CurrentBlock() or from roundstate.Seq)
-> start validating (external RPC, also clears start/stop blocks)
-> stop validating (external RPC, also clears start/stop blocks)
-> set start validating block (external RPC)
-> set stop validating block (external RPC)
```
One thing I was grappling with is the outputs
```
-> Start core validating
-> Stop core from validating
```
Both of those are function calls, so it's figuring out where this can live and have access to those functions in a clean way.

consensus/istanbul/core/core.go
```diff
@@ -455,6 +458,13 @@ func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbu
 // startNewRound starts a new round. if round equals to 0, it means to starts a new sequence
 func (c *core) startNewRound(round *big.Int) error {
 
+	// Inform the backend that a new sequence has started & bail if the backed stopped the core
+	if round.Cmp(common.Big0) == 0 {
+		if disabledCore := c.backend.UpdateReplicaState(); disabledCore {
```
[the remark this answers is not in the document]
trianglesphere: Stale reads are definitely a concern. Hopefully the blockchain object can properly handle concurrent reads and it's not a broken read.

I agree that core call update w/ it's current sequence as a parameter when running as core.

consensus/istanbul/backend/engine.go
trianglesphere: I briefed tried splitting out the configure into `injectBlockProcessors` and then having a separate `StartValidating()`. It was less mixed, but I didn't feel much cleaner. Thinking about it more, I could split and then wrap `StartValidating` in miner with a call to `IsPrimaryForSeq`

consensus/istanbul/backend/backend.go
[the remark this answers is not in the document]
trianglesphere: It's not particularly cheap, but I don't think that I call it much more than in multi proxy so it's kinda a wash.

consensus/istanbul/backend/announce.go
```diff
@@ -125,14 +125,12 @@ func (sb *Backend) announceThread() {
 		case <-checkIfShouldAnnounceTicker.C:
 			logger.Trace("Checking if this node should announce it's enode")
 
-			shouldAnnounce, err = sb.shouldSaveAndPublishValEnodeURLs()
-			if err != nil {
-				logger.Warn("Error in checking if should announce", err)
-				break
-			}
+			shouldQuery = sb.shouldParticipateInAnnounceAnnounce()
```
kevjue, line 128: How come you need to split up these periodic query and periodic announce version update?

consensus/istanbul/backend/announce.go
```diff
@@ -260,15 +282,14 @@ func (sb *Backend) startGossipQueryEnodeTask() {
 	}
 }
 
-func (sb *Backend) shouldSaveAndPublishValEnodeURLs() (bool, error) {
-
+// shouldParticipateInAnnounceAnnounce returns true if instance is an elected or nearly elected validator.
+func (sb *Backend) shouldParticipateInAnnounceAnnounce() bool {
```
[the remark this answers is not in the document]
trianglesphere: Nope, that was just a bad copy and paste.

consensus/istanbul/backend/announce.go
```diff
@@ -260,15 +282,14 @@ func (sb *Backend) startGossipQueryEnodeTask() {
...
 	// Check if this node is in the validator connection set
 	validatorConnSet, err := sb.RetrieveValidatorConnSet()
 	if err != nil {
-		return false, err
+		return false
```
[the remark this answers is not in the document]
trianglesphere: It was part of the many refactors. I ended up revert this change.

consensus/istanbul/backend/announce.go
trianglesphere: Replicas need to query to populate their validator enode table (specifically the encrypted enode urls), but don't want to update the proxie's validator assignments at the same time as the primary. Also added some more comments in the code.

consensus/istanbul/backend/handler.go
mcortesi: If i'm not mistaken this method is not called for EVERY block. Depending on how blockchain blocks are inserted, it's batched. Which means we might "miss" en epoch block...

Can we check that? i'm worried that we untapped a problem here

consensus/istanbul/backend/engine.go
```diff
@@ -560,10 +561,53 @@ func (sb *Backend) SetChain(chain consensus.ChainReader, currentBlock func() *ty
...
+
+	for {
+		select {
+		case chainEvent := <-chainEventCh:
+			if !sb.coreStarted {
+				n := new(big.Int).Add(chainEvent.Block.Number(), common.Big1)
+				sb.replicaState.NewChainHead(n)
+			}
+		case err := <-chainEventSub.Err():
+			log.Error("Error in istanbul's subscription to the blockchain's chain event", "err", err)
```
mcortesi, line 603: should we panic here?

it seems that after this happens replican won't ever become primary

consensus/istanbul/backend/engine.go
```diff
@@ -560,10 +561,53 @@ func (sb *Backend) SetChain(chain consensus.ChainReader, currentBlock func() *ty
...
+	chainHeadCh := make(chan ethCore.ChainHeadEvent, 10)
+	chainHeadSub := bc.SubscribeChainHeadEvent(chainHeadCh)
+	defer chainHeadSub.Unsubscribe()
+
+	for {
+		select {
+		case chainHeadEvent := <-chainHeadCh:
+			sb.newChainHead(chainHeadEvent.Block)
+		case err := <-chainHeadSub.Err():
+			log.Error("Error in istanbul's subscription to the blockchain's chainhead event", "err", err)
```
mcortesi, line 583: same here... newChainHead() doesn't really communicate much... metrics it's ok to skip, but current valSet looks crucial...

if we are going to miss everything we should "panic" (cleanly)

consensus/istanbul/backend/engine.go
```diff
@@ -912,6 +981,20 @@ func (sb *Backend) addParentSeal(chain consensus.ChainReader, header *types.Head
 	return writeAggregatedSeal(header, createParentSeal(), true)
 }
 
+func (sb *Backend) SetStartValidatingBlock(blockNumber *big.Int) error {
```
mcortesi, line 984: check lint errors... needs a comment

consensus/istanbul/backend/engine.go
```diff
@@ -912,6 +981,20 @@ func (sb *Backend) addParentSeal(chain consensus.ChainReader, header *types.Head
...
+	if blockNumber.Cmp(sb.currentBlock().Number()) < 0 {
+		return errors.New("blockNumber should be greater than the current block number")
+	}
+	return sb.replicaState.SetStartValidatingBlock(blockNumber)
+}
+
+func (sb *Backend) SetStopValidatingBlock(blockNumber *big.Int) error {
```
mcortesi, line 991: same

consensus/istanbul/backend/engine.go
```diff
@@ -560,10 +561,53 @@ func (sb *Backend) SetChain(chain consensus.ChainReader, currentBlock func() *ty
...
+		case chainHeadEvent := <-chainHeadCh:
+			sb.newChainHead(chainHeadEvent.Block)
+		case err := <-chainHeadSub.Err():
+			log.Error("Error in istanbul's subscription to the blockchain's chainhead event", "err", err)
+			return
+		}
+	}
+}
+
+func (sb *Backend) updateReplicaStateLoop(bc *ethCore.BlockChain) {
```
mcortesi, line 589: shouldn't this loop run only when we are on replica mode?

consensus/istanbul/backend/internal/replica/state.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
+	lvlerrors "github.com/syndtr/goleveldb/leveldb/errors"
+)
+
+type state uint64
+
+// Different possible states for validators wrt replica/primary
+const (
+	primaryPermanent state = iota
```
mcortesi, line 35: add comment to explain what they mean (the permanent/in range thing)

consensus/istanbul/backend/internal/replica/state.go
```diff
@@ -0,0 +1,342 @@
...
+	rs.stopValidatingBlock = nil
+	rs.state = replicaPermanent
+	return nil
+}
+
+// MakePrimary makes this node a primary & clears all start/stop blocks.
+func (rs *replicaStateImpl) MakePrimary() error {
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	defer rs.rsdb.StoreReplicaState(rs)
```
mcortesi, line 218: what happens if this return error?

mcortesi: even worse.... what happens if `startFn()` works but `StoreReplicaState` fails

consensus/istanbul/backend/internal/replica/state.go
```diff
@@ -0,0 +1,342 @@
...
+	rs.stopValidatingBlock = blockNumber
+
+	return nil
+}
+
+// MakeReplica makes this node a replica & clears all start/stop blocks.
+func (rs *replicaStateImpl) MakeReplica() error {
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	defer rs.rsdb.StoreReplicaState(rs)
```
mcortesi, line 201: same

consensus/istanbul/backend/internal/replica/state.go
```diff
@@ -0,0 +1,342 @@
...
+		return errors.New("Start block number should be less than the stop block number")
+	}
+
+	switch rs.state {
+	case replicaPermanent:
+		rs.state = replicaWaiting
+	case replicaWaiting:
+		// pass. Changed start block while waiting to start.
+	default:
+		// todo error
```
mcortesi, line 161: todo

consensus/istanbul/backend/internal/replica/state.go
```diff
@@ -0,0 +1,342 @@
...
+		return errors.New("Stop block number should be greater than the start block number")
+	}
+
+	switch rs.state {
+	case primaryPermanent:
+		rs.state = primaryInRange
+	case primaryInRange:
+		// pass. Changes stop block while waiting to stop.
+	case replicaPermanent:
+		// TODO: don't allow this / error
```
mcortesi, line 187: todo

consensus/istanbul/backend/backend.go
```diff
@@ -939,3 +947,16 @@ func (sb *Backend) VerifyValidatorConnectionSetSignature(data []byte, sig []byte
 		return istanbul.CheckValidatorSignature(validator.NewSet(validators), data, sig)
 	}
 }
+
+func (sb *Backend) IsPrimaryForSeq(seq *big.Int) bool {
```
mcortesi, line 951: suggest we move all functions that are mostly proxies to replicastate togheter (can be on this file o engine... not clear)

consensus/istanbul/backend/engine.go
```diff
@@ -560,10 +561,53 @@ func (sb *Backend) SetChain(chain consensus.ChainReader, currentBlock func() *ty
...
+	chainEventCh := make(chan ethCore.ChainEvent, 10)
+	chainEventSub := bc.SubscribeChainEvent(chainEventCh)
+	defer chainEventSub.Unsubscribe()
+
+	for {
+		select {
+		case chainEvent := <-chainEventCh:
+			if !sb.coreStarted {
+				n := new(big.Int).Add(chainEvent.Block.Number(), common.Big1)
+				sb.replicaState.NewChainHead(n)
```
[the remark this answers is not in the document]
mcortesi: more importantly.. newChainHead return an error which can indicate the failure to stop or start core or db save...
which we dont' handle...

consensus/istanbul/core/core.go
```diff
@@ -511,6 +513,19 @@ func (c *core) startNewRound(round *big.Int) error {
...
 
+	// Inform the backend that a new sequence has started & bail if the backed stopped the core
+	if !roundChange {
+		if primary := c.backend.IsPrimaryForSeq(newView.Sequence); !primary {
+			// We need to run UpdateReplicaState outside of the core b/c when it stops the core
+			// it runs c.handlerWg.Wait(). To empty that wait group, we need to return from this
+			// function. We unsubscribe from events to stop the core from processing more events
+			// prior to being fully shut down.
+			c.unsubscribeEvents()
+			go c.backend.UpdateReplicaState(newView.Sequence)
```
mcortesi, line 524: this calls newchainhead which returns an error...
nor updatereplicastate neither the go routine is checking for that error 

doing `go callSomeFuncThaReturnError()` is actually a smell that it's over all our code. We are silencing errors :(

consensus/istanbul/backend/engine.go
trianglesphere: One thing about a panic here, is that I think that this is the normal shut down path, though it might make sense to both add a close channel and understand what occurs with this error.

I don't think I've ran into this causing issues, so I'm less worried about this.

consensus/istanbul/backend/engine.go
trianglesphere: via `!coreStarted` it only sends events if in replica mode. We could start/stop this, but it's another thing to keep track of.

consensus/istanbul/backend/internal/replica/state.go
trianglesphere: Probably want to revert the change if the save fails. It would be nice to be able to atomically do all of these state changes at once (change replica state & write to disk), but I'm not sure it's worth the added complexity.

consensus/istanbul/backend/engine.go
```diff
@@ -560,10 +561,53 @@ func (sb *Backend) SetChain(chain consensus.ChainReader, currentBlock func() *ty
...
+	if bc, ok := chain.(*ethCore.BlockChain); ok {
+		go sb.newChainHeadLoop(bc)
+		go sb.updateReplicaStateLoop(bc)
+	}
+
 }
 
-// StartValidating implements consensus.Istanbul.StartValidating
-func (sb *Backend) StartValidating(hasBadBlock func(common.Hash) bool,
+func (sb *Backend) newChainHeadLoop(bc *ethCore.BlockChain) {
```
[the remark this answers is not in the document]
trianglesphere: I'll add the comment that it can be batched. The couple tests I did showed that I think it's much closer to real time when actively validating, and ran more into issues when syncing as a replica (problems with start, not with stop).

consensus/istanbul/backend/engine.go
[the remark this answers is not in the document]
trianglesphere: I need to think a bit more about what to do w/ errors in NewChainHead, but I'm leaning towards not even returning an error from there and trying to deal with it in new chain head. The blockchain event sender really isn't a better place to handle an error.

mcortesi (approved):
Great job joshua! Thanks for addressing over and over my comments.

Final thing, can you create issues for the few comments we decided to do on another pr (catching errors, verify batched chain head are not a problem, and maybe there are others)

AGx10k:
@trianglesphere is there a way to check that a validator `has the enode urls for the rest of the validators` ?

PR #1150, opened 2020-09-01
Add metrics: blockFinalizationTime

YoduYodu:
### Description

Add metrics blockFinalizationTime, which is the duration from the time the block being created, to its finalization.

### Tested

N/A

### Backwards compatibility

Yes


jcortejoso (approved):
lgtm

miner/worker.go
```diff
@@ -187,30 +188,34 @@ type worker struct {
 
 	// Needed for randomness
 	db *ethdb.Database
+
+	// Gauge counting the block finalization time(from created to finalized)
+	blockFinalizationTime metrics.Gauge
```
kevjue, line 193: Instead of making this a field of `worker`, can it be made as a package level variable, like this:  https://github.com/celo-org/celo-blockchain/blob/master/core/blockchain.go#L52

kevjue, line 193: How come the metrics is a `gauge` type instead of a `timer` type?

kevjue: FYI, I personally don't really know much about the two, so the question is more for my understanding.

YoduYodu: Agree, that makes more sense

miner/worker.go
```diff
@@ -84,6 +84,9 @@ const (
 var (
 	randomSeedString = []byte("Randomness seed string")
 	randomSeed       []byte
+
+	// Gauge counting the block finalization time(from created to finalized)
```
kevjue, line 88: The comment needs to be fixed.

kevjue:
Could you post a screenshot of what this metric would look like in grafana.

miner/worker.go
```diff
@@ -588,6 +592,7 @@ func (w *worker) resultLoop() {
 				logs = append(logs, receipt.Logs...)
 			}
 			// Commit block and state to database.
+			blockFinalizationTimer.UpdateSince(time.Unix(int64(block.Time()), 0))
```
kevjue, line 595: This should be updated after `WriteBlockWithState`.

PR #1151, opened 2020-09-01
Lightest sync test

hbandura:
Adds a sync test to mainnet in the lightest mode to the circleci config

### DIFF
--- a/scripts/sync_test.sh
+++ b/scripts/sync_test.sh
@@ -24,7 +24,7 @@ ATTEMPTS=5
 RETRY_SLEEP=2
 # We attempt to check it several times since, sometimes the command
 # fails with "No peers available"
-for ATTEMPT in 1 .. $ATTEMPTS; do
+for ATTEMPT in $(seq 1 $ATTEMPTS); do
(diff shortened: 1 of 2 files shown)


scripts/sync_test.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Now check what the latest block is
+ATTEMPTS=5
+RETRY_SLEEP=2
+# We attempt to check it several times since, sometimes the command
+# fails with "No peers available"
+for ATTEMPT in $(seq 1 $ATTEMPTS); do
+	echo "Attempt $ATTEMPT/$ATTEMPTS of getting the latest block timestamp" 
+	MARK=`date +%s`
+	LATEST=`build/bin/geth --datadir $DATADIR --verbosity 0 console --syncmode $MODE --exec 'eth.getBlock("latest").timestamp'`
```
mcortesi, line 30: no need to do use `--syncmode` here

hbandura: If I don't use it it reports a different block. It seems it reports the latest block downloaded in the database for the specified sync mode

PR #1154, opened 2020-09-03
Add metric: blockFinalizationTime

YoduYodu:
### Description
Add metrics blockFinalizationTime, which is the duration from the time the block being created, to it's being written to chain.

**Waiting to be merged until load-test can tell the blockFinalizationTime optimization between `master` and [this PR](https://github.com/celo-org/celo-blockchain/pull/1123)**


kevjue (approved):

miner/worker.go
```diff
@@ -83,6 +84,9 @@ const (
 var (
 	randomSeedString = []byte("Randomness seed string")
 	randomSeed       []byte
+
+	// Timer used to measure block finalization time from created to finalized
+	blockFinalizationTimer = metrics.NewRegisteredTimer("miner/block/finalization", nil)
```
kevjue, line 89: Looks like this needs to be changed to type gauge (e.g. https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/backend.go#L123).

kevjue (changes requested):

miner/worker.go
```diff
@@ -593,6 +597,7 @@ func (w *worker) resultLoop() {
 				log.Error("Failed writing block to chain", "err", err)
 				continue
 			}
+			blockFinalizationTimeGauge.Update(time.Now().UnixNano() - int64(block.Time())*1000000000)
```
kevjue, line 600: How come this is using nanoseconds scale?

YoduYodu: Two factors:
• block.Time() will return seconds.
• time.Now() will return seconds or nanoseconds.

For accuracy, I chose to use nanoseconds.

kevjue: :+1:

kevjue (approved):

PR #1158, opened 2020-09-11
Add information form epoch header chain in epoch SNARK messages

nategraf:
### Description

In the current implementation of the epoch SNARK messages, which are the data used to create proofs for the SNARKs within the [Plumo](https://docs.zkproof.org/pages/standards/accepted-workshop3/proposal-plumo_celolightclient.pdf) system, it is possible to predict far in advance what these messages will be. As a result, there exists a scenario in which, if an attacker is able to compromise a number of validators, and those validators form a quorum at any future block, the attacker may create a valid proof branching the chain to a validator set of their choice. Key to this scenario is that the attacker need not have control of the validators (or more specifically their signing oracles) at the point they branch from, and instead can use the predictability of the messages to sign constructed epoch messages for blocks far in the future. As a result, the attack is feasible for an attacker who had access to the each validator in the quorum at any point in the past.

In order to address this scenario, this PR adds the epoch block hash, and previous epoch block hash to the epoch SNARK data. Because the block hash summarizes the state root, and therefore all on-chain state including the [on-chain randomness values](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/identity/Random.sol), the block hash is considered unpredictable*. This unpredictability makes the the generation of a epoch SNARK messages  to fork from a future block infeasible.

* Note that this value is not pseudo-random because anyone with the ability to, for example, submit a transaction can bias the block hash value.

### Tested

WIP

### Backwards compatibility

Enacting this change will require a hard-fork in the validator consensus algorithm. (WIP)


consensus/istanbul/backend/backend.go
```diff
@@ -646,6 +647,15 @@ func (sb *Backend) AuthorForBlock(number uint64) common.Address {
 	return common.ZeroAddress
 }
 
+// HashForBlock returns the block hash from the canonical chain for the given number.
+// Implements istanbul.Backend.HashForBlock
+func (sb *Backend) HashForBlock(number uint64) common.Hash {
```
kobigurk, line 652: How about returning an error here if we can't find the hash? The calling method `generateEpochValidatorSetData` can return an error.

kobigurk (commented):
Concept looks great! Had some small comments and questions.

Just to clarify - the changes in the consumed libs aren't there yet, right?

consensus/istanbul/core/commit.go
```diff
@@ -40,16 +40,32 @@ func (c *core) generateCommittedSeal(sub *istanbul.Subject) (blscrypto.Serialize
...
-func (c *core) generateEpochValidatorSetData(blockNumber uint64, newValSet istanbul.ValidatorSet) ([]byte, error) {
+// Generates serialized epoch data for use in the Plumo SNARK circuit.
+// Block number and hash may be information for a pending block.
+func (c *core) generateEpochValidatorSetData(blockNumber uint64, blockHash common.Hash, newValSet istanbul.ValidatorSet) ([]byte, error) {
 	if !istanbul.IsLastBlockOfEpoch(blockNumber, c.config.Epoch) {
 		return nil, errNotLastBlockInEpoch
 	}
+
+	// Retreive the block hash for the last block of the previous epoch, and truncate each hash to 128 bits.
+	parentEpochBlockHash := c.backend.HashForBlock(blockNumber - c.config.Epoch)
```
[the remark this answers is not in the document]
nategraf: Epoch length is not currently governable. (It would be very interesting if it was.)

nategraf:
@kobigurk Question related to this PR. It seems that care is taken to make the epoch index `uint16` instead of `uint32` or `uint64`. Should `maxNonSigners` also be compressed to `uint16`?

consensus/istanbul/backend/backend.go
nategraf: Because the hash can never be zero for a known block, and we do not get an error back from the blockchain query, I've changed it to use the zero value as an error indicator at the [call site](https://github.com/celo-org/celo-blockchain/blob/c4b5fcae2e427d16b3614bd4d837f4b0af141f3c/consensus/istanbul/core/commit.go#L51-L55).

nategraf:
Here are the changes to `celo-bls-go` so far. I also did some work on the Rust portion, but it's not particularly far yet https://github.com/celo-org/celo-bls-go/pull/14

consensus/istanbul/core/commit.go
```diff
@@ -40,21 +41,41 @@ func (c *core) generateCommittedSeal(sub *istanbul.Subject) (blscrypto.Serialize
...
 	for _, v := range newValSet.List() {
 		blsPubKeys = append(blsPubKeys, v.BLSPublicKey())
 	}
+
 	maxNonSigners := uint32(newValSet.Size() - newValSet.MinQuorumSize())
-	epochData, err := blscrypto.EncodeEpochSnarkData(blsPubKeys, maxNonSigners, uint16(istanbul.GetEpochNumber(blockNumber, c.config.Epoch)))
-	if err != nil {
-		return nil, err
+
+	// TODO(victor): Replace false with a check that the hard fork is inactive.
```
trianglesphere, line 59: Actually need to do this once we have a HF mechanism defined. Otherwise as long as the randomness isn't changing mid sequence (which I don't think it is), it's good from a consensus pov.

nategraf: That's my intention. I figure we should get the code checked in now because it seems unclear when the hard-fork solution will be ready. And from the randomness perspective, this is operating on the proposal hash, which means the block is finalized, so unless I misunderstand something, it is guaranteed not to change.

trianglesphere: One thing if we check this in now, shouldn't it be `true` b/c we don't want to yet activate this hard fork.

nategraf: Oh 😅 Yes!

kobigurk:
Superseded by #1231

PR #1160, opened 2020-09-15
cache istanbul consensus message signature verification

kevjue:
### Description

This PR will cache the message signature verification for consensus istanbul messages.  This is to help with performance of the round change cert verification, which will contain a lot of duplicated messages. 

### Tested

Ran unit tests

### Backwards compatibility

Yes


consensus/istanbul/core/core.go
```diff
@@ -679,7 +684,36 @@ func (c *core) resendRoundChangeMessage() {
 }
 
 func (c *core) checkValidatorSignature(data []byte, sig []byte) (common.Address, error) {
-	return istanbul.CheckValidatorSignature(c.current.ValidatorSet(), data, sig)
+	logger := c.newLogger("func", "checkValidatorSignature")
+
+	var signerAddress common.Address
+	cacheKey := msgSigCacheKey{msgData: data, msgSig: sig}
+	cacheKeyBytes, err := rlp.EncodeToBytes(cacheKey)
```
trianglesphere, line 691: Nit: I'm a little worried about the overhead of this, but I think if there's an improvement in perf it's because `checkValidatorSignature` is used also is in `handleMsg` (via `c.validateFn`)  which checks the signature for all backlog events (& nodes are pretty aggressive about backlogging messages even in the current seq & round).

That to me seems like it occurs much more often in the happy case than round changes (which this should also help with).

trianglesphere (approved):
Implementation looks good. One comment on the larger than realized usage of this cache.

One thing that would be nice is some sort of metrics / perf data showing that this change actually improves performance.

consensus/istanbul/core/core.go
```diff
@@ -573,6 +577,7 @@ func (c *core) resetRoundState(view *istanbul.View, validatorSet istanbul.Valida
 		headBlock := c.backend.GetCurrentHeadBlock()
 		newParentCommits = newMessageSet(c.backend.ParentBlockValidators(headBlock))
 	}
+	c.msgSigCache = make(map[string]common.Address)
```
trianglesphere, line 580: QQ: Given that proxies don't mine, are we properly clearing the message cache in proxies?

kevjue: @trianglesphere - IIUC your question correctly, the proxies don't start the Istanbul core engine, and shouldn't be using a message sig cache at all.

trianglesphere: Looking through the code I think proxies do call `checkValidatorSignature` in https://github.com/celo-org/celo-blockchain/blob/0e08809b755aae0afb41e131a3e5e33816108597/consensus/istanbul/backend/handler.go#L101 I believe that proxies then use the cache.

kevjue: Good point.  The proxies are using a different `checkValidatorSignature` implementation:  https://github.com/celo-org/celo-blockchain/blob/0e08809b755aae0afb41e131a3e5e33816108597/consensus/istanbul/utils.go#L54, so the proxies are never using a cache.

There is some duplicated code as a result, since this PR does copy some of the logic in `istanbul.CheckValidatorSignature` into `core.checkValidatorSignature`.   Seems like the best way to resolve that is to modify `istanbul.CheckValidatorSignature` so that it accepts a signature cache as a parameter, and it will be non-nil when called via istanbul core, but nil when called by the proxy.

trianglesphere: makes sense. I saw the upper case `CheckValidatorSignature` but didn't realize it was being used in the proxy's `checkValidatorSignature`

PR #1162, opened 2020-09-21
Splits generic db from enodes packages

trianglesphere:
### Description

Splits generic level db wrapper into it's own package (`consensus/istanbul/backend/internal/db`)
This enables other packages to create leveldbs outside of the enode package.

### Other changes

### Tested

Unit tests.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes.


mcortesi (approved):
Maybe it's `db.New` instead of `db.NewGenericDb`. 

Also, db abstraction might be helpful outside istanbul. There's a ethdb.Database abstraction which defines a couple of interface, but not what we need, still can be an inspiration.

PR #1164, opened 2020-09-21
chore: parameter constant expression cleanup

prestwich:
### Description

Improved legibility of Celo-specific constants. In order to avoid breaking compatiblity with upstream, non-Celo-specific params have been left as-is

### Tested

Local tests. Would appreciate more manual inspection to be sure all constants are unchanged.


mcortesi (approved):
nice :)

PR #1026, opened 2020-05-04
Multiproxy

kevjue:
### Description

This PR contains the initial version of multi proxy support as well as creating a `proxy` golang subpackage for the proxy logic within the `istanbul` package.  This PR is the culmination of joint work between @kevjue and @tkporter.

#### Usage

 The command line option `proxy.proxyenodeurlpair` will be replaced with `proxy.proxyenodeurlpairs`.  The updated option will expect a comma separated list of proxy enodeURL pairs where the pair consist of the proxy's internal facing enode URL and it's external facing enode URL in that order.

E.g. 
`--proxy.proxyenodeurlpairs="enode://50e7996cd17b5ace2c8a3f5f21e4c5d564b4e45991e98b3f6d73bf5f049076db511ca20dc4acc71c9366c50f0d2d9120d4b508a1ffdd38128febb38046f0acf5@127.0.0.1:50305;enode://50e7996cd17b5ace2c8a3f5f21e4c5d564b4e45991e98b3f6d73bf5f049076db511ca20dc4acc71c9366c50f0d2d9120d4b508a1ffdd38128febb38046f0acf5@127.0.0.1:30305,enode://e85dd29368d47790013e8573102feade11b08da1ef2d6122096114c873da1179c8d571a7f3e044c0e06f77de2e4bb51a16c6a14ff4674cbbc614ce35f1aaa860@127.0.0.1:50307;enode://e85dd29368d47790013e8573102feade11b08da1ef2d6122096114c873da1179c8d571a7f3e044c0e06f77de2e4bb51a16c6a14ff4674cbbc614ce35f1aaa860@127.0.0.1:30307"
`

Proxies can also be added, removed, and listed via the RPC apis:

- `istanbul.addProxy(<internal enode>, <external enode>)`
- `istanbul.removeProxy(<internal enode>)`
- `istanbul.proxies` to view info on the current proxies, like its internal/external enodes, validators it's associated with, etc
- `admin.proxiedValidators` on a proxy to view info on the peers in the proxy p2p server (ie to ensure the proxy is peered with the proxied validator)

#### Multi proxy behavior

Each proxy will be responsible to establishing a direct connection with a set of remote validators.  The assignment of the remote validator set to each proxy is determined using a consistent hashing algorithm (https://en.wikipedia.org/wiki/Consistent_hashing) where the keys can be thought of as the keys and the proxies the slots.

[...]

PR #1166, opened 2020-09-23
Proxy Celostats fixes

gastonponti:
### Description

The proxy sends stats to celostats, previously signed by the validator

[WIP] Proxies stop sending data after some time

### Other changes

The validator node it doesn't know anymore which proxy node is the one that sends stats
(This allow us to remove a stats-proxy, and add another one with that configuration)

Adds a refactor done in the PR #1033 

### Tested

Manually

### Related issues

- Part of #1026 

### Backwards compatibility

Against the Multiproxy branch, it removes a flag.


cmd/utils/flags.go
```diff
@@ -780,11 +780,6 @@ var (
 		Usage: "Each enode URL in a pair is separated by a semicolon. Enode URL pairs are separated by a space. The format should be \"<proxy 0 internal facing enode URL>;<proxy 0 external facing enode URL>,<proxy 1 internal facing enode URL>;<proxy 1 external facing enode URL>,...\" (deprecated, use --proxy.proxyenodeurlpairs)",
 	}
 
-	ProxyCeloStatsFlag = cli.StringFlag{
```
trianglesphere, line 783: We should probably note that this flag is deprecated and doesn't do anything rather than immediately removing it as passing in undefined flags can cause problems.

trianglesphere (commented):
Looks good so far. One comment on flags. Also gofmt everything.

ethstats/ethstats.go
```diff
@@ -96,52 +97,68 @@ type blockChain interface {
...
 	if ethServ != nil {
 		engine = ethServ.Engine()
 	} else {
 		engine = lesServ.Engine()
 	}
 
 	backend := engine.(*istanbulBackend.Backend)
+	if !backend.IsProxiedValidator() {
+		// Parse the netstats connection url
+		re := regexp.MustCompile("([^:@]*)?@(.+)")
```
mcortesi, line 144: good opportunity to extract function refactor :)

ethstats/ethstats.go
```diff
@@ -546,7 +571,10 @@ func (s *Service) readLoop(conn *websocket.Conn) {
 			log.Info("Ping from websocket server", "msg", msg)
 			// Primus server might want to have a pong or it closes the connection
 			var serverTime = fmt.Sprintf("primus::pong::%d", time.Now().UnixNano()/int64(time.Millisecond))
-			conn.WriteJSON(serverTime)
+			err := conn.WriteMessage(websocket.TextMessage, []byte(serverTime))
```
mcortesi, line 574: why this change?

ethstats/ethstats.go
```diff
@@ -1035,8 +1061,8 @@ func (s *Service) reportStats(conn *websocket.Conn) error {
 	log.Trace("Sending node details to ethstats")
 
 	stats := map[string]interface{}{
-		"id":      s.node,
-		"address": etherBase,
+		"id":      s.backend.ValidatorAddress().String(),
+		"address": validatorAddress,
```
mcortesi, line 1065: if eth == nil validatorAddress is empty. Can s.eth be nil? shouldn't that imply we need to skip or something?

ethstats/ethstats.go
```diff
@@ -1035,8 +1061,8 @@ func (s *Service) reportStats(conn *websocket.Conn) error {
 	log.Trace("Sending node details to ethstats")
 
 	stats := map[string]interface{}{
-		"id":      s.node,
-		"address": etherBase,
+		"id":      s.backend.ValidatorAddress().String(),
```
mcortesi, line 1064: if last comment is solved, this could be `validatorAddress.String()`

ethstats/ethstats.go
```diff
@@ -444,7 +456,20 @@ func (s *Service) login(conn *websocket.Conn, sendCh chan *StatsPayload) error {
 	return nil
 }
 
-func (s *Service) handleDelegateSign(messageToSign *StatsPayload) error {
+func (s *Service) waitAndDelegateMessageWithTimeout(conn *websocket.Conn, sendCh chan *StatsPayload) error {
+	select {
+	case signedMessage := <-sendCh:
+		err := s.handleDelegateSend(conn, signedMessage)
```
mcortesi, line 462: can replace 
```go
		err := s.handleDelegateSend(conn, signedMessage)
		if err != nil {
			return err
    }
```
with:
```go
		return s.handleDelegateSend(conn, signedMessage)
```
and remove the `return nil` at the end

ethstats/ethstats.go
```diff
@@ -238,26 +249,33 @@ func (s *Service) loop() {
 			case <-headSub.Err():
 				break HandleLoop
 			case delegateSignMsg := <-istDelegateSignCh:
```
mcortesi, line 251: one nice (and small) refactor is to treat istDelegateCh as a "router". IMO on line 206 where we subscribe, we could just launch a go routiner like:
```go
go func() {
  ch := make(chan istanbul.MessageWithPeerIDEvent, istDelegateSignChanSize)
	subscription := s.backend.SubscribeNewDelegateSignEvent(ch)
  defer subscription.Unsubscribe()
  
  for {
    case msg := <- ch:
      // route msg either to signCh or sendCh
    case err := <- subscription.Err() // Handle subscrition errors that we are currently not doing!
    case <- quit // need to verify this, but some signal to stop (using go context module can work too)
  }
}
```

oneeman:
With these changes, I think the following fields (and all references to them) can be removed:
* `ProxyConfig.StatsHandler`
* `Proxy.statsHandler`
* `ProxyInfo.StatsHandler`

I believe the only one of them that's exposed externally somewhat is `ProxyInfo`, by the `GetProxiesInfo` method, but that method is new from the multiproxy PR, so we wouldn't be breaking anybody's expectations by removing `StatsHandler` from `ProxyInfo`.

cmd/utils/flags.go
gastonponti: Yes, I thought about that, but my concern is that the flag only lives in the multi-proxy branch. So technically was not released to everyone.
I don't want to add something that will start as deprecated. @kevjue what do you think about this?

ethstats/ethstats.go
gastonponti: This was a change of the #1033. Thought that maybe was important, and I was not seeing the why

cmd/utils/flags.go
trianglesphere: If it was just from multi proxy, I definitely agree that we should remove it.

mcortesi (approved):

ethstats/ethstats.go
```diff
@@ -188,166 +207,116 @@ func (s *Service) loop() {
...
-				if time.Duration(mclock.Now()-lastTx) < time.Second {
-					continue
+			}
+		})
+	} else {
+		group.Go(func() error {
+			// Loop reporting until termination
+			for {
+				// Resolve the URL, defaulting to TLS, but falling back to none too
+				path := fmt.Sprintf("%s/api", s.celostatsHost)
```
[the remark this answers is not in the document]
gastonponti: mmmm It's already out... Maybe an outdated file from github?

PR #1192, opened 2020-10-21
Cherry pick new dial scheduler from upstream

### LINKED ISSUE #1180: Validators/proxies initiate new dyndial connections even when they're at the max peer limit
### Expected Behavior

When a node has reached its max peers limit, it should not initiate any new dyndial connections, because those will immediately get closed due to the peer count check.

### Current Behavior

Due to a bug in the dialer, nodes with a lot of static connections (e.g. validator proxies or unproxied validators) do initiate dyndial connections even if they're at the max peers limit.  This shows in the logs as new connections constantly being established, often with the same peers within relatively short time periods.  This is likely the main thing that in practice leads to the race condition behind https://github.com/celo-org/celo-blockchain/issues/1172. 

The bug is that the value the dialer gets for `maxDynDials` really represents what should be the max of all dialed connections (see https://github.com/celo-org/celo-blockchain/blob/fd05c6b57ca09e94462723c57756a6d0a12ab278/p2p/server.go#L577-L578), but when it does the check to see if there is more room (and so whether new dyndial connections should be started) it only counts dyndial connections (see https://github.com/celo-org/celo-blockchain/blob/fd05c6b57ca09e94462723c57756a6d0a12ab278/p2p/dial.go#L154-L165), so the calculation is off.

This is an upstream bug that has been fixed by ethereum/go-ethereum#20592, and so it should be fixed as part of the merging for https://github.com/celo-org/celo-blockchain/issues/1179

oneeman:
### Description

Cherry pick the new dial scheduler from upstream.  Also cherry-picks https://github.com/ethereum/go-ethereum/pull/20688 because it is a closely-related follow-up to the dial scheduler.

Marked with "do not merge" until I add another commit, as explained below.

Merge notes:

- Upstream made static dials respect the max dialed connections limit.  This means that if there are a lot of dialed connections (static or dynamic), we are no longer guaranteed that it would try to connect to all the nodes with explicit static purpose set.  This is not desirable for us, since we rely on such a guarantee for the connections between validators/proxies and other validators/proxies.  In other words, for us static dial connections being exempt from the max dialed connection limit is more of a feature than a bug.  I will add a commit to change that behavior back (while still fixing #1180), so I am adding the "do not merge" label to this PR until then.
- Upstream removes the static map from server.go. Since we use it for purposes rather than just boolean, decided to keep it there rather than push the purposes map down into dial.go
- Upstream made srv.RemovePeer() block until the peer is completely disconnected, using doPeerOp().  Since we need to access static in order to check and modify the node's purposes and decide whether to disconnect it or not, we can't do with doPeerOp and instead have to do it in the main thread.  So I added a channel called `removestaticDone` to let us know when the peer has been disconnected fully (using the same method that they used, but in a different place now)
- Upstream changed one of the conditions for the max peers check, so that instead of being bypassed if the peer is trusted, it's bypassed if the peer is trusted or it's a static dialed connection (i.e. it now allows non-trusted static dial connections, but not non-trusted incoming connections from nodes which happen to also be listed in static).  We do the checks elsewhere, and we check using info.Network.Trusted and info.Network.Static, so we're already doing it the way they changed it to be.

### Tested

- Unit tests passing
- e2e sync test passes
- Running a node to sync with mainnet, it connects to peers and starts syncing, whether in full mode or in lightest mode.

### Related issues

- Fixes #1180

### Backwards compatibility

This only modifies how the node starts new connections to other nodes, and doesn't affect the protocols used between nodes, so there aren't backwards compatibility issues.


oneeman:
Removing the "do not merge" tag, now that I've added the commit to exempt static dial from the limits.  Unfortunately it involved considerable changes to the unit tests, but at the same time that's a good thing in that it indicates the unit tests were  testing what they should be.

p2p/server.go
```diff
@@ -833,6 +851,10 @@ func (srv *Server) run(dialstate dialer) {
 				p.RemovePurpose(purpose)
 			}
 		}
+		if ch == nil {
```
mcortesi, line 854: I know all this changes come from master, but: 

nit: i think `ch == nil` is a convoluted way to asking: `HasNoPurpose && not in peers[id]`. I would try to make it more explicit of what `ch == nil` actually means (change it to something else).

In fact, i would make `ch` and `sub` local variables to the goroutine, and thus reduce their scope. No need to have them as global to the whole function.

Also, i now realize that when removing a static we subscribe to **all events from all peers**; probably there's not easier way of doing this, but seems like an overkill.

mcortesi (approved):

p2p/server.go
oneeman: Sounds good, I'll move them and add a boolean called `disconnecting` that can then be used in the `if`

As far as the subscription, it does look like there is no way to subscribe to only certain events, so there's not much else we could do.

Issue #1106, opened 2020-07-15
Diagnose and fix peer discovery failures


kevjue:
There has been reports of nodes not being able to discover peers on startup.

A validator has created an issue that describes his discover failure instance:  https://github.com/celo-org/celo-blockchain/issues/1101

Issue #1101, opened 2020-07-08
No peers on fullnodes & proxies randomly


daithi-coombes:
### Expected Behavior

Peer discovery would work and nodes would sync.

### Current Behavior

Randomly fullnodes and peers will not sync when created first. They have been left running for over 24hrs. Also manually injecting peers (`admin.addPeer`), will increasee the peer count but p2p discovery doesn't seem to be happening.

3 out of 4 times the nodes will sync.

It appears almost as if its doing a port scan of target peers:
```
$ docker start celo-proxy && docker logs -f --tail 100 celo-proxy |& grep "Accepted connection"

TRACE[07-08|15:47:27.152] Accepted connection                      addr=34.73.132.106:40908
TRACE[07-08|15:48:02.158] Accepted connection                      addr=34.73.132.106:40914
TRACE[07-08|15:48:37.163] Accepted connection                      addr=34.73.132.106:40918
TRACE[07-08|15:49:12.168] Accepted connection                      addr=34.73.132.106:40926
TRACE[07-08|15:49:47.175] Accepted connection                      addr=34.73.132.106:40928
TRACE[07-08|15:50:22.179] Accepted connection                      addr=34.73.132.106:40936
TRACE[07-08|15:50:57.184] Accepted connection                      addr=34.73.132.106:40940
TRACE[07-08|15:51:32.190] Accepted connection                      addr=34.73.132.106:40946
```
(all above connections were rejected)

### Steps to Reproduce Behavior

Unkown at the moment

### Logs

The following can be seen in the logs (taken from a peer node):
```
TRACE[07-08|15:41:54.945] New dial task                            task="dyndial 88dbf37a6be7f92f 50.17.60.161:30303"                                                                                                         
WARN [07-08|15:41:59.789] Stats login failed                       err="login timed out"                                                                                                                                      
TRACE[07-08|15:42:04.945] Skipping dial candidate                  id=88dbf37a6be7f92f addr=50.17.60.161:30303   err="already dialing"                                                                                        
TRACE[07-08|15:42:09.945] Dial error                               task="dyndial 88dbf37a6be7f92f 50.17.60.161:30303" err="dial tcp 50.17.60.161:30303: i/o timeout"                                                          
TRACE[07-08|15:42:09.945] Dial task done                           task="dyndial 88dbf37a6be7f92f 50.17.60.161:30303"                 
[...]

PR #1194, opened 2020-10-23
Oneeman/cherry pick discovery bootnode fix

### LINKED ISSUE #929: Slow discovery
The testnets can take up to 30 minutes for the validators to peer with each other

oneeman:
### Description

This is a fix for https://github.com/ethereum/go-ethereum/issues/20630.  If a node initially fails to contract the bootnode for the purpose of discovery, there were two code paths which led to the bootnode being removed from the discovery table, leaving it empty and preventing further discovery.  These are visible in the logs @tkporter included in that issue, where it says `Removed dead node` and where it says `Too many findnode failures, dropping`.  There is an upstream PR, cherry-picked here, which fixes the second case by not dropping the node if the bucket it's in is less than half-full.  For the second case, I added a comment on the go-ethereum issue about it to see whether a PR to fix it would be welcome and what logic it should use.  In the meantime, I've included a commit in the PR to fix it, by not dropping the node if it's the only one in the bucket.

Note that there is also a task called `doRefresh()` which runs every 30 minutes and re-adds the bootnodes to the discovery table, so the inability to discover other nodes would not have been permanent

### Other changes

I've also cherry-picked two prior commits from upstream to ease the merging by avoiding out-of-order commits on the same areas of the code.

### Tested

- Unit tests pass
- Monorepo e2e sync test passes
- Running a mainnet full node, it succeeds in discovering peers, connecting to them, and starting to sync
- Running two mainnet full nodes, A and B, where B is given A as its bootnode, and B is started a minute before A is started (i.e. the setup @tkporter reported as not working).  Until A is started, B does not discover any nodes.  Once A is started, B immediately discovers multiple nodes, connects to them, and starts syncing.

### Related issues

- Fixes https://github.com/ethereum/go-ethereum/issues/20630
- This is likely not what's behind https://github.com/celo-org/celo-blockchain/issues/1101, but may have been behind other reports of discovery failing.  There is a catch-all investigation issue for it: https://github.com/celo-org/celo-blockchain/issues/1106
- Fixes #929

### Backwards compatibility

The changes here are fixes and improvements to the discover package and do not include or assume protocol changes, so there shouldn't be any question of backwards incompatibility.  Also, as mentioned in the testing section, this branch successfully discovers and syncs from mainnet peers using the mainnet bootnode.


oneeman:
Seeing some lint failures: one about a file having to end with a newline, but others about unused private functions and a field in the code from upstream.  Added the "do not merge" tag for now

kevjue:
What's the cherry-picked commit from go-ethereum?

oneeman:
The one with the fix is https://github.com/ethereum/go-ethereum/pull/21396 .  The other two I included because they were previous changes on the same code: the first was https://github.com/ethereum/go-ethereum/pull/20750 and the second was https://github.com/ethereum/go-ethereum/pull/21147 .  You can see them above, with different hashes but the commit messages will be the same.

oneeman:
Took care of the lint warnings and removed the "do not merge" label.

oneeman:
Noticed an issue that sounds like it's due to the bugs fixed in this PR: #939.  So I updated the PR description to say that it fixes that issue as well.

mcortesi (approved):
Can't say i did a deep review. I don't have a deep knowledge of the discovery refactory, and i need to take a look at it first

Still i think this is good to merge.

My reasoning is:
 * cherry picked commits are most probably valid. They come from already production code that has been running for a while on ethereum. Also, we haven't changed discovery, it's not likely they could break any of our code
 * I've reviewed our commit on top, and seems reasonable, and quite a small case concerning a corner case.

So i'm happy with merging this.

PR #1195, opened 2020-10-26
CIP-28: Split etherbase into validator & txFeeRecipient addresses

### LINKED ISSUE #1135: Split Etherbase into GatewayFeeRecipient and Validator
### Expected Behavior

Separating etherbase flags should not leave the proposer in a situation where all of the proposed blocks may not be verified by peers. The proposer should check if the fork has occurred (minClientVersion) and then enforce `txFeeRecipient == Header.Coinbase` or otherwise.

### Current Behavior

When the fork has not occurred, the proposer is not enforcing that the `txFeeRecipient` and `coinbase` addresses are the same before proposing a block, so all blocks will not verify if these flags are set differently.

This fix should be minor, but needs sufficient testing before merging.

andres-dg:
### Description

Implementation for [CIP-28](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0028.md).

>We separate the coinbase address role into two roles: validator and txFeeRecipient. The validator address will receive validation rewards, while the txFeeRecipient address will receive transaction fees. The terms etherbase and coinbase become aliases for the txFeeRecipient address.

Both parameters are set via command line
New flags:
`miner.validator` is the validator address
`txFeeRecipient` the address where all the fees payments will be transferred
If one of these flags is present, the other will be required, even if it is the same address, to force a conscious decision about it 

To maintain backwards compatibility, the `etherbase` flag if is already set, and non of the other two flags were set, it will fill that address as the `miner.validator` and the `txFeeRecipient`. The flag will eventually be deprecated.

We changed the default behaviour of the etherbase flag. Now it won't assume the first account of the keyStore as default. So beware. Either the `etherbase` flag or the pair `miner.validator` and `txFeeRecipient` **MUST** be set.

### Other changes

Split etherbase in ethstats.

### Related issues

- Fixes #1135 

### Backwards compatibility

**Hardfork Change**. Modifies block validation logic. Part of `Donut` Hardfork


core/state_processor.go
```diff
@@ -67,7 +67,12 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	}
 
 	if random.IsRunning() {
-		err := random.RevealAndCommit(block.Randomness().Revealed, block.Randomness().Committed, header.Coinbase, header, statedb)
+		author, err := p.bc.Engine().Author(header)
```
mcortesi, line 70: just to point out here, that in the old code (which some nodes will be running) if the block's Coinbase != Author it will fail.

On the new version it doesn't matter since we use header author which makes the change retrocompatible. But we need to be very carefully that we don't generate blocks BEFORE the hardfork where Coinbase != Author. Since they are assumed to be the same PREFORK.

core/vm/context.go
```diff
@@ -75,13 +75,13 @@ type ChainContext interface {
...
 
 // NewEVMContext creates a new context for use in the EVM.
-func NewEVMContext(msg Message, header *types.Header, chain ChainContext, author *common.Address) Context {
-	// If we don't have an explicit author (i.e. not mining), extract from the header
+func NewEVMContext(msg Message, header *types.Header, chain ChainContext, txFeeRecipient *common.Address) Context {
+	// If we don't have an explicit txFeeRecipient (i.e. not mining), extract from the header
 	var beneficiary common.Address
-	if author == nil {
-		beneficiary, _ = chain.Engine().Author(header) // Ignore error, we're past header validation
+	if txFeeRecipient == nil {
```
mcortesi, line 81: this looks fishy... if the header has a coinbase (which might be set to validator address in PREFORK status) here we are not using that information and just using the configured txFeeRecipient. 

I would investigate this

mcortesi: I don't get why if the header has a coinbase we ignore it. @kevjue ?

consensus/istanbul/backend/engine.go
```diff
@@ -367,8 +367,10 @@ func (sb *Backend) VerifySeal(chain consensus.ChainReader, header *types.Header)
 // Prepare initializes the consensus fields of a block header according to the
 // rules of a particular engine. The changes are executed inline.
 func (sb *Backend) Prepare(chain consensus.ChainReader, header *types.Header) error {
-	// unused fields, force to set to empty
-	header.Coinbase = sb.address
+	// If proposer has not set the `Coinbase` via the `tx-fee-recipient` flag, default to the backend address.
+	if header.Coinbase == (common.Address{}) {
```
mcortesi, line 371: look the other comment

consensus/istanbul/backend/backend.go
```diff
@@ -432,6 +432,18 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		return errInvalidProposal
 	}
 
+	// If the current block occurred before the Celo1 hard fork, check that the author and coinbase are equal.
+	if !sb.chain.Config().IsCelo1(block.Number()) {
```
mcortesi, line 436: wrong place to add this check. We don't need to add a check but to prevent generating new blocks with Coinbase != Validator, since old versions assume they are the same.

mcortesi (changes requested):

cmd/utils/flags.go
```diff
@@ -227,9 +227,9 @@ var (
 		Name:  "whitelist",
 		Usage: "Comma separated block number-to-hash mappings to enforce (<number>=<hash>)",
 	}
-	EtherbaseFlag = cli.StringFlag{
-		Name:  "etherbase",
-		Usage: "Public address for transaction broadcasting and block mining rewards (default = first account)",
+	TxFeeRecipientFlag = cli.StringFlag{
```
mcortesi, line 230: @YazzyYaz @pranaymohan we are changing (and splitting )`--etherbase` to `--tx-fee-recipient` and `--miner.validator`

1. etherbase dissapears => change to the docs? maybe just leave it to return an erro msg?
2. i'm tempted to remove this: deafult = first account behaviour. Doesn't seem to be one of those things you want to "accidentally set". What do you think?

cmd/geth/main.go
```diff
@@ -104,7 +104,10 @@ var (
 		utils.ListenPortFlag,
 		utils.MaxPeersFlag,
 		utils.MaxPendingPeersFlag,
+		utils.MinerEtherbaseFlag,
```
mcortesi, line 107: I don't see where this and Legacy one are actually defined? missing in the PR?

cmd/utils/flags.go
pranaymohan: 1. Let's change in docs for sure. Not sure I understand what you mean by error message? Another option is we can have etherbase be syntatic sugar for "I want both tx-fee and miner-fee to go to the same address" -- this way it won't break anyone who doesn't care. Also why do we keep the prefix `Miner` -- should we refactor to something else at some point, to get away from connotations of PoW?
2. +1, especially since it may lead to funds being sent elsewhere, it should probably fail if not set properly.

eth/backend.go
```diff
@@ -374,27 +375,54 @@ func (s *Ethereum) ResetWithGenesisBlock(gb *types.Block) {
 	s.blockchain.ResetWithGenesisBlock(gb)
 }
 
-func (s *Ethereum) Etherbase() (eb common.Address, err error) {
+func (s *Ethereum) Validator() (val common.Address, err error) {
```
mcortesi, line 378: here we have the logic of default to first acount that we can remove

eth/backend.go
```diff
@@ -374,27 +375,54 @@ func (s *Ethereum) ResetWithGenesisBlock(gb *types.Block) {
...
-			return etherbase, nil
+			log.Info("Validator automatically configured", "address", validator)
+			return validator, nil
 		}
 	}
-	return common.Address{}, fmt.Errorf("etherbase must be explicitly specified")
+	return common.Address{}, fmt.Errorf("validator must be explicitly specified")
+}
+
+func (s *Ethereum) TxFeeRecipient() (common.Address, error) {
```
mcortesi, line 401: here we have the logic of default to first acount that we can remove

core/vm/context.go
kevjue: Ya, agreed that this looks fishy.  I haven't been following this issue.  What's the semantics of the "coinbase" field in the header after this change?  @mcortesi @andres-dg

mcortesi: coinbase is the address where gas fees are deposited; so that's been called "txFeeRecipient" and then we have another address for the header signer (the validator address)

Now, since this is a hardfork, we need to make sure that before activation the "coinbase is always the validator" and only in the case of hardfork activation they can be different

kevjue: OK.  To answer your original question, I suspect that txFeeRecipient is used ONLY when this function is called by the proposer when creating a block proposal, and for all other times txFeeRecipient is nil.  So that would explain why the coinbase field is ignored.  If that's true, then the code here is obviously not clear about that and seems like at the very least, a comment should be added.

cmd/utils/flags.go
```diff
@@ -1127,7 +1131,12 @@ func setTxFeeRecipient(ctx *cli.Context, ks *keystore.KeyStore, cfg *eth.Config)
 			Fatalf("`etherbase` and `tx-fee-recipient` flag should not be used together. `miner.validator` and `tx-fee-recipient` constitute both of `etherbase`' functions")
 		}
 		txFeeRecipient = ctx.GlobalString(TxFeeRecipientFlag.Name)
+
+		if !ctx.GlobalIsSet(MinerValidatorFlag.Name) {
+			Fatalf("`miner.validator` flag should be set if `tx-fee-recipient` is set")
```
mcortesi, line 1136: i don't agree.. if you are a full node, you might set the tx-fee-recipient for the gateway fee, and you're not a miner, so you don't set miner.validator, why is thi s needed?

miner/worker.go
```diff
@@ -819,13 +819,23 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
 		Extra:      w.extra,
 		Time:       uint64(timestamp),
 	}
+
+	txFeeRecipient := w.txFeeRecipient
+	if !w.chainConfig.IsCelo1(header.Number) {
```
mcortesi, line 824: nit:
```go
if !w.chainConfig.IsCelo1(header.Number) && w.txFeeRecipient != w.validator {
```

consensus/istanbul/backend/engine.go
mcortesi: let see if we can remove this, we are setting the Coinbase on the `newWork` method already

cmd/utils/flags.go
andres-dg: True, missed those cases

eth/gen_config.go
```diff
@@ -1,238 +0,0 @@
-// Code generated by github.com/fjl/gencodec. DO NOT EDIT.
```
mcortesi, line 1: why is this file deleted?

andres-dg: Since I changed eth/config.go, this generated file would be different. I think I should generate it locally with [this command](https://github.com/celo-org/celo-blockchain/blob/74d0104bc0d258130677b42e1e6dac0fca88eba0/eth/config.go#L55) and push it maybe.

cmd/utils/flags.go
```diff
@@ -1095,41 +1100,75 @@ func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error
...
-func setEtherbase(ctx *cli.Context, ks *keystore.KeyStore, cfg *eth.Config) {
-	// Extract the current etherbase, new flag overriding legacy one
-	var etherbase string
+// `Validator` is the address used to sign consensus messages.
+func setValidator(ctx *cli.Context, ks *keystore.KeyStore, cfg *eth.Config) {
+	// Extract the current validator, new flag overriding legacy etherbase
+	var validator string
 	if ctx.GlobalIsSet(MinerLegacyEtherbaseFlag.Name) {
-		etherbase = ctx.GlobalString(MinerLegacyEtherbaseFlag.Name)
+		validator = ctx.GlobalString(MinerLegacyEtherbaseFlag.Name)
```
mcortesi, line 1110: can we create an issue to remove these flags in the next version?

eth/backend.go
```diff
@@ -364,27 +366,30 @@ func (s *Ethereum) ResetWithGenesisBlock(gb *types.Block) {
...
-			return etherbase, nil
-		}
+	if txFeeRecipient != (common.Address{}) {
+		return txFeeRecipient, nil
 	}
-	return common.Address{}, fmt.Errorf("etherbase must be explicitly specified")
+	return common.Address{}, fmt.Errorf("txFeeRecipient must be explicitly specified")
+}
+
+func (s *Ethereum) Etherbase() (common.Address, error) {
```
mcortesi, line 391: can we remove this?

miner/worker.go
```diff
@@ -456,7 +464,11 @@ func (w *worker) mainLoop() {
 					continue
 				}
 				w.mu.RLock()
-				coinbase := w.coinbase
+				txFeeRecipient := w.txFeeRecipient
+				if !w.chainConfig.IsDonut(w.current.header.Number) && w.txFeeRecipient != w.validator {
```
mcortesi, line 468: do we know if current.header != nil?

gastonponti: The only function that sets the `current` is `makeCurrent`, which is called only by the `Prepare` function after initialising the header.
I can't use the `header.Coinbase` in that line, because it's the case where the consensus is not running, and leaves the `header.Coinbase` with `nil`

mcortesi (approved):
Great Work1

PR #1196, opened 2020-10-26
Add SignTypedData to PrivateAccountAPI

hbandura:
Since it uses a lot of types defined in `signer/core`, which cannot be referenced from geth code due to circular references, they were extracted to a common location `common/typeddata`.


hbandura:
This has been implemented in the mobile native interface and no longer needed as API

Issue #1131, opened 2020-08-06
Revert #333 After upstream merge


mcortesi:
This upstream PR https://github.com/ethereum/go-ethereum/pull/20453 (which is actually a fix for a bug that @asaj submitted to upstream geth) should make this celo-blockchain PR unnecessary and revert-able:  https://github.com/celo-org/celo-blockchain/pull/333.

One thing to consider if the celo-blockchain PR is reverted is if it will still be backwards compatible.

_Originally posted by @kevjue in https://github.com/celo-org/celo-blockchain/pull/1115#issuecomment-665298849_

oneeman:
Closing in favor of #1213

PR #1201, opened 2020-10-29
Make primary protocol concept apply to specific versions of istanbul

oneeman:
### Description

https://github.com/celo-org/celo-blockchain/pull/333 introduced a concept of a protocol being primary, which means that it is not run alongside any other protocol on the connection between two nodes.  But, in its implementation, if any version of istanbul is marked as primary, all versions of istanbul are treated as primary.  This PR updates the way the primary protocol concept is implemented so that versions 64 and 65 primary (to maintain backwards compatibility), but newer versions will not be.  This will enable running future versions of istanbul alongside other protocols such as LES or a separate proxy-validator protocol if we were to break that away from istanbul.  And, at a later time when versions 64 and 65 need not be supported, we can remove the concept of primary protocol entirely as suggested in #1131 

### Tested

I generated 3 binaries, with some additional logging to show which protocols are being used and when a peer is disconnected:
(a) master
(b) this branch
(c) this branch, but with a new istanbul version (66) that's the same as 65 except it's not primary (I did not commit this on any branch, it was just for testing how the code for this PR will behave when we create a new istanbul version later on)

I then ran them locally in pairs (setting each one to use the other as its bootnode) to see what happens when they connect.  In all cases, they managed to connect and maintained the connection.  The protocols used were as expected:
1. when (c) and (c) connect to each other, they run istanbul 66 and les 3
2. in the other 5 cases, they run only istanbul 65

I also verified that both (b) and (c) can starting syncing to mainnet.

### Backwards compatibility

This is backwards compatible, since istanbul 64 and 65 are made primary, so when connecting to peers running older versions of celo-blockchain, istanbul will be run by itself as described in the testing section.

### DIFF
--- a/consensus/istanbul/protocol.go
+++ b/consensus/istanbul/protocol.go
@@ -29,7 +29,8 @@ const (
 // protocolName is the official short name of the protocol used during capability negotiation.
 const ProtocolName = "istanbul"
 
-// ProtocolVersions are the supported versions of the istanbul protocol.
+// ProtocolVersions are the supported versions of the istanbul protocol (first is primary).
+// (First is primary in the sense that it's the most current one supported, not in the sense of IsPrimary() below)
 var ProtocolVersions = []uint{Celo65, Celo64}
 
 // Returns whether this version of Istanbul should have Primary: true (a legacy property that was needed to work
--- a/consensus/istanbul/protocol.go
+++ b/consensus/istanbul/protocol.go
@@ -29,8 +29,16 @@ const (
 // protocolName is the official short name of the protocol used during capability negotiation.
 const ProtocolName = "istanbul"
 
-// ProtocolVersions are the supported versions of the eth protocol (first is primary).
+// ProtocolVersions are the supported versions of the istanbul protocol.
 var ProtocolVersions = []uint{Celo65, Celo64}
+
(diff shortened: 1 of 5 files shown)


consensus/istanbul/protocol.go
```diff
@@ -29,9 +29,18 @@ const (
 // protocolName is the official short name of the protocol used during capability negotiation.
 const ProtocolName = "istanbul"
 
-// ProtocolVersions are the supported versions of the eth protocol (first is primary).
+// ProtocolVersions are the supported versions of the istanbul protocol (first is primary).
```
mcortesi, line 32: i don't think primary refers to which is the current one. The current is always the biggest version number. I would remove the docs

mcortesi (approved):

consensus/istanbul/protocol.go
oneeman: I was confused by it it, at first I thought it was something we added, but then I saw it's not.  It's also present on the LES protocols, though there when they added version 3 they put it last not first.  I assume that was a mistake, and that the intention is to always place the highest version first, because in some places in the code it references using `[0]`. e.g. https://github.com/celo-org/celo-blockchain/blob/master/eth/backend.go#L559 , but there are some other places.

PR #1203, opened 2020-10-30
Save lib bls for mobile client publishing

### LINKED ISSUE #1199: Publish Mobile Client Ci Check should work
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

How can we reproduce this?

### Logs

Are there any logs?

### System information

Run `geth version`

trianglesphere:
### Description

### Other changes

### Tested

### Related issues

- Fixes #1199 

### Backwards compatibility


oneeman (approved):
LGTM.

I think it'd be nicer if we could avoid referencing the specific celo-bls-go version in these files (though that isn't new to this PR).  Then again, if we update `celo-bls-go` but forget to update the version here, it would break the CI build, and we'd notice that way that we need to update it here, too, so it wouldn't make much of a different in the end.

kobigurk:
How about using the snippet `go list -m all | grep celo-bls-go | awk '{print $2}'` to get the version instead of fixing it?

trianglesphere:
> How about using the snippet `go list -m all | grep celo-bls-go | awk '{print $2}'` to get the version instead of fixing it?

I tested it out, it looks like it's working on circle ci. Nice trick

PR #1204, opened 2020-11-01
Ethstats.go memory leak fix

zviadm:
### Description

Remove unneeded goroutine and an extra channel. Also should fix unintended goroutine leak when logins are failing.

### Tested

Testing with pull request


oneeman:
Thanks, @zviadm!  This doesn't compile, because select cases have to send to or receive from a channel.  Since the fix is small, I've opened up a separate PR with it: #1206.  Thanks again for spotting the leak and bringing it to our attention!

PR #1205, opened 2020-11-02
Fix: account proof-of-possession should fail with invalid address

### LINKED ISSUE #1119: account proof-of-possession works even with an invalid address
### Expected Behavior

Running this command `docker run -v $PWD:/root/.celo --rm -it $CELO_IMAGE --nousb account proof-of-possession $CELO_VALIDATOR_SIGNER_ADDRESS $CELO_VALIDATOR_RG_ADDRESS` should fail with a invalid address in any parameter.

### Current Behavior

It creates a signature anyway, same behavior with `--bls` flag

### System information

Using docker image `us.gcr.io/celo-testnet/celo-node:rc1`

howydev:
### Description

currently, running "geth --nousb account proof-of-possession $SIGNER_ADDRESS $ADDRESS_TO_SIGN" doesn't do a validity check on $ADDRESS_TO_SIGN as specified in #1119

### Tested

I'm using IsHexAddress in the common/types package. Since theres no new functionality i didn't think its necessary to add a test. if I assumed wrongly ill be happy to add a couple of tests

### Related issues

- Fixes #1119 

### Backwards compatibility

really small change so definitely backwards compatible

### DIFF
--- a/cmd/geth/accountcmd.go
+++ b/cmd/geth/accountcmd.go
@@ -264,7 +264,11 @@ func accountProofOfPossession(ctx *cli.Context) error {
 	ks := am.Backends(keystore.KeyStoreType)[0].(*keystore.KeyStore)
 
 	signer := common.HexToAddress(ctx.Args()[0])
-	message := common.HexToAddress(ctx.Args()[1])
+	messageString := ctx.Args()[1]
+	if !common.IsHexAddress(messageString) {
+		utils.Fatalf("Address to sign is an invalid address")
+	}
+	message := common.HexToAddress(messageString)
 
 	var err error
 	var wallet accounts.Wallet


cmd/geth/accountcmd.go
```diff
@@ -264,6 +264,9 @@ func accountProofOfPossession(ctx *cli.Context) error {
 	ks := am.Backends(keystore.KeyStoreType)[0].(*keystore.KeyStore)
 
 	signer := common.HexToAddress(ctx.Args()[0])
+	if !common.IsHexAddress(ctx.Args()[1]) {
```
tkporter, line 267: tiny nit: wdyt of putting `ctx.Args()[1]` into a variable so we only have to call `ctx.Args()[1]` twice? That way if the command args ever change in the future we don't have to change multiple places

tkporter (approved):
Thanks!! Super small nit but looks great

PR #1206, opened 2020-11-02
Fix goroutine leak in celostats login.

oneeman:
The channel needs to be buffered, so that if a timeout occurs the goroutine would still be able to terminate.

### Description

Fixes a goroutine leak spotted by @zviadm.  Note that this leak is new, and was not present on v1.1.0.

### Tested

Not tested directly, but build and unit tests pass.

### Related issues

- Fixes the issue mentioned in #1204

### Backwards compatibility

No effect on protocols, so no backwards compatibility concerns.


trianglesphere (approved):
lgtm

PR #1208, opened 2020-11-03
Fix duplicated code from bad cherry-pick conflict resolution.

oneeman:
This increment was moved a few lines lower down in the function in the upstream commit, but during the conflict resolution for #1192, I missed removing it here, so it became duplicated.

### Description

This was a bad merge on my part in PR #1192, resulted in this if block being duplicated (it is present a few lines below this), which is rectified here.

### Tested

Not specially tested, very clear fix of a bug.


kevjue (approved):
Nice catch.

mcortesi (approved):
nice

PR #1209, opened 2020-11-03
Merge upstream 1.9.11

### LINKED ISSUE #1172: p2p.Server connections not being tracked properly
`p2p.Server` uses [meteredConnection](https://github.com/celo-org/celo-blockchain/blob/master/p2p/metrics.go#L114) to monitor connections (ingress/egress connections; bandwidth usage)

Nodes on mainnet currently see a `"Metered peer count reached the limit"` warning that indicates the nodes has more the 1024 active connections, still when checking active connections, there's no more than 100 (as indicated by the max peers flag) 

This log message, seems to imply one of two things:
   1. We are not tracking close connection events properly
   2. We don't close the connections but we do loose track of them internally

Both scenarios don't look well, so we need to investigate and solve this issue

oneeman:
### Description

Merge was done in several steps, to avoid overwhelming merge conflicts.  Some notes about individual merge commits are below.

Besides that, the main thing that bears mentioning is the introduction of DNS-based discovery (was already mostly ready but unused in the code, but finalized and enabled by default here).  It is used for eth/istanbul only (not for LES), through `proto.DialCandidates`, which is then added as a source to p2pserver's `discmix` iterator.  If we wanted to start using it ourselves, we should be able to, following the instructions at https://geth.ethereum.org/docs/developers/dns-discovery-setup.  The only difference I'm aware of is that we'd need to include the right networkid (for mainnet, alfajores, or baklava) when running the `crawl` subcommand.  I tried `crawl` out, and it currently doesn't work due to a problem with the networkid flag, so I opened a separate PR to fix that issue, #1212, since the affected code is already on master and not really part of this merge PR.

`dcffb77`
- `dummyChain` in `core/vm/runtime/runtime_test.go` needed additional methods to satisfy the `ChainContext` interface, so I added them even though we also have a `mockBlockchain` type in istanbul test files
- Upstream commit `058a4ac` improved the performance of `GetHashFn()` in `core/evm.go`.  In celo-blockchain, that file no longer exists, and that function has been moved to `core/vm/runtime/runtime.go`, so I updated that to the improved version.

`049e171` (eth/65)
- For the changes to `eth/protocol.go`, I instead made the corresponding changes to `consensus/istanbul/protocol.go`.  In a later separate commit, I add a comment clarifying that Celo66 corresponds to eth/65
- In `eth/protocol_test.go`, they did not update the protocol version used for `TestForkIDSplit`.  This doesn't matter, since since the protocol change has nothing to do with that, but seemed like a good idea to have it use the newest version, so I updated it to 66
- After merging this commit, `TestSendTransactions63` and `TestSendTransactions64` fail.  This is also the case upstream, and is fixed in a later commit that was part of the same upstream PR (`9938d95`, Feb 13, 2020)

`90caa2c` (new dial scheduler)
-  Used the `--ours` merge strategy (after merging in from celo-blockchain master), since this is already on master from having been cherry picked

`9938d95` (second part of eth/65)
- This removed metrics from `eth/`, namely the files `eth/metrics.go` and `eth/fetcher/metrics.go`.  Removing `eth/metrics.go` had conflicts with ours, but they just had to do with how we renamed eth to istanbul and the versions, so not relevant to the fact that the file was deleted.  The rationale for this removal can be seen on the upstream PR: https://github.com/ethereum/go-ethereum/pull/20234#issuecomment-577221101 ("Removed the metrics from the eth package as all network packets are already covered by p2p generically, so it's better not to have to maintain this code.").  None of the metrics removed or renamed are among the ones we list in our docs (https://docs.celo.org/validator-guide/monitoring#metrics)
  - Removed files/metrics: 
    - Some of these seem to be covered in https://github.com/celo-org/celo-blockchain/blob/ba75a9ad8a16b050d84956b8bf045dca0ca9989c/eth/downloader/metrics.go: https://github.com/celo-org/celo-blockchain/blob/011f74f1ec1cdb00603b8228e6c67ab4a763dea6/eth/metrics.go#L24-L57
    - https://github.com/celo-org/celo-blockchain/blob/011f74f1ec1cdb00603b8228e6c67ab4a763dea6/eth/fetcher/metrics.go#L25-L43
  - Added metrics:
    - https://github.com/celo-org/celo-blockchain/blob/ba75a9ad8a16b050d84956b8bf045dca0ca9989c/eth/fetcher/block_fetcher.go#L46-L64
    - https://github.com/celo-org/celo-blockchain/blob/ba75a9ad8a16b050d84956b8bf045dca0ca9989c/eth/fetcher/tx_fetcher.go#L70-L97
- There are some function names ending in 64, for the old protocol version (eth/64).  In normal code: `txsyncLoop64` and `SendTransactions64`.  Did not rename them to match the istanbul version numbers instead, but in comments which mention eth/6x I mentioned which celo/6x they correspond to.  In the test code there are many to test with several protocol versions.  For those, I did update it, since there the function name is paired with a protocol number that's given as an argument to a generic test function, and so it'd be odd to have a mismatch.
- A couple of the tests relied on deterministic random numbers for the order in which transactions are requested.  Since Celo transactions have new fields to do with the fees, our hashes were different and so the indexes used in the tests had to be modified for the tests to pass.  I did this in a separate commit

`26284ec`
- They updated their CI configuration to use Go 1.13.8 instead of 1.13.6.  Looking at our CircleCI config, we use the docker image `circleci/golang:1.13` (1.13.15) for the build, for the e2e tests we use `celohq/node10-gcloud:v3` with `GO_VERSION=1.13.7`, and in our Dockerfile we use `golang:1.13-alpine`(1.13.15).  Open question: should we update GO_VERSION for the e2e tests to 1.13.15?  Any reason not to?
- Upstream enables DNS discovery by default.  Since we don't have the DNS entries for discovery, I defined the mainnet, baklava and alfajores URLs to be "", which means it will be disabled when using them, but enabled if we later on add those urls in bootnodes.go
- In `cmd/utils/flags.go`, at the end of `SetEthConfig`, there is a check for `networkId == 1` (ethereum mainnet).  I changed this to `== params.MainnetNetworkId`.

`ac72787`:
- Significant conflicts in `p2p/metrics.go`.  I made sure to preserve the metrics we've added in that file, listed below.  Of these, the two handshake metrics were being updated in `handshakeDone()`, which no longer exists.  Instead, in a separate commit, I made it update them `p2p/server.go`, where `handshakeDone()` was being called from.  Our added metrics:
  - p2p_serves_handshakes (ingressConnectWithHandshakeMeter)
  - p2p_dials_handshakes (egressConnectWithHandshakeMeter)
  - p2p/peers/discovered (discoveredPeersCounter)
  - p2p/peers/validators (activeValidatorsPeerGauge)
  - p2p/peers/proxies (activeProxiesPeerGauge)
- The changes in this commit remove the `handshakeDone()` and the `meteredPeerCount` counter, which suffered from a race condition that led to #1172.  As a result, when this PR is merged we can close that issue.

### Tested

Automated tests pass, but some are flaky.  Will look into them to see whether or not it may be anything specific to this PR and whether they also happen on upstream 1.9.11.

### Related issues

- Closes #1172

### Backwards compatibility

No breaking changes either at the protocol level or at the local use level.

### Reference links

1.9.11 release notes: https://github.com/ethereum/go-ethereum/releases/tag/v1.9.11
1.9.11 PRs: https://github.com/ethereum/go-ethereum/milestone/97?closed=1
1.9.11 commits: https://github.com/ethereum/go-ethereum/commits/master?before=6a62fe399b68ab9e3625ef5e7900394f389adc3a+35&branch=master


mcortesi (approved):
LGTM.

One minor issue... i suspect that `geth dumpgenesis` won't work for us, at least istanbul config things might be missing. We should check if this is the case. If it is i'm fine with just creating an issue to fix it, and merge anyway

oneeman:
> LGTM.
> 
> One minor issue... i suspect that `geth dumpgenesis` won't work for us, at least istanbul config things might be missing. We should check if this is the case. If it is i'm fine with just creating an issue to fix it, and merge anyway

I confirmed that dumpgenesis includes the `config.istanbul` property.  I tested for the mainnet genesis block, and the output included epoch, policy, and lookbackwindow, which are the ones we specify in the mainnet config (omitted ones like RequestTimeout get the default hard-coded values)

PR #1214, opened 2020-11-05
Fix eth_getTransactionReceipt on light clients

nategraf:
### Description

This PR includes a simple change to fix the `eth_getTransactionReceipt` API on light nodes. It turns out in the original implementation of LES transaction fetching a minor omission was made in updating the `GetTransactionReceipt` method.

https://github.com/celo-org/celo-blockchain/commit/40cdcf8c47ff094775aca08fd5d94051f9cf1dbb#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R1195

This PR updates the method to use the ODR-based light client backend instead of `rawdb`.

### Other changes

None

### Tested

Ran on `alfajores` and confirmed that `eth_getTransactionReceipt` now works.

### Related issues

- Related upstream issue: https://github.com/ethereum/go-ethereum/issues/16347

### Backwards compatibility

No concerns

PR #1217, opened 2020-11-06
Fix behavior of eth_getTransactionHash for pending transactions

nategraf:
### Description

Follow-up to https://github.com/celo-org/celo-blockchain/pull/1214, I found that returning the error when the transaction is not found is not the behavior expected by clients and leads to difficult to debug behavior. This fix aligns with upstream and client expectations.

Prior #1214 and this PR, `eth_getTransactionReceipt` did not work on light clients because the API implementation was checking for the given transaction in the local `rawdb` rather than using ODR, as it was doing for `eth_getTransaction`. As a result any query for a transaction that was not otherwise cached would always return `nil`. With these PRs, this is fixed so that the transaction is retrieved via ODR, as expected.


kevjue (approved):
LGTM.  It's consistent with upstream geth behavior.

PR #1218, opened 2020-11-06
Fix delegateMsg handle

gastonponti:
### Description

Fixes an error added in a refactor that kept an old way of receiving a message to be signed by the validator, from a proxy, to be sent to celostats

### Tested

Upload a testnet with multiple escenarios of [validators, n-proxies], validators without proxies, and their celostats server.
Monitored the data and the logs of the validators, and everything was working properly.


kevjue:
@gastonponti - Can we test this in a testnet w/ celostats.

kevjue (approved):
lgtm

PR #1219, opened 2020-11-09
mobile interface ComputeECDHSharedSecret

hbandura:


trianglesphere (approved):
Looks good to me.

The wsddn/go-ecdh library was already added to the project and seems fine: we don't use any curves from it, just it doing ECDH on a provided elliptic.

AlexBHarley:
If there are concerns about wsddn/go-ecdh, there's also no problem calling directly into the go crypto module. The operation is simple enough not to warrant going via an extra library https://github.com/wsddn/go-ecdh/blob/master/elliptic.go#L81. That was how I originally imagined implementing this

PR #1220, opened 2020-11-09
Add SignTypedData to native Keystore interface

hbandura:


mobile/accounts.go
```diff
@@ -115,6 +118,31 @@ func (ks *KeyStore) SignHash(address *Address, hash []byte) (signature []byte, _
 	return ks.keystore.SignHash(accounts.Account{Address: address.address}, common.CopyBytes(hash))
 }
 
+// SignTypedData signs EIP-712 conformant typed data
+// hash = keccak256("\x19${byteVersion}${domainSeparator}${hashStruct(message)}")
+func (ks *KeyStore) SignTypedData(account *Account, typedDataJSON []byte) ([]byte, error) {
```
trianglesphere, line 123: Is this `SignTypedData` too hard to easily use here? https://github.com/celo-org/celo-blockchain/blob/16e1c70f363451b66712a99540bc03716d93c565/signer/core/signed_data.go#L256

trianglesphere (commented):
The code looks fine. One thing that would be nice is to have some test cases with `SignTypedData`.

mobile/accounts.go
hbandura: basically i'm duplicating how that works. The issue is , that method is from an API struct, which we don't have here. But also, it's (afaik) from the clef application, and it has some hooks for UI input (for example for unblocking the account). So it can't be used verbatim. A case can be made to extract some parts in common: e.g the generation of the string to sign (it could be a method from the TypedData structure).

But since I'm not fully aware for how long will this functionality be needed in the native interface, I was reluctant to modify the `signed_data.go` file for this. I actually think that maybe for this case it makes sense, at least that part (lines 129 to 137).

trianglesphere: Sounds good. The code looks fine reading it, but some quick tests to make sure it does what we expect would be great (manual or unit is fine, but I'd like a sanity check)

hbandura: I added a simple test to check that in 'kinda' does what it should. I peeked at the clef unit tests and they are doing exactly this. It's not really testing that the signature is correct.

trianglesphere (approved):
Looks good. Once this gets merged let's let wallet know that it's in so they can test their integration with it.

PR #1222, opened 2020-11-11
Update release/1.1.0 branch for Go 1.14.x

oneeman:
### Description

This cherry-picks some commits from upstream, and adds some additional changes in a separate commit, to:

1. Become compatible with Go 1.14.x
2. Use the latest 1.14.x version in the CI and the dockerfiles

The idea is to then switch to v1.14.12 when it is released.

The changes from upstream are very minor:

1. Upgrading a dependency version
2. Fixing a test to work with 1.14

For now, I am creating this as a draft PR with "do not merge".  I believe what we should eventually do, rather than merge it to `release/1.1.0`, is create a new branch from it (e.g. `release/1.1.1`) and merge this PR onto that.  cc @trianglesphere 

### Tested

Unit tests and e2e sync tests pass.

### Backwards compatibility

No significant

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ jobs:
     - stage: lint
       os: linux
       dist: xenial
-      go: 1.13.x
+      go: 1.14.x
       env:
         - lint
       git:
@@ -35,12 +35,24 @@ jobs:
         - go run build/ci.go install
         - go run build/ci.go test -coverage $TEST_PACKAGES
 
+    - stage: build
+      os: linux
+      dist: xenial
+      go: 1.13.x
+      env:
+        - GO111MODULE=on
+      script:
+        - go run build/ci.go install
+        - go run build/ci.go test -coverage $TEST_PACKAGES
+
     # These are the latest Go versions.
     - stage: build
       os: linux
       arch: amd64
       dist: xenial
-      go: 1.13.x
+      go: 1.14.x
+      env:
+        - GO111MODULE=on
       script:
         - go run build/ci.go install
         - go run build/ci.go test -coverage $TEST_PACKAGES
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -180,10 +199,11 @@ jobs:
       env:
         - azure-android
         - maven-android
+        - GO111MODULE=on
       git:
         submodules: false # avoid cloning ethereum/tests
       before_install:
-        - curl https://dl.google.com/go/go1.13.6.linux-amd64.tar.gz | tar -xz
+        - curl https://dl.google.com/go/go1.14.2.linux-amd64.tar.gz | tar -xz
```
trianglesphere, line 206: It seems like this is split between `1.14.11` and `1.14.2` We don't use travis so this is fine, but it's also referenced in appveyor.

oneeman: Do we use appveyor?  If so, I will update it there too.  For the travis config I figured better to leave it since we don't use it so there is no need to diverge on that from upstream.

trianglesphere: Looking close I don't think so. It looks like upstream uses appveyor for windows builds which we don't have configured

oneeman:
The iOS build is indeed broken.  I have another branch where I fixed the part of getting Go 1.14 with brew, but then the build itself fails and it's not clear to me how to fix it.  There is the possibility of leaving the iOS build using 1.13 for now.

The other failures are an e2e test known to be flaky, and a timeout in the unit tests on `TestRandom` from `trie`, which does randomized testing.  I ran it locally many times and didn't encounter it (normal running time for all the `trie` tests together is 3 seconds).

oneeman:
Converted from draft to normal PR, but will leave "do not merge" until v1.14.12 is released later today and I add a commit updating the 1.14.11 download URLs

oneeman:
Added a commit updating download links to 1.14.12.  But to build the docker images, we need the `golang:1.14-alpine` image to be updated with the new version as well.

trianglesphere (approved):
lgtm

mcortesi (approved):

asaj (approved):

PR #1223, opened 2020-11-11
Fix ios build on CircleCI

oneeman:
### Description

Attempt to fix the ios build on circleci.  This successfully gets Go 1.14, but gives the following error:
```
# github.com/celo-org/celo-bls-go/bls
ld: '/private/var/folders/pt/0xykrh9j62g34sc7znh47kbc0000gn/T/gomobile-work-621612876/pkg/mod/github.com/celo-org/celo-bls-go@v0.1.6/libs/universal/libbls_snark_sys.a(bls_snark_sys-84e37a6bf817296f.186kc8ahx45hs0tv.rcgu.o)' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Most likely we should disable bitcode.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,10 +106,10 @@ jobs:
       - run:
           name: Setup Go language
           command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install go
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install go@1.14
-            HOMEBREW_NO_AUTO_UPDATE=1 brew unlink go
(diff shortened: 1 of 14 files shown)


oneeman:
Because the Go 1.14.x PR has already been merged to master, I'm closing this draft PR, in favor of #1245.

PR #1224, opened 2020-11-11
Churrito hard fork

oneeman:
### Description

1. Replace unused Celo1 fork info with `churrito` hard fork (https://forum.celo.org/t/pre-announcement-of-security-hotfix-releases-celo-blockchain-1-1-1-and-1-2-2-and-churrito-hardfork/677), and implement the EVM change.  When we cherry-pick this on top of v1.1.0, we should first cherry-pick #1173, to avoid gratuitous merge conflicts.
2. Add Donut fork info (without setting activation block numbers), for use in future PRs implementing functionality for the future Donut hard fork (until now they had been using `Celo1`, which was added before the name Donut was chosen).
3. Set Churrito activation block number for Baklava as specified in https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0024.md, and enable churrito for dev chain (`--dev`) and the test chains (used in the automated testing).

### Tested

Confirmed locally using logging:

1. With the activation block number set to `nil`, the original path (without the copy) is executed.
2. With the activation block number set to 1, the original path is executed for block 0 and then on block 1 it switches to the new path.
3. With activation block set to the future, client is able to establish connections with unupgraded peers.
4. With activation block set to the past, connections to unupgraded peers are rejected during the handshake due to the fork id check.

### Backwards Compatibility

Not backwards compatible, as this triggers a hard fork on Baklava at the specified block numbers.  Therefore, validators must and full nodes must upgrade before that time to ensure the network's health.

After the activation block number, upgraded and unupgraded peers will not be able to connect to each other, as there is a fork id check in the handshake (see https://eips.ethereum.org/EIPS/eip-2364 for details).  If the activation block number arrives and there isn't a quorum of upgraded validators, the network is likely to stall, though not necessarily immediately (because the fork id check is done during the handshake, peers who are already connected to each other won't disconnect each other because of the fork id).

### DIFF
--- a/core/vm/interpreter.go
+++ b/core/vm/interpreter.go
@@ -279,7 +279,11 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 		// if the operation clears the return data (e.g. it has returning data)
 		// set the last return to the result of the operation.
 		if operation.returns {
-			in.returnData = res
+			if in.evm.chainRules.IsChurrito {
+				in.returnData = common.CopyBytes(res)
+			} else {
+				in.returnData = res
+			}
 		}
 
 		switch {
--- a/params/config.go
+++ b/params/config.go
@@ -63,6 +63,7 @@ var (
 		PetersburgBlock:     big.NewInt(0),
 		IstanbulBlock:       big.NewInt(0),
 		ChurritoBlock:       nil,
+		DonutBlock:          nil,
 		Istanbul: &IstanbulConfig{
 			Epoch:          17280,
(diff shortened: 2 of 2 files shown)


prestwich (commented):
very cute name. fix behavior looks correct to me but I am less familiar with the HF logic

andres-dg:
Like we chatted on Slack, there's no need to delay the merge of this PR to master. The hard fork will be activated with a separate PR that sets the ChurritoBlock in the chain config.

trianglesphere (approved):
LGTM.

PR #1228, opened 2020-11-13
Only create replicaState if configured as a validator

trianglesphere:
### Description

This should reduce problems with using data-dirs from full nodes. This will not completely eliminate the problem if a data-dir is reused from a validator.

### Tested

Confirmed that full nodes don't create the replica state and that nodes configured with --mine do.

### Backwards compatibility

Backwards compatible

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -620,7 +620,7 @@ func (sb *Backend) updateReplicaStateLoop(bc *ethCore.BlockChain) {
 	for {
 		select {
 		case chainEvent := <-chainEventCh:
-			if !sb.coreStarted {
+			if !sb.coreStarted && sb.replicaState != nil {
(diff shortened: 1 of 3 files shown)


consensus/istanbul/backend/engine.go
```diff
@@ -1017,6 +1026,9 @@ func (sb *Backend) SetStartValidatingBlock(blockNumber *big.Int) error {
 
 // SetStopValidatingBlock sets the block that the validator will stop just before (exclusive range)
 func (sb *Backend) SetStopValidatingBlock(blockNumber *big.Int) error {
+	if sb.replicaState == nil {
+		return errors.New("Not configured as a validator")
```
mcortesi, line 1030: I would create an error variable with this error and resuse it. `NotAValidatorError` or something

mcortesi (approved):

PR #1239, opened 2020-11-17
Merge upstream v1.9.12

oneeman:
### Description

Merge from upstream up to v1.9.12 (release notes at https://github.com/ethereum/go-ethereum/releases/tag/v1.9.12).  Halfway through, I switched from merging groups of commits to merging each one individually (even trivial ones).

**Merge notes**:

**The merge up to `fadf84a`** includes a couple of commits with notable issues:
- `4be8840`
    - EIP-1884 and EIP-2200 were both in included in Ethereum's Istanbul fork and both included changes to gas costs, including (in both of them) changing the price of `SLOAD` from 200 (which it's been since EIP-150) to 800.
    - When they were implemented in geth, only the EIP-1884 implementation included the change to `SLOAD`, so the EIP-2200 implementation was technically incorrect.  But in practice was fine because both were activated in the same hard fork.
    - This merge includes the upstream commit `4be8840`, which updates `enable2200()` to also set the price of `SLOAD` to 800, making the EIP-2200 implementation correct even apart from EIP-1884
    - For celo-blockchain, we commented out this EIP-1884 change to `SLOAD` (as well as the ones to `BALANCE` and `EXTCODEHASH`).  Therefore, to preserve backwards compatibility, I also commented out the EIP-2200 change in `4be8840`, in a separate commit, so as to make it more explicit.  If we'd left it in, it would have caused a hard fork.
    - Note that there is also discussion about whether we should adopt the EIP-1884 changes in the upcoming Celo Donut hard-fork, to align our gas costs with Ethereum's, but that is a separate matter.
- `fadf84a`
    - This is a small breaking change to the `eth_call` RPC method, modifying its behavior when the optional `from` field is omitted (the ethereum RPC specs don't specify the behavior in that case)
    - geth's old behavior was to use its first account, or `0x00` if it has no accounts.  Parity just used `0x00`.
    - After discussion (https://github.com/ethereum/go-ethereum/issues/20685), it was decided to change geth's behavior to use `0x00`, which is done in this commit.
    - In their release notes they call it out as a breaking change and say users should always include `from` when using `eth_call`

**`4fabd9c`** is a big refactor of `les/peer.go`, splitting up the `peer` type into `serverPeer` (used by the light client to represent a peer that is a light server) and `clientPeer` (used by the light server to represent a peer that is a light client), with common fields in `peerCommons`.  The conflicts with our changes were pretty large, but not especially tricky.
- We added the fields `etherbase` and `gatewayFee`.  I moved these to `serverPeer` with a comment labeling them as "Gateway fields".  I also put the corresponding methods on `serverPeer`: `Etherbase()`, `SetEtherbase()`, `GatewayFee()`, `SetGatewayFee()`
- Methods involving server-client communication of gateway fee and etherbase:
    - in `serverPeer` (called by the client): `RequestEtherbase()`, `RequestGatewayFee()`
    - in `clientPeer` (called by the server): `SendEtherbaseRLP()`, `ReplyGatewayFee()`
- Another method we added: `WillAcceptTransaction()`, goes on `serverPeer`
- We changed `hasBlock()` to use `*uint64` rather than `uint64`, so I made that change in `serverPeer` where it's been moved to
- Upstream also split `peerSet` into `serverPeerSet` and `clientPeerSet`
    - We added `randomPeerEtherbase()`, which goes to `serverPeerSet`
    - We added `AllLightClientPeers()`, which I removed because it's unnecessary (and updated the code to instead use `clientPeerSet`'s `allPeers()`).  Light servers never added other servers to their peer set, anyway.  Originally other server peers led to an error (hence our Primary protocol concept), then later on it was modified to have them just wait and do nothing: https://github.com/celo-org/celo-blockchain/blob/bd6a4a40ea2fa883e74f26566f77416781b7198e/les/server_handler.go#L134-L138 
- `peer.headBlockInfo()` was removed.  It was being used for total difficulty and block number.  Difficulty is now done with `peer.Td()`, but there is no method for the block number.  It would be trivial to add a `.Number()` method, but the place it was being used in `les/sync.go` was using it in an unnecessary `if` (since difficulty is just block number + 1), so there was no need.
- There were considerable conflicts in `les/txrelay.go`, but straightforward to resolve.  Nevertheless, since the changes from merging in 1.9.12 are relatively small, I recommend reviewing them in their entirely.
- Besides the merge conflicts in various files in `les/`, I fixed other errors as needed, such as:
    - `peer` -> `serverPeer` in some light client functions we added
    - references to methods which were made private (e.g. `GetRequestCost()` -> `getRequestCost()`, `serverPeerSet`'s `.allPeers()`, `.len()`, etc.)
    - method with signatures updated by upstream:
        - `requestHeadersByHash()` and `requestHeadersByNumber()` had their second argument (`cost`) removed.  `cost` was being passed on to `sendRequest()` in `les/peer.go`, but `sendRequest()` wasn't using it, and so upstream removed it as an argument.  So I removed the `cost` value from where we call these functions in `les/odr_requests.go`
- `les/peer_test.go` was largely redone upstream.  Our changes were just adding a test (`TestWillAcceptTransaction()`), so I re-added it and updated it to use `serverPeer` instead of `peer`.

After merging in 1.9.12, I cherry-picked **`62cd943`** (as **`19bf7b8`** below), which is from 1.9.13, because I ran into it looking at the 1.9.13 PRs and saw that it fixes a deadlock bug in `les/peer.go` which was introduced in 1.9.12, so it didn't make much sense to leave this PR with a known bug it introduced.

### Tested

Unit tests and integration tests.

### Backwards compatibility

No backwards incompatible changes to the network protocols.  Small breaking change to the `eth_call` RPC, as mentioned above in the merge notes.


trianglesphere:
Very first pass, this seems to be getting stuck in circle ci.

the transfer test is failing with
```
#!/bin/bash -eo pipefail
export PATH=${PATH}:~/repos/golang/go/bin
./ci_test_transfers.sh local ~/repos/geth
/bin/bash: line 1: ./ci_test_transfers.sh: No such file or directory

Exited with code exit status 127
CircleCI received exit code 127
```

and the coverage test is failing with 
```
error computing cache key: template: cacheKey:1:13: executing "cacheKey" at <checksum "go.sum">: error calling checksum: open /home/circleci/repos/geth/go.sum: no such file or directory
```

trianglesphere (approved):
The test issue appeared to be flakey tests. Looking through this, the biggest code change is the LES refactor, however it seems to keep the network interface the same.

PR #1240, opened 2020-11-18
Notify proxied validator engine of new epoch

kevjue:
### Description

The proxied validator engine was not being notified of a new epoch, causing that engine to not be aware of any new validator set diffs, and consequently not assigning new validators to it's proxies and never connecting to them.

### Other changes

None

### Tested

- Unit tests
- Deployed patch to a few clabs validators and verified that they started signing all of their blocks.

### Backwards compatibility

Yes.


trianglesphere (approved):
lgtm.

PR #1242, opened 2020-11-18
Disable DNS discovery by default

oneeman:
### Description

The DNS discovery code from upstream includes the URLs to use.  Since we haven't set up the nodes tree on DNS for use by DNS discovery, we have no URLs we can use for that.  I believed that having the url set to `""` was enough to disable it, because the docs say to disable it using `--discovery.dns ""`.  However, those are not equivalent, and it was leading to a crash.  So this PR modified the initialization code to work so that if you don't specify `--discovery.dns` it will behave as if you specified `--discovery.dns ""`.

This highlights a serious gap in our integration tests, that we have no test of non-light sync using discovery:
1. The e2e tests in the monorepo all use `--nodiscover`
2. `scripts/sync_test.sh` uses only lightest mode sync, but discovery isn't the same for light/lightest mode as for the other modes

I will open up a separate issue about fixing this gap in our integration tests.

### Tested

Running `geth` before this would crash.  After this fix, it starts syncing as it should.

### Backwards compatibility

Backwards compatible.


kevjue (approved):
LGTM!

PR #1243, opened 2020-11-20
Fix argument type in native interface

hbandura:
Fix argument type error causing the android cross compile to silently ignore the ComputeECDHSharedSecret method.


gastonponti (approved):
LGTM

PR #1244, opened 2020-11-24
Fix shared secret generation

hbandura:
Fix a type conversion problem in the computeSharedSecret function in the keystore
Add a unit test that triggered the issue
Add a small doc explaining the public key format


gastonponti (approved):
LGTM

PR #1245, opened 2020-11-25
Update ios build on CI to use go 1.14.x

oneeman:
### Description

When we updated celo-blockchain to use Go 1.14.x, we left the CI ios build as is, because it was failing with an error.  This PR resolves that by:

1. Updating the ios build to use Go 1.14.x
2. (c/o @jeanregisser) Disabling bitcode for the ios build, which required creating and using a fork of https://github.com/golang/mobile (https://github.com/celo-org/mobile/tree/jeanregisser/disable-bitcode)
3. (c/o @jeanregisser) Refactoring how we include `libbls_snark_sys.a` in the build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,10 +109,11 @@ jobs:
       - run:
           name: Setup Go language
           command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install go
+            brew install go@1.14
+            brew link go@1.14
             # Check that homebrew installed the expected go version
-            if [[ "$(go version)" != "go version go1.13"* ]]; then
-              echo "go1.13 is required"
+            if [[ "$(go version)" != "go version go1.14"* ]]; then
+              echo "go1.14 is required"
               exit 1
             fi
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,9 +24,6 @@ jobs:
       - run:
           name: Build Geth
           command: go run build/ci.go install
-      - run:
-          name: Save lib bls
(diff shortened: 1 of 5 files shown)


trianglesphere:
One thing to flag is that we already have a fork of Golang sys library for building android for a different reason (I believe 64 bit emulator compatibility) https://github.com/celo-org/celo-blockchain/compare/cmcewen/mobile-1.2#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R76

Connor has a PR for that upstream, but I think it is also stuck in the google CLA stage. I think that the thinking is that we don't want to maintain a fork of Golang standard libraries in master.

oneeman:
> One thing to flag is that we already have a fork of Golang sys library for building android for a different reason (I believe 64 bit emulator compatibility) https://github.com/celo-org/celo-blockchain/compare/cmcewen/mobile-1.2#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R76
> 
> Connor has a PR for that upstream, but I think it is also stuck in the google CLA stage. I think that the thinking is that we don't want to maintain a fork of Golang standard libraries in master.

I didn't realize there was one for android too.  @jeanregisser is checking with legal about the CLA, and I don't imagine it should be a problem.  Once that's clear we can let Connor know as well for his PR.  After that, of course, it will remain to be seen whether the PRs or alternative fixes for the problems they addressed make it into the golang repos.

jeanregisser:
Yes note that gomobile is not a go library, it's a tool to create iOS/Android libs out of go code.
Other build types are untouched by this change.

trianglesphere (approved):
LGTM. Go mobile apparently is also semi-experimental, so if/when we can get changes upstreamed, let's switch back.

PR #1246, opened 2020-11-27
Disable bitcode

jeanregisser:
### Description

The work here disables bitcode so that the build works again on iOS using go 1.14
- Had to fork gomobile to patch this, PR submitted upstream https://github.com/golang/mobile/pull/59
- Move `libbls_snark_sys` handling to the Makefile and iOS step and explain why we need it

### Tested

- iOS build compiles again locally and on CircleCI
- was able to sync, send money using a local build connected to alfajores on iOS

### Backwards compatibility

Yes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,9 +24,6 @@ jobs:
       - run:
           name: Build Geth
           command: go run build/ci.go install
-      - run:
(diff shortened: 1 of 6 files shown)


oneeman (approved):
LGTM, thanks Jean!

PR #1249, opened 2020-12-01
CIP-20: Extensible Hash Functions

prestwich:
### Description

Tracking PR for ongoing work on [CIP-20](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0020.md). Initial implementations are ready for review.

I have chosen to vendor the blake2s dependency, but all other hash functions rely on the golang standard library.

TODO:
- [x] bench blake2s (or assign safe constants)
- [x] expand unit tests
- [x] Remove from Istanbul precompile mapping
- [x] Update with activation logic

### Tested

Fuzz tested for panics and correct output using [smash](https://github.com/prestwich/smash). At least 1 unit test vector for each

### Backwards compatibility

~~As written, this is a hard fork change *without activation logic*. It MUST NOT be merged as is.~~

this is no longer a hard fork, and is now activated in the donut block. It should be safe to merge


core/vm/contracts_test.go
```diff
@@ -983,3 +985,87 @@ func TestGetVerifiedSealBitmap(t *testing.T) {
...
+
+func TestCip20(t *testing.T) {
+	cip20ShaVariantTests, err := loadCip20Json()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, vector := range cip20ShaVariantTests {
+		for _, test := range vector.toPrecompiledTests() {
+			testPrecompiled("f3", test, t)
```
trianglesphere, line 1064: This is causing a nil pointer panic in the tests. The precompile that gets loaded appears to be nil. Also flagging that this selector may no longer point to the CIP-20 precompile.

prestwich: oh I had an unpushed commit that fixes this. just rebased onto master and included the fix. the problem is that `f3` is not the canonical address for CIP20. it ought to be `e2`

trianglesphere:
This looks good to me. The PR description that this does not include the HF activation logic seems like it's out of date to me.

prestwich:
> This looks good to me. The PR description that this does not include the HF activation logic seems like it's out of date to me.

Fixed. Thanks for pointing that out

trianglesphere (approved):

nategraf (changes requested):
Working on a review and I'll edit this when I am done, but I found an issue so I wanted to drop a flag here.

core/vm/cip20.go
```diff
@@ -0,0 +1,207 @@
...
+	"errors"
+
+	"crypto/sha512"
+
+	"github.com/ethereum/go-ethereum/crypto/blake2s"
+	"github.com/ethereum/go-ethereum/params"
+	"golang.org/x/crypto/sha3"
+)
+
+const (
```
nategraf, line 13: Word is quite overloaded in the current context. (e.g. it could mean machine word, EVM word, or the hash input unit here) Could you clarify what is meant here in a comment and note that the units are bytes?
```suggestion
// Token unit of length, in bytes, used for calculating gas with respect to each support hash function.
const (
```

nategraf (commented):
I was a bit confused by the usage of the term “word”, as used in the CIP and in this implementation. I inferred that this means the token input size, and not the machine-word or EVM word, but my confusion was deepened when digging into the input unit size for SHA3, which they refer to as a bitrate (rather than block size, for reasons I don’t yet understand).

Based the helpful table in Section 7 of the spec, SHA3-256 and 512 act on the input in chunks of 144 and 72 bytes respectively. Based on all of this, I feel like we should avoid using the term “word” in the CIP and in the implementation unless we are referring to an EVM word (or a machine-word). I would be supportive of either pricing the hash functions by number of EVM words input or by number of number of input chunks as defined by the bitrate, which IIUC determines the number of sponge-function invocations.

I do not think we should price SHA3-512 based on the input size divided by 64-bytes. (Part of this is that SHA3-512 requires roughly twice as many sponge-function invocations per byte for long inputs compare to SHA3-256, and part is that the number, 64-bytes, doesn’t really have much to do with the workings of SHA3-512, except in the security level and output length)

I think we would be fine to proceed as-is (i.e. I don’t have any reason to think this will be DoS vector), but it would be nice to adjust this now rather than never wanting to in the future because it’ll be a hard-fork.

Other than that note, it all looks good to me!

core/vm/cip20.go
trianglesphere: Flagging that if we want to change metering, [CIP-25](https://github.com/celo-org/celo-blockchain/pull/1256) uses a similar metering method/namings

core/vm/contracts.go
```diff
@@ -1132,3 +1132,38 @@ func (c *getVerifiedSealBitmap) Run(input []byte, caller common.Address, evm *EV
...
+		return h.RequiredGas(input[1:])
+	}
+
+	return params.InvalidCip20Gas
+}
+
+func (c *cip20HashFunctions) Run(input []byte, _ common.Address, _ *EVM, gas uint64) ([]byte, uint64, error) {
+	gas, err := debitRequiredGas(c, input, gas)
+
+	if err != nil || len(input) == 0 {
```
mcortesi, line 1155: This imply that if `len(input) ==0` you return without error. I would suggest we do error here, since it's an invalid call

core/vm/contracts.go
```diff
@@ -1132,3 +1132,38 @@ func (c *getVerifiedSealBitmap) Run(input []byte, caller common.Address, evm *EV
...
+// cip20HashFunctions is a precompile to compute any of several
+// cryprographic hash functions
+type cip20HashFunctions struct{}
+
+func (c *cip20HashFunctions) RequiredGas(input []byte) uint64 {
+	if len(input) == 0 {
+		return params.InvalidCip20Gas
+	}
+
+	if h, ok := Cip20HashesDonut[input[0]]; ok {
```
mcortesi, line 1145: we have a static reference to `Cip20HashesDonut` this can be easily fixed by adding a field to `struct cip20HashFunctions` which is the hashes map. And so in line 195 you set the donut hashes, for future hardforks we change that in the precompilesForXXX function

go.sum
```diff
@@ -70,6 +70,10 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:ma
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/dchest/blake2s v1.0.0 h1:gHCBR8ecSImY/Nwk7X0Q2KJAJcpI/HSkUAQDi8MCP4Q=
```
mcortesi, line 73: we are not using any of theses dependencies. It seems we've just copied them into our project

core/vm/cip20.go
prestwich: yeah, the overloading of word created this ugly patch in the spec. Initially I assumed the EVM authors meant hash function word when they actually meant EVM word. If we're going to redo the specification, I would suggest pricing everything by 32-byte words, but leaving prices unchanged

core/vm/contracts.go
prestwich: ah this is a bug yes. I will add a seperate block that checks for the 0 length and returns an error.

go.sum
prestwich: ran `go mod tidy`

mcortesi (approved):
Thanks!

Approved but i would check @nategraf  comment first

core/vm/cip20.go
nategraf: That would make sense. It would give a consistent and way to talk about the gas pricing. By "leave the prices unchanged" do you mean have the same cost/word with reduced word size (i.e. double the cost per byte for SHA3-512) or do you mean to keep the same cost/byte? I would be most supportive of either making the cost/byte match of SHA3-512 match SHA3-256, or be twice as large, because it seems to require twice as many sponge function invocations on average.

prestwich: my opinion is that all hash functions are severely overpriced, so I am biased towards leaving it the same as sha3-256. Because this is a precompile we have the additional CALL gas overhead that prevents abuse as well.

I have updated the PR to standardize the usage of "word" as an EVM word

nategraf (approved):
Looks good to me

prestwich:
I have just rebased this onto main, and run `go get golang.org/x/mobile@973feb4309de`.

PR #1256, opened 2020-12-07
CIP-25: Add ed25519 precompile

mkaczanowski:
### Description

This PR includes the implementation of [CIP-25](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0025.md), to perform fast and cheap verification of Ed25519 cryptographic signatures in smart contracts in general by adding a precompiled contract for Ed25519 signature verification to the EVM.

With an Ed25519 precompile, we are able to verify Ed25519 signatures origination from external, Ed25519 proof-of-stake based blockchains within Celo EVM smart contracts, allowing for proof-based trustless bridges between Celo and other blockchains.

### Other changes

None

### Tested

The change has been tested via:
* unittests
* integration tests via cURL and example solidity contract ([read more](https://github.com/celo-org/celo-blockchain/compare/master...mkaczanowski:cip25_add_ed25519_precompile_pr_ready?expand=1#commitcomment-44892851))
* benchmarks ([read more](https://github.com/celo-org/celo-blockchain/compare/master...mkaczanowski:cip25_add_ed25519_precompile_pr_ready?expand=1#commitcomment-44892809))

UPDATE: ^^ github somewhat removed the commit comments, so I re-created them into this [gist](https://gist.github.com/mkaczanowski/1db9a92b1dc99c172358c6276ca63806)

### Backwards compatibility

This PR adds a new (currently non-existing) precompile, so this change is not backward compatible.


prestwich:
hey @mkaczanowski can you do the following:

1. rebase onto latest master
2. Remove CIP25 from the Istanbul Precompile block
3. Fill in the row in the Donut Precompile block

mkaczanowski:
@prestwich thanks. I've just rebased the PR:
1. I've wrongly synced with upstream at first, so other people got subscribed (it tried to push other ~120'ish commits), sorry for noise (now it's all good, I fixed it)
2. The unittests passes
3. The benchmarks are failing, because ```benchmarkPrecompiled``` is not switched to donut yet (ie ```PrecompiledContracts{Istanbul/Donut}[common.HexToAddress(addr)]```)

please let me know if there is anything else I should do, thanks.

prestwich:
> The benchmarks are failing, because benchmarkPrecompiled is not switched to donut yet (ie PrecompiledContracts{Istanbul/Donut}[common.HexToAddress(addr)])

if you rebase onto master again, it should work.

mkaczanowski:
the change seems to be reverted:
> https://github.com/celo-org/celo-blockchain/commit/1fdcea735970c9eebdfbed357285aadcd033efeb

(no worries, this can wait. just noting this in case anyone's wondering why I am not rebasing :) )

core/vm/contracts.go
```diff
@@ -896,6 +897,40 @@ func (c *blake2F) Run(input []byte, caller common.Address, evm *EVM, gas uint64)
...
+	const sha2_512WordLength = 64
+
+	// round up to next whole word
+	lengthCeil := len(input) + sha2_512WordLength - 1
+	words := uint64(lengthCeil / sha2_512WordLength)
+	return params.Ed25519VerifyGas + params.Sha2_512BaseGas + (words * params.Sha2_512PerWordGas)
+}
+
+func (c *ed25519Verify) Run(input []byte, caller common.Address, evm *EVM, gas uint64) ([]byte, uint64, error) {
+	// Setup success/failure return values
```
trianglesphere, line 914: It looks like other precompiles are debiting gas while this precompile is not. Example from `blake2F`
```
	if err != nil {
		return nil, gas, err
	}
```

trianglesphere (changes requested):
It looks like gas isn't being paid, otherwise this looks good to me.

core/vm/contracts.go
mkaczanowski: @trianglesphere thanks, somewhat missed that.

Updated the repo accordingly

trianglesphere (approved):
lgtm

nategraf (approved):
Because this reference the same constants and formulas as CIP20 for gas calculation for SHA2-512, it would be nice if this were rebased onto https://github.com/celo-org/celo-blockchain/pull/1249.

Everything looks good to me.

core/vm/contracts.go
```diff
@@ -896,6 +897,45 @@ func (c *blake2F) Run(input []byte, caller common.Address, evm *EVM, gas uint64)
...
+	if err != nil {
+		return nil, gas, err
+	}
+
+	// Setup success/failure return values
+	var fail32byte, success32Byte = true32Byte, false32Byte
+
+	// Check if all required arguments are present
+	if len(input) < 96 {
+		return fail32byte, gas, nil
```
mcortesi, line 924: shouldn't we return an error here?

mcortesi (commented):
nit comment

core/vm/contracts.go
trianglesphere: This is the same thing as in CIP-20, I think returning an error here aborts the TX. From the spec, it seems like invalid inputs are fine and you instead return an error inside the evm. It seems like there's a little more usage of returning and out of bound error, but given that the gas metering has to work for any input, I'm pretty fine with this return convention.

prestwich: returning an error here will revert the precompile execution and the contract that called it will receive a signal that it reverted. The CIP authors decided to return a failure code instead of reverting with an error. This way, future failure codes can be added to signal precise reasons the validation failed, without being married to any solidity-level error handling system. I would think that the `UsingPrecompiles` interface to this would (by default) revert on failure

PR #1270, opened 2020-12-15
Cleanup validator uptime computation

mcortesi:
### Description

Several people had trouble following current uptime scoring logic. It looks like a pool for off by one errors. This PR attempts to do some cleanup to make it more maintainable.

* Extracts the code to its own package where before code was distributed in 3+ files.
* Add Tests 
* Add godocs to functions
* Rename functions/variables to make code easier to understand
* Introduce the type `Window` which simplifies reasoning about UP blocks
* Modify the monitoringWindow and updateUptime functions so they work with the block whose signatures they intend to monitor, vs the block whose parent block signatures we want to monitor.

We can now summarize uptime computation as:

  * `monitoringWindow(epoch):` Range of blocks from an epoch where UP status will be monitor. Represents also the maximum number of blocks within an epoch a validator can be considered UP
  * `currentLookbackWindow(blockNumber)` Current range of blocks we consider to label a validator a UP. A valiadator is considered UP when it has sign at least a block in the currentLookbackWindow
  * `validator.UpBlocks` total number of blocks within an epoch a validator has been labeled as UP
 
Then the score is:
```
uptimeScore = validator.UpBlocks / monitoringWindow(epoch).Size()
```

Finally, for each block we process, the `LastSignedBlock` for each validator is updated, and then `validator.UPBlocks` is increased if: `monitoringWindow.contains(blockNumber) && currentLookbackWindow.contains(lastSignedBlock)`

### Other changes

* Improve epoch related functions
* Add godocs to non related functions

### Tested

* Unit tests

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/uptime/store/store.go
+++ b/consensus/istanbul/uptime/store/store.go
@@ -1,0 +1,24 @@
+package store
+
+import (
+	"github.com/ethereum/go-ethereum/consensus/istanbul/uptime"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+type uptimeStoreImpl struct {
+	db ethdb.Database
+}
+
+func New(db ethdb.Database) uptime.Store {
+	return &uptimeStoreImpl{
+		db: db,
+	}
+}
+
+func (us *uptimeStoreImpl) ReadAccumulatedEpochUptime(epoch uint64) *uptime.Uptime {
+	return rawdb.ReadAccumulatedEpochUptime(us.db, epoch)
+}
+func (us *uptimeStoreImpl) WriteAccumulatedEpochUptime(epoch uint64, uptime *uptime.Uptime) {
+	rawdb.WriteAccumulatedEpochUptime(us.db, epoch, uptime)
+}
--- a/consensus/istanbul/uptime/uptime.go
+++ b/consensus/istanbul/uptime/uptime.go
@@ -1,0 +1,242 @@
+package uptime
+
+import (
+	"errors"
+	"fmt"
+	"math"
+
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/consensus/istanbul"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+// Store provides a persistent storage for uptime entries
+type Store interface {
+	ReadAccumulatedEpochUptime(epoch uint64) *Uptime
+	WriteAccumulatedEpochUptime(epoch uint64, uptime *Uptime)
+}
+
+// Uptime contains the latest block for which uptime metrics were accounted for. It also contains
+// an array of Entries where the `i`th entry represents the uptime statistics of the `i`th validator
+// in the validator set for that epoch
+type Uptime struct {
+	LatestBlock uint64
+	Entries     []UptimeEntry
+}
+
+// UptimeEntry contains the uptime score of a validator during an epoch as well as the
+// last block they signed on
+type UptimeEntry struct {
+	ScoreTally      uint64
+	LastSignedBlock uint64
+}
+
+func (u *UptimeEntry) String() string {
+	return fmt.Sprintf("UptimeEntry { scoreTally: %v, lastBlock: %v}", u.ScoreTally, u.LastSignedBlock)
+}
+
+// GetUptimeMonitoringWindow retrieves the range [first block, last block] where uptime is to be monitored
+// for a give epoch. The range is inclusive.
+// First blocks of an epoch need to be skipped since we can't assess the last `lookbackWindow` block for validators
+// as those are froma different epoch.
+// Similarly, last block of epoch is skipped since we can't obtaine the signer for it; as they are in the next block
+func GetUptimeMonitoringWindow(epochNumber uint64, epochSize uint64, lookbackWindowSize uint64) (uint64, uint64) {
+	if epochNumber == 0 {
+		panic("no monitoring window for epoch 0")
+	}
+
+	epochFirstBlock, _ := istanbul.GetEpochFirstBlockNumber(epochNumber, epochSize)
+	epochLastBlock := istanbul.GetEpochLastBlockNumber(epochNumber, epochSize)
+
+	// first block to monitor:
+	// We need to wait for the completion of the first window with the start window's block being the
+	// 2nd block of the epoch, before we start tallying the validator score for epoch "epochNumber".
+	// We can't include the epoch's first block since it's aggregated parent seals
+	// is for the previous epoch's valset.
+	firstBlockToMonitor := epochFirstBlock + 1 + (lookbackWindowSize - 1)
+
+	// last block to monitor:
+	// We stop tallying for epoch "epochNumber" at the second to last block of that epoch.
+	// We can't include that epoch's last block as part of the tally because the epoch val score is calculated
+	// using a tally that is updated AFTER a block is finalized.
+	// Note that it's possible to count up to the last block of the epoch, but it's much harder to implement
+	// than couting up to the second to last one.
+	lastBlockToMonitor := epochLastBlock - 1
+
+	return firstBlockToMonitor, lastBlockToMonitor
+}
+
+// Monitor is responsible for monitoring uptime
+// by processing blocks
+type Monitor struct {
+	epochSize      uint64
+	lookbackWindow uint64
+
+	logger log.Logger
+	store  Store
+}
+
+// NewMonitor creates a new uptime monitor
+func NewMonitor(store Store, epochSize, lookbackWindow uint64) *Monitor {
+	return &Monitor{
+		epochSize:      epochSize,
+		lookbackWindow: lookbackWindow,
+		store:          store,
+		logger:         log.New("module", "uptime-monitor"),
+	}
+}
+
+// MonitoringWindow returns the monitoring window for the given epoch in the format
+// [firstBlock, lastBlock] both inclusive
+func (um *Monitor) MonitoringWindow(epoch uint64) (uint64, uint64) {
+	return GetUptimeMonitoringWindow(epoch, um.epochSize, um.lookbackWindow)
+}
+
+// MonitoringWindowSize returns the size (in blocks) of the monitoring window
+func (um *Monitor) MonitoringWindowSize(epoch uint64) uint64 {
+	start, end := um.MonitoringWindow(epoch)
+	return end - start + 1
+}
+
+// ComputeValidatorsUptime retrieves the uptime score for each validator for a given epoch
+func (um *Monitor) ComputeValidatorsUptime(epoch uint64, valSetSize int) ([]*big.Int, error) {
+	logger := um.logger.New("func", "Backend.updateValidatorScores", "epoch", epoch)
+	logger.Trace("Updating validator scores")
+
+	// The denominator is the (last block - first block + 1) of the val score tally window
+	denominator := um.MonitoringWindowSize(epoch)
+
+	uptimes := make([]*big.Int, 0, valSetSize)
+	accumulated := um.store.ReadAccumulatedEpochUptime(epoch)
+
+	if accumulated == nil {
+		err := errors.New("Accumulated uptimes not found, cannot update validator scores")
+		logger.Error(err.Error())
+		return nil, err
+	}
+
+	for i, entry := range accumulated.Entries {
+		if i >= valSetSize {
+			break
+		}
+
+		if entry.ScoreTally > denominator {
+			logger.Error("ScoreTally exceeds max possible", "scoreTally", entry.ScoreTally, "denominator", denominator, "valIdx", i)
+			uptimes = append(uptimes, params.Fixidity1)
+			continue
+		}
+
+		numerator := big.NewInt(0).Mul(big.NewInt(int64(entry.ScoreTally)), params.Fixidity1)
+		uptimes = append(uptimes, big.NewInt(0).Div(numerator, big.NewInt(int64(denominator))))
+	}
+
+	if len(uptimes) < valSetSize {
+		err := fmt.Errorf("%d accumulated uptimes found, cannot update validator scores", len(uptimes))
+		logger.Error(err.Error())
+		return nil, err
+	}
+
+	return uptimes, nil
+}
+
+// ProcessBlock will monitor uptimeness information on the give block and add it to current epoch tally
+func (um *Monitor) ProcessBlock(block *types.Block) error {
+	// The epoch's first block's aggregated parent signatures is for the previous epoch's valset.
+	// We can ignore updating the tally for that block.
+	if istanbul.IsFirstBlockOfEpoch(block.NumberU64(), um.epochSize) {
+		return nil
+	}
+
+	// Get the bitmap from the previous block
+	extra, err := types.ExtractIstanbulExtra(block.Header())
+	if err != nil {
+		um.logger.Error("Unable to extract istanbul extra", "func", "ProcessBlock", "blocknum", block.NumberU64())
+		return errors.New("could not extract block header extra")
+	}
+	signedValidatorsBitmap := extra.ParentAggregatedSeal.Bitmap
+
+	// Get the uptime scores
+	epochNum := istanbul.GetEpochNumber(block.NumberU64(), um.epochSize)
+	uptime := um.store.ReadAccumulatedEpochUptime(epochNum)
+
+	// We only update the uptime for blocks which are greater than the last block we saw.
+	// This ensures that we do not count the same block twice for any reason.
+	if uptime == nil || uptime.LatestBlock < block.NumberU64() {
+		uptime = updateUptime(uptime, block.NumberU64(), signedValidatorsBitmap, um.lookbackWindow, epochNum, um.epochSize)
+		um.store.WriteAccumulatedEpochUptime(epochNum, uptime)
+	} else {
+		log.Trace("WritingBlockWithState with block number less than a block we previously wrote", "latestUptimeBlock", uptime.LatestBlock, "blockNumber", block.NumberU64())
+	}
+
+	return nil
+}
+
+// updateUptime receives the currently accumulated uptime for all validators in an epoch and proceeds to update it
+// based on which validators signed on the provided block by inspecting the block's parent bitmap
+//
+// It expects a blockNumber with the signatures bitmap from the previous block (parent seal)
+func updateUptime(uptime *Uptime, blockNumber uint64, bitmap *big.Int, window uint64, epochNum uint64, epochSize uint64) *Uptime {
+	if uptime == nil {
+		uptime = new(Uptime)
+		// The number of validators is upper bounded by 3/2 of the number of 1s in the bitmap
+		// We multiply by 2 just to be extra cautious of off-by-one errors.
+		validatorsSizeUpperBound := uint64(math.Ceil(float64(bitCount(bitmap)) * 2))
+		uptime.Entries = make([]UptimeEntry, validatorsSizeUpperBound)
+	}
+
+	firstBlockToMonitor, lastBlockToMonitor := GetUptimeMonitoringWindow(epochNum, epochSize, window)
+
+	// lookbackBlockNum := (blockNumber -1) - window
+
+	// signedBlockWindowLastBlockNum is just the previous block
+	signedBlockWindowLastBlockNum := blockNumber - 1
+	signedBlockWindowFirstBlockNum := signedBlockWindowLastBlockNum - (window - 1)
+
+	uptime.LatestBlock = blockNumber
+	for i := 0; i < len(uptime.Entries); i++ {
+		if bitmap.Bit(i) == 1 {
+			// update their latest signed block
+			uptime.Entries[i].LastSignedBlock = blockNumber - 1
+		}
+
+		// If we are within the validator uptime tally window, then update the validator's score
+		// if its last signed block is within the lookback window
+		if blockNumber >= firstBlockToMonitor && blockNumber <= lastBlockToMonitor {
+			lastSignedBlock := uptime.Entries[i].LastSignedBlock
+
+			// Note that the second condition in the if condition is not necessary.  But it does
+			// make the logic easier to understand.  (e.g. it's checking is lastSignedBlock is within
+			// the range [signedBlockWindowFirstBlockNum, signedBlockWindowLastBlockNum])
+			if lastSignedBlock >= signedBlockWindowFirstBlockNum && lastSignedBlock <= signedBlockWindowLastBlockNum {
+				uptime.Entries[i].ScoreTally++
+			}
+		}
+	}
+	return uptime
+}
+
+// https://stackoverflow.com/questions/19105791/is-there-a-big-bitcount/32702348#32702348
+func bitCount(n *big.Int) int {
+	count := 0
+	for _, v := range n.Bits() {
+		count += popcount(uint64(v))
+	}
+	return count
+}
+
+// Straight and simple C to Go translation from https://en.wikipedia.org/wiki/Hamming_weight
+func popcount(x uint64) int {
+	const (
+		m1  = 0x5555555555555555 //binary: 0101...
+		m2  = 0x3333333333333333 //binary: 00110011..
+		m4  = 0x0f0f0f0f0f0f0f0f //binary:  4 zeros,  4 ones ...
+		h01 = 0x0101010101010101 //the sum of 256 to the power of 0,1,2,3...
+	)
+	x -= (x >> 1) & m1             //put count of each 2 bits into those 2 bits
+	x = (x & m2) + ((x >> 2) & m2) //put count of each 4 bits into those 4 bits
+	x = (x + (x >> 4)) & m4        //put count of each 8 bits into those 8 bits
+	return int((x * h01) >> 56)    //returns left 8 bits of x + (x<<8) + (x<<16) + (x<<24) + ...
+}
--- a/consensus/istanbul/uptime/uptime.go
+++ b/consensus/istanbul/uptime/uptime.go
@@ -19,7 +19,7 @@ type Store interface {
 	WriteAccumulatedEpochUptime(epoch uint64, uptime *Uptime)
 }
 
-// Uptime contains the latest block for which uptime metrics were accounted for. It also contains
+// Uptime contains the latest block for which uptime metrics were accounted. It also contains
 // an array of Entries where the `i`th entry represents the uptime statistics of the `i`th validator
 // in the validator set for that epoch
 type Uptime struct {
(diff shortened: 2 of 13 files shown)


consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (u *UptimeEntry) String() string {
+	return fmt.Sprintf("UptimeEntry { scoreTally: %v, lastBlock: %v}", u.ScoreTally, u.LastSignedBlock)
+}
+
+// GetUptimeMonitoringWindow retrieves the range [first block, last block] where uptime is to be monitored
+// for a give epoch. The range is inclusive.
+// First blocks of an epoch need to be skipped since we can't assess the last `lookbackWindow` block for validators
+// as those are froma different epoch.
+// Similarly, last block of epoch is skipped since we can't obtaine the signer for it; as they are in the next block
```
prestwich, line 45: Nit: `obtain`

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func (u *UptimeEntry) String() string {
+	return fmt.Sprintf("UptimeEntry { scoreTally: %v, lastBlock: %v}", u.ScoreTally, u.LastSignedBlock)
+}
+
+// GetUptimeMonitoringWindow retrieves the range [first block, last block] where uptime is to be monitored
+// for a give epoch. The range is inclusive.
+// First blocks of an epoch need to be skipped since we can't assess the last `lookbackWindow` block for validators
+// as those are froma different epoch.
```
prestwich, line 44: this might be clearer with something like: "we do not monitor while the lookback window crosses the epoch boundary."

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+	// We can't include that epoch's last block as part of the tally because the epoch val score is calculated
+	// using a tally that is updated AFTER a block is finalized.
+	// Note that it's possible to count up to the last block of the epoch, but it's much harder to implement
+	// than couting up to the second to last one.
+	lastBlockToMonitor := epochLastBlock - 1
+
+	return firstBlockToMonitor, lastBlockToMonitor
+}
+
+// Monitor is responsible for monitoring uptime
```
prestwich, line 72: nit: remove line break?

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+	start, end := um.MonitoringWindow(epoch)
+	return end - start + 1
+}
+
+// ComputeValidatorsUptime retrieves the uptime score for each validator for a given epoch
+func (um *Monitor) ComputeValidatorsUptime(epoch uint64, valSetSize int) ([]*big.Int, error) {
+	logger := um.logger.New("func", "Backend.updateValidatorScores", "epoch", epoch)
+	logger.Trace("Updating validator scores")
+
+	// The denominator is the (last block - first block + 1) of the val score tally window
```
prestwich, line 109: Might be clearer if we explicitly say this is the maximum possible uptime score

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+		validatorsSizeUpperBound := uint64(math.Ceil(float64(bitCount(bitmap)) * 2))
+		uptime.Entries = make([]UptimeEntry, validatorsSizeUpperBound)
+	}
+
+	firstBlockToMonitor, lastBlockToMonitor := GetUptimeMonitoringWindow(epochNum, epochSize, window)
+
+	// lookbackBlockNum := (blockNumber -1) - window
+
+	// signedBlockWindowLastBlockNum is just the previous block
+	signedBlockWindowLastBlockNum := blockNumber - 1
```
prestwich, line 195: while we're renaming things, would this be clearer as `lookbackWindowLastBlockNum` and `lookbackWindowLastBlockNum`

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+		uptime = new(Uptime)
+		// The number of validators is upper bounded by 3/2 of the number of 1s in the bitmap
+		// We multiply by 2 just to be extra cautious of off-by-one errors.
+		validatorsSizeUpperBound := uint64(math.Ceil(float64(bitCount(bitmap)) * 2))
+		uptime.Entries = make([]UptimeEntry, validatorsSizeUpperBound)
+	}
+
+	firstBlockToMonitor, lastBlockToMonitor := GetUptimeMonitoringWindow(epochNum, epochSize, window)
+
+	// lookbackBlockNum := (blockNumber -1) - window
```
oneeman, line 192: this comment seems out of place

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+// updateUptime receives the currently accumulated uptime for all validators in an epoch and proceeds to update it
+// based on which validators signed on the provided block by inspecting the block's parent bitmap
+//
+// It expects a blockNumber with the signatures bitmap from the previous block (parent seal)
+func updateUptime(uptime *Uptime, blockNumber uint64, bitmap *big.Int, window uint64, epochNum uint64, epochSize uint64) *Uptime {
+	if uptime == nil {
+		uptime = new(Uptime)
+		// The number of validators is upper bounded by 3/2 of the number of 1s in the bitmap
+		// We multiply by 2 just to be extra cautious of off-by-one errors.
+		validatorsSizeUpperBound := uint64(math.Ceil(float64(bitCount(bitmap)) * 2))
```
oneeman, line 186: My understanding is that uptime will be nil the first time this is called in a new epoch.  If that is so, would it be possible to get the epoch's validator set size and use that instead of this upper bound?  Or could we have a separate function, called at the beginning of an epoch, initializing Uptime for that epoch with the right size for Entries?  I believe the validator set size can't increase inside an epoch, is that right?  But can it decrease?  I don't see this as a big problem, but it just doesn't feel nice to use this upper bound instead of the actual validator set size.

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+	if len(uptimes) < valSetSize {
+		err := fmt.Errorf("%d accumulated uptimes found, cannot update validator scores", len(uptimes))
+		logger.Error(err.Error())
+		return nil, err
+	}
+
+	return uptimes, nil
+}
+
+// ProcessBlock will monitor uptimeness information on the give block and add it to current epoch tally
```
oneeman, line 145: I'd rephrase this comment as:
```
// ProcessBlock uses the block's signature bitmap (which encodes who signed the parent block) to update the epoch's Uptime data
```

If I'm confused about the bitmap being about the parent block, then modify accordingly.

consensus/istanbul/uptime/store/store.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/ethereum/go-ethereum/consensus/istanbul/uptime"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+type uptimeStoreImpl struct {
+	db ethdb.Database
+}
+
+func New(db ethdb.Database) uptime.Store {
```
oneeman, line 13: Since this `New()` is only used inline inside calls to `uptime.NewMonitor()`, would it simplify things to have `NewMonitor()` take in the `db` and call this `New()` itself?  Or maybe consider removing `uptimeStoreImpl` entirely?

consensus/istanbul/utils.go
```diff
@@ -67,72 +67,56 @@ func CheckValidatorSignature(valSet ValidatorSet, data []byte, sig []byte) (comm
...
 
-// Retrieves the block number within an epoch.  The return value will be 1-based.
-// There is a special case if the number == 0.  It is basically the last block of the 0th epoch, and should have a value of epochSize
+// GetNumberWithinEpoch retrieves the block number within an epoch.
+// The return value will be 1-based; thus first block of epoch is 1, and last block of epoch is `epochSize`
+// There is a special case if the number == 0. It is basically the last block of the 0th epoch, and should have a value of epochSize
 func GetNumberWithinEpoch(number uint64, epochSize uint64) uint64 {
 	number = number % epochSize
 	if number == 0 {
 		return epochSize
```
oneeman, line 76: This feels super-wrong, to say (even as a special case) that the genesis block is that epoch's 17280th block.  But I don't think it'd be easy to change this without making a bunch of changes elsewhere, so maybe we can let sleeping dogs lie on this one.

consensus/istanbul/utils.go
```diff
@@ -67,72 +67,56 @@ func CheckValidatorSignature(valSet ValidatorSet, data []byte, sig []byte) (comm
...
 }
 
-// Retrieves the epoch number given the block number.
-// There is a special case if the number == 0 (the genesis block).  That block will be in the
-// 1st epoch.
+// GetEpochNumber retrieves the epoch number given the block number.
+// Epoch 0 is a special block that only contains the genesis block (block 0), epoch 1
+// starts at block 1
 func GetEpochNumber(number uint64, epochSize uint64) uint64 {
 	epochNumber := number / epochSize
```
oneeman, line 95: Maybe rename `epochNumber` to `epochs` or `pastEpochs` or something else.  It just seems strange to have a function `GetEpochNumber()` with a return that says `epochNumber + 1`

consensus/istanbul/uptime/uptime_test.go
```diff
@@ -0,0 +1,124 @@
...
+	tests := []struct {
+		name  string
+		args  args
+		want  uint64
+		want1 uint64
+	}{
+		{"tally on first epoch", args{1, 10, 2}, 1 + 2, 9},
+		{"tally on second epoch", args{2, 10, 2}, 11 + 2, 19},
+		{"lookback window too big", args{1, 10, 10}, 11, 9},
+		// TODO: Add test cases.
```
oneeman, line 24: Is this TODO intended to be completed before finalizing the PR?  If not, I would remove it.

consensus/istanbul/uptime/uptime_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	tests := []struct {
+		name  string
+		args  args
+		want  uint64
+		want1 uint64
+	}{
+		{"tally on first epoch", args{1, 10, 2}, 1 + 2, 9},
+		{"tally on second epoch", args{2, 10, 2}, 11 + 2, 19},
+		{"lookback window too big", args{1, 10, 10}, 11, 9},
```
oneeman, line 23: I'm skeptical about including a case with an invalid combination of lookback window size and epoch size.  Is this a case that we are supposed to support or that should lead to a fatal error somewhere?  If it's the latter, we should test elsewhere does it does lead to a fatal error, and not test it here.

consensus/istanbul/uptime/uptime_test.go
```diff
@@ -0,0 +1,124 @@
...
+		epochSize          uint64
+		lookbackWindowSize uint64
+	}
+	tests := []struct {
+		name  string
+		args  args
+		want  uint64
+		want1 uint64
+	}{
+		{"tally on first epoch", args{1, 10, 2}, 1 + 2, 9},
```
oneeman, line 21: Replace "tally" with "monitoringWindow" or something similar.

Also, I don't like the `1 + 2` and `11 + 2`, instead of `3` and `13`.  Or at least it should be consistent and instead of `9` and `19` is should be `10 - 1` and `20 - 1`) (though I don't like that either).

oneeman (changes requested):
Definitely improvement, but I think there's further we can go while we're at it.  Added a bunch of comments, you can decide which of them make sense.

consensus/istanbul/uptime/uptime.go
mcortesi: Changed to:
```
// We do not monitor while the lookback window crosses the epoch boundary; since we can only analyze window when the window
// (block range) belongs to a single block (thus initial epoch blocks are skipped)
```

consensus/istanbul/uptime/uptime.go
mcortesi: agree that `denominator` doesn't really inform anything.

It's not the max uptime score (since score it's a ratio), but the total blocks that we are monitoring

So, what about:
```
totalMonitoredBlocks

File: consensus/istanbul/utils.go (hunk 10 above)
Comment (mcortesi) [reply consensus/istanbul/utils.go:76]: agree it feels wrong, but i wouldn't dare changing this now

File: consensus/istanbul/utils.go (hunk 11 above)
Comment (mcortesi) [reply consensus/istanbul/utils.go:95]: Not sure i agree...
`epochNumber` is the result variable; so it's proper to named as this.

What could clarify this is do:
```
epochNumber = epochNumber +1
return epochNumber
```

File: consensus/istanbul/uptime/uptime.go (hunk 7 above)
Comment (mcortesi) [reply consensus/istanbul/uptime/uptime.go:185]: agree that this feels strange. I didn't ventured into changing this here, and probably won't to avoid problems, but definitely an improvment to make

File: consensus/istanbul/utils.go (hunk 11 above)
Comment (oneeman) [reply consensus/istanbul/utils.go:95]: That would be better.  We could also name the return variable:
```
func GetEpochNumber(number uint64, epochSize uint64) (epochNumber uint64) {
	epochNumber = number / epochSize
	if !IsLastBlockOfEpoch(number, epochSize) {
		epochNumber += 1
	}
}
```

There are other ways to make the special case of the last block more explicit (including with comments), but most of them don't necessarily make the code better in the end.  I almost think the most self-explanatory and readable form is when you duplicate the division:

```
func GetEpochNumber(number uint64, epochSize uint64) uint64 {
	if IsLastBlockOfEpoch(number, epochSize) {
		return number / epochSize
	} else {
		return number / epochSize + 1
	}
}
```

Comment (oneeman) [reply consensus/istanbul/utils.go:95]: But probably this is just hair-splitting...

File: consensus/istanbul/uptime/uptime.go [hunk 15]
```diff
@@ -0,0 +1,242 @@
...
+	// We can't include the epoch's first block since it's aggregated parent seals
+	// is for the previous epoch's valset.
+	firstBlockToMonitor := epochFirstBlock + 1 + (lookbackWindowSize - 1)
+
+	// last block to monitor:
+	// We stop tallying for epoch "epochNumber" at the second to last block of that epoch.
+	// We can't include that epoch's last block as part of the tally because the epoch val score is calculated
+	// using a tally that is updated AFTER a block is finalized.
+	// Note that it's possible to count up to the last block of the epoch, but it's much harder to implement
+	// than couting up to the second to last one.
```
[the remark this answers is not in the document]
mcortesi: new comment now

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+
+	epochFirstBlock, _ := istanbul.GetEpochFirstBlockNumber(epochNumber, epochSize)
+	epochLastBlock := istanbul.GetEpochLastBlockNumber(epochNumber, epochSize)
+
+	// first block to monitor:
+	// We need to wait for the completion of the first window with the start window's block being the
+	// 2nd block of the epoch, before we start tallying the validator score for epoch "epochNumber".
+	// We can't include the epoch's first block since it's aggregated parent seals
+	// is for the previous epoch's valset.
+	firstBlockToMonitor := epochFirstBlock + 1 + (lookbackWindowSize - 1)
```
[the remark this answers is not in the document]
mcortesi: it's now `firstBlockToMonitor := newWindowStartingOn(epochFirstBlock, lookbackWindowSize).End` since the first valid block is the last block of the window starting at the beginning of the epoch

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+		epochSize:      epochSize,
+		lookbackWindow: lookbackWindow,
+		store:          store,
+		logger:         log.New("module", "uptime-monitor"),
+	}
+}
+
+// MonitoringWindow returns the monitoring window for the given epoch in the format
+// [firstBlock, lastBlock] both inclusive
+func (um *Monitor) MonitoringWindow(epoch uint64) (uint64, uint64) {
```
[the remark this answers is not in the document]
mcortesi: refactored everything to use a window type

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+		uptime.Entries = make([]UptimeEntry, validatorsSizeUpperBound)
+	}
+
+	firstBlockToMonitor, lastBlockToMonitor := GetUptimeMonitoringWindow(epochNum, epochSize, window)
+
+	// lookbackBlockNum := (blockNumber -1) - window
+
+	// signedBlockWindowLastBlockNum is just the previous block
+	signedBlockWindowLastBlockNum := blockNumber - 1
+	signedBlockWindowFirstBlockNum := signedBlockWindowLastBlockNum - (window - 1)
```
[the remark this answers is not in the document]
mcortesi: using a window now

consensus/istanbul/uptime/uptime.go
```diff
@@ -173,11 +166,8 @@ func (um *Monitor) ProcessBlock(block *types.Block) error {
 	return nil
 }
 
-// updateUptime receives the currently accumulated uptime for all validators in an epoch and proceeds to update it
-// based on which validators signed on the provided block by inspecting the block's parent bitmap
-//
-// It expects a blockNumber with the signatures bitmap from the previous block (parent seal)
-func updateUptime(uptime *Uptime, blockNumber uint64, bitmap *big.Int, window uint64, epochNum uint64, epochSize uint64) *Uptime {
+// updateUptime update accumulated uptime given a block and its validator's signatures bitmap
+func updateUptime(uptime *Uptime, blockNumber uint64, bitmap *big.Int, lookbackWindowSize uint64, monitoredWindow Window) *Uptime {
```
oneeman, line 170: Maybe rename the argument to `monitoringWindow` to be consistent?  I think `monitored` is fine when it comes to blocks within the window (as below)

consensus/istanbul/uptime/uptime.go
```diff
@@ -173,11 +166,8 @@ func (um *Monitor) ProcessBlock(block *types.Block) error {
 	return nil
 }
 
-// updateUptime receives the currently accumulated uptime for all validators in an epoch and proceeds to update it
-// based on which validators signed on the provided block by inspecting the block's parent bitmap
-//
-// It expects a blockNumber with the signatures bitmap from the previous block (parent seal)
-func updateUptime(uptime *Uptime, blockNumber uint64, bitmap *big.Int, window uint64, epochNum uint64, epochSize uint64) *Uptime {
+// updateUptime update accumulated uptime given a block and its validator's signatures bitmap
```
oneeman, line 169: update -> updates the

consensus/istanbul/uptime/uptime.go
```diff
@@ -30,20 +30,21 @@ type Uptime struct {
...
 
 func (u *UptimeEntry) String() string {
-	return fmt.Sprintf("UptimeEntry { scoreTally: %v, lastBlock: %v}", u.ScoreTally, u.LastSignedBlock)
+	return fmt.Sprintf("UptimeEntry { upBlocks: %v, lastBlock: %v}", u.UpBlocks, u.LastSignedBlock)
 }
 
-// GetUptimeMonitoringWindow retrieves the range [first block, last block] where uptime is to be monitored
-// for a give epoch. The range is inclusive.
+// MonitoringWindow retrieves the block window where uptime is to be monitored
+// for a given epoch. The range is inclusive.
```
oneeman, line 43: Could drop mention of inclusivity, since Window itself is inclusive. But fine to leave it here.

consensus/istanbul/uptime/uptime.go
```diff
@@ -30,20 +30,21 @@ type Uptime struct {
...
 // We do not monitor while the lookback window crosses the epoch boundary; since we can only analyze window when the window
 // (block range) belongs to a single block (thus initial epoch blocks are skipped)
 // Similarly, last block of epoch is skipped since we can't obtain the signer for it; as they are in the next block
```
oneeman, line 46: I would remove these three lines of comments, because the inline comments below explain it better and more correctly.

consensus/istanbul/uptime/uptime.go
```diff
@@ -52,21 +53,18 @@ func GetUptimeMonitoringWindow(epochNumber uint64, epochSize uint64, lookbackWin
...
-	// We stop tallying for epoch "epochNumber" at the second to last block of that epoch.
-	// We can't include that epoch's last block as part of the tally because the epoch val score is calculated
-	// using a tally that is updated AFTER a block is finalized.
-	// Note that it's possible to count up to the last block of the epoch, but it's much harder to implement
-	// than couting up to the second to last one.
-	lastBlockToMonitor := epochLastBlock - 1
-
-	return firstBlockToMonitor, lastBlockToMonitor
+	// Last 2 blocks from the epoch are removed from the window
+	// lastBlock     => it's parentSeal is on firstBlock of next epoch
```
oneeman, line 62: it's -> its

consensus/istanbul/uptime/uptime.go
```diff
@@ -164,7 +156,8 @@ func (um *Monitor) ProcessBlock(block *types.Block) error {
 	// We only update the uptime for blocks which are greater than the last block we saw.
 	// This ensures that we do not count the same block twice for any reason.
 	if uptime == nil || uptime.LatestBlock < block.NumberU64() {
-		uptime = updateUptime(uptime, block.NumberU64(), signedValidatorsBitmap, um.lookbackWindow, epochNum, um.epochSize)
+		uptime = updateUptime(uptime, block.NumberU64()-1, signedValidatorsBitmap, um.lookbackWindow, um.MonitoringWindow(epochNum))
+		uptime.LatestBlock = block.NumberU64()
```
oneeman, line 160: Downside of doing this is here is the tests can't cover `LatestBlock`, but I guess that's not too bad.  The reason for moving it here was because of the `-1`?

oneeman (commented):
Much improved.  Left some minor comments here and there.  Also, `TestUptimeStorage` is failing due to stray references to `ScoreTally`.  That made me wonder: since the uptime store uses the db, would renaming a field cause problems when it comes to backwards compatibility?

Also, github says there are now merge conflicts (my guess is it's since 1.9.12 is merged to master now) 😬

consensus/istanbul/uptime/uptime.go
mcortesi: The `-1` and the fact that here is here it's actually used:

```
if latestBlock < currentBlock {
  do something
  latestBlock = currentblock
}  
```

consensus/istanbul/utils.go
mcortesi: went with the last version proposed....
totally hair-splitting :P

consensus/istanbul/uptime/uptime.go
mcortesi: End up removing the comment, in favor of inline comments in the function. Check @oneeman comment

consensus/istanbul/uptime/window.go
```diff
@@ -22,14 +22,14 @@ func newWindowStartingOn(start uint64, size uint64) Window {
...
 type Window struct {
 	Start uint64
 	End   uint64
 }
 
 // Size returns the size of the window.
-// Window block ranges are inclusive
 func (w Window) Size() uint64 { return w.End - w.Start + 1 }
 
 // Contains indidicates if a block number is contained within the window
```
oneeman, line 35: indidicates -> indicates

mcortesi:
> Much improved. Left some minor comments here and there. Also, `TestUptimeStorage` is failing due to stray references to `ScoreTally`. That made me wonder: since the uptime store uses the db, would renaming a field cause problems when it comes to backwards compatibility?
> 
> Also, github says there are now merge conflicts (my guess is it's since 1.9.12 is merged to master now) 😬

Fixed the stray reference.

About the field renaming, it doesn't cause problems. since RLP (the encoding) it's unaware of the names, only the structure of the type..

here's a test I've just did to verify this
```
func TestUptimeStorageChange(t *testing.T) {
	db := NewMemoryDatabase()
	epoch := uint64(0)

	// Create a test uptime to move around the database and make sure it's really new
	if entry := ReadAccumulatedEpochUptime(db, epoch); entry != nil {
		t.Fatalf("Non existent uptime returned: %v", entry)
	}

	expectedStoredUptime := &uptime.Uptime{
		Entries: []uptime.UptimeEntry{
			{UpBlocks: 0, LastSignedBlock: 1},
			{UpBlocks: 2, LastSignedBlock: 2},
			{UpBlocks: 8, LastSignedBlock: 8},
		},
		LatestBlock: 5,
	}

	type OldUptimeEntry struct {
		// Numbers of blocks validator is considered UP within monitored window
		UpBlocks        uint64
		LastSignedBlock uint64
	}
	type OldUptime struct {
		LatestBlock uint64
		Entries     []OldUptimeEntry
	}

	// Write and verify the uptime in the database
	uptime := &OldUptime{
		Entries: []OldUptimeEntry{
			{UpBlocks: 0, LastSignedBlock: 1},
			{UpBlocks: 2, LastSignedBlock: 2},
			{UpBlocks: 8, LastSignedBlock: 8},
		},
		LatestBlock: 5,
	}

	data, err := rlp.EncodeToBytes(uptime)
	if err != nil {
		t.Fatalf("Failed to RLP encode updated uptime %v", err)
	}
	if err := db.Put(uptimeKey(epoch), data); err != nil {
		t.Fatalf("Failed to store updated uptime: %v", err)
	}

	if entry := ReadAccumulatedEpochUptime(db, epoch); entry == nil {
		t.Fatalf("Stored uptime not found")
	} else if !reflect.DeepEqual(entry, expectedStoredUptime) {
		t.Fatalf("Retrieved uptime mismatch: have %v, want %v", entry, uptime)
	}
	// Delete the uptime and verify the execution
	DeleteAccumulatedEpochUptime(db, epoch)
	if entry := ReadAccumulatedEpochUptime(db, epoch); entry != nil {
		t.Fatalf("Deleted uptime returned: %v", entry)
	}
}
```

mcortesi:
Rebased to master to fix conflicts

oneeman (approved):
LGTM, very nice!

consensus/istanbul/uptime/uptime_test.go
mcortesi: i'll review this on when finishing https://github.com/celo-org/celo-blockchain/pull/1136

consensus/istanbul/uptime/uptime.go
```diff
@@ -0,0 +1,242 @@
...
+
+	// last block to monitor:
+	// We stop tallying for epoch "epochNumber" at the second to last block of that epoch.
+	// We can't include that epoch's last block as part of the tally because the epoch val score is calculated
+	// using a tally that is updated AFTER a block is finalized.
+	// Note that it's possible to count up to the last block of the epoch, but it's much harder to implement
+	// than couting up to the second to last one.
+	lastBlockToMonitor := epochLastBlock - 1
+
+	return firstBlockToMonitor, lastBlockToMonitor
```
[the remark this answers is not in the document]
mcortesi: i'll review this on when finishing https://github.com/celo-org/celo-blockchain/pull/1136

PR #1273, opened 2020-12-15
Only report missed blocks by the local address

### LINKED ISSUE #1248: Warning message "Elected but didn't sign block" is outputted by the proxy
**Describe the bug**
The warning message "Elected by didn't sign block" is outputted by both the validator and it's proxies, when the validator missed signing those blocks, as reported by david.trustworks (discord handle).

**Expected behavior**
Only the validator should output that message.  This can cause confusion on the validator operator, thinking that the proxy is misconfigured as a validator.

**Additional context**
gocelo version: v1.2.2-beta

nategraf:
### Description

As described in #1248, the proxy currently reported about missed blocks for it's validator peer.
This is confusing and my mislead the node operator to believe that their proxy is misconfigured to
run as a validator.

In this PR, the metrics check is changed to only report the local address and not on the peer
validator address. As a result, these lines will not appear on the proxy (unless it actually _is_
misconfigured to run as a valdiator)

### Related issues

- Fixes #1248

### Backwards compatibility

Metrics about the validator will not longer be reported by the proxy. If a node operator was relying
on this in there monitoring setup, the metrics will be unavailable when they update.

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -776,13 +776,10 @@ func (sb *Backend) RefreshValPeers() error {
 }
 
 func (sb *Backend) ValidatorAddress() common.Address {
-	var localAddress common.Address
 	if sb.IsProxy() {
-		localAddress = sb.config.ProxiedValidatorAddress
-	} else {
-		localAddress = sb.Address()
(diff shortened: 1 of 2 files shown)


kevjue (approved):
LGTM!

PR #1275, opened 2020-12-15
Check if the transactor's balance covers the Tx

### LINKED ISSUE #810: Fix transaction failed to buy gas
Created issue from https://github.com/celo-org/celo-blockchain/pull/809#discussion_r366637087

So this is an error because if a transaction is being processed here, it is assumed that it was already validated by the transaction pool, in which case this should never happen. We are breaking that assumption in the `EstimateGas` function. I think the best way to handle this would be to run the transaction through transaction pool validation, either in its entirety or just to validate that the transaction can pay its fees, before applying the state transition.
Here is the `validateTx` function on `TxPool`, which you obviously can't call directly as it stands https://github.com/celo-org/celo-blockchain/blob/f86d6733274bb43377584cb9475a2fe46e17f368/core/tx_pool.go#L607
And here is the part that validates balances to ensure fees can be paid https://github.com/celo-org/celo-blockchain/blob/f86d6733274bb43377584cb9475a2fe46e17f368/core/tx_pool.go#L1201
Two courses of action would either be to export the former, or just call the latter if that turns out to be inappropriate. The call should probably occur here-ish https://github.com/celo-org/celo-blockchain/blob/f86d6733274bb43377584cb9475a2fe46e17f368/internal/ethapi/api.go#L740

If this seems onerous, please leave this log statement as is and copy the above into an issue to address later. And feel free to assign me to it.

gastonponti:
### Description

In the `DoCall` for rpc calls, when it's not estimating, check if the transactors balance can cover the whole transaction

### Related issues

- Fixes #810 

### Backwards compatibility

yes

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -819,7 +819,6 @@ func DoCall(ctx context.Context, b Backend, args CallArgs, blockNrOrHash rpc.Blo
 	// which will always be in Gold. This allows the default price to be set for the proper currency.
 	// TODO(asa): Remove this once this is handled in the Provider.
 	if gasPrice.Sign() == 0 || gasPrice.Cmp(big.NewInt(0)) == 0 {
-		// TODO(mcortesi): change SuggestGastPriceInCurrent so it doesn't return an error
 		gasPrice, err = b.SuggestPriceInCurrency(ctx, args.FeeCurrency, header, state)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


gastonponti:
Closing PR to avoid noice. Even if the case is possible, is not relevant enough, and it's going to be changed in the v1.9.13 from upstream.

The issue #810 was making reference of another problem already fixed in the PR #825

PR #1278, opened 2020-12-18
random beacon cache changes

kevjue:
### Description

The changes in this PR will make all validators have they randomness cache set before starting mining.  It does this by two ways:

1) Whenever a block is synced or finalized via the consensus engine, the node will check to see if it authored that block, and will cache the block's parent hash (which is needed to recover the block's randomness in the future).
2) Before a node starts syncing, it will check to see if its last randomness commitment (saved on the random smart contract) has a cached entry.  If not, then it will do a reverse search of the blockchain to find the last randomness's block and then cache the block's parent hash.

### Other changes

Refactored some of randomness logic.  Specifically

1) Moved generation of randomness logic from package `contract_comm/random` to `consensus/istanbul`.
2) Moved specification of randomness cache leveldb location from package `contract_comm/random` to `consensus/istanbul`.
3) Moved saving and retrieving to/from randomness cache logic from `contract_comm/random` to `core/rawdb/accessors_chain.go`.
4) Moved the randomness seed logic from `miner/worker.go` to `consensus/istanbul` package.  

### Tested

1) e2e tests (note that the `when a validator's data directory is deleted` test case in the e2e sync test will verify that restoring the randomness cache works.)
2) unit tests

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2338,44 +2338,53 @@ func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscr
 	return bc.scope.Track(bc.blockProcFeed.Subscribe(ch))
 }
 
+// RecoverRandomnessCache will do a search for the block that was used to generate the given commitment.
+// Specifically, it will find the block that this node authored and that block's parent hash is used to
+// created the commitment.  The search is a reverse iteration of the node's local chain starting at
+// the block where it's hash is the given commitmentBlockHash.
 func (bc *BlockChain) RecoverRandomnessCache(commitment common.Hash, commitmentBlockHash common.Hash) error {
-	if istEngine, isIstanbul := bc.engine.(consensus.Istanbul); isIstanbul {
+	istEngine, isIstanbul := bc.engine.(consensus.Istanbul)
+	if !isIstanbul {
+		return nil
+	}
 
-		blockHashIter := commitmentBlockHash
-		var parentHash common.Hash
-		for {
-			blockHeader := bc.GetHeaderByHash(blockHashIter)
+	log.Info("Recovering randomness cache entry", "commitment", commitment.Hex(), "initial block search", commitmentBlockHash.Hex())
 
-			// We got to the genisis block, so this goroutine didn't find the latest
-			// block authored by this validator.
-			if blockHeader.Number.Uint64() == 0 {
-				return errCommitmentNotFound
-			}
+	blockHashIter := commitmentBlockHash
+	var parentHash common.Hash
+	for {
+		blockHeader := bc.GetHeaderByHash(blockHashIter)
 
-			blockAuthor, err := istEngine.Author(blockHeader)
-			if err != nil {
-				log.Error("Error is retrieving block author", "block number", blockHeader.Number.Uint64(), "block hash", blockHeader.Hash(), "error", err)
-				return err
-			}
-
-			if blockAuthor == istEngine.ValidatorAddress() {
-				parentHash = blockHeader.ParentHash
-				break
-			}
-
-			blockHashIter = blockHeader.ParentHash
+		// We got to the genesis block, so search didn't find the latest
+		// block authored by this validator.
+		if blockHeader.Number.Uint64() == 0 {
+			return errCommitmentNotFound
 		}
 
-		// Calculate the randomness commitment
-		// The calculation is stateless (e.g. it's just a hash operation of a string), so any passed in block header and state
-		// will do. Will use the previously fetched current header and state.
-		_, randomCommitment, err := istEngine.GenerateRandomness(parentHash)
+		blockAuthor, err := istEngine.Author(blockHeader)
 		if err != nil {
-			log.Error("Couldn't generate the randomness from the parent hash", "parent hash", parentHash, "err", err)
+			log.Error("Error is retrieving block author", "block number", blockHeader.Number.Uint64(), "block hash", blockHeader.Hash(), "error", err)
 			return err
 		}
 
-		rawdb.WriteRandomCommitmentCache(bc.db, randomCommitment, parentHash)
+		if blockAuthor == istEngine.ValidatorAddress() {
+			parentHash = blockHeader.ParentHash
+			break
+		}
+
+		blockHashIter = blockHeader.ParentHash
 	}
+
+	// Calculate the randomness commitment
+	// The calculation is stateless (e.g. it's just a hash operation of a string), so any passed in block header and state
+	// will do. Will use the previously fetched current header and state.
+	_, randomCommitment, err := istEngine.GenerateRandomness(parentHash)
+	if err != nil {
+		log.Error("Couldn't generate the randomness from the parent hash", "parent hash", parentHash, "err", err)
+		return err
+	}
+
+	rawdb.WriteRandomCommitmentCache(bc.db, randomCommitment, parentHash)
+
 	return nil
 }
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1321,7 +1321,8 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 
 	// We are going to update the uptime tally.
 	// TODO find a better way of checking if it's istanbul
-	if _, isIstanbul := bc.engine.(consensus.Istanbul); isIstanbul {
+	var randomCommitment common.Hash
+	if istEngine, isIstanbul := bc.engine.(consensus.Istanbul); isIstanbul {
 
 		if hash := bc.GetCanonicalHash(block.NumberU64()); (hash != common.Hash{} && hash != block.Hash()) {
 			log.Error("Found two blocks with same height", "old", hash, "new", block.Hash())
@@ -1355,6 +1356,21 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 				log.Trace("WritingBlockWithState with block number less than a block we previously wrote", "latestUptimeBlock", uptime.LatestBlock, "blockNumber", block.NumberU64())
 			}
 		}
+
+		blockAuthor, err := istEngine.Author(block.Header())
+		if err != nil {
+			log.Warn("Unable to retrieve the author for block", "blockNum", block.NumberU64(), "err", err)
+		}
+
+		if blockAuthor == istEngine.ValidatorAddress() {
+			// Calculate the randomness and commitment
+			_, randomCommitment, err = istEngine.GenerateRandomness(block.ParentHash(), block.Header(), state)
+
+			if err != nil {
+				log.Error("Couldn't generate the randomness for the block", "blockNum", block.NumberU64(), "err", err)
+				return NonStatTy, err
+			}
+		}
 	}
 
 	// Calculate the total difficulty of the block
@@ -1376,6 +1392,9 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	rawdb.WriteBlock(blockBatch, block)
 	rawdb.WriteReceipts(blockBatch, block.Hash(), block.NumberU64(), receipts)
 	rawdb.WritePreimages(blockBatch, state.Preimages())
+	if (randomCommitment != common.Hash{}) {
+		rawdb.WriteRandomCommitmentCache(blockBatch, randomCommitment, block.ParentHash())
+	}
 	if err := blockBatch.Write(); err != nil {
 		log.Crit("Failed to write block into disk", "err", err)
 	}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1393,6 +1393,8 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	rawdb.WriteReceipts(blockBatch, block.Hash(), block.NumberU64(), receipts)
 	rawdb.WritePreimages(blockBatch, state.Preimages())
 	if (randomCommitment != common.Hash{}) {
+		// Note that the random commitment cache entry is never transferred over to the freezer,
+		// unlike all of the other saved data within this batch write
 		rawdb.WriteRandomCommitmentCache(blockBatch, randomCommitment, block.ParentHash())
 	}
 	if err := blockBatch.Write(); err != nil {
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -74,6 +74,7 @@ var (
 	blockPrefetchInterruptMeter = metrics.NewRegisteredMeter("chain/prefetch/interrupts", nil)
 
 	errInsertionInterrupted = errors.New("insertion is interrupted")
+	errCommitmentNotFound   = errors.New("randomness commitment not found")
 )
 
 const (
@@ -1364,7 +1365,7 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 
 		if blockAuthor == istEngine.ValidatorAddress() {
 			// Calculate the randomness commitment
-			_, randomCommitment, err = istEngine.GenerateRandomness(block.ParentHash(), block.Header(), state)
+			_, randomCommitment, err = istEngine.GenerateRandomness(block.ParentHash())
 
 			if err != nil {
 				log.Error("Couldn't generate the randomness for the block", "blockNum", block.NumberU64(), "err", err)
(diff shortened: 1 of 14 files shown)


miner/worker.go
```diff
@@ -867,40 +862,44 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
...
+
+		lastCommitment, err := random.GetLastCommitment(w.coinbase, w.current.header, w.current.state)
+		if err != nil {
+			log.Error("Failed to get last commitment", "err", err)
+			return
+		}
+
+		lastRandomnessParentHash := rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
+		if (lastRandomnessParentHash == common.Hash{}) {
+			log.Error("Failed to get last randomness cache entry", "err", err)
```
trianglesphere, line 878: Coupled with when we write to the cache, this change makes it impossible to use a non-validator chaindata as a backup (I suspect that a replica node might save to the randomness cache, but I'm not positive yet).

core/blockchain.go
```diff
@@ -1376,6 +1392,9 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	rawdb.WriteBlock(blockBatch, block)
 	rawdb.WriteReceipts(blockBatch, block.Hash(), block.NumberU64(), receipts)
 	rawdb.WritePreimages(blockBatch, state.Preimages())
+	if (randomCommitment != common.Hash{}) {
+		rawdb.WriteRandomCommitmentCache(blockBatch, randomCommitment, block.ParentHash())
```
trianglesphere, line 1396: If my understanding is correct, a validator must process the last block they proposed to include the randomness in the DB?

miner/worker.go
```diff
@@ -186,10 +181,10 @@ type worker struct {
 	resubmitHook func(time.Duration, time.Duration) // Method to call upon updating resubmitting interval.
 
 	// Needed for randomness
-	db *ethdb.Database
+	db ethdb.Database
```
trianglesphere, line 184: For my own curiosity, why did this switch from a pointer to a literal?

trianglesphere (commented):
I think that this fixes the backwards iteration issue because we bail out in commit new work, but this kinda looks like a validator will not propose a block until it rolls back it's head block and re-syncs or it's swapped out for a replica that was able to do that rollback and save the last randomness.


My understanding is that the validator needs to sync the randomness with the block state, but is there a way to do a backwards iteration and save on startup? Would that break something with atomic state?

trianglesphere:
Separately this breaks all of the e2e tests with the following as the last output (seems to time out):
```
   Deploying 'Random'
   ------------------
   > transaction hash:    0xc62a41b1569a16224df46bf749d66ccc0794f47c0239797e96f8528f7865a716
- Blocks: 0            Seconds: 0
   > Blocks: 0            Seconds: 0
   > contract address:    0xdDe36eF5b705E05fF72851E93Ca5Cc6f593A8461
   > block number:        300
   > block timestamp:     1608568122
   > account:             0x47e172F6CfB6c7D01C1574fa3E2Be7CC73269D95
   > balance:             3010022.399200419
   > gas used:            2729453
   > gas price:           100 gwei
   > value sent:          0 ETH
   > total cost:          0.2729453 ETH

  Setting initial Random implementation on proxy

- Saving migration to chain.

Error: Transaction was not mined within750 seconds, please make sure your transaction was properly sent. Be aware that it might still be mined!    at PromiEvent (/home/circleci/repos/celo-monorepo/node_modules/truffle/build/webpack:/packages/contract/lib/promievent.js:9:1)
    at TruffleContract.setCompleted (/home/circleci/repos/celo-monorepo/node_modules/truffle/build/webpack:/packages/contract/lib/execute.js:169:1)
    at Migration._deploy (/home/circleci/repos/celo-monorepo/node_modules/truffle/build/webpack:/packages/migrate/migration.js:93:1)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

kevjue:
> I think that this fixes the backwards iteration issue because we bail out in commit new work, but this kinda looks like a validator will not propose a block until it rolls back it's head block and re-syncs or it's swapped out for a replica that was able to do that rollback and save the last randomness.
> 
> My understanding is that the validator needs to sync the randomness with the block state, but is there a way to do a backwards iteration and save on startup? Would that break something with atomic state?

Totally agree with you points.  I've been adding to this PR where the it will do the backwards iteration (if the randomness cache is not already saved) right before starting mining.

miner/worker.go
kevjue: `ethdb.Database` is actually an interface type where the concrete value of it is a pointer.  So `db` was originally a pointer to a pointer.

Also, all other uses throughout the gocelo code uses the `ethdb.Database` type as a literal.

miner/worker.go
kevjue: Ya, the reverse iteration logic that I am going to add will address the case of a validator using chaindata created from another node.

core/blockchain.go
kevjue: Yes, they will process it to find out the resulting state.  TBH, I'm not really understanding your concern about the block processing and this code.

miner/worker.go
kevjue: BTW, the `db` variable was one that clabs originally added.

trianglesphere: Makes sense

core/blockchain.go
trianglesphere: This was me working out that we still need the backward iteration, but just at startup time, and that this handles the normal case operation of saving the randomness when processing blocks.

miner/miner.go
```diff
@@ -192,3 +204,138 @@ func (miner *Miner) SetEtherbase(addr common.Address) {
...
+	if currentHeader.Number.Uint64() > 0 {
+		blockHashIter := currentHeader.Hash()
+
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				blockHeader := bc.GetHeaderByHash(blockHashIter)
+
+				// We got to the genisis block, so this goroutine didn't find the latest
```
mcortesi, line 261: typo "genesis"

miner/miner.go
```diff
@@ -192,3 +204,138 @@ func (miner *Miner) SetEtherbase(addr common.Address) {
...
+	// If there was no previous commitment from this validator, then return true.
+	if (lastCommitment == common.Hash{}) {
+		return true
+	}
+
+	if (rawdb.ReadRandomCommitmentCache(miner.db, lastCommitment) != common.Hash{}) {
+		return true
+	}
+
+	// goroutine communication channels and waitgroup
```
mcortesi, line 242: this second part of the function looks weird...

it reads like:
```
firstBlock = race(
  getMostRecentFromCoinbaseInHistory()
  getFirstBlockFronCoinbaeInFuture()
)
generateAndSaveCommitement(firstBlock)

```

now, because of the race, this doesn't look to be deterministic, hence weird.

miner/miner.go
```diff
@@ -192,3 +204,138 @@ func (miner *Miner) SetEtherbase(addr common.Address) {
...
+
+	// Stacked defers are exececuted in LIFO order
+	defer wg.Wait()
+	defer close(stopCh)
+
+	if currentHeader.Number.Uint64() > 0 {
+		blockHashIter := currentHeader.Hash()
+
+		wg.Add(1)
+		go func() {
```
mcortesi, line 256: i would extract this two function into their own functions: `getMostRecentBlockFromAuthor(errgroup)` and `waitForNextBlockFromAuthor(errgroup)`

and use an errgroup to handle coordination (https://godoc.org/golang.org/x/sync/errgroup) 

Also, try move evertyhing to the `random.go` file; so to confine all code related to randomness to a single place. We can even not make it a part of `miner` object, and just pass everything as parameter to a static function

miner/miner.go
```diff
@@ -192,3 +204,138 @@ func (miner *Miner) SetEtherbase(addr common.Address) {
 func (self *Miner) SubscribePendingLogs(ch chan<- []*types.Log) event.Subscription {
 	return self.worker.pendingLogsFeed.Subscribe(ch)
 }
+
+// commitmentCacheSaved will check to see if the last commitment cache entry is saved.
+// If not, it will search for it and save it.
+func (miner *Miner) commitmentCacheSaved(istEngine consensus.Istanbul) bool {
```
mcortesi, line 210: function name suggest is just a query, no side effect or write. I would change the name to make that more apparent.

core/blockchain.go
```diff
@@ -1376,6 +1392,11 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	rawdb.WriteBlock(blockBatch, block)
 	rawdb.WriteReceipts(blockBatch, block.Hash(), block.NumberU64(), receipts)
 	rawdb.WritePreimages(blockBatch, state.Preimages())
+	if (randomCommitment != common.Hash{}) {
```
mcortesi, line 1395: i'm hesitatnt to have this logic inside `blockchain`, would it make sense to have a blockchain listener that computes/writes the commitment for every block?

In fact, we could have a more cohesive type with all random related functions:

 * Listen to new Blocks to generate/write commitment
 * Reconstruct commitment cache
 * Get current / genearte current

miner/miner.go
mcortesi, line 242: i think here i would have a new function, something like "regenerateCache"... and a good log message stating taht we are doing such... (and that it might take a while)

kevjue: How would you suggest this is changed?

Both `getters` of the race is necessary.

kevjue: nvm, i see some more of you comments below.

trianglesphere: Do we need to wait for new blocks at this point? During sync start, mining gets stopped with `Mining aborted due to sync`, but as it reads in the new blocks that it doesn't have, it will check if it should save the randomness. Once it's done syncing or fails to sync, it runs this function prior to starting mining again.

core/blockchain.go
kevjue: What's your hesitancy of putting this in `blockchain`?

The changes are following much of the same design that `blockchain` already implements.  E.g. the totalDifficulty is also a "cache", in that it's possible to reconstruct from the past chain, but it's computed at the time of block insert and saved in it's own leveldb location.

miner/miner.go
kevjue: I actually was considering putting some of this logic within `istanbul`, but decided against it.  The reason is that I believe Istanbul should only be responsible for defining the randomness beacon protocol (e.g. when/how is the randomness generated and verified). 

As for the randomness cache, the only users of it is the `miner` package, as the cache is not an essential part of the randomness protocol, but more of an optimization that the `miner` can use.  So with that in mind, restoration of the cache seems to fit within the `miner` package.  I'm happy to create a new `random.go` file within `miner` package, though.

core/blockchain.go
trianglesphere: +1 for keeping this in blockchain. I think keeping this code block here is much easier to understand than having a dedicated listener.

trianglesphere:
I deployed `64494cc9ce6a1717de45b1a61d618d1f17de394c` to baklava, and the fix got the validator to start signing again.

core/blockchain.go
mcortesi: My hesitation is about coupling...  or using other maxims: Separation of Concerns / Single Responsibility Principle

Now blockchain "knows" about randomness cache, (it's coupled to it). Same with miner.

And then, the "responsibility" of keeping the cache sane is split in a few places. There's the logic of the cache (read/write), there's the logic of check & recovery on miner, and there's the logic of generation with each new block on blockchain.

Now, in any future where a maintainer makes a change to those modules, it need to have the same logical context and awareness about this. And that's tricky, since it's just one of the many responsibilities these modules have.

That's why my suggestion is to have "something" that represents the randomness cache, and it's responsible of keeping it sane, recovering it and providing access to it. That "something" can even be an interface, thus it's easy to mock and test things in isolation.

@trianglesphere @kevjue what do you think?

miner/miner.go
mcortesi: I have the same doubt as @trianglesphere . I understood that when this function is called we are "already synced" since that's the reason of the `update()` function. (but i remember it's a tricky one, since it only triggers once)

And it would seems that even if that's not the case, the code in blockchain would still handle it, since it reacts on new blocks

core/blockchain.go
kevjue: WDYT of keeping the randomness cache within the blockchain object (basically it would be a "primitive" of the block) and move the restoration logic from `miner` into `blockchain` (However, `miner` will still invoke the restoration logic if needed).

As for the actual insertion of the cache (at least for the normal case, as opposed to the restoration case), I still think that it should be done within the block insertion function (basically how it's implemented now), since it's inserted as part of a leveldb transaction along with all of the other parts of the block, since inserting all block related data atomically is one high level goal that we wanted to attain.

miner/miner.go
kevjue: This handles the case of when the the partition head block (made up of the node and it's peers when the sync `downloader.DoneEvent` is emitted) is behind a block that the node authored (let's call this the "future authored block"). Note that this probably is a very rare scenario.

When this scenario happens, two things can happen.

1) The node's reverse iteration is completed before the "future authored block" is synced.  The node will then start istanbul.  Even if this partition's aggregate head block is behind other validators, this is already handled via the istanbul engine.
2) The "future authored block" is synced before the reverse iteration is completed.  In this case, the syncing of that block will stop the reverse iteration since it has become irrelevant, and then istanbul will be started.

So waiting for new blocks is basically to handle case 2.  Having said all that, that case is probably going to happen very infrequently, and the added benefit (not having to wait for the reverse iteration to complete) seems marginal.  So seems like it's not worth the added complexity.  WDYT?

trianglesphere: So thinking about what we store, it's a lookup of `commitment -> block hash` where the block hash is mangled to create the commitment. I think that we have multiple of these commitments in the eth db. When looking up the last commitment made, we check a smart contract, so I think that order does not matter here, only that we do have the last commitment saved in the level db.

It's true that we want to bail out of reverse iteration as an optimization if we find a "future authored block", but shouldn't the future block record its commitment when we process it.

core/blockchain.go
trianglesphere: I'm fine with bundling lines approx 1360 to 1370 into some manager object/primitive, but I agree with Kevin that `rawdb.WriteRandomCommitmentCache(...)` should stay here.

mcortesi: I like it!

miner/miner.go
kevjue: @trianglesphere - Ya, you're correct. It's unnecessarily recalculating the cache entry when it gets the "future authored block".

WDYT of the proposal of removing that handling of new block logic altogether. It does add alot of complex logic (which means higher chance for bugs) for minor benefit?

trianglesphere: That sounds good to me. Seem like it's rare enough and doesn't make it incorrect.

mcortesi: +1

core/blockchain.go
```diff
@@ -2336,3 +2337,45 @@ func (bc *BlockChain) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscript
 func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscription {
 	return bc.scope.Track(bc.blockProcFeed.Subscribe(ch))
 }
+
+func (bc *BlockChain) RecoverRandomnessCache(commitment common.Hash, commitmentBlockHash common.Hash) error {
+	if istEngine, isIstanbul := bc.engine.(consensus.Istanbul); isIstanbul {
```
mcortesi, line 2342: very personal nit :P 

I prefer exiting fast, and avoiding nesting here:
```go
istEngine, isIstanbul := bc.engine.(consensus.Istanbul)
if !isIstanbul {
   return 
}
...
```

core/blockchain.go
```diff
@@ -2336,3 +2337,45 @@ func (bc *BlockChain) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscript
...
+
+		blockHashIter := commitmentBlockHash
+		var parentHash common.Hash
+		for {
+			blockHeader := bc.GetHeaderByHash(blockHashIter)
+
+			// We got to the genisis block, so this goroutine didn't find the latest
```
mcortesi, line 2349: typo: `genesis`

mcortesi: also, not a goroutine anymore

core/blockchain.go
```diff
@@ -2336,3 +2337,45 @@ func (bc *BlockChain) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscript
 func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscription {
 	return bc.scope.Track(bc.blockProcFeed.Subscribe(ch))
 }
+
+func (bc *BlockChain) RecoverRandomnessCache(commitment common.Hash, commitmentBlockHash common.Hash) error {
```
mcortesi, line 2341: go doc

core/blockchain.go
```diff
@@ -2336,3 +2337,45 @@ func (bc *BlockChain) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscript
...
+	if istEngine, isIstanbul := bc.engine.(consensus.Istanbul); isIstanbul {
+
+		blockHashIter := commitmentBlockHash
```
mcortesi, line 2344: i would add a log message here, it's quite a rare thing, and can take a decent amount of time, so an `INFO` msg would be nice, to inform the user what's happening

mcortesi (approved):

PR #1157, opened 2020-09-04
CIP-30: BLS12-377 Precompiles

kilic:
### Description

This pr adds BLS12-377 elliptic curve operations as precompiles. Precompiles are added in same fashion with [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) except the mapping functions.

#### Background

Elliptic curve library is adapted from [BLS12-381](https://github.com/kilic/bls12-381) and [EIP-1962](https://github.com/saitima/eip1962) works. x86 optimised field implementation is generated with [fp](https://github.com/kilic/fp) and native go implementation is generated with [goff](https://github.com/ConsenSys/goff)

### Other changes

Gas table for these new precompiles set is not decided yet. Rather than actual numbers placeholder constants are placed.

### Tests

Underlying elliptic curve library is heavily tested in terms of mathematical properties. 

Original library BLS12-381 is subjected to the [security audit](https://github.com/kilic/bls12-381/issues/19) under [drand project](https://drand.love/blog/2020/08/10/drand-launches-v1-0/) and also made its way to be backend of EIP-2537 go-ethereum precompiles[*](https://github.com/ethereum/go-ethereum/pull/21018) for the next HF.

Cross and fuzz testing are not applied for the adapted library yet, also _static call_ test vectors are not generated.

### Backwards compatibility

Precompiles are included next to Celo specific precompile bundle. It is assumed that precompile addresses incrementally follow the others.

PR #1282, opened 2021-01-05
BW6 Precompiles

kilic:
### Description

This PR implements [EIP-3026](https://eips.ethereum.org/EIPS/eip-3026) which defines BW6 elliptic curve operations as precompiles.

#### Background

x86 optimised field implementation is generated with [fp](https://github.com/kilic/fp) and native go implementation is generated with [goff](https://github.com/ConsenSys/goff)

### Other changes

Gas table for these new precompiles set is not decided yet. Rather than actual numbers placeholder constants are placed.

### Tests

Underlying elliptic curve library is heavily tested in terms of mathematical properties.  Test vectors are generated with [zexe library](https://github.com/kilic/zexe/blob/eip3026/algebra/src/bw6_761/curves/eip3026_test.rs)

### Backwards compatibility

Precompiles are included next to Celo specific precompile bundle. It is assumed that precompile addresses incrementally follow the [BLS12-377 precompiles](https://github.com/celo-org/celo-blockchain/pull/1157).


mcortesi:
close for now

PR #1288, opened 2021-01-08
Remove istanbul flags that must not be reconfigured

mcortesi:
### Description

Parameters:
  * lookbackWindow
  * epochSize
  * blockTime
  * requestTimeout for istanbul request
  * proposerPolicy

are configured on the genesis block (chainConfig). 

They can also be configured by command line flags. This is problematic, since we can't change `epochSize`, `lookbackWindow` or `blockTime` on already existent blockchain. For `requestTimeout` and `proposerPolicy` almost the same applies, if validators have different values for these parameters, then unwanted behaviour might appear

With that in mind, this PR removes the command line flags to set those parameters. 

#### Note about current usage

Current code in master, doesn't actually uses the cmdline flags values. This is because flags are read and pass to `istanbul.Config` object in `setIstanbul`; but later on on `eth.New()`, the function [CreateConsensusEngine](https://github.com/celo-org/celo-blockchain/blob/8ec69a83aa480883b4b4d719e8c4d645b1173fed/eth/backend.go#L272) is called that will use `chainConfig` (configuration obtained from the genesis block) and override those values (given they are defined in chainConfig, where they should)

Currently, docs.celo.org specifies those variables in Running a Validator guide for mainnet and baklava. 

For mainnet, the [guide](https://docs.celo.org/getting-started/mainnet/running-a-validator-in-mainnet#deploy-a-validator-machine) established:
```
--istanbul.blockperiod=5 --istanbul.requesttimeout=3000
```
but instead of using these values, the [genesis block](https://storage.googleapis.com/genesis_blocks/rc1) is used; which has the same values :)

For **baklava**; the story is different, since the [guide](https://docs.celo.org/getting-started/baklava-testnet/running-a-validator-in-baklava#deploy-a-validator-machine) established a `requesttimeout` of 3000, but the [genesis](https://storage.googleapis.com/genesis_blocks/baklava) block states 10000 as the timeout

### Tested

Unit tests

### Backwards compatibility

* Changes cmdline flags (should deprecate them first)
* Implies modifying the doc

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -67,6 +67,8 @@ var (
 		Istanbul: &IstanbulConfig{
 			Epoch:          17280,
 			ProposerPolicy: 2,
+			BlockPeriod:    5,
+			RequestTimeout: 3000,
 			LookbackWindow: 12,
 		},
 	}
@@ -111,6 +113,8 @@ var (
 		Istanbul: &IstanbulConfig{
 			Epoch:          17280,
 			ProposerPolicy: 2,
+			BlockPeriod:    5,
+			RequestTimeout: 10000,
 			LookbackWindow: 12,
 		},
 	}
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -724,22 +724,22 @@ var (
 	// Istanbul settings
 	IstanbulRequestTimeoutFlag = cli.Uint64Flag{
 		Name:  "istanbul.requesttimeout",
-		Usage: "Timeout for each Istanbul round in milliseconds",
+		Usage: "Timeout for each Istanbul round in milliseconds (deprecated, value obtained from genesis config)",
 		Value: eth.DefaultConfig.Istanbul.RequestTimeout,
 	}
 	IstanbulBlockPeriodFlag = cli.Uint64Flag{
 		Name:  "istanbul.blockperiod",
-		Usage: "Default minimum difference between two consecutive block's timestamps in seconds",
+		Usage: "Default minimum difference between two consecutive block's timestamps in seconds  (deprecated, value obtained from genesis config)",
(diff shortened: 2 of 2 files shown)


params/config.go
```diff
@@ -111,6 +113,8 @@ var (
 		Istanbul: &IstanbulConfig{
 			Epoch:          17280,
 			ProposerPolicy: 2,
+			BlockPeriod:    5,
+			RequestTimeout: 10000,
```
oneeman, line 117: I'm not sure about setting the timeout for Baklava to 10000, given what we discovered when we looked into how the configuration works.  It was 10000 initially due to the value in the genesis json, but when validators starting using `--baklava`, it overwrote it in the stored configuration with the default which is 3000.  I verified this today as follows:
1. add logging all over the place to see exactly how the timeout parameter is passed around and what it is in the end
1. run `geth init` using the baklava genesis
2. with the same datadir, run geth without `--baklava`, verify the timeout it uses is 10000
3. with the same datadir, run geth with `--baklava`, verify the timeout is uses is now 3000

At this point, because of Churrito, we know everyone is using `--baklava`, so everyone has 3000.  With this change, it will overwrite it back to 10000.  Is that explicitly the intention, to go back to 10000 to agree with the genesis json?  If not, I would prefer to use 3000 here.

oneeman (commented):
Looks good, except see the comment below about the Baklava RequestTimeout value.

params/config.go
mcortesi: I'm convinced!. If everyone is running with --baklava, then let's use `3000` here, and modify the genesis.json file just in case... (or not... i need to check if the number is part of the genesis block hash)

mcortesi:
This set of change was included as part of #1136 . Closing it

PR #1293, opened 2021-01-11
Add patch for mobile libs

gastonponti:
### Description

Adds a patch that will be applied for the mobile building, adding the required libs to build successfully the client

### Tested

Manually


mcortesi (approved):
Nice. Maybe update README build instructions for mobile?

jeanregisser (approved):
Thanks so much! This will make our life easier 👍

PR #1299, opened 2021-01-14
Add a golang build tag 'mobile'

hbandura:
Add a golang build tag 'mobile' to prevent specific types and functions to be included in the mobile build.

UPDATE:

Upon closer look at how the mobile client is being used from the monorepo mobile package, it seems that at least the eth rpc api is being used from ipc, and even the http server api can be used in a debug mode.

I'm updating this pr to only add the mobile tag until a more detailed research on which rpc methods are used from the mobile end


eth/backend_apis_mobile.go
```diff
@@ -0,0 +1,9 @@
+// +build mobile
+
+package eth
+
+import "github.com/ethereum/go-ethereum/rpc"
+
+func (s *Ethereum) APIs() []rpc.API {
+	return []rpc.API{}
```
mcortesi, line 8: don't we want the eth_ apis?

eth/api.go
```diff
@@ -1,3 +1,4 @@
+// +build !mobile
```
mcortesi, line 1: what's in here, why is this removeD?

go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	github.com/tyler-smith/go-bip39 v1.0.1-0.20181017060643-dbb3b84ba2ef
 	github.com/wsddn/go-ecdh v0.0.0-20161211032359-48726bab9208
 	golang.org/x/crypto v0.0.0-20200311171314-f7b00557c8c4
-	golang.org/x/mobile v0.0.0-20200801112145-973feb4309de // indirect
+	golang.org/x/mobile v0.0.0-20201217150744-e6ae53a27f4f // indirect
```
mcortesi, line 63: what's this?

eth/backend_apis_mobile.go
hbandura: I believe no api's are being used from the mobile client, but I'll double check

eth/api.go
hbandura: api implementation. in particular some methods of privatedebugapi and others

go.mod
hbandura: this is actually happening in master too if I run make android

oneeman:
I think using build tags to exclude code is fine.  Where it gets iffy is in cases where we have to move code to separate file in order to exclude or include it, as I see had to be done in one case here.  In such cases, we'd really have to quantify how much space that will ultimately save in the android/ios bundles (which I believe are compressed) and whether it's worth the additional conflicts later on during upstream merged (any time you move content from one file to another you will get a conflict whenever that is changed upstream and have to make the same changes yourself in the new location).

From playing around with the 'goweight' tool, I get the impression much of the binary size comes from the runtime and libraries we can't really avoid, so if we're making non-trivial changes to reduce the size we should know how much we're gaining from them.

hbandura:
> I think using build tags to exclude code is fine. Where it gets iffy is in cases where we have to move code to separate file in order to exclude or include it, as I see had to be done in one case here. In such cases, we'd really have to quantify how much space that will ultimately save in the android/ios bundles (which I believe are compressed) and whether it's worth the additional conflicts later on during upstream merged (any time you move content from one file to another you will get a conflict whenever that is changed upstream and have to make the same changes yourself in the new location).
> 
> From playing around with the 'goweight' tool, I get the impression much of the binary size comes from the runtime and libraries we can't really avoid, so if we're making non-trivial changes to reduce the size we should know how much we're gaining from them.

Yes, it's definitely a tradeoff. This PR I think reduced the compressed size like a 5% or so. It's not huge, but I believe more work like this would bring more improvements. To avoid (huge) merging issues I'm trying to not actually touch code but just in some cases (like this one) move a function to a different place, but I agree it still makes merging a little bit more difficult.

go.mod
mcortesi: check with @gastonponti who has done some work on mobile build

mcortesi:
> > I think using build tags to exclude code is fine. Where it gets iffy is in cases where we have to move code to separate file in order to exclude or include it, as I see had to be done in one case here. In such cases, we'd really have to quantify how much space that will ultimately save in the android/ios bundles (which I believe are compressed) and whether it's worth the additional conflicts later on during upstream merged (any time you move content from one file to another you will get a conflict whenever that is changed upstream and have to make the same changes yourself in the new location).
> > From playing around with the 'goweight' tool, I get the impression much of the binary size comes from the runtime and libraries we can't really avoid, so if we're making non-trivial changes to reduce the size we should know how much we're gaining from them.
> 
> Yes, it's definitely a tradeoff. This PR I think reduced the compressed size like a 5% or so. It's not huge, but I believe more work like this would bring more improvements. To avoid (huge) merging issues I'm trying to not actually touch code but just in some cases (like this one) move a function to a different place, but I agree it still makes merging a little bit more difficult.

To me there's a trade off between improving the code vs making the merge process more complex. I believe we should make the changes when we see there's a sufficient reason for that. In this case, i think improving that 5% seems reasonable enough to do the change

go.mod
oneeman: I suspect this doesn't matter, given replace we have for it below:
```
// Use our fork which disables bitcode
replace golang.org/x/mobile => github.com/celo-org/mobile v0.0.0-20201127114005-6a1221213dcf	replace golang.org/x/mobile => github.com/celo-org/mobile v0.0.0-20201127114005-6a1221213dcf
```

oneeman (approved):
LGTM!

oneeman:
Approved, but now I see there is a small merge conflict. Once that's resolved, it should be good to merge in.

mcortesi:
we wont' be doing this for the time being. Might be reopened in the fututre

PR #1303, opened 2021-01-15
Move TypedData from signer.core to shared.signer

hbandura:
By adding SignTypedData to the mobile interface, an indirect reference to protocol buffers was added, which gets dragged into the mobile binary. To prevent this I moved the only part that's needed (The TypedData type and a few other subtypes) to a shared location where both the mobile interface and the clef signer core package can access it.

This is not the only indirect reference to protocol buffers, it's only one in two, but I will remove the other one in a different PR.


oneeman:
Approved, but now I noticed a small merge conflict.  Should be good to merge after that's resolved.

oneeman:
Approved, but now I see there is a small merge conflict.  Once that's resolved, it should be good to merge in.

oneeman (approved):
LGTM

PR #1250, opened 2020-12-02
CIP-26: Get validator BLS key precompile

mrsmkl:
### Description

https://github.com/celo-org/celo-proposals/pull/84

Depends on CIP22 PR (https://github.com/celo-org/celo-blockchain/pull/1231) because it needs the latest celo-bls-go.

### Other changes

To speed it up, added the uncompressed BLS key to the snapshots. (Many possibilities how this should be done)

### Tested

Added unit tests.

### Backwards compatibility

Will require hardfork.

PR #1307, opened 2021-01-20
CIP-26 precompile

mrsmkl:
### Description

https://github.com/celo-org/celo-blockchain/pull/1250

Recover if cannot load the round state.
Changed so that the BLS keys are now uncompressed when the snapshot is created.

### Tested

I ran some tests to make sure that stored snapshots include the uncompressed keys.
Also checked that it won't get stuck if there's an old roundstate.


consensus/istanbul/core/core.go
```diff
@@ -618,8 +618,15 @@ func (c *core) createRoundState() (RoundState, error) {
 		roundState, err = c.rsdb.GetRoundStateFor(lastStoredView)
 
 		if err != nil {
-			logger.Error("Failed to fetch lastStoredRoundState", "err", err)
-			return nil, err
+			logger.Warn("Failed to fetch lastStoredRoundState", "err", err)
+			if err == leveldb.ErrNotFound {
```
kobigurk, line 622: Can we maybe extract the common functionality of re-creating the round state to some function?

consensus/istanbul/validator/default.go
```diff
@@ -28,12 +28,27 @@ import (
...
 type defaultValidator struct {
 	address      common.Address
 	blsPublicKey blscrypto.SerializedPublicKey
+	uncompressed []byte
+}
+
+func newValidatorFromDataWithCache(data *istanbul.ValidatorDataWithCache) *defaultValidator {
+	uncompressed := data.Uncompressed
+	if len(uncompressed) == 0 {
+		log.Warn("no uncompressed")
```
kobigurk, line 44: Maybe it's be nice to print the address. To be honest, it's more of a DEBUG rather than WARN imo.

consensus/istanbul/validator/default.go
```diff
@@ -28,12 +28,27 @@ import (
...
+		uncompressed = nil
```
kobigurk, line 45: Maybe we should uncompress here directly? It should be rare after some time, right?

consensus/istanbul/validator.go
```diff
@@ -36,6 +37,31 @@ type ValidatorData struct {
...
+	tmp := struct {
+		Address      common.Address
+		BLSPublicKey blscrypto.SerializedPublicKey
+		Uncompressed []byte
+	}{}
+	err = json.Unmarshal(data, &tmp)
+	if err != nil {
+		return err
+	} else if tmp.Uncompressed == nil && tmp.BLSPublicKey != (blscrypto.SerializedPublicKey{}) {
+		return fmt.Errorf("Cached key missing")
```
kobigurk, line 56: Will this "get fixed" later? Will anyone catch this error and try to re-initialize it correctly?
If not, maybe it's better to directly fill in the uncompressed part?

mrsmkl: It will be caught when snapshot is loaded. It's better to error here, that way the old snapshot will get replaced with the one that has the uncompressed key.

kobigurk: Can you show where the retry happens? I had a hard time tracing it

mrsmkl: It tries to load if here https://github.com/celo-org/celo-blockchain/blob/mrsmkl/precompile-bls-key/consensus/istanbul/backend/engine.go#L861
If it fails, it will create a new one eventually.

consensus/istanbul/validator.go
```diff
@@ -44,12 +70,19 @@ type Validator interface {
...
 	// can be use used to instantiate a validator with DeserializeValidator()
 	Serialize() ([]byte, error)
 
 	// AsData returns Validator representation as ValidatorData
 	AsData() *ValidatorData
+
+	// AsData returns Validator representation as ValidatorData
+	AsDataWithCache() *ValidatorDataWithCache
+
+	CacheUncompressed()
```
mcortesi, line 85: tbh, this looks rather ugly. We have a data type `ValidatorSet` where some implementation of it `defaultSet` has a derived field that it caches. Which is an implementation detail, not a design one.

Do all implementations need to have a cache?

I understand this is convenient, for the purppose of snapshot caching this values.. but it's ugly for a encapsulation/separation of cocerns PoV.

Still, given we are rather on a hurry, and i don't have a better proposal for the moment, we can go ahead. but still.. i want to make that clear

core/vm/contracts.go
```diff
@@ -945,6 +945,75 @@ func (c *getValidator) Run(input []byte, caller common.Address, evm *EVM, gas ui
...
+	validators := evm.Context.Engine.GetValidators(new(big.Int).Sub(blockNumber, common.Big1), common.Hash{})
+
+	// Ensure index, which is guaranteed to be non-negative, is valid.
+	if index.Cmp(big.NewInt(int64(len(validators)))) >= 0 {
+		return nil, gas, ErrValidatorsOutOfBounds
+	}
+
+	validator := validators[index.Uint64()]
+	uncompressedBytes := validator.BLSPublicKeyUncompressed()
+	if len(uncompressedBytes) == 0 {
```
mcortesi, line 995: following the same rationale as my previous comments. I don't think this check should be done here. It's validator's responsability to return the actual uncompressed version of the bls key when asked for it with `BLSPublicKeyUncompressed()`.

If you need to make sure it's already cache, take measures before, as you already do in snapshot

core/vm/contracts.go
```diff
@@ -945,6 +945,75 @@ func (c *getValidator) Run(input []byte, caller common.Address, evm *EVM, gas ui
...
+		log.Warn("Uncompressed BLS key wasn't cached", "address", validator.Address())
+		validator.CacheUncompressed()
+		uncompressedBytes = validator.BLSPublicKeyUncompressed()
+	}
+	if len(uncompressedBytes) != 192 {
```
mcortesi, line 1000: shouldn't this check be done the moment we compute the uncompressed key??

are all the others callers left to do the validation themselves, as if `.BLSPublicKeyUncompressed()` could return an invalid version of this???

consensus/istanbul/backend/snapshot.go
```diff
@@ -70,6 +70,7 @@ func loadSnapshot(epoch uint64, db ethdb.Database, hash common.Hash) (*Snapshot,
 
 // store inserts the snapshot into the database.
 func (s *Snapshot) store(db ethdb.Database) error {
+	s.ValSet.CacheUncompressed()
```
mcortesi, line 73: CacheUncompressed() rigth now computes the uncompressed key wether it's already set or not... which it should. It should only compute it when it's not already cached

consensus/istanbul/backend/snapshot.go
```diff
@@ -156,17 +157,21 @@ func (s *Snapshot) validators() []istanbul.ValidatorData {
 	return validator.MapValidatorsToData(s.ValSet.List())
 }
 
+func (s *Snapshot) validatorsWithCache() []istanbul.ValidatorDataWithCache {
```
mcortesi, line 160: i wouldn't have this method if you're only going to call it from a single place

consensus/istanbul/validator/default.go
```diff
@@ -251,16 +291,27 @@ func (valSet *defaultSet) AsData() *istanbul.ValidatorSetData {
 	}
 }
 
+func (valSet *defaultSet) AsDataWithCache() *istanbul.ValidatorSetDataWithCache {
```
mcortesi, line 294: do we need this method? it seems the whole reason for the change is for snapshot to cache and persist validators with the uncompressed keys. Now, snapshot's marshallJSON doesn't use this method, nor it marshalls directly the set, instead it marshalls the validatorList only.

consensus/istanbul/validator/default.go
```diff
@@ -237,7 +277,7 @@ func (valSet *defaultSet) RemoveValidators(removedValidators *big.Int) bool {
 func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 	valSet.validatorMu.RLock()
 	defer valSet.validatorMu.RUnlock()
-	newValSet := NewSet(MapValidatorsToData(valSet.validators))
+	newValSet := NewSetWithCache(MapValidatorsToDataWithCache(valSet.validators))
```
mcortesi, line 280: do we need to change this method?? it's not used by snapshot

consensus/istanbul/validator/default.go
```diff
@@ -251,16 +291,27 @@ func (valSet *defaultSet) AsData() *istanbul.ValidatorSetData {
...
+	return &istanbul.ValidatorSetDataWithCache{
+		Validators: MapValidatorsToDataWithCache(valSet.validators),
+		Randomness: valSet.randomness,
+	}
+}
+
 // JSON Encoding -----------------------------------------------------------------------
 
-func (val *defaultSet) MarshalJSON() ([]byte, error) { return json.Marshal(val.AsData()) }
+func (val *defaultSet) MarshalJSON() ([]byte, error) {
```
mcortesi, line 305: do we need tho change this method? it's not used by snapshot, and it breaks simmetry. If i marshall as json it's on thing, if a marshall with RLP it's another

consensus/istanbul/validator/default.go
```diff
@@ -251,16 +291,27 @@ func (valSet *defaultSet) AsData() *istanbul.ValidatorSetData {
...
+	return json.Marshal(val.AsDataWithCache())
+}
 
 func (val *defaultSet) UnmarshalJSON(b []byte) error {
```
mcortesi, line 309: same

consensus/istanbul/backend/snapshot.go
```diff
@@ -185,7 +190,7 @@ func (s *Snapshot) UnmarshalJSON(b []byte) error {
 	s.Epoch = j.Epoch
 	s.Number = j.Number
 	s.Hash = j.Hash
-	s.ValSet = validator.NewSet(j.Validators)
+	s.ValSet = validator.NewSetWithCache(j.Validators)
```
mcortesi, line 193: here's something strange. Not related to this PR actually.

A ValidatorSet is actually a `([]validators, randomness)`, but when we "save" it we don't save the randomness, and thus on unmarshall we don't have it, nor we care.

Which makes me wonder if we really want a validatorSet or if we should save the randomenss.

mcortesi (changes requested):
Left some comments!

mcortesi (changes requested):
Left some comments!

consensus/istanbul/validator.go
mrsmkl: Moved the check to a different place.

consensus/istanbul/validator/default.go
mrsmkl: It's used by Snapshot.copy(), so it's needed...

core/vm/contracts.go
mrsmkl: It will return a string with length 0 if the compressed BLS public key is invalid. Of course, this should never be the case, but I think it's better to not crash with out of bounds access here.

consensus/istanbul/backend/snapshot.go
mrsmkl: Looks like randomness is only needed in the round state to determine next proposer, so it's not needed in the snapshots.

consensus/istanbul/validator/default.go
```diff
@@ -253,7 +308,10 @@ func (valSet *defaultSet) AsData() *istanbul.ValidatorSetData {
 
 // JSON Encoding -----------------------------------------------------------------------
 
-func (val *defaultSet) MarshalJSON() ([]byte, error) { return json.Marshal(val.AsData()) }
+func (val *defaultSet) MarshalJSON() ([]byte, error) {
+	log.Warn("Here")
```
mcortesi: remove

consensus/istanbul/validator.go
```diff
@@ -120,6 +133,9 @@ type ValidatorSet interface {
 	// Copy validator set
 	Copy() ValidatorSet
 
+	CacheUncompressed()
```
mcortesi, line 136: Add Comments to both here and in Validator.

Also, rename `CacheUncompressed` to `CacheUncompressedBLSKey`

core/state/database.go
```diff
@@ -98,18 +98,18 @@ type Trie interface {
 
 // NewDatabase creates a backing store for state. The returned database is safe for
 // concurrent use, but does not retain any recent trie nodes in memory. To keep some
-// historical state in memory, use the NewDatabaseWithCache constructor.
+// historical state in memory, use the NewDatabaseWithBLSKeyCache constructor.
```
mcortesi: ??????????????

core/state/database.go
```diff
@@ -98,18 +98,18 @@ type Trie interface {
...
-	return NewDatabaseWithCache(db, 0)
+	return NewDatabaseWithBLSKeyCache(db, 0)
 }
 
-// NewDatabaseWithCache creates a backing store for state. The returned database
+// NewDatabaseWithBLSKeyCache creates a backing store for state. The returned database
 // is safe for concurrent use and retains a lot of collapsed RLP trie nodes in a
 // large memory cache.
-func NewDatabaseWithCache(db ethdb.Database, cache int) Database {
+func NewDatabaseWithBLSKeyCache(db ethdb.Database, cache int) Database {
```
mcortesi: ???????????

core/blockchain.go
```diff
@@ -221,7 +221,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 		cacheConfig:    cacheConfig,
 		db:             db,
 		triegc:         prque.New(nil),
-		stateCache:     state.NewDatabaseWithCache(db, cacheConfig.TrieCleanLimit),
+		stateCache:     state.NewDatabaseWithBLSKeyCache(db, cacheConfig.TrieCleanLimit),
```
mcortesi: ??????????

core/genesis.go
```diff
@@ -171,7 +171,7 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, genesis *Genesis, override
 	// We have the genesis block in database(perhaps in ancient database)
 	// but the corresponding state is missing.
 	header := rawdb.ReadHeader(db, stored, 0)
-	if _, err := state.New(header.Root, state.NewDatabaseWithCache(db, 0), nil); err != nil {
+	if _, err := state.New(header.Root, state.NewDatabaseWithBLSKeyCache(db, 0), nil); err != nil {
```
mcortesi: ????????

crypto/bls/bls.go
```diff
@@ -252,3 +252,15 @@ func SerializedSignatureFromBytes(serializedSignature []byte) (SerializedSignatu
...
+func UncompressKey(serialized SerializedPublicKey) []byte {
+	publicKey, err := bls.DeserializePublicKeyCached(serialized[:])
+	if err != nil {
+		return nil
+	}
+	uncompressedBytes, err := publicKey.SerializeUncompressed()
+	if err != nil {
+		return nil
+	}
+	return uncompressedBytes
```
mcortesi, line 265: maybe here we can check that len is 192? that means we know for a fact that if we return something here it's correct.

Also, in that sense, it might make sense for this function to return an error... since `nil` doesn't really work as a good error value. I don't see any place checking if the returned value by this function is nil and reacting to that...

so we are effectively silently ignoring errors in this pr

eth/api_tracer.go
```diff
@@ -147,7 +147,7 @@ func (api *PrivateDebugAPI) traceChain(ctx context.Context, start, end *types.Bl
 
 	// Ensure we have a valid starting state before doing any work
 	origin := start.NumberU64()
-	database := state.NewDatabaseWithCache(api.eth.ChainDb(), 16) // Chain tracing will probably start at genesis
+	database := state.NewDatabaseWithBLSKeyCache(api.eth.ChainDb(), 16) // Chain tracing will probably start at genesis
```
mcortesi: ???????????????

eth/api_tracer.go
```diff
@@ -641,7 +641,7 @@ func (api *PrivateDebugAPI) computeStateDB(block *types.Block, reexec uint64) (*
 	}
 	// Otherwise try to reexec blocks until we find a state or reach our limit
 	origin := block.NumberU64()
-	database := state.NewDatabaseWithCache(api.eth.ChainDb(), 16)
+	database := state.NewDatabaseWithBLSKeyCache(api.eth.ChainDb(), 16)
```
mcortesi: ????????????????????

consensus/istanbul/validator.go
mrsmkl: Perhaps we should change all precompiles that depend on snapshots to only work for a limited amount of epochs, (then we can keep them all in memory). Then we could remove `ValidatorDataWithCache`. To remove `CacheUncompressed` from here, we could just store them in the valset instead, or the `defaultSet` could contain a list of `defaultValidator`...

light/postprocess.go
```diff
@@ -143,7 +143,7 @@ func NewChtIndexer(db ethdb.Database, odr OdrBackend, size, confirms uint64, ful
 		diskdb:      db,
 		odr:         odr,
 		trieTable:   trieTable,
-		triedb:      trie.NewDatabaseWithCache(trieTable, 1), // Use a tiny cache only to keep memory down
+		triedb:      trie.NewDatabaseWithBLSKeyCache(trieTable, 1), // Use a tiny cache only to keep memory down
```
mcortesi: ????????????????

light/postprocess.go
```diff
@@ -264,7 +264,7 @@ func NewBloomTrieIndexer(db ethdb.Database, odr OdrBackend, parentSize, size uin
 		diskdb:     db,
 		odr:        odr,
 		trieTable:  trieTable,
-		triedb:     trie.NewDatabaseWithCache(trieTable, 1), // Use a tiny cache only to keep memory down
+		triedb:     trie.NewDatabaseWithBLSKeyCache(trieTable, 1), // Use a tiny cache only to keep memory down
```
mcortesi: ??????????

params/protocol_params.go
```diff
@@ -138,6 +138,7 @@ const (
 	FractionMulExpGas           uint64 = 50     // Cost of performing multiplication and exponentiation of fractions to an exponent of up to 10^3.
 	ProofOfPossessionGas        uint64 = 350000 // Cost of verifying a BLS proof of possession.
 	GetValidatorGas             uint64 = 1000   // Cost of reading a validator's address.
+	GetValidatorBLSGas          uint64 = 1000   // Cost of reading a validator's BLS public key.
```
mcortesi, line 141: how did we obtained this value?? cc  @prestwich @kobigurk

trie/database.go
```diff
@@ -276,13 +276,13 @@ func expandNode(hash hashNode, n node) node {
 // its written out to disk or garbage collected. No read cache is created, so all
 // data retrievals will hit the underlying disk database.
 func NewDatabase(diskdb ethdb.KeyValueStore) *Database {
-	return NewDatabaseWithCache(diskdb, 0)
+	return NewDatabaseWithBLSKeyCache(diskdb, 0)
```
mcortesi: ??????

trie/database.go
```diff
@@ -276,13 +276,13 @@ func expandNode(hash hashNode, n node) node {
...
 }
 
-// NewDatabaseWithCache creates a new trie database to store ephemeral trie content
+// NewDatabaseWithBLSKeyCache creates a new trie database to store ephemeral trie content
 // before its written out to disk or garbage collected. It also acts as a read cache
 // for nodes loaded from disk.
-func NewDatabaseWithCache(diskdb ethdb.KeyValueStore, cache int) *Database {
+func NewDatabaseWithBLSKeyCache(diskdb ethdb.KeyValueStore, cache int) *Database {
```
mcortesi: ??????????????

consensus/istanbul/backend/snapshot.go
```diff
@@ -63,13 +64,19 @@ func loadSnapshot(epoch uint64, db ethdb.Database, hash common.Hash) (*Snapshot,
 	if err := json.Unmarshal(blob, snap); err != nil {
 		return nil, err
 	}
+
+	if !snap.ValSet.HasBLSKeyCache() {
```
mcortesi, line 68: why fail? why not just cache it here? cc @kobigurk

consensus/istanbul/validator/default.go
```diff
@@ -237,11 +283,20 @@ func (valSet *defaultSet) RemoveValidators(removedValidators *big.Int) bool {
 func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 	valSet.validatorMu.RLock()
 	defer valSet.validatorMu.RUnlock()
-	newValSet := NewSet(MapValidatorsToData(valSet.validators))
+	newValSet := NewSetFromDataWithBLSKeyCache(MapValidatorsToDataWithBLSKeyCache(valSet.validators))
```
mcortesi, line 286: i'd suggest we instead have `Validator.Copy()` that will copy the uncompressed value if it has one. And so, this method doesn't really have to handle with the cache.

And then you just do:
```go
	return &defaultSet{
    validators: copiedValidators,
    randomness: randomness
  }
```

If not, what we are doing is actually transforming all validators to data, and then back again to valiators.. no sense on doing that

params/protocol_params.go
kobigurk: @mrsmkl's reasoning was that since it's cached, the cost is the same as `GetValidatorGas`, because it's just reading a value from a similar place.

consensus/istanbul/backend/snapshot.go
kobigurk: Caching + storing makes sense to me...

crypto/bls/bls.go
mrsmkl: I'll just pass the error instead. The only place this value is currently used is in the precompile, where it checks that the length is correct (note that `len(nil) == 0`).

params/protocol_params.go
mrsmkl: The benchmarking code is in this repo: https://github.com/mrsmkl/b12-sol/

consensus/istanbul/backend/snapshot.go
mrsmkl: It fails so that it's cached in the DB.

consensus/istanbul/validator.go
```diff
@@ -48,8 +48,10 @@ type Validator interface {
 	// Address returns address
 	Address() common.Address
 
+	// BLSPublicKey returns the BLS public key (compressed format)
 	BLSPublicKey() blscrypto.SerializedPublicKey
 
+	// BLSPublicKey returns the BLS public key (uncompressed format)
```
mcortesi, line 54: typo

mcortesi (approved):
Great for going over all my comments!!! Looks much better now. There's a typo on one godoc that missing, but we're ready to go

PR #1313, opened 2021-01-25
Command to create genesis block with all migration on it

mcortesi:
### Description

This PR adds a new command `mycelo` that will create a `genesis.json` file compatible with `celo-blockchain`. This genesis will already have all the migrations and contracts already on it. Thus, it eliminates the need to run the migrations altogheter.

To do this, the PR creates an in memory `stateDb` and runs EVM calls on it.

TODO complete

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


mcortesi:
replaced by #1377

PR #1336, opened 2021-01-29
Better gas estimation from upstream (partial merge from 1.9.14)

oneeman:
### Description

This includes some more merges from upstream which are at the beginning of the 1.9.14 work.  Opening up a PR for it, even though there's a lot more left from 1.9.14, because it includes the gas estimation API improvement which we want for the mycelo feature @mcortesi has been working on.  The upstream commits can be seen at: https://github.com/ethereum/go-ethereum/commits/master?after=6d74d1e5f762e06a6a739a42261886510f842778+69&branch=master .  This PR includes commits from `7a63faf` (April 16, 2020) to `2ec7232` (April 22)

My merge notes, highlighting points of interest:

- `648b0cb` removes flags to override istanbul and muir glacier, as well as the `SetupGenesisBlockWithOverride()` function that uses them to actually override the config.  I followed upstream in removing `SetupGenesisBlockWithOverride()`.  If we want to add an override flag for a future hard-fork, we can reinstitute it using the same pattern.
- `c60c0c9` upgrades the fastcache dependency, which allowed upstream to remove cespane/xxhash/v2 as an indirect dependency (since it's instead explicitly required by fastcache).  We are using xxhash (but v1) in `validator_assignment.go`, but there is no reason to use v1 rather than v2.  So, in a separate commit, I switched to using v2 which is newer and matches what fastcache uses.
- `b9df7ec` bundles a few changes: (a) it refactors the EVM message processing API to use a new `ExecutionResult` type containing the gas used, return data, and any EVM error, (b) distinguishing between EVM execution errors (e.g. running out of gas or a revert), which are part of `ExecutionResult` (and which now include the revert reason if there is one), and pre-execution consensus errors (e.g. incorrect nonce, insufficient funds for fee, etc.), which are turned as errors from `TransitionDb()`, (c) uses the above changes to improve the gas estimation API by returning different errors when the problem is running out of gas vs the transaction always failing (and returning the reason for it failing, including the revert reason if there is one)
    - Conflict summary: Many conflicts, mostly (a) around various error definitions which were scattered being consolidated in  `core/error.go` and `core/vm/errors.go` (and in some cases being converted to public), and (b) `core/state_transition.go`
    - The error `insufficientBalanceForGas` has been replaced by `ErrInsufficientFunds` (which encompasses not enough balance for value and paying for gas).  In our case we instead have `errInsufficientBalanceForFees`.  Since it might be helpful to distinguish the error when the problem is not enough CELO vs not enough of another fee currency, I decided to keep this error (and move it to `core/error.go`), but renamed it to say Funds instead of Balance in a separate commit, to be consistent with the other errors are insufficient funds.
        - However, I found that there is already `ErrNonWhitelistedFeeCurrency` in `core/tx_pool.go`.  Since `tx_pool.go` already uses other errors from `core/error.go` (e.g. ErrNonceTooLow), I decided to deduplicate by removing the one in `tx_pool.go` in favor of the one in `core/error.go`.
    - The main place with non-trivial conflicts was `TransitionDb()` in `core/state_transition.go`, and this warrants detailed code review of this file's before and after, due to the critical nature of the code here.
        - I moved `ErrGasPriceDoesNotExceedMinimum`, `errInsufficientBalanceForFees`, and `errNonWhilelistedFeeCurrency` to `core/error.go`, which is where the errors from pre-evm-execution checks are now located, and made them public like the others.
        - I noticed that the 'out of gas' errors in `IntrinsicGas()` are actually uint64 overflows (because in upstream they changed the error to reflect that), and that after them we add to them using `GetIntrinsicGasForAlternativeFeeCurrency()`, but don't have an overflow check for that, so I added one in a separate commit.
        - our `preCheck()` differs in that it doesn't buy gas, but does check gas price against the gas price minimum, and the gas is bought later in `payFees()`
    - In `core/vm/emv.go`, we are importing `"errors"` as `goErrors`.  This was originally to avoid a conflict with `contract_comm/errors`, but that is no longer imported here, so the alias isn't necessary.  The upstream commit also adds an import of "errors", so it seemed like a good time to remove the alias, which I did in a separate commit.
    
### Backwards Compatibility

Errors from `eth_call` and `eth_estimateGas` will now give more informative errors, including the revert reason if the result was an EVM revert.


mcortesi (approved):
Looks good to me!

I was surprised to find that `TransitionDb() (*ExecutionResult, error)` in case of a vm error returns a `nil` error. 

And thus, any caller of `ApplyMessage()` should check (if desired) `result.Failed()` to see if there was an error.

But that has been always the case, so no changes on that on this PR

oneeman:
Merged master into this branch and resolved the conflicts (`go.mod` and `go.sum`)

PR #1342, opened 2021-02-03
Change module name to `celo-org/celo-blockchain`

lukasz-zimnoch:
This draft pull request is intended to open a discussion about changing the module name from `ethereum/go-ethereum` to `celo-org/celo-blockchain`. The main motivation behind that change is opening a possibility for seamless integration of `celo-blockchain` with projects already depending on the original `go-ethereum` module. Specifically, a project with a single dependency can do a single `replace` in their `go.mod` file but a project already having a dependency to `ethereum/go-ethereum` can not import Celo without this change.


mcortesi:
closed in favor of https://github.com/celo-org/celo-blockchain/pull/1375

PR #1343, opened 2021-02-03
Add a cache to StateProcessor

### LINKED ISSUE #1253: Remove redundant block proposal processing when adding block to local blockchain
### Expected Behavior

A validator will process a block 2 or 3 times (3 times when it's the proposer, and twice for all other validators).  The processing instances are:

1). When the block proposal is created (only the block proposer does this).
2). When the block proposal is verified.
3). When the block is inserted into the local blockchain.

The validators should only process a block once and cache the resulting state.  Then use that cached state when inserting the block into the local blockchain.

### Current Behavior

The validators process the block multiple times as described in the previous section.

### System information

This occurs for all version <= 1.2.2.  (As of Dec 3. 2020.  This may need to be updated until this issue is released into a version).

YoduYodu:
### Description

Implement a CachingStateProcessor to avoid redundancy.
Document:
https://www.notion.so/Block-Processing-Optimization-9c997b2fe2e04b879df3cd261c3ed6d8

- [x] Add a cache to StateProcessor to avoid redundant processing.
- [x] Replace `miner/worker.go`'s local cache `pendingTasks`
- [ ] Add unit tests

### Other changes

Renaming & removing unused code

### Tested

Pending

### Related issues

- Fixes #1253 

### Backwards compatibility

Yes

### DIFF
--- a/consensus/consensustest/mockprotocol.go
+++ b/consensus/consensustest/mockprotocol.go
@@ -373,7 +373,7 @@ func (e *MockEngine) SealHash(header *types.Header) (hash common.Hash) {
 func (e *MockEngine) Seal(chain consensus.ChainReader, block *types.Block, results chan<- *types.Block, stop <-chan struct{}) error {
 	header := block.Header()
 	select {
-	case results <- block.WithSeal(header):
+	case results <- block.WithHeader(header):
 	default:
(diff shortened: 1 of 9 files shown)


mcortesi:
Closed in favor of #1394

PR #1349, opened 2021-02-08
Adds script to build and publish a private docker image

trianglesphere:
### Description

Adds script to build and publish a private docker image
The image is uploaded to gcr.io/celo-testnet which is only accessible to cLabs.

### DIFF
--- a/scripts/publish_private_docker_image.sh
+++ b/scripts/publish_private_docker_image.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+COMMIT_SHA=$(git rev-list  HEAD | head -n 1)
+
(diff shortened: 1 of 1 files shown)


oneeman (approved):
Very nice!  I'm wondering about the permissions for the push.  Presumably this will give you an error if you don't have the permissions, which is not bad.  If there were an easy way to check the permissions first before doing the push, that'd be best.  But from some quick googling I haven't found anything.

For PRs, if the template doesn't fit, just delete the template sections rather than leaving them in template form.

trianglesphere:
Yea, it'd be nicer to have an explicit permission check, but it's probably fine as is.

PR #1375, opened 2021-02-10
Rename module to `celo-blockchain`

### LINKED ISSUE #1344: Rename our packages to not be called github.com/ethereum/go-ethereum/*
## Description

In `go.mod`, rename our package to:

```
module github.com/celo-org/celo-blockchain
```

And update all the imports which say `github.com/ethereum/go-ethereum` to say `github.com/celo-org/celo-blockchain`.

Additionally, we should have some protection against accidentally importing anything from go-ethereum due to new imports added as part of a merge.  This could be part of CI, or maybe part of the build.  The key point is to ensure we are not pulling in go-ethereum packages.  Probably grepping `go.mod` would be sufficient, but this needs to be verified.

Or, as suggested by @trianglesphere, there may be a way to automate the renaming of imports using gofmt.  See https://spf13.com/post/go-fmt/

## Context

The current naming, besides being inaccurate, prevents third-party projects from adding both go-ethereum and celo-blockchain as dependencies if they need to.

trianglesphere:
### Description

This changes the module name from ethereum/go-ethereum to celo-org/celo-blockchain. The main motivation behind that change is opening a possibility for seamless integration of celo-blockchain with projects already depending on the original go-ethereum module. Specifically, a project with a single dependency can do a single replace in their go.mod file but a project already having a dependency to ethereum/go-ethereum can not import Celo without this change.

This PR was initially opened by @lukasz-zimnoch of keep. I made some adjustments to docs and merged in master.

### Tested

N/A

### Related issues

- Fixes #1344

### Backwards compatibility

External contributors will need to change their import path. Some temporary hiccups while developers move everything over.


interfaces.go
```diff
@@ -22,8 +22,8 @@ import (
 	"errors"
 	"math/big"
```
lukasz-zimnoch, line 24: The main package will be still `ethereum`. Can we also rename it to `celo` (line 18)?

trianglesphere: Yep

oneeman: I'm less enthusiastic about this change.  The resulting conflicts in code when merging form upstream will not be trivial like ones from imports (and code changes more often than imports).  Would leaving the alias as ethereum cause any problems?  After all, any other project that imports from celo-blockchain can alias it to celo themselves.

trianglesphere: Ok, we're renaming it externally as `celo`, but keeping it as `ethereum` internally. Practically the internal rename should have no effect on other users.

interfaces.go
```diff
@@ -15,15 +15,15 @@
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
 // Package ethereum defines interfaces for interacting with Ethereum.
```
nategraf, line 17:
```suggestion
// Package celo defines interfaces for interacting with Celo.
```

nategraf (approved):
When I saw this PR, I thought to myself: "Ah! The perfect PR to hide a backdoor! No one would ever expect a reviewer to look at every file!" And so I did! Ha! No exploits....

I did find one place where we should replace "Ethereum" with "Celo", so theres that

PR #1380, opened 2021-02-12
Improves load bot

trianglesphere:
### Description

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


cmd/mycelo/main.go
```diff
@@ -426,11 +444,12 @@ func loadBot(ctx *cli.Context) error {
...
-		Accounts:         env.DeveloperAccounts(),
-		Amount:           big.NewInt(10000000),
-		TransactionDelay: 1 * time.Second,
+		Accounts:              env.DeveloperAccounts(),
+		Amount:                big.NewInt(10000000),
+		TransactionsPerSecond: env.Config.LoadTestTPS,
+		ClientCount:           1,
 		ClientFactory: func() (*ethclient.Client, error) {
-			return ethclient.Dial("http://localhost:8545")
+			return ethclient.Dial(env.IPC())
```
mcortesi, line 452: i used http since i didn't know the IPC support multiple connections or how well handles that

mycelo/env/paths.go
```diff
@@ -21,3 +21,7 @@ func (p paths) envJson() string {
 func (p paths) validatorDatadir(idx int) string {
 	return path.Join(p.Workdir, fmt.Sprintf("validator-%02d", idx))
 }
+
+func (p paths) validatorIPC() string {
```
mcortesi, line 25: i would use index here, for consistency

mycelo/loadbot/bot.go
```diff
@@ -60,19 +85,35 @@ func runBot(ctx context.Context, acc env.Account, sleepTime time.Duration, clien
...
+				fmt.Printf("Error sending transaction: %v\n", err)
+			}
 			return fmt.Errorf("Error sending transaction: %w", err)
 		}
-		fmt.Printf("cusd transfer generated: from: %s to: %s amount: %s\ttxhash: %s\n", acc.Address.Hex(), recipient.Hex(), value.String(), tx.Transaction.Hash().Hex())
+		// fmt.Printf("cusd transfer generated: from: %s to: %s amount: %s\ttxhash: %s\n", acc.Address.Hex(), recipient.Hex(), value.String(), tx.Transaction.Hash().Hex())
 
-		printJSON(tx)
+		// printJSON(tx)
+		_, err = tx.WaitMined(ctx)
```
mcortesi, line 100: i 'm not fully reviewing the whole logic.... one idea i had for this is to have another goroutine that will fetch every block, and inspect the txs...  if you have a "sentTxs" and then you check the block's tx you can easily see when they are mined... and you do all of that in a single routine...

to learn about sentTxs, every tx produce can send them to a channel.

cmd/mycelo/main.go
trianglesphere: The answer is that IPC is a lot better than HTTP. HTTP works well up to maybe 10 or 20 TPS and then just falls over. I've also done work to limit the number of clients used (I had to tune that with HTTP, but one IPC client supports as many TPS as needed).

mycelo/loadbot/bot.go
trianglesphere: I'm not really sure what the idea is. This is also setup to deal with the case that the TPS is set higher than the network can handle (say when the block gas limit is below what the desired TPS should be).

trianglesphere:
closing for #1382 b/c of rebases

PR #1381, opened 2021-02-12
Disable reorgs (also no need for sidechains)

### LINKED ISSUE #1190: Disable Reorgs (quick version)

gastonponti:
### Description

Disable reorgs. As collateral, no sidechains are needed anymore

### Other changes

No sidechains

### Tested

Add a few tests, to check if in the case of being necessary, to set the head to a previous cannonical block, and add a different block with that block as a parent (this was a particular case of the reorg, but in our case, it would be only needed in the case of an idle chain, or a 1/3+1 attack) 

### Related issues

- Fixes #1190 

### Backwards compatibility

Yes


core/blockchain.go
```diff
@@ -72,8 +71,8 @@ var (
 	blockValidationTimer = metrics.NewRegisteredTimer("chain/validation", nil)
 	blockExecutionTimer  = metrics.NewRegisteredTimer("chain/execution", nil)
 	blockWriteTimer      = metrics.NewRegisteredTimer("chain/write", nil)
-	blockReorgAddMeter   = metrics.NewRegisteredMeter("chain/reorg/drop", nil)
-	blockReorgDropMeter  = metrics.NewRegisteredMeter("chain/reorg/add", nil)
+	blockReorgAddMeter   = metrics.NewRegisteredMeter("chain/reorg/add", nil)
```
mcortesi, line 74: nice catch!

core/blockchain.go
```diff
@@ -301,7 +300,9 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	// The first thing the node will do is reconstruct the verification data for
 	// the head block (ethash cache or clique voting snapshot). Might as well do
 	// it in advance.
-	bc.engine.VerifyHeader(bc, bc.CurrentHeader(), true)
+	if err := bc.engine.VerifyHeader(bc, bc.CurrentHeader(), true); err != nil {
```
mcortesi, line 303: nice catch bis!

core/blockchain.go
```diff
@@ -1276,9 +1277,10 @@ func (bc *BlockChain) writeKnownBlock(block *types.Block) error {
 
 	current := bc.CurrentBlock()
 	if block.ParentHash() != current.Hash() {
-		if err := bc.reorg(current, block); err != nil {
-			return err
-		}
+		// if err := bc.reorg(current, block); err != nil {
+		// 	return err
+		// }
+		return errors.New("Not saving block. Reorgs are disabled")
```
mcortesi, line 1283: externalize the error into a module variable, and reuse.

Also, i think linting expecte error messages to start with lowercase

core/blockchain.go
```diff
@@ -1298,13 +1300,14 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
+	if hash := bc.GetCanonicalHash(block.NumberU64()); (hash != common.Hash{} && hash != block.Hash()) {
+		log.Error("Found two blocks with same height", "old", hash, "new", block.Hash())
+		return NonStatTy, errors.New("Not saving block. Reorgs are disabled")
```
mcortesi, line 1305: same

core/blockchain.go
```diff
@@ -1610,8 +1623,8 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 	// First block is pruned, insert as sidechain and reorg only if TD grows enough
 	case err == consensus.ErrPrunedAncestor:
```
mcortesi, line 1624: when can this happen?

core/blockchain.go
```diff
@@ -1758,6 +1771,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 				"elapsed", common.PrettyDuration(time.Since(start)),
 				"txs", len(block.Transactions()), "gas", block.GasUsed(),
 				"root", block.Root())
+			log.Error("We are not handling Side chains")
```
mcortesi, line 1774: maybe we should remove `SideStatTy` since it seems it's not possible anymore

core/blockchain.go
```diff
@@ -1800,8 +1814,8 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 // switch over to the new chain if the TD exceeded the current chain.
 func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator) (int, error) {
```
mcortesi, line 1815: does this methos makes sense now?

core/blockchain.go
```diff
@@ -1298,13 +1300,14 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
+	if hash := bc.GetCanonicalHash(block.NumberU64()); (hash != common.Hash{} && hash != block.Hash()) {
```
mcortesi, line 1303: shouldn't insertChain have a similar check? Thinking here is that we want to avoid processing the block if it's already non cannonical.

core/headerchain.go
```diff
@@ -145,10 +145,15 @@ func (hc *HeaderChain) WriteHeader(header *types.Header) (status WriteStatus, er
...
 		hash     = header.Hash()
 		number   = header.Number.Uint64()
-		localTd  *big.Int
 		externTd *big.Int
+		// localTd  *big.Int
 	)
 
+	if cannonicalHash := hc.GetCanonicalHash(number); (cannonicalHash != common.Hash{} && cannonicalHash != hash) {
+		log.Error("Found two headers with same height", "old", cannonicalHash, "new", hash)
+		return NonStatTy, errors.New("Not saving header. Reorgs are disabled")
```
mcortesi, line 154: externalize error

mcortesi (commented):
left some comments

core/blockchain.go
gastonponti: removing the reorg func, this is not used anymore and the lint fails, so I've removed it

core/blockchain.go
gastonponti: Removed, also with the reorg method

core/blockchain.go
gastonponti: This happens when it tries to verify the block header's transaction, it has the header but not the state.
To be honest, I didn't know if it's possible to not have the state of a really old block, but what I did is to cut the chain insertion in that moment.
That case was calling to:
`return bc.insertSideChain(block, it)`
which I removed, but, you are right, maybe I could still throw the error, and not just cut the iteration

core/blockchain.go
gastonponti: Removed

core/blockchain.go
gastonponti: The insertChain, basically skip all the blocks that are already in the db, and then, starts to apply a `bc.writeKnownBlock(block)` to every block, that knows, but the head is pointing to a smaller block (this could be due to a rollback). This `writeKnownBlock` checks if the parent is a canon one.
As this parts of the basis that the old rolled back block was written, it means that it already passed for the validation that at the same level the block is the only one available (the new canon), that's why we are not checking it. It's basically a head update in the case of sending the same chain after a rollback, and at that level it can't exist a canon because the last canon is the parent that you are checking.
After that, the chainInsert, with the unknown blocks, calls the `WriteBlockWithState`
which checks that for the same level, no canon already exists.

So, TL;DR; we are checking that, because we are using the same function inside insertChain

core/blockchain.go
gastonponti: Re added but just for the case of importing a chain, with previously pruned blocks

core/blockchain_repair_test.go
```diff
@@ -132,252 +126,10 @@ func TestShortFastSyncingRepair(t *testing.T) {
 	// Expected head block     : G
 	testRepair(t, &rewindTest{
 		canonicalBlocks:    8,
-		sidechainBlocks:    0,
 		freezeThreshold:    16,
 		commitBlock:        0,
 		pivotBlock:         uint64ptr(4),
 		expCanonicalBlocks: 8,
-		expSidechainBlocks: 0,
```
asaj, line 140: Why do these values need to be deleted?

core/blockchain_sethead_test.go
```diff
@@ -248,280 +220,11 @@ func TestShortFastSyncingSetHead(t *testing.T) {
 	// Expected head block     : G
 	testSetHead(t, &rewindTest{
 		canonicalBlocks:    8,
-		sidechainBlocks:    0,
-		freezeThreshold:    16,
```
asaj, line 252: Ditto, why remove anything other than `sidechainBlocks` and `expSideChainBlocks`?

core/headerchain.go
```diff
@@ -145,10 +145,14 @@ func (hc *HeaderChain) WriteHeader(header *types.Header) (status WriteStatus, er
 	var (
 		hash     = header.Hash()
 		number   = header.Number.Uint64()
-		localTd  *big.Int
 		externTd *big.Int
 	)
 
+	if cannonicalHash := hc.GetCanonicalHash(number); (cannonicalHash != common.Hash{} && cannonicalHash != hash) {
```
asaj, line 151: Can you adjust the function comment? It talks a lot about re-orgs and side chains which ofc is no longer relevant

core/headerchain.go
```diff
@@ -175,71 +175,55 @@ func (hc *HeaderChain) WriteHeader(header *types.Header) (status WriteStatus, er
...
-	// If the total difficulty is higher than our known, add it to the canonical chain
-	// Second clause in the if statement reduces the vulnerability to selfish mining.
-	// Please refer to http://www.cs.cornell.edu/~ie53/publications/btcProcFC.pdf
-	reorg := externTd.Cmp(localTd) > 0
-	if !reorg && externTd.Cmp(localTd) == 0 {
-		if header.Number.Uint64() < head {
-			reorg = true
-		} else if header.Number.Uint64() == head {
-			reorg = mrand.Float64() < 0.5
+	// If the header can be added into canonical chain, adjust the
```
asaj, line 178: I don't understand what is going on here, can you help me understand?

core/headerchain.go
```diff
@@ -175,71 +175,55 @@ func (hc *HeaderChain) WriteHeader(header *types.Header) (status WriteStatus, er
...
-		for i := number + 1; ; i++ {
-			hash := rawdb.ReadCanonicalHash(hc.chainDb, i)
-			if hash == (common.Hash{}) {
+	// Overwrite any stale canonical number assignments
+	var (
+		headHash   = header.ParentHash
+		headNumber = header.Number.Uint64() - 1
+		headHeader = hc.GetHeader(headHash, headNumber)
+	)
+	for rawdb.ReadCanonicalHash(hc.chainDb, headNumber) != headHash {
```
asaj, line 199: What is this for loop doing? Wouldn't we always expect to be adding on to HEAD? Why do we need to iterate backwards through the chain?

core/blockchain_repair_test.go
gastonponti: We are explicitly failing if something is inserted as a sidechain, and these checks, basically ask for the amount of blocks as a sidechain.
I thought of having all of those with 0, but, as I'm trying to remove the concept of sidechain, if I let those, in the future I think that will only generate some noice. This way it's a clearer test

core/blockchain_sethead_test.go
gastonponti: That's git showing some deletes in a weird manner.
Basically, we had tests with only the canonical chain, and others that were the same, but adding the sidechain.
I removed those with the sidechain, and github instead of remove the entire test, cuts the first part of this one, and paste the last part of the other. But basically is removing the unused sidechain parameters, and then github collapsing things in a strange way

core/blockchain_test.go
```diff
@@ -1998,6 +1171,119 @@ func testInsertKnownChainData(t *testing.T, typ string) {
...
+	testInsertDifferentChainDataAfterReset(t, "headers", 32, 64)
+}
+func TestInsertDifferentReceiptChainAfterReset(t *testing.T) {
+	testInsertDifferentChainDataAfterReset(t, "receipts", 32, 64)
+}
+func TestInsertDifferentBlocksAfterReset(t *testing.T) {
+	testInsertDifferentChainDataAfterReset(t, "blocks", 32, 64)
+}
+
+func testInsertDifferentChainDataAfterReset(t *testing.T, typ string, setHeadBlock int, chainsBlocksNumber int) {
```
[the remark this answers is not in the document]
gastonponti: I'll add a comment yes.

Basically I'm testing if I can rollback our chain to a specific block, and create a new canonical from there. This covers the scenarios where we must set the head behind, and continue with a new chain with everything working (for example if we had a 1/3+1 attack and we want to "re-do" the epoch with different validators)

piersy:
Hey @gastonponti the CI coverage step is failing with a very relevant looking failure.

```
--- FAIL: TestForkIDSplit (0.50s)
    protocol_test.go:240: fork ID rejection error mismatch: have p2p: read or write on closed message pipe, want Fork ID rejected - local incompatible or needs update
FAIL
```

core/blockchain.go
```diff
@@ -72,14 +71,14 @@ var (
...
-	blockReorgAddMeter   = metrics.NewRegisteredMeter("chain/reorg/drop", nil)
-	blockReorgDropMeter  = metrics.NewRegisteredMeter("chain/reorg/add", nil)
 
 	blockPrefetchExecuteTimer   = metrics.NewRegisteredTimer("chain/prefetch/executes", nil)
 	blockPrefetchInterruptMeter = metrics.NewRegisteredMeter("chain/prefetch/interrupts", nil)
 
 	errInsertionInterrupted = errors.New("insertion is interrupted")
 	errCommitmentNotFound   = errors.New("randomness commitment not found")
+	errParentNotCanonical   = errors.New("parent not canonical, reorgs disabled")
+	errAlreadyCanonical     = errors.New("already a canonical block in same level, reorgs disabled")
```
piersy, line 81: I would suggest not using error values such as this, using error values tends to lead to having errors devoid of useful information.

In the case of errParentNotCanonical, what would be useful, would be to know the hashes and numbers of the expected canonical parent and the actual canonical head block. If you see this popping up in the logs that information will massively narrow down the potential scope of the problem.

The only argument for using error values is if we wanted to check for them explicitly, but these are never checked. Also there are problems with explicitly checking error values, this [article](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) explains the issues around using error values in depth and suggests better alternatives.

In this case I just removing these values and returning an error that contains relevant contextual information would be fine.

core/blockchain.go
```diff
@@ -301,7 +299,10 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	// The first thing the node will do is reconstruct the verification data for
 	// the head block (ethash cache or clique voting snapshot). Might as well do
 	// it in advance.
-	bc.engine.VerifyHeader(bc, bc.CurrentHeader(), true)
+	currentHeader := bc.CurrentHeader()
+	if err := bc.engine.VerifyHeader(bc, currentHeader, true); currentHeader.Number.Cmp(big.NewInt(0)) > 0 && err != nil {
```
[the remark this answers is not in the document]
piersy: Rather than modifying the production code so that tests pass, I think it would be preferable to update the tests so that they run with valid data. Is that something we could do?

It seems that this code as is would allow for a blockchain instance to be constructed with an invalid genesis block.

core/blockchain.go
```diff
@@ -942,7 +944,6 @@ type WriteStatus byte
 const (
 	NonStatTy WriteStatus = iota
 	CanonStatTy
```
piersy, line 946: I'm just wondering, what is the meaning of `NonStatTy` and `CanonStatTy`?

core/blockchain.go
```diff
@@ -1252,33 +1253,14 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
...
 func (bc *BlockChain) writeKnownBlock(block *types.Block) error {
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
 	current := bc.CurrentBlock()
 	if block.ParentHash() != current.Hash() {
-		if err := bc.reorg(current, block); err != nil {
-			return err
-		}
+		return errParentNotCanonical
```
piersy, line 1263: As per my previous comment about error values, I think something like this would make a much more useful error.

```suggestion
		return fmt.Errorf("failed to write block %s, expected parent hash %s, current head %s",
			block.ShortString(),
			block.ParentHash().ShortString(),
			current.ShortString())
```
Supporting functions.
```go
// ShortString returns a hex representation of the first 8 bytes of the hash. Its intent is to aid debugging by providing enough of the hash that it can be easily reconciled with logs containing the full hash but is also small enough to keep the logs human readable.
func (h Hash) ShortString() string { return hexutil.Encode(h[:8]) }

// ShortString returns the block number and short hash separated b a colon.
func (b *Block) ShortString() string { return fmt.Sprintf("%d:%s", b.NumberU64(), b.Hash().ShortString()) }
```

core/blockchain_test.go
```diff
@@ -1998,6 +1171,119 @@ func testInsertKnownChainData(t *testing.T, typ string) {
...
+
+	engine := mockEngine.NewFaker()
+
+	db := rawdb.NewMemoryDatabase()
+	genesis := new(Genesis).MustCommit(db)
+
+	originalBlocks, originalReceipts := GenerateChain(params.TestChainConfig, genesis, engine, db, chainsBlocksNumber, func(i int, b *BlockGen) {
+		b.SetCoinbase(common.Address{1})
+	})
+	// Total difficulty is higher to create a different chain.
```
piersy, line 1195: How is the total difficulty higher here, and why is it important?

core/blockchain_test.go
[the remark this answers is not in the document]
piersy:
```suggestion
// testInsertDifferentChainDataAfterReset checks that after rolling back the chain to a specific block
// a new canonical chain can be imported.
func testInsertDifferentChainDataAfterReset(t *testing.T, typ string, setHeadBlock int, chainsBlocksNumber int) {
```

piersy:
Hey @gastonponti I left a few comments/suggestions.

core/blockchain_test.go
```diff
@@ -1736,144 +1042,6 @@ func TestIncompleteAncientReceiptChainInsertion(t *testing.T) {
...
-		header = chain.GetHeader(header.ParentHash, number-1)
-	}
-}
-
-// Tests that importing a sidechain (S), where
-// - S is sidechain, containing blocks [Sn...Sm]
-// - C is canon chain, containing blocks [G..Cn..Cm]
-// - A common ancestor is placed at prune-point + blocksBetweenCommonAncestorAndPruneblock
-// - The sidechain S is prepended with numCanonBlocksInSidechain blocks from the canon chain
-func testSideImport(t *testing.T, numCanonBlocksInSidechain, blocksBetweenCommonAncestorAndPruneblock int) {
```
piersy, line 1800: Do we have a test that verifies that attempts to import a side chain now fail?

core/blockchain.go
gastonponti: CanonStatTy is that the block added is part of the cannonical chain
NonStatTy is used for the blocks that had some error
After this PR there was the SideStatTy and even after, a SplitStatTy

I could remove both, and just change the state for a boolean (valid or not) but I thought that it would be easier to follow, thinking in the merges from upstream

core/blockchain_test.go
gastonponti: Haha, yes, sorry
Changing the comment to:
`// Block offsetTime different to create a different chain.`
Basically if I didn't change something in the generation of the block, it generates the same exactly chain

core/blockchain_test.go
gastonponti: Yes, in ll the `testInsertKnownChainData`
check the line :1152 of this file (https://github.com/celo-org/celo-blockchain/pull/1381/files#diff-d78b498ba524953c810b53b2f1e866245d052a063863ed820c9b4852f8762d69R1152 I think is the way to link it)

core/blockchain.go
gastonponti: Agree. Changing it with your other recommendation. Thanks!

gastonponti:
> Hey @gastonponti the CI coverage step is failing with a very relevant looking failure.
> 
> ```
> --- FAIL: TestForkIDSplit (0.50s)
>     protocol_test.go:240: fork ID rejection error mismatch: have p2p: read or write on closed message pipe, want Fork ID rejected - local incompatible or needs update
> FAIL
> ```

Mmmmmm, run it locally and it worked, then added a commit with a lint fix and it worked. Have to check if it is a flaky thing that happened before or not.

core/blockchain_test.go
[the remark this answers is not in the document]
gastonponti: adding it

core/headerchain.go
gastonponti: Agree. Headerchain also refactored to avoid unnecessary checks

mcortesi:
Left a few comments, i'm going to deep dive into blockchain.go next

mcortesi (commented):
the comments, finally

cmd/geth/genesis_test.go
```diff
@@ -17,35 +17,39 @@
...
...43c1bde8a2e1d555d911cd1d594bc236d14fbbe74b4d3bc132309b861dcf1c527c594ab3233a00d1e03b3e0c7c9912421b56f498901ef9421d87e9445d3b355807c91937e70e6f9914f2a1d94f6a964ad845a8a8a4a50b884c140ab8c3ed8252cf901eab86056631e24a6ec7fb89b5d967b6e7169eed00c0d802b9050d5a749b9d69a2c70d047b5add1d78646c4c3d579e7da7691001f0c832e9b26dd02b960b610ebfa63cd3a9dc0d6da8abfd8c16ffc5679debc682216c2934af7dc93d82e4a42564b4d80b86065be0e12d1ac62a4ddf326cb0dde804d8c1aab376dff994c9db706c6a5c0d5ba57319105491658f9ae5d5a4f48752b006137148123d8a526b2237d7284d5f0ab664f186ef0e7649d586a9754f54ca042539b3ba7ea25d0209f639cce30262081b860f1dade6f52a125a236a546a5bf5468ef8c95da980a51ee2ea595919e80bb56b3941bd17315b2681a411b6f7b6a2aaa01815a62fff57a14cae0cbef5a540dbd34f098ae18c07f93137eefc25132ac1971c8e74f2ddf24ceeeece87dcd18f19500b8601e4c01b96874cbc25fe98a4a8300035865d02724ec7b62d1662eba6b49777aaad51b73eb228d136d8d1f436e391e7e01d2c597cfcc17465a3aa1951d610360365bb116ab759887d5a74064c663aeeb3facf55ddc9a212e9f06b9925d27011700b860306ff63a1746b45b891a33d8ad7a02c92bf0ec628499f5308c55b15e7b5a2ab3d97113ef669c5ae446cb69b965c28d0056c780a4f2462b989bcb15380cc71fcf163d7e7be97c8c2a70b72e8273ff87f7d249f3c552fe7ecc28331f4ce90d92808080c3808080c3808080"
```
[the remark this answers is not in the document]
gastonponti: That was an extraData generated by mycelo. As the block insertion now checks that is a valid block, our genesis block, requires a valid extraData field.

core/blockchain.go
mcortesi: Seems to me that this check is already being done on `InsertChain` and thus this check should be moved to `WriteBlockWithState()` since that's the only other caller of this private method

core/blockchain.go
```diff
@@ -1434,13 +1417,17 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
+	if hash := bc.GetCanonicalHash(block.NumberU64()); (hash != common.Hash{} && hash != block.Hash()) {
+		log.Error("Found two blocks with same height", "old", hash, "new", block.Hash())
+		return NonStatTy, fmt.Errorf(
+			"fail to insert block %s: already a canonical block %s in same level",
+			block.ShortString(),
```
mcortesi, line 1424: 👎 shortstrings

core/blockchain.go
```diff
@@ -1551,58 +1536,22 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
...
-		}
-		status = CanonStatTy
-	} else {
-		status = SideStatTy
-	}
-	// Set new head.
-	if status == CanonStatTy {
-		bc.writeHeadBlock(block)
-	}
+	bc.writeHeadBlock(block)
```
mcortesi, line 1539: since we do this on a separate "atomic operation", it can happen that we save blokc, we save state, but we fail to mark it as canoncial.

mcortesi (commented):
More comments!

gastonponti:
@mcortesi, I don't totally agree with this one https://github.com/celo-org/celo-blockchain/pull/1381#discussion_r641848689 (It does not allow me to answer in the same thread)

The idea of having that check in the insertChain is to avoid processing unnecessary things that will change the cache and even part of the state, knowing in that moment, that we will eventually "cut" that.
It's more of a fail fast for the insertChain.
The check in the private `writeBlockWithState`, is in the exact place were that block will be saved, and basically perform the checks to avoid it if it's necessary. If I extract the check into the other public, I'll be avoiding one check, yes, but if someone uses this private one, will also require to perform another check.

What I'm trying to say, is that in my opinion, that check should be performed in the private function, and in the case of the insertChain function, as we want a fail check, we duplicate it for a fail fast scenario.

(I can extract that check in another private function to have both consuming the same)

PR #1382, opened 2021-02-16
Improves load bot

### LINKED ISSUE #1371: MyCelo: Specify TPS for load-bot instead of tx time or #developer accounts
## Description

Specify TPS for load-bot instead of tx time or #developer accounts.
Tx frequency is hardcoded to 1 second, and it runs a tx for every developer account at that rate.
Instead specify tx per second in the env file for the load bot, this make tests more repeatable by having the same number of funded
accounts, but easy to vary the load on the system.

trianglesphere:
### Description

This makes several changes to make the load bot easier to use. 

* Use `--tps` to specify the target transactions per second
*  The load bot will wait for txs to be mined before sending the next TX (this enables setting a TPS higher
    that what the block gas limit can support).
* Use `transferWithComment` and cUSD fee currency in transactions to use about 100k gas per tx
* Use a single IPC eth client by default. This handles high transactions load better than local HTTP clients.
* Adds `--blockgaslimit` to env creation commands.
* Use `--verbosity >= 4` with the load bot to get verbose transaction output.

### Other changes

Add validator IPC path from env functions.

### Tested

Tested load bot locally with 1 and multiple validators.

### Related issues

- Fixes #1371


mycelo/genesis/base_config.go
```diff
@@ -42,7 +42,7 @@ func BaseConfig() *Config {
 			Symbol:                      "cUSD",
 			Decimals:                    18,
 			Rate:                        fixed("1"),
-			InflationFactorUpdatePeriod: bigInt(2 * Year),
+			InflationFactorUpdatePeriod: bigInt(1 * Week),
```
mcortesi, line 45: why 1 week? i understand that we want to turnf inflation off, so 2 years is far away in the future

mycelo/loadbot/bot.go
```diff
@@ -40,42 +41,71 @@ func Start(ctx context.Context, cfg *Config) error {
...
+	// developer accounts / TPS = duration in seconds.
+	// Need the fudger factor to get up a consistent TPS at the target.
+	delay := time.Duration(int(float64(len(cfg.Accounts)*1000/cfg.TransactionsPerSecond)*0.95)) * time.Millisecond
+	startDelay := delay / time.Duration(len(cfg.Accounts))
+
+	for i, acc := range cfg.Accounts {
+		// Spread out client load across different diallers
+		client := cfg.Clients[i%len(cfg.Clients)]
+
+		err := waitFor(ctx, startDelay)
```
mcortesi, line 53: i would have this delay inside the worker go routine... looks cleaner

Also, if the first go routine ends with error, you should be able to notice it right away... but you only check that on `group.Wait()`, so you only check that after doing all the delays for every routine... (i'm not being 100% true here, since waitFor will notice it becaue of the `ctx` argument.. i think... but still shady logic)

mycelo/loadbot/bot.go
```diff
@@ -40,42 +41,71 @@ func Start(ctx context.Context, cfg *Config) error {
...
 
-func runBot(ctx context.Context, acc env.Account, sleepTime time.Duration, client bind.ContractBackend, nextTransfer func() (common.Address, *big.Int)) error {
+func runBot(ctx context.Context, acc env.Account, verbose bool, sleepTime time.Duration, client bind.ContractBackend, nextTransfer func() (common.Address, *big.Int)) error {
 	abi := contract.AbiFor("StableToken")
 	stableToken := bind.NewBoundContract(env.MustAddressFor("StableToken"), *abi, client)
 
 	transactor := bind.NewKeyedTransactor(acc.PrivateKey)
 	transactor.Context = ctx
 
+	stableTokenAddress := common.HexToAddress("0xd008")
```
mcortesi, line 74: you should use `env.MustProxyAddressFor("StableToken")`. bring changes from master, line 69 is broken here

mycelo/loadbot/bot.go
```diff
@@ -40,42 +41,71 @@ func Start(ctx context.Context, cfg *Config) error {
...
 
-		recipient, value := nextTransfer()
-		tx, err := stableToken.TxObj(transactor, "transfer", recipient, value).Send()
+		_, err = tx.WaitMined(ctx)
 		if err != nil {
-			return fmt.Errorf("Error sending transaction: %w", err)
+			if err != context.Canceled {
+				fmt.Printf("Error waiting for tx: %v\n", err)
+			}
+			return fmt.Errorf("Error waitin for tx: %w", err)
```
mcortesi, line 97: typo "waiting"

mycelo/loadbot/bot.go
```diff
@@ -40,42 +41,71 @@ func Start(ctx context.Context, cfg *Config) error {
...
+			return fmt.Errorf("Error sending transaction: %w", err)
+		}
+		if verbose {
+			fmt.Printf("cusd transfer generated: from: %s to: %s amount: %s\ttxhash: %s\n", acc.Address.Hex(), recipient.Hex(), value.String(), tx.Transaction.Hash().Hex())
+			printJSON(tx)
 		}
 
-		recipient, value := nextTransfer()
-		tx, err := stableToken.TxObj(transactor, "transfer", recipient, value).Send()
+		_, err = tx.WaitMined(ctx)
```
mcortesi, line 92: if you want for the tx to be mined, then you can't guarantee TPS. a tx will be mined on the next block at most, thus you can only have X transactions per block, where X is the number of dev accounts

why do you want to wait for them to be mined?

mcortesi: And also, it really depends on blocktime.. if blocktime is 5s then you can get X txs per 5 seconds

mycelo/env/types.go
```diff
@@ -17,5 +17,5 @@ type Config struct {
 	InitialValidators  int      `json:"initialValidators"`  // Number of initial validators
 	ValidatorsPerGroup int      `json:"validatorsPerGroup"` // Number of validators per group in the initial set
 	DeveloperAccounts  int      `json:"developerAccounts"`  // Number of developers accounts
-
+	LoadTestTPS        int      `json:"loadTestTPS"`        // Transactions per second to target in the load test
```
mcortesi, line 20: i don't get why we need this parameter, nor where this is used.

In any case, it doesn't seem to be a good fit for env.Config. I would put here things that are about the genessis block creation, accounts etc.... loadTestTPS seems to be a load-bot cmd only config variable

cmd/mycelo/main.go
```diff
@@ -422,19 +442,25 @@ func validatorRun(ctx *cli.Context) error {
...
-		Accounts:         env.DeveloperAccounts(),
-		Amount:           big.NewInt(10000000),
-		TransactionDelay: 1 * time.Second,
-		ClientFactory: func() (*ethclient.Client, error) {
-			return ethclient.Dial("http://localhost:8545")
-		},
+		Accounts:              env.DeveloperAccounts(),
+		Amount:                big.NewInt(10000000),
+		TransactionsPerSecond: env.Config.LoadTestTPS,
+		Clients:               []*ethclient.Client{client},
```
mcortesi, line 463: something i'm a bit dubious here is that i don't know if the ipc client connection is able to handle several concurrent requests; and if instead it just sequence requests.

RPC is a request -> response protocol. So if have 100 goroutines using the SAME ipc connection; are requests queued one after the other, or do they happen concurrently.

I suspect they are queued, and hence you need to wait for the first request to finish to process the second request.

That's why initially, i used multiple http connections, since every connection was independente, they can happen concurrently, and then it's a question on how many requests can the rpc server handle concurrently.

cmd/mycelo/main.go
```diff
@@ -110,6 +114,16 @@ var gethPathFlag = cli.StringFlag{
 	Usage: "Path to geth binary",
 }
 
+var loadTestTPSFlag = cli.IntFlag{
+	Name:  "tps",
+	Usage: "Transactions per second to target in the load test",
+}
+
+var verboseFlag = cli.BoolFlag{
```
mcortesi, line 122: notice that there's already a `mycelo --verbosity=5 load-bot` option. So might be confusing.

That is added on line 42 `debug.Setup()`

mcortesi (commented):
Added some comments, please clean up the PR description

cmd/mycelo/main.go
trianglesphere: The IPC client has been significantly more stable that HTTP connections (even when I would play with the number of connections). If you go 1:1 with HTTP connections to dev accounts, you can easily saturate the RPC server with too many connections (and you get `err connection refused`). Even limiting it to 50 or 100 HTTP clients, it'd run fine for a bit, and then eventually have an issue and be unable to send the transaction.

mycelo/genesis/base_config.go
trianglesphere: I'll move this to the load test template, it's to more closely replicate mainnet conditions (so gas usage per tx will match)

cmd/mycelo/main.go
trianglesphere: Switched to requiring `--verbosity 5` for verbose output

mycelo/env/types.go
trianglesphere: Removed

mycelo/loadbot/bot.go
trianglesphere: I've set up the default load test to have 10k accounts. I wait for the tx b/c if I specify a higher TPS than the block gas limit, I want to slow down the sending of transactions to be similar to the block gas limit.

mycelo/loadbot/bot.go
trianglesphere: I tried to implement this, but it significantly reduced the stability of the actual TPS to move it into the load bot function. Doing a proper start time might enable the go routine to be more accurate (`wait until`)

mycelo/genesis/base_config.go
mcortesi: are you sure that's the case? my guess is taht inflation is "off" o mainnet...

mycelo/loadbot/bot.go
mcortesi: An idea here, probably for the future is to have: `number of tx inflight` per account. So an account can stop sending tx when there's too many tx inflight.

you can do something like that, by watching blocks and reading the txhashes that were included.

cmd/mycelo/main.go
mcortesi: ok, a small change then is to use all the validators ipc connections. so you have a subset of clients connected to each validator

mcortesi (approved):
Approved! please fix PR description (remove template lines)

mycelo/genesis/base_config.go
trianglesphere: It's not quite off, but `updatePeriod: 1 year, 6 months, 2 days, 12 hours` on mainnet. I'll switch this back.

cmd/mycelo/main.go
trianglesphere: Yea, I can use all validator connections. During my tests I actually only run it with one validator, but that makes more sense.

PR #1385, opened 2021-02-18
Small fixes for mycelo validator-run command.

oneeman:
- Add --nodiscover flag (we're using static nodes)
- Use interrupt signal to quit geth, to prevent db corruption leading to stall on restart

### Description

Fixes a couple of problems I ran into while using mycelo testnets:

1. Since we aren't specifying either bootnodes nor `--nodiscover`, it tries to use the mainnet bootnodes.  This commit adds `--nodiscover` to prevent that
2. When quitting from validator-run, it was using a kill signal which meant the validators didn't have time to shut down cleanly.  This led to database corruption.  When starting again it would then rewind (on all validators), leading the network to stall permanently since the validators don't have the necessary state in their db.  This PR makes it instead use the interrupt signal, which lets the node shut down cleanly.

### Tested

- Confirmed that it's no longer using the bootnode
- Confirmed that the network stall on restart no longer happens

### DIFF
--- a/mycelo/cluster/node.go
+++ b/mycelo/cluster/node.go
@@ -156,6 +156,7 @@ func (n *Node) Run(ctx context.Context) error {
 		"--syncmode", "full",
 		"--mine",
 		"--allow-insecure-unlock",
+		"--nodiscover",
 		"--nat", "extip:127.0.0.1",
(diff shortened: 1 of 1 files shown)


trianglesphere (approved):
LGTM (I've actually got a local branch with the interrupt change)

PR #1386, opened 2021-02-18
mycelo: Fix load bot & add --init to validator-run

mcortesi:
### Description

3 small changes:

 * Add `--init` to `mycelo validator-run`; to avoid a call to mycelo
 * Fix load-bot: It was using the impl address instead of the proxy address
 * Rename `env.XXAddressFor` functions so it's clear if you're asking for the impl or the proxy address. 

### Tested

Run mycelo & loadbot


trianglesphere (approved):
LGTM if it's working when tested

PR #1394, opened 2021-02-24
Skip Block processing if already processed during consensus

### LINKED ISSUE #1253: Remove redundant block proposal processing when adding block to local blockchain

YoduYodu:
### Description

There are 3 different cases where a block is "processed" multiple times, the processing is computationally expensive so it can be improved for scalability.

1. When creating a proposal(a block), in `worker.go`.  (Proposer)
2. When verifying the proposal, either from itself as proposer, or from other validators as non-proposer during consensus, in preprepare.go::handlePreprepare() (proposer, non-proposer)
3. When adding a new block to the blockchain from fetcher & downloader, at blockchain.InsertChain() (proposer, non-proposer)

For non-proposers, we could cache the block processing in case 2, then avoid re-process it in case 3. This can save us duplicate block processing for once.

Detailed doc: https://www.notion.so/Duplicate-Block-Processing-9c997b2fe2e04b879df3cd261c3ed6d8

### Tested

CI tests ✅ 
[load test w/ 20 validators](https://console.cloud.google.com/cloud-build/builds;region=global/99403b72-07af-4f29-b70d-c99cb7e8f2aa?authuser=0&project=celo-testnet) ✅ 
[load test w/ 100 validators]
[multiversion test](https://console.cloud.google.com/cloud-build/builds;region=global/305e33bb-f025-4918-8129-03e43b461d24?authuser=0&project=celo-testnet) (master & this PR)✅ 

### Related issues

- Fixes #1253 

### Backwards compatibility

Yes


consensus/istanbul/core/core.go
```diff
@@ -377,7 +378,13 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
-		if err := c.backend.Commit(proposal, aggregatedSeal, aggregatedEpochValidatorSetSeal); err != nil {
+
+		result, isCached := c.current.GetStateProcessResult(proposal.Hash())
+		if !isCached {
```
mcortesi, line 383: in case we don't find in the cache (for example, if nodes goes offline and then online again); we should call `Verify()` again, instead of failing

consensus/istanbul/core/core.go
```diff
@@ -63,11 +64,11 @@ type CoreBackend interface {
 
 	// Commit delivers an approved proposal to backend.
 	// The delivered proposal will be put into blockchain.
-	Commit(proposal istanbul.Proposal, aggregatedSeal types.IstanbulAggregatedSeal, aggregatedEpochValidatorSetSeal types.IstanbulEpochValidatorSetSeal) error
+	Commit(proposal istanbul.Proposal, aggregatedSeal types.IstanbulAggregatedSeal, aggregatedEpochValidatorSetSeal types.IstanbulEpochValidatorSetSeal, stateProcessResult *core2.StateProcessResult) error
 
 	// Verify verifies the proposal. If a consensus.ErrFutureBlock error is returned,
 	// the time difference of the proposal and current time is also returned.
-	Verify(istanbul.Proposal) (time.Duration, error)
+	Verify(istanbul.Proposal) (time.Duration, *core2.StateProcessResult, error)
```
hbandura, line 71: nit: it would seem the `time.Duration` return value is related to the error value. Maybe swap the order for `result, duration, error` ?

consensus/istanbul/core/errors.go
```diff
@@ -86,4 +86,6 @@ var (
 	errInvalidValidatorAddress = errors.New("failed to find an existing validator by address")
 	// Invalid round state
 	errInvalidState = errors.New("invalid round state")
+	// errStateProcessResultNotFound is returned when result is not found in cache
+	errStateProcessResultNotFound = errors.New("failed to retrieve errStateProcessResultNotFound")
```
hbandura, line 90: That error message line sounds a bit weird

consensus/istanbul/core/roundstate.go
```diff
@@ -464,6 +472,57 @@ func (rs *roundStateImpl) GetProposalVerificationStatus(proposalHash common.Hash
...
+		//if stateProcessResult, isCached = rs.stateProcessResults[proposalHash]; isCached {
+		//	stateProcessResult = deepCopy(stateProcessResult)
+		//}
+	}
+
+	return
+}
+
+//deepCopy will deep copy a StateProcessResult.
+func deepCopy(result *core2.StateProcessResult) *core2.StateProcessResult {
```
hbandura, line 501: This sounds like it could be a method on the StateProcessResult struct

consensus/istanbul/backend/engine.go
```diff
@@ -643,10 +643,11 @@ func (sb *Backend) updateReplicaStateLoop(bc *ethCore.BlockChain) {
 	}
 }
 
-// SetBlockProcessors implements consensus.Istanbul.SetBlockProcessors
-func (sb *Backend) SetBlockProcessors(hasBadBlock func(common.Hash) bool,
+// SetCallBack implements consensus.Istanbul.SetCallBacks
+func (sb *Backend) SetCallBacks(hasBadBlock func(common.Hash) bool,
```
hbandura, line 647: Why this name change?

YoduYodu: To give it a more appropriate name, since this func injects several call back functions, rather than "block processors".

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +498,26 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		return nil
 	}
 
-	if sb.broadcaster != nil {
-		sb.broadcaster.Enqueue(fetcherID, block)
+	// If no cached result, we import block through fetcher, which requires processing the block one more time.
+	// Else if cache hit, we directly write the block.
+	if result == nil {
```
trianglesphere, line 503: What happens if both `result` and `broadcaster` are `nil`? Is that a case that could occur or is it not supposed to occur at all?

mcortesi: Good catch!

IMO, it should fail, today it seems to panic due to trying to use the `nil` result

consensus/istanbul/backend/engine.go
mcortesi: +1 to the change

consensus/istanbul/core/core.go
```diff
@@ -377,7 +378,9 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
-		if err := c.backend.Commit(proposal, aggregatedSeal, aggregatedEpochValidatorSetSeal); err != nil {
+
+		result, _ := c.current.GetStateProcessResult(proposal.Hash())
```
mcortesi, line 382: why are we ignoreing the error?

consensus/istanbul/backend/backend.go
YoduYodu: Great catch! That is a case that could occur.
I fixed it.

consensus/istanbul/core/core.go
YoduYodu: That's a `isCached` bool, not an error.
Since in the next line the `result` is going to be passed to `c.backend.Commit` anyway, and if the `result` is `nil`, we'll ignore it.

piersy:
Hey @YoduYodu I recently joined clabs, was talking with @timmoreton and he suggested I take a look at this PR.

I'm assuming that this change is being done to improve the performance of the node and I was wondering if there any tests that show that this change actually did have a positive impact on the performance?

YoduYodu:
> ing with @timmoreton and he suggested I take a look at this PR.
> 
> I'm assuming that this change is being done to improve the performance of the node and I was wondering if there any tests that show that this change actually did have a positive impact on the performance?

Hey @piersy 👋  Welcome onboard!
Yes, your assumption is right. This is to avoid duplicate block processing during consensus.
As for the tests, we previously run test net on GCP and monitor certain metrics like https://github.com/celo-org/celo-blockchain/blob/68e2ab747b40ce215a916cde7fdc77613366dcd0/les/metrics.go#L84 but that was not dev friendly.
So now we are working on `mycelo`, a local test net. By adding transactions to it, we can get better idea how changes really work. I believe @mcortesi has better idea on `mycelo`.

light/lightchain.go
```diff
@@ -406,11 +406,11 @@ func (lc *LightChain) InsertHeaderChain(chain []*types.Header, checkFreq int, co
 		status, err := lc.hc.WriteHeader(header)
 
 		switch status {
-		case core.CanonStatTy:
+		case types.CanonStatTy:
```
piersy, line 409: Why the move from core to types? This change will probably introduce quite a few extra conflicts when merging from upstream, and since the status is only used in the `core` and `light` packages, moving them to types doesn't seem to add any value.

consensus/istanbul/backend/backend.go
piersy: Hey @YoduYodu, Is it even possible for result to be nil here? I've looked through the code a bit and it seems to me that to be calling Commit a node must have either proposed the block or verified it and in both of those cases they would have processed the block, meaning that It shouldn't be possible to reach this point with a nil result.

consensus/istanbul/core/roundstate.go
```diff
@@ -464,6 +471,28 @@ func (rs *roundStateImpl) GetProposalVerificationStatus(proposalHash common.Hash
...
+	defer rs.mu.Unlock()
+
+	if rs.stateProcessResults == nil {
+		rs.stateProcessResults = make(map[common.Hash]*core2.StateProcessResult)
+	}
+
+	rs.stateProcessResults[proposalHash] = stateProcessResult
+}
+
+func (rs *roundStateImpl) GetStateProcessResult(proposalHash common.Hash) (stateProcessResult *core2.StateProcessResult, isCached bool) {
```
piersy, line 485: I think it would be cleaner here to not return the bool `isCached`. Given that `stateProcessResult` is a pointer, lack of state can be communicated by returning nil. Maps in go provide the multi value index operation because in the case of a map with non pointer values you need to be able to determine the difference between the zero value and no value at all, but this is not the case here.

piersy:
Hey @YoduYodu very happy to be here, thanks for clarifying :) @mcortesi is it possible to run mycelo over this PR to measure its performance impact?

trianglesphere:
> @mcortesi is it possible to run mycelo over this PR to measure its performance impact?

You could certainly run mycelo on this PR. The hard part is observing the difference is performance. If you look at the individual components, a fully loaded block (13 mil gas) verification is order 300ms (range 200-400ms), and a state write is 10~20ms. The trick is that you need to observe two different code paths (with an adaptive sleep in the middle) on a non proposer (best to use a sticky proposer policy for this test).

piersy:
Hey @trianglesphere thanks for the info. I guess I don't really have enough in depth knowledge to fully understand the approach to testing you are suggesting.

I was just thinking it would be nice to be able to know that this PR improved block processing times by some amount.

consensus/istanbul/backend/backend.go
YoduYodu: @piersy In an edge case where the node goes offline and then back online, we could get a nil result when calling `Commit`

light/lightchain.go
YoduYodu: Because `WriterStatus` is introduced into `consensus` package, here:
https://github.com/celo-org/celo-blockchain/blob/f3d394618ceeed33d54fbd8140223c1c31bfc086/consensus/consensus.go#L185
it causes import cycle if not moved to types.
An alternative way is to remove `WriteStatus`, only return error. Since we don't use `WriteStatus` in worker anyway.
wdyt?

consensus/istanbul/backend/backend.go
piersy: @YoduYodu Is that because we are not saving the state in `core.rsSaveDecorator`?  And if so, is it possible we could save the state in  `core.rsSaveDecorator`?

consensus/istanbul/backend/test_utils.go
```diff
@@ -85,14 +85,16 @@ func newBlockChainWithKeys(isProxy bool, proxiedValAddress common.Address, isPro
...
+		b.SetCallBacks(blockchain.HasBadBlock,
 			func(block *types.Block, state *state.StateDB) (types.Receipts, []*types.Log, uint64, error) {
 				return blockchain.Processor().Process(block, state, *blockchain.GetVMConfig())
 			},
 			func(block *types.Block, state *state.StateDB, receipts types.Receipts, usedGas uint64) error {
 				return blockchain.Validator().ValidateState(block, state, receipts, usedGas)
+			},
+			func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB, emitHeadEvent bool) (status types.WriteStatus, err error) {
+				return blockchain.WriteBlockWithState(block, receipts, logs, state, emitHeadEvent)
 			})
```
piersy, line 97: I think this could be reduced to:

```suggestion
			blockchain.Validator().ValidateState,
			blockchain.WriteBlockWithState)
```

light/lightchain.go
piersy: Yeah that seems like a good approach, It looks like the status from the  public method `Blockchain.WriteBlockWithState` is never used, its only used internally inside `Blockchain`.

consensus/istanbul/backend/test_utils.go
piersy: But actually if we change the callback method to not return the status as you suggested above we can't keep this simplification verbatim.

consensus/istanbul/backend/backend.go
mcortesi: That is correct @piersy. This state is in memory, thus it's loss when restarting. We could try to persist it to disk. But i wouldn't recommend that approach. The `stateDB` is not really a `value object`, more of a `transaction object`, so not really designed for persistence. Trying to do that would be quite hard and probably a good source of future bugs. 

And all of this, just to catch an edge case and avoid re-processing a block at that time. So, IMO it's not worth it.

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +498,27 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
...
+		sb.logger.Warn("state process cache miss, enqueuing the block to fetcher", "number", block.Number(), "hash", block.Hash())
+		if sb.broadcaster != nil {
+			sb.broadcaster.Enqueue(fetcherID, block)
+		}
+		sb.logger.Warn("Cannot import consensus block", "number", block.Number(), "hash", block.Hash())
+		return nil
+	}
+
+	if block := sb.chain.GetBlock(block.Hash(), block.NumberU64()); block != nil {
```
mcortesi, line 512: Is there really an issue with calling `writeBlockWithState` for an already persisted block? If that's the case we have a bigger problem.

This check here, won't save us, since there's a race condition on it. And the proposer already has the same race condition.

consensus/istanbul/core/core.go
```diff
@@ -377,7 +378,9 @@ func (c *core) commit() error {
 			c.waitForDesiredRound(nextRound)
 			return nil
 		}
-		if err := c.backend.Commit(proposal, aggregatedSeal, aggregatedEpochValidatorSetSeal); err != nil {
+
+		result := c.current.GetStateProcessResult(proposal.Hash())
```
mcortesi, line 382: Add a comment explaning when this can be `nil`

consensus/istanbul/core/core.go
```diff
@@ -769,14 +772,18 @@ func (c *core) verifyProposal(proposal istanbul.Proposal) (time.Duration, error)
...
-		duration, err := c.backend.Verify(proposal)
+		result, duration, err := c.backend.Verify(proposal)
 		logger.Trace("proposal verify return values", "duration", duration, "err", err)
 
 		// Don't cache the verification status if it's a future block
 		if err != consensus.ErrFutureBlock {
 			c.current.SetProposalVerificationStatus(proposal.Hash(), err)
 		}
 
+		if result != nil {
```
mcortesi, line 783: it seems to be that if `err == nil` then `result != nil`. If that doesn't hold, it's a bug. 

So my suggestion is to either: 
 a) remove the if here. 
 b) replace it by checking if `result == nil` and returning an error there, clearly stating it's a bug.

light/lightchain.go
mcortesi: +1 to keeping `types.WriteStatus` in blockchain, and not exposing it as a result of `WriteBlockWithState`

core/types/block.go
```diff
@@ -430,3 +430,12 @@ func (b *Block) Hash() common.Hash {
 }
 
 type Blocks []*Block
+
+// WriteStatus status of write
+type WriteStatus byte
```
mcortesi, line 435: As mentioned elsewhere, i would keep this inside `blockchain` and avoid exposing it to the outside the package.

core/state_processor.go
```diff
@@ -155,3 +155,10 @@ func ApplyTransaction(config *params.ChainConfig, bc vm.ChainContext, txFeeRecip
 
 	return receipt, err
 }
+
+// StateProcessResult represents processing results from StateProcessor.
+type StateProcessResult struct {
```
mcortesi, line 160: Even if it makes sense here from a semantics perspecitve. It's not really used here. 

I would move this type inside instabul modules, to the same place where we define the interface for `Verify()` and `Commit()` methods

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +498,27 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		return nil
 	}
 
-	if sb.broadcaster != nil {
-		sb.broadcaster.Enqueue(fetcherID, block)
+	// If no cached result, we import block through fetcher, which requires processing the block one more time.
+	// Else if cache hit, we directly write the block.
+	if result == nil {
+		sb.logger.Warn("state process cache miss, enqueuing the block to fetcher", "number", block.Number(), "hash", block.Hash())
+		if sb.broadcaster != nil {
```
mcortesi, line 505: I'd prefer to totally remove the `broadcaster` dependency. It's a very different code path for the block thatn directly calling `writeBlockWithState()`. 

So, my recommendation is to call `Verify` here as a way of obtaining the `result`

mcortesi:
Lest some comments!

consensus/istanbul/core/core.go
YoduYodu: `err` could be an Error other than ErrFutureBlock, in which case `if result != nil` is necessary

core/state_processor.go
YoduYodu: Put it in `consensus/istanbul/core/roundstate.go`, that's where it's used in the lowest level.

consensus/istanbul/backend/backend.go
YoduYodu: What really matters in `broadcaster.Enqueue()` is process the block and call `writeBlockWithState()`, I think calling `process` and `writeBlockWithState()` will be enough. Calling `Verify` seems a overkill

consensus/istanbul/backend/backend.go
piersy: Hey @mcortesi I hadn't realised that the `StateProcessResult` has a `StateDB` as one of its fields. I agree it's not worth it here.

My concern about re-processing the block is not that it might happen as an edge case but rather that, a future erroneous code change could cause the result to always be nil and we wouldn't be alerted to that unless we looked at the logs.

consensus/istanbul/backend/backend.go
piersy: Hey @mcortesi, I'm wondering how does the block get propagated if we are not calling `broadcaster.Enqueue`? As I understand, `broadcaster.Enqueue` eventually results in a call to `ProtocolManager.BroadcastBlock` as does returning a block on the `resultsCh` in the worker, but I don't see how the block gets propagated if we are just calling `writeBlockWithState`

piersy, line 505: I think this log is slightly out of place
```suggestion
		if sb.broadcaster != nil {
				sb.logger.Warn("state process cache miss, enqueuing the block to fetcher", "number", block.Number(), "hash", block.Hash())
```

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +498,27 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
...
+		return nil
+	}
+
+	if block := sb.chain.GetBlock(block.Hash(), block.NumberU64()); block != nil {
+		sb.logger.Info("Duplicate block to commit", "number", block.Number(), "hash", block.Hash())
+		return nil
+	}
+	if err := sb.writeBlockWithState(block, result.Receipts, result.Logs, result.State, true); err != nil {
+		sb.logger.Error("Failed writeBlockWithState", "err", err)
+		return err
```
piersy, line 518: This seems like it is going to be racy, between the call to `GetBlock` and the call to `writeBlockWithState` a block may have been written.

So I think we could just skip the call to `GetBlock`. It would be good to be able to tell from the error returned from `writeBlockWithState` if the block was a duplicate or if it was an inconsistent block (same height but different hash). I think inconsistent blocks should never happen (I.E it implies consensus has been broken) so I guess we would want to stop processing in this case.

It looks like at present inconsistent blocks are ignored in write block with state.
https://github.com/celo-org/celo-blockchain/blob/ebd9817f69659a1ce9731f5aef60a023d9d09a7f/core/blockchain.go#L1304-L1306

So I think if an inconsistent block were encountered it would just overwrite the pre-existing block at that height, which doesn't seem like a good approach.

consensus/istanbul/core/core.go
```diff
@@ -769,13 +772,14 @@ func (c *core) verifyProposal(proposal istanbul.Proposal) (time.Duration, error)
...
 
-		duration, err := c.backend.Verify(proposal)
+		result, duration, err := c.backend.Verify(proposal)
 		logger.Trace("proposal verify return values", "duration", duration, "err", err)
 
 		// Don't cache the verification status if it's a future block
 		if err != consensus.ErrFutureBlock {
 			c.current.SetProposalVerificationStatus(proposal.Hash(), err)
 		}
+		c.current.SetStateProcessResult(proposal.Hash(), result)
```
nategraf, line 782: If this is a future block, setting the result in the cache here causes an inconsistent state because the block evaluated in the future will achieve a result that does not match what is in the cache (i.e. it will result a full state, but the cache holds `nil`). I would recommend either writing to the cache only if the error is non-nill, or only if it is not `consensus.ErrFutureBlock`

consensus/istanbul/core/roundstate.go
```diff
@@ -98,6 +102,9 @@ type roundStateImpl struct {
 	// to be persisted.
 	proposalVerificationStatus map[common.Hash]error
 
+	// Cache for StateProcessResult in this sequence.
+	stateProcessResults map[common.Hash]*StateProcessResult
```
nategraf, line 106: This cache implementation does not have any cache invalidation and will slowly leak memory. IIUC the cache only needs to hold the most recent value. If this is true, I recommend changing this to be a single result object, along with it's block hash. (i.e. have it be a singleton cache) If more values are needed, I recommend building a simple circular buffer implementation that stores the results for up to `k` blocks, indexed by `blockNumber % k`.

nategraf (commented):
I left some lower-level comments, but most of my concerns stem from writing to the chain database directly from the consensus engine. I cite a specific concern about performance in one of my comments, but more generally, it punches a hole through many of the abstraction layers we'd rather patch up.

Instead of passing in a `writeBlockWithState` function, I recommend that we augment the existing `results` channel passed into the `Seal` function to also send the block processing results.
https://github.com/celo-org/celo-blockchain/blob/633fb0d1fb45c145171d2811e9d61d848360e0a6/consensus/consensus.go#L100

In the current implementation, the worker contains all the logic necessary to write blocks to the database and broadcast them, as it does for PoW, but it's logic does not work well in our case because the result of a `Seal` operation may not be the block that was passed in. (This strange interaction is another example of why we should rip out the miner and worker, but that is a battle for another PR) If we pass the block processing results along with the block though, it becomes seemless for the worker to handle committing the block to the DB and broadcasting it. It also allows use to avoid the consensus engine getting further tangled with the blockchain code, which it already has too much interaction with.

I also believe this is why Istanbul currently uses the convoluted method of importing blocks through the broadcaster. If we pass processing results through the `results` channel, we could use it in all cases instead of going through the broadcaster. (We would also need to call `Verify` from `Commit` in cases where the processing results as unavailable, as Mariano suggested)
https://github.com/celo-org/celo-blockchain/blob/633fb0d1fb45c145171d2811e9d61d848360e0a6/consensus/istanbul/backend/backend.go#L489-L499

Also, I noticed none of the tests use the cached path. Please include some tests that cover the new code paths.

mcortesi:
> I left some lower-level comments, but most of my concerns stem from writing to the chain database directly from the consensus engine. I cite a specific concern about performance in one of my comments, but more generally, it punches a hole through many of the abstraction layers we'd rather patch up.
> 
> Instead of passing in a `writeBlockWithState` function, I recommend that we augment the existing `results` channel passed into the `Seal` function to also send the block processing results.
> https://github.com/celo-org/celo-blockchain/blob/633fb0d1fb45c145171d2811e9d61d848360e0a6/consensus/consensus.go#L100
> 
> In the current implementation, the worker contains all the logic necessary to write blocks to the database and broadcast them, as it does for PoW, but it's logic does not work well in our case because the result of a `Seal` operation may not be the block that was passed in. (This strange interaction is another example of why we should rip out the miner and worker, but that is a battle for another PR) If we pass the block processing results along with the block though, it becomes seemless for the worker to handle committing the block to the DB and broadcasting it. It also allows use to avoid the consensus engine getting further tangled with the blockchain code, which it already has too much interaction with.
> 
> I also believe this is why Istanbul currently uses the convoluted method of importing blocks through the broadcaster. If we pass processing results through the `results` channel, we could use it in all cases instead of going through the broadcaster. (We would also need to call `Verify` from `Commit` in cases where the processing results as unavailable, as Mariano suggested)
> https://github.com/celo-org/celo-blockchain/blob/633fb0d1fb45c145171d2811e9d61d848360e0a6/consensus/istanbul/backend/backend.go#L489-L499
> 
> Also, I noticed none of the tests use the cached path. Please include some tests that cover the new code paths.

There's another PR where we've pass the result to the commit channel and that created weird behaviours when testing it. The reason IMO, is that worker's result ch is the result of a `Seal()` call, and using it to return something that is not the a `Seal` operation is really really weird. And very hard to follow, so i would strongly suggest against that.

Puting that aside, i agree that we need to decouple and the the engine should not now about blockchain. My suggestion to Tong on 1:1 we had was to change `broadcaster`, `writeBlockWithState` and `emit NewBlock miner event` with a single callback.

something like `onNewBlockMined func(...)` that we should call with `go onNewBlockMined()` so not to block on that (we didn't block before)

The idea behind that, the engine is this "thing" that "produces blocks", and whenever it does, it "notifies" others. But it doesn't really care what happens with them. So it's not responsible, nor aware of blockchain insertion, etc. You can do that decoupling by defining the right boundaries.

In fact, you take a look, right not it doesn't actually depende on blockchain, it depends on a callback named `writeBlockWithState()`, it's subtle, but not the same. The coupling is on the naming, and the expectation. It's semantic, not syntactic.

If i change the name, and make it more general, then responsability for coupling that "event" to an actual "blockchain insert" lies on the engine's user, not on the engine.

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +492,15 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		return nil
 	}
 
-	if sb.broadcaster != nil {
-		sb.broadcaster.Enqueue(fetcherID, block)
+	// Note that result should NOT be nil, one edge case is when the validator re-start in the middle of a sequence.
+	// When happening, we re-verify the proposal and write it.
+	if result == nil {
+		sb.logger.Error("state process cache miss, verifying the block again", "number", block.Number(), "hash", block.Hash())
```
mcortesi, line 498: several things here:

1. if `result != nil` it should still call `onNewConsensusBlock`
2. The log message implies that `backend` knows about the existence of a `cache`. That's a way of semantic coupling. It shouldn't be aware of any cache.
3. If `result ==nil` and `sb.Verify()` fails, we should return and error. As a rule of thumb, we should never silently ignore errors.

consensus/istanbul/core/core.go
mcortesi: +1

consensus/istanbul/core/roundstate.go
mcortesi: Not sure if that's really needed. on `StartNewSequence` clear this map. So you only cache values for the current sequence. You could argue that if there are many rounds, and each round implies a new block, then this could get bigger. But not sure if we want to worry about that (or at least we would have other things to worry if that happened)

miner/worker.go
```diff
@@ -288,12 +288,19 @@ func (w *worker) start() {
...
-				return w.chain.Validator().ValidateState(block, state, receipts, usedGas)
+			w.chain.Validator().ValidateState,
+			func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB) {
+				_, err := w.chain.WriteBlockWithState(block, receipts, logs, state, true)
+				if err != nil {
+					log.Error("Failed writeBlockWithState in onNewConsensusBlock", "blockNumber", block.Number(), "hash", block.Hash(), "err", err)
+					return
+				}
+				log.Info("Successfully imported new block in onNewConsensusBlock", "number", block.Number(), "hash", block.Hash(), "elapsed")
+				w.mux.Post(core.NewMinedBlockEvent{Block: block})
```
mcortesi, line 303: this also returns an error, and we are silently ignoring it here. 

We cant' do much here, but at least we should log it

mcortesi:
Also, we don't need `broadcaster.Enqueue` anymore, so i would remove it from here:
https://github.com/celo-org/celo-blockchain/blob/68e2ab747b40ce215a916cde7fdc77613366dcd0/consensus/protocol.go#L29

consensus/istanbul/core/core.go
YoduYodu: I handled it here: https://github.com/celo-org/celo-blockchain/blob/d1d3805c15fdc754a8c948f3fcf345eac30ad01e/consensus/istanbul/backend/backend.go#L497
we get a non-nil result from the cache only when error is nil.

mcortesi (approved):
LGTM!! Great Work!

consensus/istanbul/core/roundstate.go
nategraf: Oh, great! I missed didn't notice that when I was looking for cache invalidation logic. 👍

consensus/istanbul/backend/backend.go
piersy: The comment here is confusing to me. If the result should not be nil, then we shouldn't allow it to be nil, we should panic. If however it is a valid case that the result can be nil, then the comment needs to be changed, it could simply explain the case where the result is nil.

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +492,16 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		return nil
 	}
 
-	if sb.broadcaster != nil {
-		sb.broadcaster.Enqueue(fetcherID, block)
+	// Note that result should NOT be nil, one edge case is when the validator re-start in the middle of a sequence.
+	// When happening, we re-verify the proposal and write it.
+	if result == nil {
```
piersy, line 497: The comment here is confusing to me. If the result should not be nil, then we shouldn't allow it to be nil, we should panic. If however it is a valid case that the result can be nil, then the comment needs to be changed, it could simply explain the case where the result is nil.

consensus/istanbul/backend/backend.go
```diff
@@ -497,9 +492,16 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
...
+		sb.logger.Error("Verifying the block again", "number", block.Number(), "hash", block.Hash())
+		if result, _, err = sb.Verify(proposal); err != nil {
+			sb.logger.Info("Verified proposal", "err", err)
+			return err
+		}
 	}
+	go sb.onNewConsensusBlock(block, result.Receipts, result.Logs, result.State)
```
piersy, line 504: So now `onNewConsensusBlock` is calling `blockchain.WriteBlockWithState` which as far as I know does not broadcast the block (doesn't result in a call to `ProtocolManager.BroadcastBlock`), so I'm wondering how does the block get propagated to peers in this case, and how does that affect performance?

mcortesi: It does now:

```go
func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB) {
  if err := w.chain.InsertPreprocesedBlock(block, receipts, logs, state); err != nil {
	  if err == core.ErrNotHeadBlock {
		  log.Warn("Tried to insert duplicated produced block", "blockNumber", block.Number(), "hash", block.Hash(), "err", err)
	  } else {
		  log.Error("Failed to insert produced block", "blockNumber", block.Number(), "hash", block.Hash(), "err", err)
	  }
	  return
  }
  log.Info("Successfully produced new block", "number", block.Number(), "hash", block.Hash())
  
  if err := w.mux.Post(core.NewMinedBlockEvent{Block: block}); err != nil {
	  log.Error("Error when posting NewMinedBlockEvent", "err", err)
  }
  })
```

mcortesi:
> One thing that is unclear to me is whether we need to include a check to prevent writing duplicate blocks. If the block gets imported from the network before it is written by this new method, it will result in the same block being sent twice to `BlockChain.WriteBlockWithState`. One concern of mine is that the uptime monitor might count the block twice, which could lead to inconsistencies between validators and possibly stalling the chain. Do we know whether or not this could be an issue?
> 
> Also, the tests all pass in a `nil` state processing result. It would be good to have some tests that pass in a non-nil result.

Added a check to preven adding adding a block that it's not the next block on the chain. So, not only duplicated, but forks.

Also renamed `WriteBlockWithState` to `InsertPreprocessedBlock()` to make it more meaningfull, and removed unnecessary parameters & returns.

consensus/istanbul/backend/backend.go
mcortesi: Addressed on https://github.com/celo-org/celo-blockchain/pull/1394/commits/e5556b16439b86a66524eb2c11c91cee8a338088

nategraf (approved):
Looks good to me! Very excited for these changes!

consensus/istanbul/backend/test_utils.go
```diff
@@ -85,14 +85,16 @@ func newBlockChainWithKeys(isProxy bool, proxiedValAddress common.Address, isPro
...
+		b.SetCallBacks(blockchain.HasBadBlock,
 			func(block *types.Block, state *state.StateDB) (types.Receipts, []*types.Log, uint64, error) {
 				return blockchain.Processor().Process(block, state, *blockchain.GetVMConfig())
 			},
-			func(block *types.Block, state *state.StateDB, receipts types.Receipts, usedGas uint64) error {
-				return blockchain.Validator().ValidateState(block, state, receipts, usedGas)
+			blockchain.Validator().ValidateState,
+			func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB) {
+				if err := blockchain.InsertPreprocessedBlock(block, receipts, logs, state); err != nil {
+					panic("could not InsertPreprocessedBlock")
```
piersy, line 95:
```suggestion
					panic(fmt.Sprintf("could not InsertPreprocessedBlock: %v", err))
```

miner/worker.go
```diff
@@ -288,12 +288,25 @@ func (w *worker) start() {
...
 			},
-			func(block *types.Block, state *state.StateDB, receipts types.Receipts, usedGas uint64) error {
-				return w.chain.Validator().ValidateState(block, state, receipts, usedGas)
+			w.chain.Validator().ValidateState,
+			func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB) {
+				if err := w.chain.InsertPreprocessedBlock(block, receipts, logs, state); err != nil {
+					if err == core.ErrNotHeadBlock {
+						log.Warn("Tried to insert duplicated produced block", "blockNumber", block.Number(), "hash", block.Hash(), "err", err)
+					} else {
+						log.Error("Failed to insert produced block", "blockNumber", block.Number(), "hash", block.Hash(), "err", err)
```
piersy, line 301: Its not super clear to me the meaning of all the errors that could be returned here, but I would think some of those errors indicate a consensus breakdown or programming error and for those errors we should stop processing.

For example if the randomness cannot be generated, that looks like it is a serious failure, not one we would want to keep processing on top of.

YoduYodu: https://github.com/celo-org/celo-blockchain/blob/68e2ab747b40ce215a916cde7fdc77613366dcd0/eth/fetcher/block_fetcher.go#L682-L686

The broadcast this PR replaced had this same scenario here, that insertChain error doesn't need to(can't) be handled

PR #1405, opened 2021-02-24
Disable GPM updates during mycelo load tests

### LINKED ISSUE #1400: Debug MyCelo load test out of funds
## Description

Figure out why and then fix why mycelo accounts run out of funds after about 3-5 minutes of load tests w/ 1 second blocks

## Context

Enables longer load tests.

trianglesphere:
### Description

Disables gas price updating in the load test. Prior to this change the gas price was going so high that transactions cannot go through after a couple hundred full blocks
This also sends transactions in a round robin fashion (Initial attempt to fix). This is probably not strictly needed to fix the issue.

### Tested

- [x] Manual ~10 min test
- [x] k8s 30 mins test

### Related issues

- Fixes #1400

### DIFF
--- a/cmd/mycelo/templates.go
+++ b/cmd/mycelo/templates.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/celo-org/celo-blockchain/common"
+	"github.com/celo-org/celo-blockchain/common/fixed"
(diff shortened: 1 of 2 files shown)


oneeman (approved):
Makes sense.

PR #1406, opened 2021-03-01
Adds flag to limit pending txs

trianglesphere:
### Description

Limits the number of pending txs sent to --maxpending
This does not completely prevent a slowdown on the testnet, but it stops it
from becoming a cascading failure.

### Tested

Yes

### DIFF
--- a/cmd/mycelo/main.go
+++ b/cmd/mycelo/main.go
@@ -472,9 +472,9 @@ func loadBot(ctx *cli.Context) error {
 	return loadbot.Start(runCtx, &loadbot.Config{
 		Accounts:              env.DeveloperAccounts(),
 		Amount:                big.NewInt(10000000),
-		TransactionsPerSecond: ctx.Int("tps"),
+		TransactionsPerSecond: ctx.Int(loadTestTPSFlag.Name),
 		Clients:               clients,
 		Verbose:               verbose,
-		MaxPending:            ctx.Uint64("maxpending"),
+		MaxPending:            ctx.Uint64(loadTestMaxPendingFlag.Name),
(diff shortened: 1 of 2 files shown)


cmd/mycelo/main.go
```diff
@@ -469,5 +475,6 @@ func loadBot(ctx *cli.Context) error {
 		TransactionsPerSecond: ctx.Int("tps"),
 		Clients:               clients,
 		Verbose:               verbose,
+		MaxPending:            ctx.Uint64("maxpending"),
```
mcortesi, line 478: we should start using `loadTestMaxPendingFlag.Name` instead of a string

mycelo/loadbot/bot.go
```diff
@@ -74,6 +81,14 @@ func runBot(ctx context.Context, acc env.Account, verbose bool, sleepTime time.D
 	transactor.FeeCurrency = &stableTokenAddress
 
 	for {
+		pendingLock.Lock()
```
mcortesi, line 84: i would check golang atomic value, i thing they are exactly what you need
https://gobyexample.com/atomic-counters

mcortesi (approved):
left some improvements ideas

mycelo/loadbot/bot.go
trianglesphere: The first implementation did in fact use them. I had issue doing an atomic comparison though.

cmd/mycelo/main.go
trianglesphere: Switched this and `loadTestTPSFlag` as well

PR #1410, opened 2021-03-02
Adds revert reason on eth_call and simulatedBackend

mcortesi:
### Description

Adds smart contract revert reason to the error message when a call to a smart contract fails.

For the case:
 * `eth_call` RPC
 * `SimulatedBackend.CallContract()`

### Tested

Manual testing

### DIFF
--- a/accounts/abi/bind/backends/simulated.go
+++ b/accounts/abi/bind/backends/simulated.go
@@ -350,6 +350,18 @@ func (b *SimulatedBackend) CallContract(ctx context.Context, call ethereum.CallM
 		return nil, err
 	}
 	res, err := b.callContract(ctx, call, b.blockchain.CurrentBlock(), state)
+
+	if res != nil && res.Err != vm.ErrOutOfGas {
+		if len(res.Revert()) > 0 {
+			ret, err := abi.UnpackRevert(res.Revert())
+			if err != nil {
+				return nil, fmt.Errorf("execution reverted: %#x", res.Revert())
+			} else {
+				return nil, fmt.Errorf("execution reverted: %s", ret)
+			}
+		}
+	}
+
 	if err != nil {
 		return nil, err
 	}
--- a/accounts/abi/bind/backends/simulated.go
+++ b/accounts/abi/bind/backends/simulated.go
@@ -353,12 +353,10 @@ func (b *SimulatedBackend) CallContract(ctx context.Context, call ethereum.CallM
 
 	if res != nil && res.Err != vm.ErrOutOfGas {
 		if len(res.Revert()) > 0 {
-			ret, err := abi.UnpackRevert(res.Revert())
-			if err != nil {
-				return nil, fmt.Errorf("execution reverted: %#x", res.Revert())
-			} else {
+			if ret, err := abi.UnpackRevert(res.Revert()); err == nil {
 				return nil, fmt.Errorf("execution reverted: %s", ret)
 			}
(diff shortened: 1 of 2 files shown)


accounts/abi/bind/backends/simulated.go
```diff
@@ -350,6 +350,18 @@ func (b *SimulatedBackend) CallContract(ctx context.Context, call ethereum.CallM
...
 	res, err := b.callContract(ctx, call, b.blockchain.CurrentBlock(), state)
+
+	if res != nil && res.Err != vm.ErrOutOfGas {
+		if len(res.Revert()) > 0 {
+			ret, err := abi.UnpackRevert(res.Revert())
+			if err != nil {
+				return nil, fmt.Errorf("execution reverted: %#x", res.Revert())
+			} else {
+				return nil, fmt.Errorf("execution reverted: %s", ret)
+			}
```
gastonponti, line 361: nit (?)
```suggestion
			if ret, err := abi.UnpackRevert(res.Revert()); err == nil {
				return nil, fmt.Errorf("execution reverted: %s", ret)
			}
			return nil, fmt.Errorf("execution reverted: %#x", res.Revert())
```

internal/ethapi/api.go
```diff
@@ -878,6 +878,17 @@ func (s *PublicBlockChainAPI) Call(ctx context.Context, args CallArgs, blockNrOr
...
 	}
+
+	if result != nil && result.Err != vm.ErrOutOfGas {
+		if len(result.Revert()) > 0 {
+			ret, err := abi.UnpackRevert(result.Revert())
+			if err != nil {
+				return nil, fmt.Errorf("execution reverted: %#x", result.Revert())
+			} else {
+				return nil, fmt.Errorf("execution reverted: %s", ret)
+			}
```
gastonponti, line 889: Same
```suggestion
			if ret, err := abi.UnpackRevert(result.Revert()); err == nil {
				return nil, fmt.Errorf("execution reverted: %s", ret)
			}
			return nil, fmt.Errorf("execution reverted: %#x", result.Revert())
```

accounts/abi/bind/backends/simulated.go
```diff
@@ -350,6 +350,18 @@ func (b *SimulatedBackend) CallContract(ctx context.Context, call ethereum.CallM
 		return nil, err
 	}
 	res, err := b.callContract(ctx, call, b.blockchain.CurrentBlock(), state)
+
+	if res != nil && res.Err != vm.ErrOutOfGas {
```
gastonponti, line 354: out of gas is handled in the original flow, right?

gastonponti (approved):
LGTM
Some nit and a question

accounts/abi/bind/backends/simulated.go
mcortesi: if `err == ErrOutOfGas` then it's not a revert :)

oneeman:
Looking at how web3 handles this (https://github.com/ChainSafe/web3.js/commit/2948927a18a2424621743294a80ea3ee5bf0e1e9#diff-9050e060908a6957eb4e25bb2f4b8930f5097e55e846e0e8194ff5fa3a3440a4R621-R630), I think we should cherry-pick the upstream PR for this (https://github.com/ethereum/go-ethereum/pull/21083).  Because web3 expects the revert reason to either be in `err.data` (geth v1.9.15 and higher) or as the return value (geth v1.9.13 and lower).

PR #1412, opened 2021-03-04
Remove Ethereum flags that don't apply to Celo

### LINKED ISSUE #1064: Remove ethereum flags
### Expected Behavior

All celo-blockchain flags should either function as they do in upstream geth, be removed, or return a "not implemented yet" error. This is to reduce confusion for external devs familiar with upstream geth. 

### Current Behavior

Certain flags (--fakepow, --rinkeby, --dev etc.) are not applicable to celo-blockchain or are currently non-functional. 

See also https://github.com/celo-org/celo-blockchain/issues/1063

How can we reproduce this?

### Logs

Are there any logs?

### System information

Run `geth version`

YoduYodu:
### Description

All celo-blockchain flags should either function as they do in upstream geth, or be removed.
This is to reduce confusion for external devs familiar with upstream geth.

Certain flags are removed in this PR.
--fakepow
--pcscdpath

### Other changes

Examed following commands:

`geth import`
`geth export`
`geth copydb`
`geth dump`
`geth inspect`

### Tested

Yes

### Related issues

- Fixes #1064

### Backwards compatibility

Yes

### DIFF
--- a/cmd/clef/README.md
+++ b/cmd/clef/README.md
@@ -32,7 +32,6 @@ GLOBAL OPTIONS:
    --chainid value         Chain id to use for signing (1=mainnet, 3=Ropsten, 4=Rinkeby, 5=Goerli) (default: 1)
    --lightkdf              Reduce key-derivation RAM & CPU usage at some expense of KDF strength
    --nousb                 Disables monitoring for and managing USB hardware wallets
-   --pcscdpath value       Path to the smartcard daemon (pcscd) socket file (default: "/run/pcscd/pcscd.comm")
    --rpcaddr value         HTTP-RPC server listening interface (default: "localhost")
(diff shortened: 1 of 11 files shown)


mcortesi:
we should keep `--dev`

mcortesi (commented):
SmartCard removal looks good.

About FakePow, i'm worried that exportChain, importChain, copy and all the other commands might fail. They are probably already failing before this PR. 

So what i would do, is to check who calls `MakeChain()` and try running those commands. See what we find.

YoduYodu:
@mcortesi I've test all 5 commands that involves `MakeChain()`, all valid, but cli instructions seem un-maintained.

mcortesi (approved):

PR #1416, opened 2021-03-04
Update mycelo to work with e2e tests in celo-monorepo

oneeman:
### Description

Makes some changes to mycelo to make it usable in conjunction with tooling from celo-monorepo.  In particular, this branch has been tested and works with the blockchain parameters e2e test (celo-monorepo PR to come as well).  The key changes involved in this were:

1. Using override types with `gencodec` in order to serialize `big.Int` values as strings rather than numbers, because (a) javascript can't handle such large numbers without losing precision, (b) javascript serializes them using scientific notation, which `big.Int`'s deserialization doesn't accept.
2. Adding support (via a boolean config flag) for using the first validator account as the admin instead of the admin account.  This is because in celo-monorepo it's done this way and the admin account isn't used at all.
3. Creating a new `monorepo` template in mycelo (which uses this validator-instead-of-admin feature)

There is remaining work on the monorepo side in order to support overwriting the migration configuration in `genesis-config.json`, which may be necessary for some of the tests and also maybe for non-test uses later on.  It also remains to make it work with the other e2e tests there.  I didn't try all of them, but at least a couple have some issues (and in one case it seems to be due to a bug in celo-blockchain... more details to come at a later date when I've understood it fully).

### Other changes

* Added cEUR balances to developer accounts (previously had only cUSD and CELO)
* Changed various mycelo genesis config variables from `big.Int` to `uint64`.  My rule of thumb was to use `uint64` if they're fields that could never realistically be larger than Javascript's `Number.MAX_SAFE_INTEGER` (around 9*10^15), which is itself less than the max `uint64`.  Practically, this means using `uint64` just about everything except currencies denominated in Wei (and fixidity decimals, which are already using `fixed`).  As an example, for fields denominated in seconds this maximum value is 285 million years, which is plenty.
* Factored out some common code in the mycelo templates (before adding the new `monorepo` template)
* Trim trailing whitespace in the README (VS Code did it without me noticing until later)

### Tested

* Works in conjunction with corresponding changes in monorepo, such that the blockchain parameters e2e test can now run using mycelo for genesis generation
* Verified that `genesis.json` generated by the existing templates hasn't changes (except for the developer accounts now also getting cEUR)
* Verified that normal commands involved in setting up a testnet work.
* Verified that all the new generated files (`mycelo/genesis/gen_*.go`) are correct by running `go generate ./...` and seeing no changes to any of them.

### Backwards compatibility

Changes to the `genesis-config.json`, but they should be backwards compatible since when parsing it back from json we support either numbers or strings.


mycelo/genesis/base_config.go
```diff
@@ -26,12 +26,10 @@ func BaseConfig() *Config {
...
-			TobinTaxStalenessThreshold: bigInt(3153600000),
-			TobinTax:                   common.Big0,
-			TobinTaxReserveRatio:       common.Big0,
-			DailySpendingRatio:         bigIntStr("50000000000000000000000"),
-			FrozenDays:                 nil,
-			FrozenGold:                 nil,
+			TobinTaxStalenessThreshold: 3153600000,
+			TobinTax:                   fixed("0"),
+			TobinTaxReserveRatio:       fixed("0"),
+			DailySpendingRatio:         fixed("0.05"),
```
mcortesi, line 32: are all this `fixidity` numbers on the smarts contracts?

mycelo/genesis/genesis_state.go
```diff
@@ -580,16 +580,16 @@ func (ctx *deployContext) deployValidators() error {
...
-			ctx.genesisConfig.Validators.ValidatorScoreExponent,
+			new(big.Int).SetUint64(ctx.genesisConfig.Validators.ValidatorLockedGoldRequirements.Duration),
+			new(big.Int).SetUint64(ctx.genesisConfig.Validators.ValidatorScoreExponent),
 			ctx.genesisConfig.Validators.ValidatorScoreAdjustmentSpeed.BigInt(),
-			ctx.genesisConfig.Validators.MembershipHistoryLength,
-			ctx.genesisConfig.Validators.SlashingPenaltyResetPeriod,
-			ctx.genesisConfig.Validators.MaxGroupSize,
-			ctx.genesisConfig.Validators.CommissionUpdateDelay,
-			ctx.genesisConfig.Validators.DowntimeGracePeriod,
+			new(big.Int).SetUint64(ctx.genesisConfig.Validators.MembershipHistoryLength),
```
mcortesi, line 588: nit: tempted to ask for a `bigInt(uint64)` function, to remove bolierplate code :)

mcortesi (approved):
Nice! Love that we are removing most of the big.Int, and the we now have an extra template...

mycelo keeps improving!

mycelo/genesis/base_config.go
oneeman: They are, though I will check to make sure I'm understanding correctly why in some of them there is a `wrap` and in others not. 
 
https://github.com/celo-org/celo-monorepo/blob/0564826758fc171a36bf4131093e52773fad5254/packages/protocol/contracts/stability/Reserve.sol#L100-L101 and 
https://github.com/celo-org/celo-monorepo/blob/0564826758fc171a36bf4131093e52773fad5254/packages/protocol/contracts/stability/Reserve.sol#L158

PR #1421, opened 2021-03-08
Adds passthrough flag to mycelo

trianglesphere:
### Description

1. Fixes help flag nil pointer panic
2. Internal code cleanup
3. Adds pass through flags to geth. This can be used to activate profiling the validator.

### DIFF
--- a/cmd/mycelo/main.go
+++ b/cmd/mycelo/main.go
@@ -75,7 +75,7 @@ var gethPathFlag = cli.StringFlag{
 }
 
 var gethExtraFlagsFlag = cli.StringFlag{
-	Name:  "extraFlags",
+	Name:  "extraflags",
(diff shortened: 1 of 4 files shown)


cmd/mycelo/main.go
```diff
@@ -72,12 +74,23 @@ var gethPathFlag = cli.StringFlag{
 	Usage: "Path to geth binary",
 }
 
+var gethExtraFlagsFlag = cli.StringFlag{
+	Name:  "extraFlags",
```
oneeman, line 78: Maybe name it `extraflags` to be consistent with `maxpending` and `blockgaslimit`

oneeman (approved):
Looks good!  I left a nitpicky comment about one of the flags' naming.

PR #1429, opened 2021-03-10
Split extra flags on whitespace

trianglesphere:
### Description

Split extra flags on whitespace and then pass the array.
This enables passing multiple extra flags through to geth.


oneeman (approved):
LGTM.

PR #1430, opened 2021-03-10
Short circuit in currency compare

trianglesphere:
### Description

Properly dereference the fee currency point such that currency.Cmp short circuits when comparing non-native fee currencies.

### Tested

https://colab.research.google.com/drive/1duSJnIAxUs-3U8YQeEzPGZy8xnpwgWGC#scrollTo=4MyDZZHg_eZh

Tested with all cUSD and a mix of cUSD and CELO fee currencies.

### Related issues

- Partial fix of https://github.com/celo-org/celo-labs/issues/837

### Backwards compatibility

Yes.


mcortesi (approved):
Nice!

PR #1431, opened 2021-03-11
Implement CIP35 for Donut hard fork

oneeman:
### Description

This PR implements CIP35 as specified at https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0035.md.

Summary of changes:

* `Transaction` type has a new field `EthCompatible` to distinguish between normal transactions (with the Celo-only fields) and ethereum-compatible ones (without those fields)
* When decoding transactions of RLP, it identifies the type based on the number of fields the RLP list has
* When encoding transactions to RLP, it encodes them to have either 9 or 12 fields depending on the value of `EthCompatible`
* When generating or verifying transaction signatures, the signing data used is determined by the value of `EthCompatible`
* When serializing to or deserializing from JSON, the `EthCompatible` field is encoded as `ethCompatible`.
* When validating transactions (whether in the tx pool or in the state transition function), reject them if either
   (a) `EthCompatible` is true but Donut has not been activated,
   (b) `EthCompatible` is true but they have non-trivial values for any of the Celo-only fields (which can't happen if decoding from RLP, but can if decoding from JSON), or
   (c) Donut has been activated but the transaction doesn't use replay protection (EIP-155)

### Other changes

* Fix an upstream bug in RLP decoding (fixed upstream by https://github.com/ethereum/go-ethereum/pull/22336)
* Remove the `rlp:nil` tag from `txdata.GatewayFee`.  This tag is not respected for `*big.Int` values, and so its presence was misleading.
* Point the CI to use the celo-monorepo branch `oneeman/cip35-e2e-tests`.  This was not strictly speaking necessary, since I have a PR on celo-monorepo pointing at this branch, the tests will execute there.  Regardless, this change will be reverted to point it back to master

### Tested

* Unit tests for serialization an deserialization of transactions (for both RLP and JSON)
* e2e tests (see https://github.com/celo-org/celo-monorepo/pull/7418) which cover combinations of scenarios such as pre-Donut vs post-Donut, eth-compatible vs normal transactions, sending to a full node or a light client, including vs not including Celo-only fields, sending using `eth_sendTransaction` vs using `eth_sendRawTransaction`.

### Backwards compatibility

* `Transaction` and other types like it (e.g. `RPCTransaction`) have a new field.  However, as the zero value for boolean fields is `false` (which means normal Celo transactions), no changes should be required for software importing `celo-blockchain` packages as a library.
* When Donut is activated, a new transaction type will be valid which isn't currently
* When Donut is activated, transactions without EIP-155 replay protection (i.e. using the chain id) will be rejected.  However, all known Celo wallets and all up to date Ethereum wallets use replay protection, so this isn't likely to affect anyone.
* Before and after Donut, transactions serialized to JSON will have a new boolean field, `ethCompatible`.  This should not break existing software as additional fields are generally ignored during JSON parsing.  In particular, contractkit is not affected, as may be seen in the e2e tests.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -349,6 +349,8 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	// First check this message satisfies all consensus rules before
 	// applying the message. The rules include these clauses
 	//
+	// 0. If the message is from an eth-compatible transaction, that we support those
+	//    and that none of the non-eth-compatible fields are present
 	// 1. the nonce of the message caller is correct
 	// 2. the gas price meets the minimum gas price
 	// 3. caller has enough balance (in the right currency) to cover transaction fee
@@ -356,6 +358,14 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	// 5. the purchased gas is enough to cover intrinsic usage
 	// 6. there is no overflow when calculating intrinsic gas
 	// 7. caller has enough balance to cover asset transfer for **topmost** call
+
+	// Clause 0
+	if st.msg.EthCompatible() && !st.evm.ChainConfig().IsDonut(st.evm.BlockNumber) {
+		return nil, ErrEthCompatibleTransactionsNotSupported
+	}
+	if st.msg.EthCompatible() && !(st.msg.FeeCurrency() == nil && st.msg.GatewayFeeRecipient() == nil && st.msg.GatewayFee().Sign() == 0) {
+		return nil, ErrEthCompatibleTransactionIsntCompatible
+	}
 
 	// Check clauses 1-2
 	if err := st.preCheck(); err != nil {
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -363,8 +363,8 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	if st.msg.EthCompatible() && !st.evm.ChainConfig().IsDonut(st.evm.BlockNumber) {
 		return nil, ErrEthCompatibleTransactionsNotSupported
 	}
-	if st.msg.EthCompatible() && !(st.msg.FeeCurrency() == nil && st.msg.GatewayFeeRecipient() == nil && st.msg.GatewayFee().Sign() == 0) {
-		return nil, ErrEthCompatibleTransactionIsntCompatible
+	if err := vm.CheckEthCompatibility(st.msg); err != nil {
+		return nil, err
 	}
 
 	// Check clauses 1-2
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -255,7 +255,8 @@ type TxPool struct {
 	signer      types.Signer
 	mu          sync.RWMutex
 
-	istanbul bool // Fork indicator whether we are in the istanbul stage.
+	istanbul      bool // Fork indicator whether we are in the istanbul stage.
+	ethCompatible bool // Fork indicator whether we support eth-compatible transactions
 
 	currentState  *state.StateDB // Current state in the blockchain head
 	pendingNonces *txNoncer      // Pending state tracking virtual nonces
@@ -577,6 +578,12 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
+	if tx.EthCompatible() && !pool.ethCompatible {
+		return ErrEthCompatibleTransactionsNotSupported
+	}
+	if tx.EthCompatible() && !(tx.FeeCurrency() == nil && tx.GatewayFeeRecipient() == nil && tx.GatewayFee().Sign() == 0) {
+		return ErrEthCompatibleTransactionIsntCompatible
+	}
 	// Reject transactions over defined size to prevent DOS attacks
 	if uint64(tx.Size()) > txMaxSize {
 		return ErrOversizedData
@@ -1272,6 +1279,7 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	// Update all fork indicator by next pending block number.
 	next := new(big.Int).Add(newHead.Number, big.NewInt(1))
 	pool.istanbul = pool.chainconfig.IsIstanbul(next)
+	pool.ethCompatible = pool.chainconfig.IsDonut(next)
 }
 
 // promoteExecutables moves transactions that have become processable from the
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -255,8 +255,8 @@ type TxPool struct {
 	signer      types.Signer
 	mu          sync.RWMutex
 
-	istanbul      bool // Fork indicator whether we are in the istanbul stage.
-	ethCompatible bool // Fork indicator whether we support eth-compatible transactions
+	istanbul bool // Fork indicator whether we are in the istanbul stage.
+	donut    bool // Fork indicator for the Donut fork.
 
 	currentState  *state.StateDB // Current state in the blockchain head
 	pendingNonces *txNoncer      // Pending state tracking virtual nonces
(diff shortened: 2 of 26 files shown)


core/types/transaction.go
```diff
@@ -31,6 +31,10 @@ import (
 
 //go:generate gencodec -type txdata -field-override txdataMarshaling -out gen_tx_json.go
 
+const (
+	ETH_COMPATIBLE_TX_NUM_FIELDS = 9
```
mcortesi, line 35: My understanding is that naming convention is just camelcase, no upcase por constants

core/types/transaction.go
```diff
@@ -61,6 +65,64 @@ type txdata struct {
...
+
+	// This is only used when marshaling to JSON.
+	Hash *common.Hash `json:"hash" rlp:"-"`
+
+	// Whether this is an ethereum-compatible transaction (i.e. with FeeCurrency, GatewayFeeRecipient and GatewayFee omitted)
+	EthCompatible bool `rlp:"-"`
+}
+
+func toEthCompatibleTxData(data txdata) ethcompatibletxdata {
+	return ethcompatibletxdata{
```
mcortesi, line 94: i'd use named fields to create the struct. too many fields to count index and then validate it's the correct one

core/types/transaction.go
```diff
@@ -61,6 +65,64 @@ type txdata struct {
...
+		data.Payload,
+		data.V,
+		data.R,
+		data.S,
+		data.Hash,
+		false,
+	}
+}
+
+func fromEthCompatibleTxData(data ethcompatibletxdata) txdata {
```
mcortesi, line 109: same

core/types/transaction_signing.go
```diff
@@ -154,18 +154,30 @@ func (s EIP155Signer) SignatureValues(tx *Transaction, sig []byte) (R, S, V *big
...
+		return rlpHash([]interface{}{
+			tx.data.AccountNonce,
+			tx.data.Price,
+			tx.data.GasLimit,
+			tx.data.Recipient,
+			tx.data.Amount,
+			tx.data.Payload,
+			s.chainId, uint(0), uint(0),
+		})
+	} else {
```
mcortesi, line 167: golang lint rules would ask you to remove the `else` claus and just return

core/types/transaction_signing.go
```diff
@@ -219,17 +231,28 @@ func (fs FrontierSigner) SignatureValues(tx *Transaction, sig []byte) (r, s, v *
...
+	if tx.data.EthCompatible {
+		return rlpHash([]interface{}{
+			tx.data.AccountNonce,
+			tx.data.Price,
+			tx.data.GasLimit,
+			tx.data.Recipient,
+			tx.data.Amount,
+			tx.data.Payload,
+		})
+	} else {
```
mcortesi, line 243: same

core/state_transition.go
```diff
@@ -357,6 +359,14 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	// 6. there is no overflow when calculating intrinsic gas
 	// 7. caller has enough balance to cover asset transfer for **topmost** call
 
+	// Clause 0
+	if st.msg.EthCompatible() && !st.evm.ChainConfig().IsDonut(st.evm.BlockNumber) {
+		return nil, ErrEthCompatibleTransactionsNotSupported
+	}
+	if st.msg.EthCompatible() && !(st.msg.FeeCurrency() == nil && st.msg.GatewayFeeRecipient() == nil && st.msg.GatewayFee().Sign() == 0) {
```
mcortesi, line 366: i'd the `st.msg.FeeCurrency() == nil && st.msg.GatewayFeeRecipient() == nil && st.msg.GatewayFee().Sign() == 0` to a method inside `msg` since they are instrinsic to the `msg` and not to `TransitionDb`. 

Maybe a `Valid() bool` msg or `Validate() error`

core/tx_pool.go
```diff
@@ -1272,6 +1282,7 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	// Update all fork indicator by next pending block number.
 	next := new(big.Int).Add(newHead.Number, big.NewInt(1))
 	pool.istanbul = pool.chainconfig.IsIstanbul(next)
+	pool.donut = pool.chainconfig.IsDonut(next)
```
mcortesi, line 1285: do we call `validateTx` after this is enabled? Or old tx will continue to existt on the txpool?

internal/ethapi/api.go
```diff
@@ -1392,6 +1396,10 @@ type SendTxArgs struct {
 
 // setDefaults is a helper function that fills in default values for unspecified tx fields.
 func (args *SendTxArgs) setDefaults(ctx context.Context, b Backend) error {
+	// Reject if Celo-only fields set when EthCompatible is true
+	if args.EthCompatible && !(args.FeeCurrency == nil && args.GatewayFeeRecipient == nil && (args.GatewayFee == nil || args.GatewayFee.ToInt().Sign() == 0)) {
```
mcortesi, line 1400: again we have duplicated "validation logic" here. I understand part of the problem is that `msg` is an interface and there are several implementers? Not sure.

signer/fourbyte/validation.go
```diff
@@ -32,6 +33,10 @@ import (
 func (db *Database) ValidateTransaction(selector *string, tx *core.SendTxArgs) (*core.ValidationMessages, error) {
 	messages := new(core.ValidationMessages)
 
+	// Reject if EthCompatible is true but the transaction has non-nil-or-0 values for the non-eth-compatible fields
+	if tx.EthCompatible && !(tx.FeeCurrency == nil && tx.GatewayFeeRecipient == nil && tx.GatewayFee.ToInt().Sign() == 0) {
```
mcortesi, line 37: duplicated validation logic

mcortesi (commented):
Left some comments

nategraf (commented):
Gave this a look and didn't have any comments beyond what @mcortesi already added 👍

core/types/transaction_signing.go
oneeman: Turns out we don't have golint (or any drop-in replacement) enabled in our golangci config.  If I add it, it gives a ton of errors of different sorts (including some of this rule, e.g. https://github.com/celo-org/celo-blockchain/blob/master/rpc/client.go#L392-L396)

Personally, I am not a fan of this if/else lint rule.  I remember when I first used golint I didn't like it and also didn't like that golint doesn't let you disable individual rules.  But if you like that rule I can change them.

core/tx_pool.go
oneeman: Good point!  `validateTx()` is only called when adding a transaction.  For later on it only checks a subset of what `validateTx()` checks, and this should be checked as well.

internal/ethapi/api.go
oneeman: Yeah, there are several different types (and they don't even all have the same underlying fields, e.g. `SendTxArgs.GatewayFee` is `hexutil.Big` which is a wrapper around `big.Int`.  But I like the idea of moving it out to a method.  It will remove at least once case of duplication.  And, even where it doesn't, it would still be nicer.

core/types/transaction.go
oneeman: Fixed

core/types/transaction.go
oneeman: Done.  I also renamed the `ethcompatibletxdata` type to make it clearer what it's for, and removed the `Hash` and `EthCompatible` fields and tags which weren't necessary for it.

core/types/transaction.go
oneeman: Done (see above)

core/state_transition.go
oneeman: I removed some of the duplication and moved the logic to methods or stand-alone functions.  This meant moving the error from `core` to `types` to avoid circular input or duplicating the error definition.

core/tx_pool.go
oneeman: I added a commit making it so that unprotected transactions are removed when Donut is activated.  I still need to add a test for it, though.

oneeman:
Added a test for `pool.handleDonutActivation`.  Ready for final review

trianglesphere (approved):
What's here looks good to me. Looks like there's a tx pool conflict, and the CI branches need to be straightened out, but otherwise everything looks ready.

oneeman:
> What's here looks good to me. Looks like there's a tx pool conflict, and the CI branches need to be straightened out, but otherwise everything looks ready.

Thanks!  Merged in master and resolved the conflict.

oneeman:
Pointed the CI back to monorepo master.  This means the new e2e tests won't run, but they will after the corresponding PR on monorepo is merged (I believe this is a chicken-and-egg scenario, but I could be wrong)

PR #1432, opened 2021-03-12
Adds istanbul.getLookbackWindow rpc

mcortesi:
Now we are able to obtain the current or a block's lookback window from
the node's perspective.

Fixes web3ext mappings, to properly parse block numbers

### DIFF
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -43,17 +43,21 @@ type API struct {
 
 // getHeaderByNumber retrieves the header requested block or current if unspecified.
 func (api *API) getHeaderByNumber(number *rpc.BlockNumber) (*types.Header, error) {
-	if number == nil || *number == rpc.LatestBlockNumber || *number == rpc.PendingBlockNumber {
-		head := api.chain.CurrentHeader()
-		if head == nil {
-			return nil, errUnknownBlock
-		}
-		return head, nil
+	var header *types.Header
+	if number == nil || *number == rpc.LatestBlockNumber {
+		header = api.chain.CurrentHeader()
+	} else if *number == rpc.PendingBlockNumber {
+		return nil, fmt.Errorf("can't use pending block within istanbul")
 	} else if *number == rpc.EarliestBlockNumber {
-		return api.chain.GetHeaderByNumber(0), errUnknownBlock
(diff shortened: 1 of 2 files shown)


consensus/istanbul/backend/api.go
```diff
@@ -42,6 +42,21 @@ type API struct {
...
 // getHeaderByNumber retrieves the header requested block or current if unspecified.
+func (api *API) getHeaderByNumber(number *rpc.BlockNumber) (*types.Header, error) {
+	if number == nil || *number == rpc.LatestBlockNumber || *number == rpc.PendingBlockNumber {
+		head := api.chain.CurrentHeader()
+		if head == nil {
+			return nil, errUnknownBlock
+		}
+		return head, nil
+	} else if *number == rpc.EarliestBlockNumber {
+		return api.chain.GetHeaderByNumber(0), errUnknownBlock
```
oneeman, line 53: This `errUnknownblock` doesn't sound right.  I guess it should also check against nil and return error if appropriate

consensus/istanbul/backend/api.go
```diff
@@ -42,6 +42,21 @@ type API struct {
...
+	}
+
+	return api.chain.GetHeaderByNumber(uint64(*number)), nil
```
oneeman, line 56: Shouldn't this check the value and return `nil, errUnknownBlock` if it's `nil`?  Sounds like this function will need three nil checks.  Maybe set the result to a variable and then at the end do a single check against nil and then return according to that?

consensus/istanbul/backend/api.go
```diff
@@ -42,6 +42,21 @@ type API struct {
 }
 
 // getHeaderByNumber retrieves the header requested block or current if unspecified.
+func (api *API) getHeaderByNumber(number *rpc.BlockNumber) (*types.Header, error) {
+	if number == nil || *number == rpc.LatestBlockNumber || *number == rpc.PendingBlockNumber {
```
oneeman, line 46: I guess you shouldn't really call this with pending.  Not sure if it's better to treat it the same as latest or to return an error.  Probably doesn't matter much either way.

oneeman (commented):
Added some comments on `getHeaderByNumber()`.

For the methds which were missing the formatter, how were they failing before?

oneeman (approved):
LGTM!

PR #1435, opened 2021-03-15
Refactor load bot to reduce CPU Usage

trianglesphere:
### Description

Refactors the load bot to reduce CPU Usage. This does it by having one thread that is on a ticker and firing off single run go-rountine transfers.

Previously there where `developer account` go routines being ran, and each was sleeping until it should send its transaction (through a couple of calculations to ensure that the TPS was correct). It appears to be a semi-known issue that ticker's (and fewer of them) seem to perform better than sleeping.

### Tested

Tested on a gcp vm instance. At 5s block times, order 50-100 TPS, cpu usage dropped from a couple hundred percent (out of 400% b/c 4 core machine), to 10-15% sustained. (The old version actually does ok at ~25 TPS, but gets much worse as the load increases)

### DIFF
--- a/mycelo/loadbot/bot.go
+++ b/mycelo/loadbot/bot.go
@@ -21,12 +21,6 @@ var maxPending uint64
 var pending uint64
 var pendingLock sync.Mutex
 
-// Range represents an inclusive big.Int range
-type Range struct {
-	From *big.Int
-	To   *big.Int
-}
-
 // Config represent the load bot run configuration
 type Config struct {
 	Accounts              []env.Account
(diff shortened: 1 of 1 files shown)


mcortesi (approved):
LGTM!

The only change i would make, is from a previous change: I'd remove static/module variables in favor of instance variables. So `pending`, `pendingLock` can be part of a `LoadGenerator` type.

trianglesphere:
> The only change i would make, is from a previous change: I'd remove static/module variables in favor of instance variables. So `pending`, `pendingLock` can be part of a `LoadGenerator` type.

I switched it to this and actually found a bug in my refactor

PR #1440, opened 2021-03-19
Fixing dev mode and eth_call

mrsmkl:
### Description

Changed it so that validator and tx fee recipient do not have to be specified when running in dev mode.
Also looks like eth_call rpc call failed when there's no specified "from", added a special case for that.

### Tested

Only manual testing.

### Backwards compatibility


oneeman (changes requested):
The dev mode changes look good.  I guess we don't really use it so we didn't notice it was broken by the split etherbase PR.

But I think we shouldn't make the state transition changes.  `eth_call` can use gas price 0 (and defaults to 0 if it's not specified).  So the error due to not being able to pay fees only happens if the caller actively specifies a non-zero gas price and can't afford to pay it.  And in that case the error is correct, so I don't favor modifying that.

mcortesi:
too old, closing

PR #1442, opened 2021-03-19
Cherry pick tx pool fixes from v1.2.3 and v1.2.4

oneeman:
### Description

Cherry-pick to master the security fixes from v1.2.3 and v1.2.4.

### Tested

* The changes themselves have already been tested and deployed in v1.2.4.
* Verified that this diff is like the diff between 1.2.2 and 1.2.4 except for line number differences and the versions.

### Backwards compatibility

No incompatibility.


trianglesphere (approved):
LGTM, the patch is the same.

PR #1443, opened 2021-03-19
Refactor contract comm internal caller interface

trianglesphere:
### Description

The core of this PR is the addition of this interface:
```
// SystemContractCaller runs core contract calls. The implementation provides the state on which the calls are ran
type SystemContractCaller interface {
	AddressCaller
	RegistryCaller
	GetRegisteredAddress(registryId common.Hash) (*common.Address, error)
}

// AddressCaller makes a call with a specified address
type AddressCaller interface {
	MakeStaticCallWithAddress(contractAddress common.Address, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64) (uint64, error)
	MakeCallWithAddress(contractAddress common.Address, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int) (uint64, error)
}

// RegistryCaller looks up the contract address from the registry ID and then makes an address cal
type RegistryCaller interface {
	MakeStaticCall(registryId common.Hash, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64) (uint64, error)
	MakeCall(registryId common.Hash, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int) (uint64, error)
}
```

Previously we had a similar interface, but explicitly passed a `header, statedb` tuple to each function. It was optionally nil to signify to use the current state. I've then created a couple implementations of the `SystemContractCaller` where 1 uses the current block chain state and where 1 uses the passed in state.

The external interface has not changed, so we still rely on the EVM singleton. Long term the goal would be to construct a SystemContractCaller for the current state or for a specified state and pass it as a parameters to the contract comm functions.

In terms of prep for building the cache, this lets me simplify what was `contract_comm/evm.go` (now `contract_comm/caller.go` and `contract_comm/types.go` to more easily add the memoized versions of the static call.

### Tested

Builds, lint, unit test


contract_comm/types.go
```diff
@@ -0,0 +1,137 @@
...
+// SetIEVMHandler must be called prior to using this (to maintain backwards compatibility)
+func NewCaller(header *types.Header, state vm.StateDB) SystemContractCaller {
+	if header == nil || state == nil || reflect.ValueOf(state).IsNil() {
+		return NewCurrentStateCaller()
+	}
+	var chain vm.ChainContext
+	if internalEvmHandlerSingleton != nil {
+		chain = internalEvmHandlerSingleton.chain
+	} else {
+		chain = nil
```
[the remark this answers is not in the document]
trianglesphere: I'd like to move towards failing at `NewCaller` and `NewCurrentStateCaller`, but I want to do that after we pass the callers through to the contract comm interface - that way we create fewer callers and have to deal with less error handling.

`createEVM` basically has to return an error (in practice once we can guarantee we have a chain reference it won't fail, but it's possible to error on `chain.State()`), so for now it does double duty, and will be less overloaded once I'm able to make the other changes.

piersy:
Hi @trianglesphere could you provide a bit more context for this PR?

Why do we want this new interface? What is the purpose of the refactor? What is the external interface?

trianglesphere:
> What is the purpose of the refactor? 

This is partially cleaning up what was `contract_comm/evm.go` so I could add a memoized version of `MakeStaticCall` and partially setting up further cleanup around the current external interface to contract comm.

> What is the external interface?

One example of the external interface is this function from `contract_comm/election/election.go`
```
GetElectedValidators(header *types.Header, state vm.StateDB) ([]common.Address, error)
```
We provide then create the EVM (execution environment) from `header`, `state`, and the `IEVMHandlerSingleton`. If `header` and `state` are nil (actually it's or, but the logic is weird if one is nil but the other is not), it gets the current header and state from the `IEVMHandlerSingleton`.

> Why do we want this new interface? 

It lets us move away from a singleton providing access to the current blockchain state to the EVM by wrapping up the `state` and `header` parameters into a `ContractCaller` interface before handing it off to the contract comm layer.


For the ticket associated with it, it'd be https://github.com/celo-org/celo-labs/issues/837. I've split the PRs into two to fix this issue: this one does what should be a straightforwards refactor and the second implements the cache logic (https://github.com/celo-org/celo-blockchain/pull/1444)

contract_comm/blockchain_parameters/blockchain_parameters.go
```diff
@@ -113,15 +113,18 @@ func init() {
 func GetMinimumVersion(header *types.Header, state vm.StateDB) (*params.VersionInfo, error) {
 	version := [3]*big.Int{big.NewInt(0), big.NewInt(0), big.NewInt(0)}
 	var err error
-	_, err = contract_comm.MakeStaticCall(
+	comm, err := caller.NewCaller(header, state)
```
nategraf, line 116: Just from a naming perspective, this is a bit odd to me. Why call the package `caller` and the `Caller` object `comm` instead of the other way around?

contract_comm/caller.go
```diff
@@ -0,0 +1,119 @@
...
+// Contract Communicator Implementation
+func (c contractCommunicator) MakeStaticCallWithAddress(contractAddress common.Address, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64) (uint64, error) {
+	evm, err := c.builder.createEVM()
+	if err != nil {
+		return 0, err
+	}
+	return evm.StaticCallFromSystem(contractAddress, abi, funcName, args, returnObj, gas)
+}
+
+func (c contractCommunicator) MakeCallWithAddress(contractAddress common.Address, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int) (uint64, error) {
```
nategraf, line 50: My initial thought in reading `MakeCallWithAddress` was that the call would be made _from_ the given address rather than to it. You might consider renaming this to `MakeCallToAddress`

contract_comm/caller.go
```diff
@@ -0,0 +1,119 @@
...
+		if err == ccerrors.ErrSmartContractNotDeployed {
+			log.Debug("Contract not yet registered", "function", funcName, "registryId", hexutil.Encode(registryId[:]))
+			return common.Address{}, err
+		} else if err == ccerrors.ErrRegistryContractNotDeployed {
+			log.Debug("Registry contract not yet deployed", "function", funcName, "registryId", hexutil.Encode(registryId[:]))
+			return common.Address{}, err
+		} else {
+			log.Error("Error in getting registered address", "function", funcName, "registryId", hexutil.Encode(registryId[:]), "err", err)
+			return common.Address{}, err
+		}
```
nategraf, line 115: Seems like we should have lifted the return statement out of the conditionals before.
```suggestion
		if err == ccerrors.ErrSmartContractNotDeployed {
			log.Debug("Contract not yet registered", "function", funcName, "registryId", hexutil.Encode(registryId[:]))
		} else if err == ccerrors.ErrRegistryContractNotDeployed {
			log.Debug("Registry contract not yet deployed", "function", funcName, "registryId", hexutil.Encode(registryId[:]))
		} else {
			log.Error("Error in getting registered address", "function", funcName, "registryId", hexutil.Encode(registryId[:]), "err", err)
		}
		return common.Address{}, err
```

contract_comm/types.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Creates a new EVM on demand.
+type evmBuilder interface {
+	// createEVM creates an EVM. May fail due to `chain.State()`
+	createEVM() (*vm.EVM, error)
+}
+
+// Private struct that implements SystemContractCaller.
+type contractCommunicator struct {
```
nategraf, line 57: I think we should keep the struct definition with the implementation, and would recommend we move this to `caller.go`

contract_comm/types.go
```diff
@@ -0,0 +1,137 @@
...
+
+// specificStateCaller implements evmBuilder for a given state. This allows core contract calls against and arbitrary state
+type specificStateCaller struct {
+	header *types.Header
+	state  vm.StateDB
+	chain  vm.ChainContext
+}
+
+func (c specificStateCaller) createEVM() (*vm.EVM, error) {
+	if internalEvmHandlerSingleton == nil || c.chain == nil {
```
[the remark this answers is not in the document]
nategraf: I also found it odd that we are checking for `internalEvmHandlersingleton` here since it is not used. What is the reason for that?

contract_comm/types.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// The EVM Context requires a msg, but the actual field values don't really matter for this case.
+	// Putting in zero values.
+	context := vm.NewEVMContext(emptyMessage, header, internalEvmHandlerSingleton.chain, nil)
+	return vm.NewEVM(context, state, internalEvmHandlerSingleton.chain.Config(), *internalEvmHandlerSingleton.chain.GetVMConfig()), nil
+}
+
+// NewCaller creates a caller object that acts on the supplied statedb in the environment of the header, state
+// SetIEVMHandler must be called prior to using this (to maintain backwards compatibility)
```
nategraf, line 116: Can you elaborate a bit more here on why it would break compatibility to ignore `internalEvmHandlerSingleton` here?

contract_comm/types.go
```diff
@@ -0,0 +1,135 @@
...
+func NewCaller(header *types.Header, state vm.StateDB) (SystemContractCaller, error) {
```
nategraf, line 117: Related to my comment above, I'd recommend moving `NewCaller` and `NewCurrentStateCaller` into `caller.go`

contract_comm/blockchain_parameters/blockchain_parameters.go
trianglesphere: It probably makes sense to switch it. It's like that as a hold over from a couple earlier changes.

contract_comm/types.go
trianglesphere: It's backward compatibility with the external contract comm interface of `header, statedb`. I'd like to get away from that and instead provide the contract comm functions with a caller object that can do that instead.

contract_comm/types.go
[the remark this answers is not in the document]
trianglesphere: It's weirdness around the fact that the specific state caller needs the current chain config. Removing this check would be part of completely removing the singleton and `header, state` interface.

trianglesphere:
Not quite sure what the final approach will be, but it does not appear to be this.

PR #1444, opened 2021-03-19
Adds memoization to static core contract calls

trianglesphere:
### Description

This sits on top of https://github.com/celo-org/celo-blockchain/pull/1443

This extends the `SystemContractCaller` with memoized versions of the StaticCall functions. This uses the state root of the statedb as part of the key and then the transaction data as the other part. It will only use the cache when it thinks that the statedb is clean. This also provides a memoized version of `GetRegisteredAddress`. `MakeMemoizedStaticCall` uses the memoized `GetRegisteredAddress` when doing the registry lookup.

In the memoized static call it will deep copy the resulting bytes before returning them or placing them in the cache to ensure that a consumer does not change the value for a later user.

For testing this provides the flag `vm.checkcache=bool`. If true the memoized static call will re-compute the result of the static call and compare it against the cached value. It will log an error if the value is not the same.

### Tested

Verified that I got the number of expected cache hits and misses during a load test.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/837


trianglesphere:
Not quite sure what the final approach will be, but it does not appear to be this.

PR #1445, opened 2021-03-22
Verify commit message BLS signatures in aggregate

### LINKED ISSUE #1434: Verify bls signatures in aggregate rather than individually
Currently verifying bls signatures is a bottleneck for consensus - https://www.notion.so/clabsco/Aggregated-BLS-signature-verification-for-commit-messages-a837c2aac87040cbb08b9fc518c4f701

In order to be able to verify bls signatures in aggregate the structure of the code needs to be modified to support this, since at present bls signatures are verified individually, allowing us to gate their addition to the round state. If we move to verifying the signatures in aggregate we need a way to be able to identify which signatures are bad if the aggregate is bad and also a way to remove or mark the bad messages so that they do not break subsequent attempts to verify the signature in aggregate.

To be considered done we need to run a testnet and see that it continues to function in the face of validators sending messages with bad bls signatures.

### LINKED ISSUE #1474: Benchmark performance enhancement provided by verifying commit bls signatures in aggregate
## Description

Benchmark the performance impact of verifying commit bls signatures in aggregate.

## Output

Numbers showing how the commit step times have changed and also numbers showing how the network throughput has changed.

piersy:
### Description

This PR modifies the processing of consensus commit messages so that instead of individually having their BLS signatures verified they now have their signatures verified in aggregate (takes approximately the same time as a single signature verification regardless of the number of signatures aggregated see [here](https://www.notion.so/clabsco/Aggregated-BLS-signature-verification-for-commit-messages-a837c2aac87040cbb08b9fc518c4f701) for more depth).

The reason for doing this is to improve performance, since BLS signature verification is costly and this was identified as a bottleneck in the commit phase of consensus.

The aggregate signature verification occurs once a quorum of commit messages is received by a node, if that fails, the node reverts to checking commit message signatures individually and then building an aggregate from those.

A side effect of this approach is that now, commit messages cannot count towards the transition from preprepared to prepared, because the only point at which we know we have valid commit messages is when we are completing the commit phase. This may impact the performance of the consensus algorithm because it may delay nodes in reaching the prepared phase.

### Other changes

Added a benchmark of BLS signature verification.

### Tested

Tested with a local mycelo testnet, both this branch and a branch where validators introduce bad signatures, see the python notebook [here](https://colab.research.google.com/drive/1LIT-mw8mN7XnSo554ukaL7WwyCIiO-Br#scrollTo=bFMW0TQ2XBPC) for more detail and instructions to recreate. The results showed a small improvement in overall performance.

Also performance tested on 3 large scale networks with 100 validators (80 with proxies). I used 3 different branches for this test:

* [baseline](https://github.com/celo-org/celo-blockchain/tree/piersy/benchmark%2Fbls-aggregate-verification_baseline) - The branch point.
* [good](https://github.com/celo-org/celo-blockchain/tree/piersy/benchmark%2Fbls-aggregate-verification_good) - Essentially these changes but with extra timing metrics.
* [bad-signatures](https://github.com/celo-org/celo-blockchain/tree/piersy/benchmark%2Fbls-aggregate-verification_bad-signatures) - A version where validators produce invalid signatures roughly 1/8th of the time.

None of these tests had any transaction load being applied since the performance enhancement is an improvement of the consensus handling code, so leaving out transactional load gives a clearer view of the gains provided by this change.

I measured 3 metrics for this testing, sleep time (how long do nodes sleep before processing the next block), block period (time between blocks) and consensus time (time between node sending first prepare message to the point they start the next round).

Results of this testing are:

| Branch | Sleep time | Block period | Consensus time | 
| ------------- | ------------- | ------------- | ------------- |
| baseline  | 4.452s  | 5.0375s  | 597ms  |
| good | 4.718s   | 5.042  | 283ms  |
| bad-signatures | 4.490  |5.035s | 1.128s  |

So we can see that we have a saving of 314ms in consensus time from the baseline to the good implementation but that translates into 266ms increase in sleep time. Strangely for the bad-signatures despite consensus time almost doubling (an extra ~530ms)  compared to the baseline the sleep time only reduced by ~40ms compared to the baseline.

Block period doesn't seem to be much affected, although oddly seems to be longer when performing the only aggregate signature verification.

## Related issues

- Fixes #1474
- Fixes #1434 
### Backwards compatibility

This should be backwards compatible, no external/public APIs have changed and the block structure or validation rules have not changed.

### DIFF
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -21,7 +21,6 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"errors"
-	"fmt"
 	"io"
 	"math"
 	"time"
@@ -346,74 +345,6 @@ func (sb *Backend) pruneAnnounceDataStructures() error {
 	return nil
 }
 
-// ===============================================================
-//
-// define the IstanbulQueryEnode message format, the QueryEnodeMsgCache entries, the queryEnode send function (both the gossip version and the "retrieve from cache" version), and the announce get function
-
-type encryptedEnodeURL struct {
-	DestAddress       common.Address
-	EncryptedEnodeURL []byte
-}
-
-func (ee *encryptedEnodeURL) String() string {
-	return fmt.Sprintf("{DestAddress: %s, EncryptedEnodeURL length: %d}", ee.DestAddress.String(), len(ee.EncryptedEnodeURL))
-}
-
-type queryEnodeData struct {
-	EncryptedEnodeURLs []*encryptedEnodeURL
-	Version            uint
-	// The timestamp of the node when the message is generated.
-	// This results in a new hash for a newly generated message so it gets regossiped by other nodes
-	Timestamp uint
-}
-
-func (qed *queryEnodeData) String() string {
-	return fmt.Sprintf("{Version: %v, Timestamp: %v, EncryptedEnodeURLs: %v}", qed.Version, qed.Timestamp, qed.EncryptedEnodeURLs)
-}
-
-// ==============================================
-//
-// define the functions that needs to be provided for rlp Encoder/Decoder.
-
-// EncodeRLP serializes ar into the Ethereum RLP format.
-func (ee *encryptedEnodeURL) EncodeRLP(w io.Writer) error {
-	return rlp.Encode(w, []interface{}{ee.DestAddress, ee.EncryptedEnodeURL})
-}
-
-// DecodeRLP implements rlp.Decoder, and load the ar fields from a RLP stream.
-func (ee *encryptedEnodeURL) DecodeRLP(s *rlp.Stream) error {
-	var msg struct {
-		DestAddress       common.Address
-		EncryptedEnodeURL []byte
-	}
-
-	if err := s.Decode(&msg); err != nil {
-		return err
-	}
-	ee.DestAddress, ee.EncryptedEnodeURL = msg.DestAddress, msg.EncryptedEnodeURL
-	return nil
-}
-
-// EncodeRLP serializes ad into the Ethereum RLP format.
-func (qed *queryEnodeData) EncodeRLP(w io.Writer) error {
-	return rlp.Encode(w, []interface{}{qed.EncryptedEnodeURLs, qed.Version, qed.Timestamp})
-}
-
-// DecodeRLP implements rlp.Decoder, and load the ad fields from a RLP stream.
-func (qed *queryEnodeData) DecodeRLP(s *rlp.Stream) error {
-	var msg struct {
-		EncryptedEnodeURLs []*encryptedEnodeURL
-		Version            uint
-		Timestamp          uint
-	}
-
-	if err := s.Decode(&msg); err != nil {
-		return err
-	}
-	qed.EncryptedEnodeURLs, qed.Version, qed.Timestamp = msg.EncryptedEnodeURLs, msg.Version, msg.Timestamp
-	return nil
-}
-
 // getValProxyAssignments returns the remote validator -> external node assignments.
 // If this is a standalone validator, it will set the external node to itself.
 // If this is a proxied validator, it will set external node to the proxy's external node.
@@ -580,32 +511,19 @@ func (sb *Backend) generateQueryEnodeMsg(version uint, enodeQueries []*enodeQuer
 		logger.Trace("No encrypted enodeURLs were generated, will not generate encryptedEnodeMsg")
 		return nil, nil
 	}
-	queryEnodeData := &queryEnodeData{
+
+	msg := istanbul.NewMessage(&istanbul.QueryEnodeData{
 		EncryptedEnodeURLs: encryptedEnodeURLs,
 		Version:            version,
 		Timestamp:          getTimestamp(),
-	}
-
-	queryEnodeBytes, err := rlp.EncodeToBytes(queryEnodeData)
-	if err != nil {
-		logger.Error("Error encoding queryEnode content", "QueryEnodeData", queryEnodeData.String(), "err", err)
-		return nil, err
-	}
-
-	msg := &istanbul.Message{
-		Code:      istanbul.QueryEnodeMsg,
-		Msg:       queryEnodeBytes,
-		Address:   sb.Address(),
-		Signature: []byte{},
-	}
-
+	}, sb.Address())
 	// Sign the announce message
 	if err := msg.Sign(sb.Sign); err != nil {
 		logger.Error("Error in signing a QueryEnode Message", "QueryEnodeMsg", msg.String(), "err", err)
 		return nil, err
 	}
 
-	logger.Debug("Generated a queryEnode message", "IstanbulMsg", msg.String(), "QueryEnodeData", queryEnodeData.String())
+	logger.Debug("Generated a queryEnode message", "IstanbulMsg", msg.String(), "QueryEnodeData", msg.QueryEnodeMsg().String())
 
 	return msg, nil
 }
@@ -617,10 +535,10 @@ type enodeQuery struct {
 }
 
 // generateEncryptedEnodeURLs returns the encryptedEnodeURLs to be sent in an enode query.
-func (sb *Backend) generateEncryptedEnodeURLs(enodeQueries []*enodeQuery) ([]*encryptedEnodeURL, error) {
+func (sb *Backend) generateEncryptedEnodeURLs(enodeQueries []*enodeQuery) ([]*istanbul.EncryptedEnodeURL, error) {
 	logger := sb.logger.New("func", "generateEncryptedEnodeURLs")
 
-	var encryptedEnodeURLs []*encryptedEnodeURL
+	var encryptedEnodeURLs []*istanbul.EncryptedEnodeURL
 	for _, param := range enodeQueries {
 		logger.Debug("encrypting enodeURL", "externalEnodeURL", param.enodeURL, "publicKey", param.recipientPublicKey)
 		publicKey := ecies.ImportECDSAPublic(param.recipientPublicKey)
@@ -630,7 +548,7 @@ func (sb *Backend) generateEncryptedEnodeURLs(enodeQueries []*enodeQuery) ([]*en
 			return nil, err
 		}
 
-		encryptedEnodeURLs = append(encryptedEnodeURLs, &encryptedEnodeURL{
+		encryptedEnodeURLs = append(encryptedEnodeURLs, &istanbul.EncryptedEnodeURL{
 			DestAddress:       param.recipientAddress,
 			EncryptedEnodeURL: encEnodeURL,
 		})
@@ -672,7 +590,7 @@ func (sb *Backend) handleQueryEnodeMsg(addr common.Address, peer consensus.Peer,
 		return errUnauthorizedAnnounceMessage
 	}
 
-	var qeData queryEnodeData
+	var qeData istanbul.QueryEnodeData
 	err = rlp.DecodeBytes(msg.Msg, &qeData)
 	if err != nil {
 		logger.Warn("Error in decoding received Istanbul QueryEnode message content", "err", err, "IstanbulMsg", msg.String())
@@ -681,7 +599,7 @@ func (sb *Backend) handleQueryEnodeMsg(addr common.Address, peer consensus.Peer,
 
 	logger = logger.New("msgAddress", msg.Address, "msgVersion", qeData.Version)
 
-	// Do some validation checks on the queryEnodeData
+	// Do some validation checks on the istanbul.QueryEnodeData
 	if isValid, err := sb.validateQueryEnode(msg.Address, &qeData); !isValid || err != nil {
 		logger.Warn("Validation of queryEnode message failed", "isValid", isValid, "err", err)
 		return err
@@ -776,7 +694,7 @@ func (sb *Backend) answerQueryEnodeMsg(address common.Address, node *enode.Node,
 // message. This is to force all validators that send a queryEnode message to
 // create as succint message as possible, and prevent any possible network DOS attacks
 // via extremely large queryEnode message.
-func (sb *Backend) validateQueryEnode(msgAddress common.Address, qeData *queryEnodeData) (bool, error) {
+func (sb *Backend) validateQueryEnode(msgAddress common.Address, qeData *istanbul.QueryEnodeData) (bool, error) {
 	logger := sb.logger.New("func", "validateQueryEnode", "msg address", msgAddress)
 
 	// Check if there are any duplicates in the queryEnode message
@@ -1267,19 +1185,10 @@ func (sb *Backend) generateEnodeCertificateMsgs(version uint) (map[enode.ID]*ist
 	}
 
 	for _, externalNode := range externalEnodes {
-		enodeCertificate := &istanbul.EnodeCertificate{
-			EnodeURL: externalNode.URLv4(),
-			Version:  version,
-		}
-		enodeCertificateBytes, err := rlp.EncodeToBytes(enodeCertificate)
-		if err != nil {
-			return nil, err
-		}
-		msg := &istanbul.Message{
-			Code:    istanbul.EnodeCertificateMsg,
-			Address: sb.Address(),
-			Msg:     enodeCertificateBytes,
-		}
+		msg := istanbul.NewMessage(
+			&istanbul.EnodeCertificate{EnodeURL: externalNode.URLv4(), Version: version},
+			sb.Address(),
+		)
 		// Sign the message
 		if err := msg.Sign(sb.Sign); err != nil {
 			return nil, err
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -32,7 +32,6 @@ import (
 	"github.com/celo-org/celo-blockchain/crypto/ecies"
 	"github.com/celo-org/celo-blockchain/p2p"
 	"github.com/celo-org/celo-blockchain/p2p/enode"
-	"github.com/celo-org/celo-blockchain/rlp"
 )
 
 // ==============================================
@@ -587,17 +586,11 @@ func (sb *Backend) handleQueryEnodeMsg(addr common.Address, peer consensus.Peer,
 		return errUnauthorizedAnnounceMessage
 	}
 
-	var qeData istanbul.QueryEnodeData
-	err = rlp.DecodeBytes(msg.Msg, &qeData)
-	if err != nil {
-		logger.Warn("Error in decoding received Istanbul QueryEnode message content", "err", err, "IstanbulMsg", msg.String())
-		return err
-	}
-
+	qeData := msg.QueryEnodeMsg()
 	logger = logger.New("msgAddress", msg.Address, "msgVersion", qeData.Version)
 
 	// Do some validation checks on the istanbul.QueryEnodeData
-	if isValid, err := sb.validateQueryEnode(msg.Address, &qeData); !isValid || err != nil {
+	if isValid, err := sb.validateQueryEnode(msg.Address, msg.QueryEnodeMsg()); !isValid || err != nil {
 		logger.Warn("Validation of queryEnode message failed", "isValid", isValid, "err", err)
 		return err
 	}
@@ -753,17 +746,6 @@ func (sb *Backend) regossipQueryEnode(msg *istanbul.Message, msgTimestamp uint, 
 	sb.lastQueryEnodeGossiped[msg.Address] = time.Now()
 
 	return nil
-}
-
-func (sb *Backend) generateVersionCertificate(version uint) (*istanbul.VersionCertificate, error) {
-	vc := &istanbul.VersionCertificate{
-		Version: version,
-	}
-	err := vc.Sign(sb.Sign)
-	if err != nil {
-		return nil, err
-	}
-	return vc, nil
 }
 
 func (sb *Backend) gossipVersionCertificatesMsg(versionCertificates []*istanbul.VersionCertificate) error {
@@ -985,7 +967,7 @@ func (sb *Backend) setAndShareUpdatedAnnounceVersion(version uint) error {
 	}
 
 	// Generate and gossip a new version certificate
-	newVersionCertificate, err := sb.generateVersionCertificate(version)
+	newVersionCertificate, err := istanbul.NewVersionCertificate(version, sb.Sign)
 	if err != nil {
 		return err
 	}
@@ -1082,11 +1064,7 @@ func (sb *Backend) handleEnodeCertificateMsg(_ consensus.Peer, payload []byte) e
 	}
 	logger = logger.New("msg address", msg.Address)
 
-	var enodeCertificate istanbul.EnodeCertificate
-	if err := rlp.DecodeBytes(msg.Msg, &enodeCertificate); err != nil {
-		logger.Warn("Error in decoding received Istanbul Enode Certificate message content", "err", err, "IstanbulMsg", msg.String())
-		return err
-	}
+	enodeCertificate := msg.EnodeCertificate()
 	logger.Trace("Received Istanbul Enode Certificate message", "enodeCertificate", enodeCertificate)
 
 	parsedNode, err := enode.ParseV4(enodeCertificate.EnodeURL)
@@ -1137,10 +1115,7 @@ func (sb *Backend) SetEnodeCertificateMsgMap(enodeCertMsgMap map[enode.ID]*istan
 
 	// Verify that all of the certificates have the same version
 	for _, enodeCertMsg := range enodeCertMsgMap {
-		var enodeCert istanbul.EnodeCertificate
-		if err := rlp.DecodeBytes(enodeCertMsg.Msg.Msg, &enodeCert); err != nil {
-			return err
-		}
+		enodeCert := enodeCertMsg.Msg.EnodeCertificate()
 
 		if enodeCertVersion == nil {
 			enodeCertVersion = &enodeCert.Version
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -766,9 +766,9 @@ var versionCertificateSalt = []byte("versionCertificate")
 
 // versionCertificate is a signed message from a validator indicating the most
 // recent version of its enode.
-type versionCertificate istanbul.VersionCertificateEntry
+type versionCertificate istanbul.VersionCertificate
 
-func newVersionCertificateFromEntry(entry *istanbul.VersionCertificateEntry) *versionCertificate {
+func newVersionCertificateFromEntry(entry *istanbul.VersionCertificate) *versionCertificate {
 	return &versionCertificate{
 		Address:   entry.Address,
 		PublicKey: entry.PublicKey,
@@ -791,7 +791,7 @@ func (vc *versionCertificate) Sign(signingFn func(data []byte) ([]byte, error)) 
 
 // RecoverPublicKeyAndAddress recovers the ECDSA public key and corresponding
 // address from the Signature
-func (vc *versionCertificate) RecoverPublicKeyAndAddress() error {
+func RecoverPublicKeyAndAddress(vc istanbul.VersionCertificate) error {
 	payloadToSign, err := vc.payloadToSign()
 	if err != nil {
 		return err
@@ -833,8 +833,8 @@ func (vc *versionCertificate) DecodeRLP(s *rlp.Stream) error {
 	return nil
 }
 
-func (vc *versionCertificate) Entry() *istanbul.VersionCertificateEntry {
-	return &istanbul.VersionCertificateEntry{
+func (vc *versionCertificate) Entry() *istanbul.VersionCertificate {
+	return &istanbul.VersionCertificate{
 		Address:   vc.Address,
 		PublicKey: vc.PublicKey,
 		Version:   vc.Version,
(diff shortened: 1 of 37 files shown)


consensus/istanbul/core/core.go
```diff
@@ -215,33 +218,46 @@ func PrepareCommittedSeal(hash common.Hash, round *big.Int) []byte {
 	return buf.Bytes()
 }
 
-// GetAggregatedSeal aggregates all the given seals for a given message set to a bls aggregated
-// signature and bitmap
-func GetAggregatedSeal(seals MessageSet, round *big.Int) (types.IstanbulAggregatedSeal, error) {
-	bitmap := big.NewInt(0)
-	committedSeals := make([][]byte, seals.Size())
-	for i, v := range seals.Values() {
+func (c *core) PrepareEpochValidatorSetSeal(comSub *istanbul.CommittedSubject, newValSet istanbul.ValidatorSet) (seal, sealExtraData []byte, cip22 bool, err error) {
```
trianglesphere, line 221: Dangling function?

consensus/istanbul/core/commit.go
```diff
@@ -207,62 +218,135 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
+				logger.Warn("Error on commit, waiting for desired round", "reason", "getAggregatedSeal", "err", err, "desired_round", nextRound)
+				c.waitForDesiredRound(nextRound)
+				return nil
+			}
+
+			aggregatedEpochValidatorSetSeal.Bitmap = epochBitmap
+			aggregatedEpochValidatorSetSeal.Signature = epochAggregate
 		}
 
-	} else if c.current.GetPrepareOrCommitSize() >= minQuorumSize && c.current.State().Cmp(StatePrepared) < 0 {
```
trianglesphere, line 250: GH appears to be mangling this, but it looks like we no longer consider commit message for this transition, so I think it makes sense to remove it from the commit function.

consensus/istanbul/core/commit.go
```diff
@@ -207,62 +218,135 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
+		return types.IstanbulAggregatedSeal{}, fmt.Errorf("failed verify aggregate seal: %v", err)
+	}
+	return aggregatedSeal, nil
+}
+
+// removeInvalidCommittedSeals individually verifies the committed seal on each
+// commit message and discards commits with invalid committed seals.
+//
+// Note that when commit messages are discarded they are not removed from the
+// round state database, this should not pose a problem however because if this
```
trianglesphere, line 334: Nit: maybe "the deletion/removal is not persisted to disk"? When I first read this I assumed that it was not removed from the commit set and that confused me for a bit.

consensus/istanbul/core/commit.go
trianglesphere: Maybe I was looking at an old version and this isn't here anymore.

consensus/istanbul/core/commit.go
```diff
@@ -207,62 +218,135 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
+// step is reached again they will again be discarded.
+func (c *core) removeInvalidCommittedSeals() {
+
+	commits := c.current.Commits()
+
+	for _, msg := range commits.Values() {
```
trianglesphere, line 340: Maybe worth doing a binary search / parallelizing this.

trianglesphere:
Oh, random nit: could the PR title be updated before merging.

consensus/istanbul/backend/backend.go
```diff
@@ -462,7 +462,8 @@ func (sb *Backend) GetValidators(blockNumber *big.Int, headerHash common.Hash) [
 	return validatorSet.List()
 }
 
-// Commit implements istanbul.Backend.Commit
+// Commit implements istanbul.Backend.Commit, unless this is committing the
+// last block of an epoch the aggregatedEpochValidatorSetSeal will be empty.
```
nategraf, line 466: I found this a bit hard top parse because I read it as a single sentence the first few times.
```suggestion
// Commit implements istanbul.Backend.Commit.
// Unless this is committing the last block of an epoch the aggregatedEpochValidatorSetSeal will be empty.
```

consensus/istanbul/core/commit.go
```diff
@@ -167,6 +169,16 @@ func (c *core) handleCommit(msg *istanbul.Message) error {
 	return c.handleCheckedCommitForCurrentSequence(msg, commit)
 }
 
+// handleCheckedCommitForPreviousSequence adds messages for the previous
+// sequence to the parent commit set, if the subject digest of msg does not
+// match that of the previous block or it was not sent by one of the previous
+// block's validators an error is returned. If this is the last block of the
+// epoch then the epoch seal will also be validated.
```
nategraf, line 176: I found this comment a bit hard to parse at first. Changing up the punctuation would help.
```suggestion
// handleCheckedCommitForPreviousSequence adds messages for the previous
// sequence to the parent commit set. 
// If the subject digest of msg does not match that of the previous block or it
// was not sent by one of the previous block's validators, an error is returned.
// If this is the last block of the epoch then the epoch seal will also be validated.
```

consensus/istanbul/core/commit.go
```diff
@@ -177,10 +189,9 @@ func (c *core) handleCheckedCommitForPreviousSequence(msg *istanbul.Message, com
 	if validator == nil {
 		return errInvalidValidatorAddress
 	}
-	if err := c.verifyCommittedSeal(commit, validator); err != nil {
-		return errInvalidCommittedSeal
-	}
```
nategraf, line 182: I didn't see anywhere in the code that a bad signature introduced the parent seal at this point would be handled. I'd be worried that this means a validator can compromise liveness by sending a bad commit for the previous sequence every round.

consensus/istanbul/core/commit.go
```diff
@@ -207,62 +218,135 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
+		return err
 	}
 
 	newValSet, err := c.backend.NextBlockValidators(c.current.Proposal())
 	if err != nil {
 		return err
 	}
 
+	// Verifies the individual seal for this commit message for the epoch
+	// block, no-op unless this is the last block of an epoch.
```
nategraf, line 232: We should optimize this to use the same aggregation techniques.

consensus/istanbul/core/commit.go
```diff
@@ -207,62 +218,135 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
+			if c.current.Commits().Size() < c.current.ValidatorSet().MinQuorumSize() {
+				// Wait for more commits
+				return nil
+			}
+			aggregatedSeal, err = c.generateAggregateCommittedSeal()
+		}
+		// If there is still an error then sit this round out, we can't continue.
+		if err != nil {
+			nextRound := new(big.Int).Add(c.current.Round(), common.Big1)
+			logger.Warn("Error on commit, waiting for desired round", "reason", "failed to aggregate commit seals", "err", err, "desired_round", nextRound)
```
nategraf, line 273: IIUC, this can only happen if `removeInvalidCommittedSeals` fails and puts us in a state where there are a quorum of commits, but they do not verify. If this is the case, we will want to log a louder error. Arguably, we might `panic`, but I'm not sure that's wise.

consensus/istanbul/core/commit.go
```diff
@@ -207,62 +218,135 @@ func (c *core) handleCheckedCommitForCurrentSequence(msg *istanbul.Message, comm
...
+func (c *core) removeInvalidCommittedSeals() {
+
+	commits := c.current.Commits()
+
+	for _, msg := range commits.Values() {
+		var commit *istanbul.CommittedSubject
+		msg.Decode(&commit)
+		err := c.verifyCommittedSeal(commit, c.current.GetValidatorByAddress(msg.Address))
+		if err != nil {
+			commits.Remove(msg.Address)
```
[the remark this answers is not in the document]
nategraf: +1 we should definitly do that as we will want to do something in response, possibly Governance slashing.

consensus/istanbul/core/commit.go
nategraf: In rank of the optimizations I'd like to see implemented, we should consider:
1. Streaming aggregation
2. Binary search for bad signatures
3. Parallization

I'd say 1 is most important to implement now, as it implies non-trivial code changes to how this feature works. 2 and 3 should be fairly easy though and are self-contained, so it would be nice to implement those as well.

nategraf, line 340: Because this method does not have any way to record the verification status of a given signature, it is possible for this to do `O(n^2)` verifications in the worst case. It is also not forward compatible with incremental verification, whereby we verify chunks of BLS signatures as they arrive.

The most obvious solution would be to add a structure to the round state to record "unverified commits" and only move signatures into the "commit" message set when they have passed a BLS signature check. Another solution would be to keep a structure in memory only that caches the verification check and use it here to skip verification for messages that are in that set.

consensus/istanbul/core/core.go
```diff
@@ -362,63 +378,19 @@ func (c *core) commit() error {
...
-	return nil
-}
-
-// GetAggregatedEpochValidatorSetSeal aggregates all the given seals for the SNARK-friendly epoch encoding
-// to a bls aggregated signature. Returns an empty signature on a non-epoch block.
-func GetAggregatedEpochValidatorSetSeal(blockNumber, epoch uint64, seals MessageSet) (types.IstanbulEpochValidatorSetSeal, error) {
-	if !istanbul.IsLastBlockOfEpoch(blockNumber, epoch) {
-		return types.IstanbulEpochValidatorSetSeal{}, nil
+	if proposal == nil {
+		return nil
```
nategraf, line 382: It seems odd that a `nil` proposal can be committed. Unless there is an explanation for why this would happen during normal operation, I'd like us to add a warning here.
```suggestion
		logger.Warn("Committed nil proposal")
		return nil
```

consensus/istanbul/core/roundstate.go
```diff
@@ -358,32 +342,19 @@ func (rs *roundStateImpl) TransitionToPrepared(quorumSize int) error {
...
-			i++
-		}
-	}
-	if i != quorumSize {
-		return errFailedCreatePreparedCertificate
 	}
+
 	rs.preparedCertificate = istanbul.PreparedCertificate{
 		Proposal:                rs.preprepare.Proposal,
 		PrepareOrCommitMessages: messages,
```
nategraf, line 356: It seems like we should probably change the field in this type from `PrepareOrCommitMessages` to `Prepares`

consensus/istanbul/core/core.go
trianglesphere: I think proposal can be nil if the node never receives the pre-prepare, though it also should get stuck on the accept pre-prepare stage

consensus/istanbul/core/core.go
piersy: Yes thanks!

consensus/istanbul/core/commit.go
piersy: Yeah probably, this was one of the last changes I made.

consensus/istanbul/core/commit.go
piersy: How about:
```
// Note that commit messages are discarded from memory but the removal is not
// persisted to disk, this should not pose a problem however because if this
// step is reached again they will again be discarded.

```

piersy:
> 0 ms back . Super nice work. I had a couple nits, but the only big question is about of optimizing the worse case (well and I guess crypto support as well for the random linear combination).
> While you tested with 12% bad validators, which I think is good amount of bad signatures, I think it only takes one bad signature to hit worse cast performance with the current linear search.

Hey @trianglesphere do you have any idea why in the bad signatures scenario the consensus time increased by 530ms but sleep time was only reduced by 38ms? It's almost 10 times less than the increase in consensus time.

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
piersy: +1

trianglesphere:
> > 0 ms back . Super nice work. I had a couple nits, but the only big question is about of optimizing the worse case (well and I guess crypto support as well for the random linear combination).
> > While you tested with 12% bad validators, which I think is good amount of bad signatures, I think it only takes one bad signature to hit worse cast performance with the current linear search.
> 
> Hey @trianglesphere do you have any idea why in the bad signatures scenario the consensus time increased by 530ms but sleep time was only reduced by 38ms? It's almost 10 times less than the increase in consensus time.

Not really sure. Off the top of my head I'd look into the sleep that we do when constructing new blocks though that only sleeps in second increments... 
https://github.com/celo-org/celo-blockchain/blob/ede6fa241231e232213739217a33311993957540/miner/worker.go#L847-L852

Second thought is that sleep + consensus > block cycle for the bad signature case. I skimmed your notebook and saw negative sleep times: I wonder if there is some wraparound data going on that results in a sleep average/median that is off.

consensus/istanbul/core/commit.go
piersy: This is handled here https://github.com/celo-org/celo-blockchain/blob/d759429e620d0098f76f35ef73648dca6c2a6df3/consensus/istanbul/backend/engine.go#L983-L1024

The process to build the parent seal is to take the previous block seal as a starting point and combine any new commit messages encountered while processing the current block, if the aggregate seal does not verify then the previous block seal is used. So if there is one bad signature in any late arriving commits no late arriving commits will be added to the parent seal.

trianglesphere: hmm, this is technically correct behavior, but it makes it possible for a single bad validator to "mark" the "late" (actually anything past quorum) validators as down for the block. There's definitely overhead to it, but I think it makes sense to do the check again on the late arriving commits.

consensus/istanbul/core/commit.go
piersy: Can do, I think It could be a significant amount of work, do we want to make it part of this PR?

consensus/istanbul/core/core.go
nategraf: IIUC it would still be a bug if that were to happen, correct?

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
trianglesphere: So we'd need to extend slashing to be able to do programatic slashing based on bad BLS signatures. I worry that adding the bitmap would then require proof of the bad signature to be tied to the block (but if you don't have the bad bitmap, you could submit the bad signature).

Just logging the address would enable governance slashing / bringing the bad behavior to light.

consensus/istanbul/core/core.go
trianglesphere: If we tried to commit a nil proposal, I think it's a bug.

consensus/istanbul/core/commit.go
piersy: So I was just maintaining the previous logic here, this is the case where there is something else wrong with the aggregation that is not an invalid signature. I would consider this a programming error and I would be inclined to panic here but I didn't want to go tweaking this unrelated logic as part of this PR.

consensus/istanbul/core/commit.go
piersy: @nategraf Just to clarify, streaming aggregation is not a performance optimisation beyond what has been done here, it would probably cause a minor reduction in performance but would let us know that all signatures are individually valid?

piersy: Good point.

consensus/istanbul/core/commit.go
nategraf: I think your changes make this more obviously a programming error if it happens, and I think we should raise this message to an error level right now.
```suggestion
			logger.Error("Error on commit, waiting for desired round", "reason", "failed to aggregate commit seals", "err", err, "desired_round", nextRound)
```

consensus/istanbul/core/commit.go
nategraf: I think so yes, because without that work we will suffer (further) issues on epoch blocks if we raise the number of validators

consensus/istanbul/core/core.go
piersy: I would be in favour of panicking here again, but I didn't want to go tweaking logic unrelated to the PR. I think I must have decided that committing a nil proposal at least warranted a warning though.

consensus/istanbul/core/roundstate.go
piersy: Yes, thanks!

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
piersy: How would logging the address enable governance slashing? How do you verify someone's log?

piersy:
> Second thought is that sleep + consensus > block cycle for the bad signature case. I skimmed your notebook and saw negative sleep times: I wonder if there is some wraparound data going on that results in a sleep average/median that is off.

Interesting, but wouldn't that show up as a significant increase in block period, the block period for the bad signature case is practically the same as the baseline (slightly shorter in fact)

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
trianglesphere: Good question. A single log on its own couldn't really be trusted. I think it could be combined with multiple logs from multiple validators, looking at missed signatures (a validator with a bad signature would not be included). It'd also let us know if we need to modify the client also log out the bad BLS signature to then programmatically slash.

I'm starting to think of logging as an incomplete solution, but without it we don't even know that we have a problem.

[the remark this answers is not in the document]
nategraf: If possible, we should log here the raw message including signature. With that we would have strong evidence of misbehavior

consensus/istanbul/core/commit.go
piersy: @nategraf I'm going to see about changing this to the streaming model.

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
nategraf: I understand your hesitation about bloating the logs, but don't think it really applies here. We should only log if there is misbehavior, which would be an extreme event which we would want to respond to and address quickly. In this investigation, the signature would be absolutely invaluable, and unless it persisted and retrievable in another way, I feel pretty strongly that we should include it here.

We would expect this this to never happen, so I am personally fine with this solution, having it logged out and taking the manual investigation from there. I don't think this represents a item of work that needs much beyond that as what we really want is a credible threat of response to ward off validators trying to mess with this part of the system. I don't think we need to expand this into a separate item of work.

consensus/istanbul/core/commit.go
nategraf: Retraction on the statement of parallization being easy. As we discussed elsewhere, we probably shouldn't open that can of worms right now.

consensus/istanbul/backend/announce.go
```diff
@@ -1305,11 +1064,7 @@ func (sb *Backend) handleEnodeCertificateMsg(_ consensus.Peer, payload []byte) e
 	}
 	logger = logger.New("msg address", msg.Address)
 
-	var enodeCertificate istanbul.EnodeCertificate
-	if err := rlp.DecodeBytes(msg.Msg, &enodeCertificate); err != nil {
-		logger.Warn("Error in decoding received Istanbul Enode Certificate message content", "err", err, "IstanbulMsg", msg.String())
-		return err
-	}
+	enodeCertificate := msg.EnodeCertificate()
```
nategraf, line 1067: We need some way to handle the case where the message is valid, but not the type we expect. (i.e. someone sends a prepare message instead of an enode certificate.) In this code as it stands, we would actually be vulnerable to an attack where the peer sends an unexpected message, and a nil pointer panic would occur on line 1070.

consensus/istanbul/backend/announce.go
```diff
@@ -1199,12 +967,12 @@ func (sb *Backend) setAndShareUpdatedAnnounceVersion(version uint) error {
 	}
 
 	// Generate and gossip a new version certificate
-	newVersionCertificate, err := sb.generateVersionCertificate(version)
+	newVersionCertificate, err := istanbul.NewVersionCertificate(version, sb.Sign)
```
nategraf, line 970: Its a bit weird to me that this `New*` function takes a signers. It is also different than other messages IIUC. Is there a reason this not done is a separate Sign function, as `istanbul.Message` implements https://github.com/celo-org/celo-blockchain/blob/948566e0d22fbb047be512d53c5e3918b0a19826/consensus/istanbul/types.go#L491?

consensus/istanbul/types.go
```diff
@@ -343,6 +425,65 @@ type Message struct {
...
+		message.Code = EnodeCertificateMsg
+		message.enodeCertificate = t
+	case []*VersionCertificate:
+		message.Code = VersionCertificatesMsg
+		message.versionCertificates = t
+	case *ValEnodesShareData:
+		message.Code = ValEnodesShareMsg
+		message.valEnodeShareData = t
+	default:
+		panic(fmt.Sprintf("attempt to construct Message unrecognised inner message type %T", t))
```
nategraf, line 479: Is there any way to make this a compile-time error rather than a panic?

consensus/istanbul/types.go
```diff
@@ -343,6 +425,65 @@ type Message struct {
...
+	}
+	bytes, err := rlp.EncodeToBytes(innerMessage)
+	if err != nil {
+		panic(fmt.Sprintf("attempt to serialise inner messgae of type %T failed", innerMessage))
```
nategraf, line 483: Are we absolutely sure this cannot occur under any normal circumstances? If there is any way for a crafted message from an attacker to trigger this code path, then this becomes a DoS attack. Looking through the `rlp` codebase, there seem to be a number of ways this can result in an error. Most of them involve unserializable types, and would be caught on the first invocation of this function in a test or testnet, but it's unclear if that is always the case, or if a crafted message could surface an error that went otherwise unnoticed.

Another option here is obviously to return the `error` and allow the caller to handle it. I can see the appeal of cleanliness from this approach, but it does make me a bit nervous. I am curious to hear more about why we should panic here.

consensus/istanbul/types.go
```diff
@@ -343,6 +425,65 @@ type Message struct {
...
+	// instance, which is set depends on the Message.Code.
+	committedSubject    *CommittedSubject
+	prePrepare          *Preprepare
+	prepare             *Subject
+	roundChange         *RoundChange
+	queryEnode          *QueryEnodeData
+	forwardMessage      *ForwardMessage
+	enodeCertificate    *EnodeCertificate
+	versionCertificates []*VersionCertificate
+	valEnodeShareData   *ValEnodesShareData
```
nategraf, line 440: This seems to be a prime use-case for interfaces. Is there a reason to prefer this approach over piecing this out into an interface and various implementations?

nategraf (commented):
I didn't have enough time today for a full review of everything that has changed, but here are some comments on what I did get to.

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
piersy: How would we actually respond to and adress this issue? If we saw this in the logs what would we do?

consensus/istanbul/core/commit.go
piersy: I've updated this and a few similar cases to use Error

consensus/istanbul/core/commit.go
piersy: This is now done

consensus/istanbul/core/commit.go
piersy: This is now done

consensus/istanbul/core/commit.go
piersy: I've updated this comment, its slightly different now though because the function has changed.

consensus/istanbul/core/commit.go
piersy: Caching of verifications is now done.

consensus/istanbul/core/roundstate.go
piersy: Done

consensus/istanbul/backend/announce.go
piersy: The reason for providing the sign function at construction time is that, we never want an unsigned message. If we didn't sign it at construction time then it opens up the possibility that we could gossip or store an unsigned message and also the possibility that we could waste resources signing a message that has already been signed.

The separate Sign function on istanbul.Message also exhibits these problems, so it would be good to make signing part of construction there as well, and from a quick inspection of the code looks like it would not be so difficult to accomplish.

consensus/istanbul/types.go
piersy: I can't see any neat way to convert this into a compile time error, can you?

consensus/istanbul/types.go
piersy: I'm rarely absolutely sure of anything, but I'd say in this case I'm pretty sure.

At a high level the intuition is that we should never be passing random types to NewMessage, since nodes only use  NewMessage when they need to convey something to the network and at that point they would call NewMessage with a specific type.

If you look at where NewMessage is being called you can see that it is always called with specific types, (i.e there is never a case where we pass a parameter of type `interface{}` to NewMessage, which means that regardless of what an attacker may be able to do, we can be sure that the set of types that NewMessage can be called with cannot be modified.

How can we be sure that those types are supported? We can just look at the code and make sure that they are supported, or we can have tests that that exercise all usages of NewMessage, this is preferable.

The normal case in  which this would be triggered is where NewMessage is called with an unsupported type, which implies that there is an error in our code (i.e there is no way to recover from this error without changing the code) in those cases I think panic is a reasonable course of action.

Code can panic, its an unavoidable part of the go language, there is no way to totally avoid them in your code. I think we should just ensure that we trust our tests to uncover any potential panics. It seems that this is the approach we have used thus far successfully, given that any pointer access, array/slice access and channel closing/sending can result in a panic. So I think we can use that same approach in this case.

consensus/istanbul/types.go
piersy: Could you give an example, I'm not seeing how to use an interface here?

trianglesphere: Maybe you could do
```
type Message interface {
       Msg()       []byte
      Address()  common.Address // The sender address
      Signature() []byte         // Signature of the Message using the private key associated with the "Address" 
}
```
And then have a struct for message that fulfills the interface.

One general comment I have by looking at this code is that while I don't know how easy it is, I'd like to be able to have have a `parseMessage` which returns something I can do a typeswitch on to handle each message (so each handler is guaranteed by the type system to have the correct type). Unfortunately go doesn't make this particularly easy.

consensus/istanbul/types.go
trianglesphere: The way to make this a compile time error would to have a `func NewTypeMessage(msg Type, sender common.Address) *Msg ` for each type. It's probably a similar amount of total code, but way more functions.

piersy: @trianglesphere Thanks, I had considered that, but I was thinking it would be quite a lot more code. Thinking about it more I'm not so sure that it would be. I could open a PR on this to see what that change looks like.

consensus/istanbul/types.go
piersy: @trianglesphere, I'm not really seeing what that interface gets you, it seems to be just repeating what is already available on an instance of instanbul.Message minus the `Code` field.
```go
type Message struct {
	Code      uint64
	Msg       []byte
	Address   common.Address // The sender address
	Signature []byte         // Signature of the Message using the private key associated with the "Address" field
}

```

consensus/istanbul/backend/announce.go
piersy: Thanks for pointing this out!

I hadn't realised that the message code is duplicated between the istanbul.Message and the p2p.Msg. In fact the codes are quite tangled some are used just at the p2p level some are used at just the istanbul level and some are used in both levels. So I think the problem here is the case when the p2p message code should correlate to the istanbul message code but it doesn't. What we can do in this case is decode the istanbul.Message and check that its code matches the p2p code, and chuck the message if it doesn't match. I think that would be a way to solve the issue you have pointed out here. Or alternatively we could just process the message based on its inner code and not worry about it matching the p2p code.

consensus/istanbul/types.go
nategraf: What find uncomfortable about this factoring is that there are a number of private fields, one for each type of message, and the reliance on mutations. It also doesn't provide a clear way to make sure a message is a particular type, which can result in bugs, like the `nil` pointer error I mentioned in another comment.

consensus/istanbul/backend/announce.go
nategraf: Checking the message code during RLP decoding makes the most sense to me. The caller should have an idea of what message they expect to get back, and it should be an error if the payload is that that type.

consensus/istanbul/backend/announce.go
nategraf: It doesn't seems so crazy to be that construction and signing are done in two steps. Transactions are a great example of why that is useful, and we have proposed more abstract methods of signing, such as utilizing a threshold scheme for each validator. So I personally favor the two step approach, but only weakly.

consensus/istanbul/types.go
nategraf: I don't fully agree with the general principle that the best answer to a code error is always to `panic`, but I agree in this case. Thanks for the explanation.

consensus/istanbul/core/commit.go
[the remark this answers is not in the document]
nategraf: If we saw these logs we would want to investigate to determine whether this could have been some kind of bug or malfunction (e.g. wrong signing key being used) or if there is evidence that this is intentional. If intentional, we would publish our findings, using the signature as a core piece of evidence, and potentially follow it up with a Governance slashing proposal.

piersy:
Closing this as the library for handling streaming bls verification https://github.com/celo-org/celo-bls-go is not ready and the crypto team has no immediate bandwidth to work on it.

PR #1448, opened 2021-03-23
mycelo: support new etherbase and tx-fee-recipient flags

oneeman:
### Description

This PR updates mycelo to use the new `--miner.validator` and `--tx-fee-recipient` flags instead of `--etherbase` if the geth version supports it.

### Other changes

1. Add a new account type for tx fee recipient (a corresponding change should be made in celo-monorepo, to allow it to be used on GCP testnets and to be consistent)
2. Add the bootnode account type, which exists in celo-monorepo but was missing in mycelo, resulting in the subsequent account type values being off by one.
3. Enable the istanbul RPC on the validators nodes (e.g. to get the lookback window)

### Tested

Confirmed to work with both master and v1.2.4

### Backwards compatibility

As some of the account type values have been shifted by one, you could get the wrong value when using `mycelo env account` to generate account information (for account types >= 3) for a mycelo environment that was generated before this PR.  However, mycelo environments are generally not long-lived.  To work around this, either use a new mycelo environment, or generate the account info yourself rather than with mycelo.

### DIFF
--- a/mycelo/cluster/cluster.go
+++ b/mycelo/cluster/cluster.go
@@ -75,13 +75,13 @@ func (cl *Cluster) ensureNodes() []*Node {
 		cl.nodes = make([]*Node, len(validators))
 		for i, validator := range validators {
 			nodeConfig := &NodeConfig{
-				GethPath:   cl.config.GethPath,
-				ExtraFlags: cl.config.ExtraFlags,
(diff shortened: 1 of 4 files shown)


trianglesphere (approved):
needs a gofmt, but otherwise looks good to me.

PR #1458, opened 2021-03-25
Mc/contracts comm refactor1

mcortesi:
### Description

This is a serious of refactors on how we handle EVM calls within the EVM.

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


mcortesi:
Replaced by a sequence of PRs:
#1460 #1461 #1462 #1463 #1465

PR #1459, opened 2021-03-25
Adds a flag to hardcode gas estimation in the loadbot

### LINKED ISSUE #1452: Add flag to stop mycelo from estimating gas
## Description

Add flag to stop mycelo from estimating gas.

## Context

It appears that the RPC handler is starting to have an effect on the validator resource usage when doing mycelo load tests. The biggest factor appears to be gas estimation.

This could also be solved with a more complex network layout that has full node(s) servicing the RPC clients.

trianglesphere:
### Description

Adds a flag to hardcode gas estimation in the loadbot.
Also disables donut for the load test environment. This is because the load bot does not yet implement replay protection.

### Other changes

* Add the `txpool` API to the list of RPC APIs mycelo validator nodes enable.

### Tested

Skips gas estimation when the flag is given. Yields a fairly large reduction in resources on the validator node.

### Related issues

- Fixes #1452
- Should be followed by https://github.com/celo-org/celo-blockchain/issues/1488


oneeman:
While there is a PR open for mycelo, I added a commit to add the the `txpool` API to the list of RPC API's mycelo nodes enable.  Will update the description to mention that as well.

cmd/mycelo/templates.go
```diff
@@ -154,6 +154,10 @@ func (e loadtestEnv) createGenesisConfig(env *env.Environment) (*genesis.Config,
 	genesisConfig.GasPriceMinimum.TargetDensity = fixed.MustNew("0.9999")
 	genesisConfig.GasPriceMinimum.AdjustmentSpeed = fixed.MustNew("0")
 
+	genesisConfig.Hardforks = genesis.HardforkConfig{
```
mcortesi, line 157: do you want this too?

mycelo/loadbot/bot.go
```diff
@@ -109,6 +110,9 @@ func runTransaction(ctx context.Context, lg *LoadGenerator, acc env.Account, non
 
 	stableTokenAddress := env.MustProxyAddressFor("StableToken")
 	transactor.FeeCurrency = &stableTokenAddress
+	if skipEstimation {
+		transactor.GasLimit = 110000 // 110lk gas for stable token transfer is pretty reasonable. It's just under 100k in practice
```
mcortesi, line 114: you can externalize this a constant

mcortesi (approved):
looks good, left some nits

cmd/mycelo/templates.go
trianglesphere: yes. we can't load test donut until we enable replay protection in the load bot (https://github.com/celo-org/celo-blockchain/issues/1488), so this lets us load test until we complete the issue

PR #1460, opened 2021-03-25
EVM: Removes Knowledge of TobinTax

mcortesi:
### Description

EVM should be unaware of tobin tax. From the EVM perspective there is
just Context.Transfer() function.

This commit, moves evm.TobinTransfer to replace context.Transfer
function. In so, it removes unnecessary input and output parameters.

### Tested

CI / unit test

### DIFF
--- a/core/vm/context.go
+++ b/core/vm/context.go
@@ -17,6 +17,8 @@
 package vm
 
 import (
+	"encoding/binary"
+	goerrors "errors"
 	"math/big"
 
 	"github.com/celo-org/celo-blockchain/common"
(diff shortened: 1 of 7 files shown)


core/vm/context.go
```diff
@@ -185,3 +188,57 @@ func VerifySealFn(ref *types.Header, chain ChainContext) func(*types.Header) boo
...
+	if numerator.Cmp(denominator) == 1 {
+		return nil, nil, nil, goerrors.New("Tobin tax numerator greater than denominator")
+	}
+	return numerator, denominator, reserveAddress, nil
+}
+
+// TobinTransfer performs a transfer that may take a tax from the sent amount and give it to the reserve.
+// If the calculation or transfer of the tax amount fails for any reason, the regular transfer goes ahead.
+// NB: Gas is not charged or accounted for this calculation.
+func TobinTransfer(evm *EVM, sender, recipient common.Address, amount *big.Int) {
```
piersy, line 222: Hey @mcortesi, this PR doesn't seem to be actually removing knowledge of tobin tax at the package level, I think it would be nice if we could really decouple the vm from this special logic.

I created a PR showing how this could work here - #1463

mcortesi:
Replaced by #1465

PR #1461, opened 2021-03-25
EVM: Removes static call to GetRegisteredAddress

mcortesi:
### Description

This commit removes the use of a static call to obtain a registry
address. This was used on the Transfer precompile to check that the only
caller to the contract is the GoldToken contract.

With this commit, there's a GetRegisteredAddress function on the
evm.Context that is injected at the moment of creating.

For now, this function is hardcoded, but later it will be received as a
parameter to the vm.NewContext()

### Tested

ci & unit tests

### DIFF
--- a/core/vm/dynamic_params.go
+++ b/core/vm/dynamic_params.go
@@ -33,6 +33,14 @@ const (
 
 var getAddressForFuncABI, _ = abi.JSON(strings.NewReader(getAddressForABI))
 
+func GetRegisteredAddress(evm *EVM, registryId common.Hash) (common.Address, error) {
+	addr, err := GetRegisteredAddressWithEvm(registryId, evm)
+	if err != nil {
+		return common.ZeroAddress, nil
+	}
+	return *addr, nil
+}
+
 // TODO(kevjue) - Re-Enable caching of the retrieved registered address
 // See this commit for the removed code for caching:  https://github.com/celo-org/geth/commit/43a275273c480d307a3d2b3c55ca3b3ee31ec7dd.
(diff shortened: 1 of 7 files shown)


core/vm/dynamic_params.go
```diff
@@ -33,6 +33,14 @@ const (
 
 var getAddressForFuncABI, _ = abi.JSON(strings.NewReader(getAddressForABI))
 
+func GetRegisteredAddress(evm *EVM, registryId common.Hash) (common.Address, error) {
+	addr, err := GetRegisteredAddressWithEvm(registryId, evm)
+	if err != nil {
+		return common.ZeroAddress, nil
+	}
+	return *addr, nil
```
piersy, line 41: Why the switch from pointer to non pointer return value here? I think the pointer approach is actually superior to the non pointer approach because a function that tries to use a nil pointer without checking it will panic, whereas a function that tries use a zero address without checking it, may succeed (E.G a send to a zero address will have no problem, the user might take issue with it though).


Also looking at the next [PR](#1462) in this chain of PRs I can see that the change caused significant code churn so that is another reason against making the change.

That is also a reason I think that these PRs would be better presented as a single PR split into nicely formed commits, since having them separated like this makes it difficult to assess the true impact of a change, and subsequently making a change here requires it to be propagated to the upstream PRs, which just adds a lot of overhead.

trianglesphere: +1 to making this one PR. The commit are already nicely formatted, so I've mainly been looking at the last PR and then looking at all changes from master on it.

mcortesi:
Replaced by #1465

PR #1462, opened 2021-03-25
EVM module: Removes GetRegistedAddressWithEVM

mcortesi:
### Description

Simple refactor, replaces calls to `GetRegistedAddressWithEVM` to the new
version `GetRegistedAddress`.

As a side effect, return type was change from `*common.Address` to
`common.Address` which might be solving a bug, since we found that we are
comparing two pointers when we should be comparing values. (on
`currency.Convert()`)


mcortesi:
Replaced by #1465

PR #1463, opened 2021-03-26
Move knowledge of tobin out of vm package

piersy:
### Description

This shows an approach to remove all knowledge of tobin tax from the vm package.

It's not finished, we would need to updated all current instances of runtime.Config to contain the tobin transfer function, but this should be a mechanical change.


mcortesi:
Replaced by #1465

PR #1464, opened 2021-03-26
Removes notion of system call & abi on the evm

mcortesi:
### Description

celo-blockchain added `StaticCallFromSystem` and `CallFromSystem` which
also implied knowledge of the abi module and of the system caller
(address zero)

Both concepts don't belong in the EVM and are better placed on a higher
layer that "uses" the EVM.

This commit extracts those methods, into the `contracts` module.
Decouples the evm from those concept and leaves it closer to upstream's
EVM.

Additionally, `GetRegisteredAddress` is moved to the contracts module.
Decoupling the evm module from the concept of "Registry" and of a
"Registry Contract".

This commit, introduces a new parameter to the vm.NewContext():
ContractsContext which is an interface that provides a
`GetRegisteredAddress` method for use within the EVM. It is a interface
instead of a single function since `getTobinTax` will also be included
there in a follow up PR.

### Tested

CI & unit tests


mcortesi:
Replaced by #1465

PR #1465, opened 2021-03-26
Decouples EVM from the contract layer

mcortesi:
### Description

**Note for reviewers**: It's easier to review commit after commit, to check the validity of the transformations.

This PR decouples the EVM from contracts.

1) `TobinTransfer`: before it was part of the EVM implementation, now it's `context.Transfer`. Making the EVM unaware of the tobin tax

2) `vm.GetRegisteredAddressWithEVM` was moved to `contracts` module. The vm module should not be aware of contracts. It now returns `common.address` instead of `*common.address`. This also helps fixes a bogus address comparison on `currency.Convert`

3) transfer precompile now relies ton `context.GetRegisteredAddress` instead of `vm.GetRegisteredAddress`. Decoupling it from contracts.

4) Moved `evm.StaticCallFromSystem` and `evm.CallFromSystem` to the `contracts` package. Removing the notion of `systemCaller`, `abi` packages and contracts from the EVM.

5) Move `vm.NewContext` to `vm/context.New`. Rationale behind this, is that on creation we introduce a dependency with contracts (tobinTax, registry); and at the same time the `vm` package should be unaware of such details. By moving the creation of `vm.Context` that separation is more clear. This follows the recommendation from @piersy in https://github.com/celo-org/celo-blockchain/pull/1463

6) Extract `ComputeTobinTax` to the `contracts` package

### Tested

* ci
* unit tests

### Related issues

- Part of EVM refactoring and cleanup

### Backwards compatibility

Yes


contracts/tobin_tax.go
```diff
@@ -0,0 +1,53 @@
+package contracts
```
piersy, line 1: It would be good to have a package doc for this package. At the moment it looks like a bit of a mishmash and its not really clear to me what is the functionality it encapsulates.

Looking at the public api from the godoc it looks like splitting it into 2 packages could be a good idea, a tobin package and a different package, I'm not really sure that contracts makes sense as a package for these system call functions, but then again without package doc its hard to understand the purpose.
```
func CallFromSystem(evm *vm.EVM, contractAddress common.Address, abi abipkg.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int) (uint64, error)
func ComputeTobinTax(evm *vm.EVM, sender common.Address, transferAmount *big.Int) (tax *big.Int, taxRecipient common.Address, err error)
func GetRegisteredAddress(evm *vm.EVM, registryId common.Hash) (common.Address, error)
func StaticCallFromSystem(evm *vm.EVM, contractAddress common.Address, abi abipkg.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64) (uint64, error)
func TobinTax(evm *vm.EVM, sender common.Address) (numerator *big.Int, denominator *big.Int, reserveAddress common.Address, err error)
```

les/api_backend.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/celo-org/celo-blockchain/core/state"
 	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/core/vm"
+	vmcontext "github.com/celo-org/celo-blockchain/core/vm/context"
```
piersy, line 33: I would go for just calling this package vmcontext, duplicating package names can be confusing especially when those package names are from the standard library since clashes become even more likely.

core/vm/runtime/env.go
```diff
@@ -17,14 +17,16 @@
 package runtime
 
 import (
+	"github.com/celo-org/celo-blockchain/contracts"
 	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/core/vm/context"
```
piersy, line 22: The `vm` package is still coupled to the contract layer, this is importing the context package which is importing the contracts package.

To break this coupling I would move the `Context` object back into the `vm` package because that is where it lives in geth so moving it out of the `vm` package is a departure from geth.

contract_comm/evm.go
```diff
@@ -25,8 +25,10 @@ import (
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/common/hexutil"
 	"github.com/celo-org/celo-blockchain/contract_comm/errors"
+	"github.com/celo-org/celo-blockchain/contracts"
 	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/core/vm/context"
```
trianglesphere, line 31: nit: I think importing this as `vmcontext` would help make it clear what it is (`context := context.New(...)` feels ambiguous)

contracts/tobin_tax.go
trianglesphere: I think it can be split into 1) system calls / EVM communication
```
CallFromSystem(....)
StaticCallFromSystem(...)
GetRegisteredAddress(...) // this is a more involved static call to system
```
and 2) Tobin Tax
```
ComputeTobinTax(...) // don't need TobinTax to be public
```

core/state_prefetcher.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/celo-org/celo-blockchain/core/state"
 	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/core/vm/context"
```
trianglesphere, line 27: Another place where `vmcontext` would be nice

core/vm/context/context.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/celo-org/celo-blockchain/common"
+	"github.com/celo-org/celo-blockchain/consensus/istanbul"
+	"github.com/celo-org/celo-blockchain/contracts"
+	"github.com/celo-org/celo-blockchain/core/types"
+	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/log"
+)
+
+// New creates a new context for use in the EVM.
+func New(msg vm.Message, header *types.Header, chain vm.ChainContext, txFeeRecipient *common.Address) vm.Context {
```
trianglesphere, line 15: I like that you removed `TobinTransfer` and the evm callers from `evm.go`, but I think moving all of the context will make upstream harder to merge.

piersy: We could re-instate inside evm.go the NewEVMContext from upstream and then additionally have this version here which we will use. That would help push the vm package closer to geth

contracts/utils.go
```diff
@@ -0,0 +1,88 @@
+package contracts
```
trianglesphere, line 1: These calls feel more like `contract_comm` or `contracts/comm` or even `vm/abicaller` as a package. These are not associated with any particular contract and are used by every contract call.

les/api_backend.go
mcortesi: Yeah, had the same feeling. I always feel quite constrained by golang package names, since only the last part of the path matters...  and so you end up doing `vm/vmcontext` which duplicates the name, but makes it clear...

i'll do that change. @trianglesphere this should also solve your import comments

contracts/tobin_tax.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/params"
+)
+
+var (
+	ErrTobinTaxZeroDenominator  = errors.New("Tobin tax denominator equal to zero")
+	ErrTobinTaxInvalidNumerator = errors.New("Tobin tax numerator greater than denominator")
+)
+
+func TobinTax(evm *vm.EVM, sender common.Address) (numerator *big.Int, denominator *big.Int, reserveAddress common.Address, err error) {
```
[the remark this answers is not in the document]
mcortesi: i though it was nice to be able to ask for the current tax value (thinking it as a contract interface).

core/vm/context/context.go
mcortesi: I though on this. To me It's a balance. By removing almost all our changes to `evm.go` i made much closer to upstream and simpler to merge. Now our changes are centralized in the `context.go`. I'd argue that's better since you isolated your changed to a single place. And it's not a very large file, so i don't see it very challenging.

I also though of leaving the old NewEVMContext and creating a new on top, but i don't like that, because the old version will just create an invalid version of a vm.Context and i'd like the construct to validate the context is actually correct. I might be encouraging human error if you use the old version by mistake. So i prefer to remove that version alltogheter.

wdyt?

contracts/tobin_tax.go
mcortesi: I agree that tobin tax belongs to its own module, and imo same applies to registry access (but there's a circular dependency there)

tobin tax code also uses a hardcoded way of doing a contract call; that would be nice to modify

I didn't do those changes to avoid making the PR bigger. But i can at least move tobin tax for now

contracts/utils.go
mcortesi: i like the idea of `vm/abicaller` or `vm/something` idea. There's already a `vm/runtime` that provides a simplified interface to the evm (we use it on mycelo).

But i'd like to first get https://github.com/celo-org/celo-blockchain/pull/1468 merged if that's possible.

contracts/tobin_tax.go
trianglesphere: system call may not be the best name, but but one definition is a call not associated with an externally owned account (i.e. systemcaller address is `0x0`). The static calls are read only calls to some notion of the current state. The `CallFromSystem` modifies the provided state as part of the consensus protocol (like paying out validator rewards or updating the gas price minimum)

contracts/tobin_tax.go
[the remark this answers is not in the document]
trianglesphere: It's an interesting question, but especially in the old version was a little surprising to see it next to the system callers when the question wasn't be asked. It's fine to leave as, but I lean to the side of make it public if/when we start asking that question.

core/state_transition.go
```diff
@@ -452,26 +452,26 @@ func (st *StateTransition) distributeTxFees() error {
 		gatewayFeeRecipient = &common.ZeroAddress
 	}
 
-	governanceAddress, err := vm.GetRegisteredAddressWithEvm(params.GovernanceRegistryId, st.evm)
+	governanceAddress, err := vm.GetRegisteredAddress(st.evm, params.GovernanceRegistryId)
 	if err != nil && err != commerrs.ErrSmartContractNotDeployed && err != commerrs.ErrRegistryContractNotDeployed {
```
gastonponti, line 457: nit
```
if err != nil {
   if err != commerrs.ErrSmartContractNotDeployed && err != commerrs.ErrRegistryContractNotDeployed {
      return err
   else {
      log.Trace("Cannot credit gas fee to community fund: refunding fee to sender", "error", err, "fee", baseTxFee)
      governanceAddress = common.ZeroAddress
      refund.Add(refund, baseTxFee)
      baseTxFee = new(big.Int)
   }
}
```

core/vm/context/context.go
trianglesphere: I agree that I don't like having two `NewEVMContext` functions. Even if there are conflicts in merging upstream in the `evm.go:NewEVMContext`, it still feels easier than manually porting over the changes to the new file.

core/vm/contracts.go
```diff
@@ -634,7 +634,7 @@ func (c *transfer) Run(input []byte, caller common.Address, evm *EVM, gas uint64
 			return nil, gas, ErrInsufficientBalance
 		}
 
-		gas, err = evm.TobinTransfer(evm.StateDB, from, to, gas, value)
+		evm.Context.Transfer(evm, from, to, value)
```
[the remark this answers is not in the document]
gastonponti: mmm this comment was pointing to why we are not returning an error in the transfer function any more, but got lost in the review by commits

core/vm/context/context.go
piersy: Agreed

core/state_transition.go
piersy: nitnit ;)
```go
if err != nil {
   if err != commerrs.ErrSmartContractNotDeployed && err != commerrs.ErrRegistryContractNotDeployed {
      return err
   }
   log.Trace("Cannot credit gas fee to community fund: refunding fee to sender", "error", err, "fee", baseTxFee)
   governanceAddress = common.ZeroAddress
   refund.Add(refund, baseTxFee)
   baseTxFee = new(big.Int)
}
```

core/vm/contracts.go
```diff
@@ -599,7 +599,7 @@ func (c *transfer) Run(input []byte, caller common.Address, evm *EVM, gas uint64
 	if err != nil {
 		return nil, gas, err
 	}
-	celoGoldAddress, err := GetRegisteredAddressWithEvm(params.GoldTokenRegistryId, evm)
+	celoGoldAddress, err := evm.Context.GetRegisteredAddress(evm, params.GoldTokenRegistryId)
```
piersy, line 602: I think it would be an improvement to move transfer, and all of our extra precompiled contracts out of the `vm` package.

core/vm/contracts.go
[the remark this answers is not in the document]
mcortesi: no change intended. The TobinTransfer impl alwars returned with `err == nil` and with `gas == gas input parameter` so i removed the parameter and the output parameters

core/vm/contracts.go
mcortesi: yeah, ideally i would want to minimize the changes to vm module and do everything outside of it. But i didn't want to get too greedy on my first refactor of it....

An interesting exercise is to do a diff between 1.9.13 (our current upstream version) and our code... I'm pretty sure there are things that we added in the past and we actually don't use

core/state_transition.go
mcortesi: done

core/vm/context/context.go
mcortesi: bad news....:P `vm/context.go` doesn't really exists on upstream :P 

It seems we've already moved it from `core/evm.go` to `core/vm/context.go`! With that in mind, i think i'll just move all the remainig code within vm/context.go to the context pkg

piersy: I think it makes sense to move the ChainContext and Message to this package because this is where they are used and they are not used at all in the vm package

core/vm/opcodes.go
```diff
@@ -385,7 +385,7 @@ var opCodeToString = map[OpCode]string{
 func (op OpCode) String() string {
 	str := opCodeToString[op]
 	if len(str) == 0 {
-		return fmt.Sprintf("opcode 0x%x not defined", int(op))
+		return fmt.Sprintf("Missing opcode 0x%x", int(op))
```
trianglesphere, line 388: This is causing a test to break

contracts/reserve/tobin_tax.go
```diff
@@ -0,0 +1,61 @@
...
+var (
+	ErrTobinTaxZeroDenominator  = errors.New("tobin tax denominator equal to zero")
+	ErrTobinTaxInvalidNumerator = errors.New("tobin tax numerator greater than denominator")
+)
+
+type Ratio struct {
+	Numerator, Denominator *big.Int
+}
+
+func (r *Ratio) Apply(value *big.Int) *big.Int {
```
trianglesphere, line 23: Is this used anywhere?

trianglesphere: Wait, see it used below, but this feels like a weird thing to export.

core/vm/dynamic_params.go
```diff
@@ -33,6 +33,14 @@ const (
 
 var getAddressForFuncABI, _ = abi.JSON(strings.NewReader(getAddressForABI))
 
+func GetRegisteredAddress(evm *EVM, registryId common.Hash) (common.Address, error) {
+	addr, err := GetRegisteredAddressWithEvm(registryId, evm)
+	if err != nil {
+		return common.ZeroAddress, nil
```
oneeman: Is this `nil` intentional?  The current behavior (from `GetRegisteredAddressWithEvm`) is to return the error (which is then checked and returned if non-nil in `(c *transfer) Run()`.

oneeman: Looks like this is modified in a later commit, so nevermind.

core/state_transition.go
```diff
@@ -452,26 +452,26 @@ func (st *StateTransition) distributeTxFees() error {
...
 		return err
 	} else if err != nil {
 		log.Trace("Cannot credit gas fee to community fund: refunding fee to sender", "error", err, "fee", baseTxFee)
-		governanceAddress = &common.ZeroAddress
+		governanceAddress = common.ZeroAddress
```
oneeman, line 493: Is this line redundant now? If is it implicit that `GetRegisteredAddress` will always return `ZeroAddress` if it returns an error?

contract_comm/currency/currency.go
```diff
@@ -118,7 +118,7 @@ func ConvertToGold(val *big.Int, currencyFrom *common.Address) (*big.Int, error)
 		return val, err
 	}
 
-	if currencyFrom == celoGoldAddress {
+	if *currencyFrom == celoGoldAddress {
```
oneeman, line 121: This definitely looks it was a bug.  But I don't think we have any code path that would hit that bug on the actual networks, because the places we call this function are only called with whitelisted fee currencies, and the Celo token contract isn't in the whitelist.  If it were added to the whitelist, we could hit this bug and the result would be that an error would be returned (due to divide by zero) and the transaction would not be added to the pool (but it would work if the transaction instead use feeCurrency: nil)

contract_comm/evm.go
```diff
@@ -57,12 +58,12 @@ func MakeStaticCallWithAddress(scAddress common.Address, abi abi.ABI, funcName s
 	return makeCallFromSystem(scAddress, abi, funcName, args, returnObj, gas, nil, header, state, true)
 }
 
-func GetRegisteredAddress(registryId [32]byte, header *types.Header, state vm.StateDB) (*common.Address, error) {
+func GetRegisteredAddress(registryId [32]byte, header *types.Header, state vm.StateDB) (common.Address, error) {
```
piersy, line 61: Why the switch from pointer to non pointer return value here? I think the pointer approach is actually superior to the non pointer approach because a function that tries to use a nil pointer without checking it will panic, whereas a function that tries use a zero address without checking it, may succeed (E.G a send to a zero address will have no problem, the user might take issue with it though).

core/vm/contracts.go
piersy: Cool, added an issue here #1475

contracts/reserve/tobin_tax.go
mcortesi: why weird to export?

TobinTax returns a Ratio, so if TobinTax is public, so is Ratio... and so is `Apply`. idea here to avoid other need to do the math themselves.

core/state_transition.go
mcortesi: my rule of thumb is that if `GetRegisteredAddress` returns an error, them make no assumptions on the other value.

Here it seems "easy" to assume it's zeroValue... but somewhere over the `GetRegisteredAddres` we pass and address variable pointer to an internal funtion that unmarshall a result there.. .what if unmarshalling ends unexpectectly mid process..

contract_comm/evm.go
mcortesi: I disagree... If you're a caller and you disregard the `err` return value.. then good luck! You're already violating the contract. Since the contract of any golang function clearly states that if `err!=nil` disregard the value.

Also, if you see @oneeman comment https://github.com/celo-org/celo-blockchain/pull/1465#discussion_r605266455. We're actually making the mistake of using pointer values for comparisons, thus only comparing if they reference to the same variable, not if they are the same value. I worry much more about that kind of error, that of someone disregarding the err return value

oneeman: > Since the contract of any golang function clearly states that if err!=nil disregard the value.

Is this explicitly stated somewhere? `io.Reader` (https://golang.org/pkg/io/#Reader) violates this, and so I'm wondering how established this convention is.

I ask out of curiosity, not suggesting we switch back to using a pointer.

core/vm/context/context.go
```diff
@@ -156,11 +126,10 @@ func TobinTransfer(evm *vm.EVM, sender, recipient common.Address, amount *big.In
 	}
 
 	if amount.Cmp(big.NewInt(0)) != 0 {
-		numerator, denominator, reserveAddress, err := getTobinTax(evm, sender)
+		tax, taxCollector, err := contracts.ComputeTobinTax(evm, sender, amount)
```
oneeman: Maybe `taxRecipient` instead of `taxCollector`?

contract_comm/evm.go
mcortesi: I've checked on https://golang.org/doc/effective_go#functions but didn't find anything. 

I'd argue is only natural to assume that at least stated otherwise. If a function returns an error, it can happen at several moment on the code flow, does unless the function returns a default value or a "processed up to this point" value, then it seems hard to make any assumption on the value when `err != nil`

contract_comm/currency/currency.go
mcortesi: totally... it was a bug ;)

core/vm/context/context.go
mcortesi: i moved them to the actual place they were on upstream

trianglesphere (approved):
Looks good to go to me. I've got a couple last nits, but I think it's generally fine as is.

I would have liked some of the smaller function signature changes like the change from `*common.Address` -> `common.Address` to be split out. (Ditto for `GetHeader()` -> `Header`). I like these changes, but it makes it harder to review the rest of the changes; however at this point it doesn't seem worth pulling those changes apart.

On a naming / code organization note: I lean away from putting our interface to the core contracts in the `contracts` package. I think `core_contracts` or `contract_comm` would be more descriptive. `contracts/utils.go` is also not a particularly descriptive name for what it contains.`

mcortesi:
Thanks @trianglesphere i'll make those changes on a subsequent PR, which i'm already working on for contracts interface refactor

PR #1466, opened 2021-03-26
Upstream merge up to 1.9.14

hbandura:
Some notes from @oneeman about the flag changes:

- `c989bca`:
    - It renames many flags (keeping the old versions as deprecated).  Besides that, it moves the deprecated flags to a separate file (`flags_legacy.go`) and groups them by when they were deprecated, to ease removal later on.  
    - It also improves the UX around deprecated flags by (a) issuing warnings if deprecated flags are used, (b) not showing deprecated flags in the help message unless you use `--show-deprecated-flags` (new), (c) renaming (in code) the deprecated flag constants to start with `Legacy`.  
   - There were considerable conflicts, but straightforward.  Some notable points:
        - We had already removed `LightLegacyServFlag` and `LightLegacyPeersFlag`.  So the parts about `LegacyLightServFlag` and `LegacyLightPeersFlag` (the new names) had to be removed.
    - The following flags were renamed:
        - `--rpc` becomes `--http`
        - `--rpc*` becomes `--http.*` for [addr, port, corsdomain, vhosts, api]
        - `--ws*` becomes `--ws.*` for [addr, port, api, origins]
        - `--pprof*` becomes `--pprof.*` for [addr, port, memprofilerate, blockprofilerate, cpuprofile]
        - `--gpo*` becomes `--gpo.*` (gas price oracle, deleted by us)
    - In a separate commit, I made corresponding changes to Celo-only flags:
        - `--console*` to `--console.*` for [format, output]
        - deprecated by us (moved to legacy flag file, and warnings added if they're used and there was no warning):
            - `--miner.etherbase`
            - `--ethstats`
            - `--istanbul.*` for [requesttimeout, blockperiod, proposerpolicy, lookbackwindow]
            - `proxy.proxyenodeurlpair`


trianglesphere:
Were there any upstream commits that had significant conflicts or should be looked at more closely?
cc @oneeman

core/rlp_test.go
```diff
@@ -1,201 +0,0 @@
-// Copyright 2019 The go-ethereum Authors
```
trianglesphere: Did this file get deleted in upstream?

hbandura: If I'm not mistaken it was deleted by us previously, not sure why it's showing up here. I may have accidentally added it and re-deleted it.

hbandura:
> One CI note: it looks like `TestShortRepair - core` is failing. I don't believe this to be a known flakey test.
I'm currently working on this one. It seems a goroutine is blocked trying to write to a channel that noone is reading from (the freezer quit channel).

core/rlp_test.go
trianglesphere: This file was added in upstream during this commit, and is still in upstream, it should be re-added unless there's a specific reason why the test is no longer required.

trianglesphere (changes requested):
The deletion of `core/rlp_test.go` needs to be reverted or be explained why it should be deleted.

On the topic of tracking upstream changes, consensus has a quite nice template that they use (see here: https://github.com/ConsenSys/quorum/pull/1142). I'm not sure if it's manual or not, but having visibility into potential conflicts is nice.

signer/core/signed_data.go
```diff
@@ -216,57 +211,6 @@ func (api *SignerAPI) SignTypedData(ctx context.Context, addr common.MixedcaseAd
 	return signature, nil
 }
 
-func parseInteger(encType string, encValue interface{}) (*big.Int, error) {
```
trianglesphere, line 219: Why was this removed? It's still in upstream master

hbandura: because in our codebase this was moved to `shared/signer/signed_data.go`. During the merge it was accidentally duplicated

core/rlp_test.go
hbandura: Yeap, I readded it and modified it a bit to work

hbandura:
> Were there any upstream commits that had significant conflicts or should be looked at more closely?
> cc @oneeman

Not that I can recall in the ones I did myself, I think `ethstats/ethstats.go` is the only one. Overall it was just a matter of getting things to compile

cmd/geth/main.go
```diff
@@ -137,16 +136,16 @@ var (
...
 		utils.NoCompactionFlag,
 		utils.EWASMInterpreterFlag,
 		utils.EVMInterpreterFlag,
 		utils.CheckStaticCallCacheFlag,
 		configFileFlag,
-		utils.IstanbulRequestTimeoutFlag,
-		utils.IstanbulBlockPeriodFlag,
-		utils.IstanbulProposerPolicyFlag,
-		utils.IstanbulLookbackWindowFlag,
+		utils.LegacyIstanbulRequestTimeoutFlag,
```
trianglesphere, line 145: These also need to be flagged.

internal/debug/flags.go
```diff
@@ -60,26 +60,26 @@ var (
...
 	}
 	pprofPortFlag = cli.IntFlag{
-		Name:  "pprofport",
+		Name:  "pprof.port",
 		Usage: "pprof HTTP server listening port",
 		Value: 6060,
 	}
 	pprofAddrFlag = cli.StringFlag{
-		Name:  "pprofaddr",
+		Name:  "pprof.addr",
```
trianglesphere, line 68: The flag changes need to be recorded for when we make the next release so we can warn operators that we will be following upstream in deprecating certain flags.

trianglesphere (commented):
Test fixes look good to me.
This needs a PR description + flagging that we've changing some of the flags (it's following upstream, but it is important for node operators to know).
Once that's done, I think this is ready to merge.

trianglesphere (approved):

PR #1467, opened 2021-03-26
Cache exchange rates when creating a block 

mcortesi:
### Description

Adds a CurrencyComparator that caches exchange rates.


oneeman (approved):
Great!

PR #1468, opened 2021-03-28
Refactors contract call logic (prep for caching)

mcortesi:
**Note for the reviewer**. This PR is better reviewed by commit. Each commit takes a single step and should be easy enough to follow

### Description

This PR is preparation work for adding a caching layer to contract calls.

It modifies `contracts` and `contract_comm` packages to:

 * Add the concept of `contract.Backend` interface which the EVM already implements, but can be implemented by others (cache for example)
 * Adds variants to contract_comm utility functions that expect a `Backend` instead of `header & state`. It maintains backwards compatibility.
 * Simplifies logic, by removing generic function and adding a little duplication. In this PR's author mind, the code now it's simpler to read than before since code clearly shows the sequence of steps required to do a call (create backend, obtain contract address, pack arguments, make call, unpack result)
 * Variant for `MakeStaticCall` is named `Query` to clearly state this is a read only call
 * Variant for `MakeCall` is named `Execute` to imply it's call with write permissions.

### Other changes

 * As part of the refactor, execution time metering is moved to `contracts` package from `contract_comm` package. This implies that we will be metering all calls, and not just the ones done through `contract_comm`

### Tested

 * unit tests
 * ci

### Backwards compatibility

Changes should be backward compatible. Several considerations have been made to keep compatibility, even when that implied leaving a potentially problematic behaviour.

 * Each call uses it's own EVM when used from contract_comm, and a shared one elsewhere

### DIFF
--- a/contract_comm/evm.go
+++ b/contract_comm/evm.go
@@ -19,7 +19,6 @@ package contract_comm
 import (
 	"math/big"
 	"reflect"
-	"time"
 
 	"github.com/celo-org/celo-blockchain/accounts/abi"
 	"github.com/celo-org/celo-blockchain/common"
@@ -30,7 +29,6 @@ import (
 	"github.com/celo-org/celo-blockchain/core/vm"
 	"github.com/celo-org/celo-blockchain/core/vm/vmcontext"
 	"github.com/celo-org/celo-blockchain/log"
-	"github.com/celo-org/celo-blockchain/metrics"
 )
 
 var (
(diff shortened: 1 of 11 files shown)


contract_comm/evm.go
```diff
@@ -133,25 +169,18 @@ func SetInternalEVMHandler(chain vm.ChainContext) {
 	}
 }
 
-func makeCallWithContractId(registryId [32]byte, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int, header *types.Header, state vm.StateDB, static bool) (uint64, error) {
-	scAddress, err := GetRegisteredAddress(registryId, header, state)
+func resolveAddressForCall(backend contracts.Backend, registryId common.Hash, method string) (common.Address, error) {
```
trianglesphere, line 172: I really like this cleanup

contracts/backend.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/celo-org/celo-blockchain/params"
+)
+
+type evmBackend struct {
+	vmConfig    vm.Config
+	vmContext   vm.Context
+	chainConfig *params.ChainConfig
+	state       vm.StateDB
+
+	dontMeterGas bool
```
trianglesphere, line 17: Having played with this before, I think that an `evm builder` is the right level of abstraction for making these calls. I don't have a solution for this, but I find the `dontMeterGas` that's only used for `GetRegisteredAddress` that also calls functions of the same name (except on a `evm` instead of `ev`) a little confusing. It's also dragging down a boolean flag through several layers of code.

contracts/utils.go
```diff
@@ -4,85 +4,150 @@ import (
...
 	}
-	return handleABICall(evm, abi, funcName, args, returnObj, call)
 }
 
-var (
-	errorSig     = []byte{0x08, 0xc3, 0x79, 0xa0} // Keccak256("Error(string)")[:4]
-	abiString, _ = abipkg.NewType("string", "", nil)
-)
+type ExecOpts struct {
+	Backend Backend
```
trianglesphere, line 43: This is probably more my mental model, but I think of passing a contract/contract info to execute on a backend rather than binding a contract to a backend or binding the contract and passing in the backend.

contract_comm/evm.go
```diff
@@ -43,31 +41,95 @@ type InternalEVMHandler struct {
 	chain vm.ChainContext
 }
 
-func MakeStaticCall(registryId [32]byte, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, header *types.Header, state vm.StateDB) (uint64, error) {
-	return makeCallWithContractId(registryId, abi, funcName, args, returnObj, gas, nil, header, state, true)
+func MakeStaticCall(registryId common.Hash, abi abi.ABI, method string, args []interface{}, returnObj interface{}, gas uint64, header *types.Header, state vm.StateDB) (uint64, error) {
```
trianglesphere, line 44: It'd be nice to go through and actually add a function doc string to these.

contract_comm/evm.go
```diff
@@ -43,31 +41,95 @@ type InternalEVMHandler struct {
...
+	backend, err := createBackend(header, state)
+	if err != nil {
+		return 0, err
+	}
+
+	return Query(backend, registryId, &abi, method, args, returnObj, gas)
+}
+
+func Query(backend contracts.Backend, registryId common.Hash, abi *abi.ABI, method string, args []interface{}, returnObj interface{}, gas uint64) (uint64, error) {
```
trianglesphere, line 53: New functions definitely should be documented.

contract_comm/evm.go
```diff
@@ -43,31 +41,95 @@ type InternalEVMHandler struct {
...
+
+	contractAddress, err := resolveAddressForCall(backend, registryId, method)
+	if err != nil {
+		return 0, err
+	}
+
+	contract := contracts.NewContract(abi, contractAddress, contracts.SystemCaller)
+	gasLeft, err := contract.Query(contracts.QueryOpts{MaxGas: gas, Backend: backend}, returnObj, method, args...)
```
trianglesphere, line 61: I find the repeated use of the same function name (especially inside itself) a little confusing. It's a common pattern in this PR, but I don't think it gets us much in terms of code re-use (the actual functions) or clarity (which name am I looking at).

contracts/backend.go
```diff
@@ -0,0 +1,60 @@
...
+func NewEVMBackend(vmContext vm.Context, state vm.StateDB, chainConfig *params.ChainConfig, vmConfig vm.Config) Backend {
+	return &evmBackend{
+		vmConfig:    vmConfig,
+		vmContext:   vmContext,
+		chainConfig: chainConfig,
+		state:       state,
+	}
+}
+
+// Call implements Backend.Call
```
trianglesphere, line 29: Saying `evmBackend` implements `contracts/utils.go:Backend` and then documenting these functions would be more clear.

trianglesphere (commented):
There's some much needed `contract_comm/evm.go` refactors which are great.
At a lower level as we clean this up we should document the existing functions and document our new functions as well.

At a higher level I think I see two things
1. This refactor brings seems to bring in the idea of binding contracts (and then only uses that for the registry instead rebinding at each call)
2. There's a couple places where we have boolean flags dragged across several layers of interfaces

Addressing 2 first, I think that we used to directly cross the layers which wasn't particularly great, but I also think some of the layers are a little unwieldy.

For 1, I tend to think of executing calls against a specific evm/statedb pair (which is also the external interface), and find it a little jarring to suddenly switch to bound contracts as we jump down the stack.

contracts/utils.go
```diff
@@ -4,85 +4,147 @@ import (
...
+	MaxGas  uint64
+	Value   *big.Int
```
oneeman, line 45: I think either `Value` doesn't belong here, or `method` and `args` should also be included here, because they're pieces of the same thing (what the call is which we are doing).  To a lesser degree I think the same goes for `MaxGas` (either it doesn't belong here or `method` and `args` should be here too)... to a lesser degree because it could be considered more of a constraint on how the call is to be done rather than what the call is which should be done.

contract_comm/evm.go
```diff
@@ -43,31 +41,95 @@ type InternalEVMHandler struct {
...
+
+	contract := contracts.NewContract(abi, contractAddress, contracts.SystemCaller)
+	gasLeft, err := contract.Execute(contracts.ExecOpts{MaxGas: gas, Backend: backend, Value: value}, returnObj, method, args...)
+
+	if err != nil {
+		log.Error("Error when invoking evm function", "err", err, "function", method, "address", contractAddress, "args", args, "gas", gas, "gasLeft", gasLeft)
+		return gasLeft, err
+	}
+
 	return gasLeft, err
```
oneeman, line 100: Either make this `gasLeft, nil` or make the corresponding one in `Query` (line 68) also be `gasLeft, err` (in which case we don't need `return`'s inside the `if`'s.  Either way is fine, but nice to be consistent.

contracts/utils.go
```diff
@@ -4,85 +4,150 @@ import (
...
+	GetStateDB() vm.StateDB
 }
 
-func CallFromSystem(evm *vm.EVM, contractAddress common.Address, abi abipkg.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int) (uint64, error) {
-	call := func(transactionData []byte) ([]byte, uint64, error) {
-		return evm.Call(systemCaller, contractAddress, transactionData, gas, value)
+func NewContract(contractAbi *abi.ABI, contractAddress, defaultCaller common.Address) *Contract {
+	return &Contract{
+		abi:     contractAbi,
+		address: contractAddress,
```
oneeman, line 37: There is something a little strange about the address being here, since the address (except in the case of the registry, where it's hard-coded) is really part of the state and thus part of the backend that the contract will be used with.  But I don't see another way given that one of the uses is `MakeStaticCallWithAddress` which needs to be callable with ERC20 tokens referenced by address rather than contracts listed in the registry.

contracts/utils.go
```diff
@@ -4,85 +4,150 @@ import (
...
+type QueryOpts struct {
+	Backend Backend
+	MaxGas  uint64
+}
 
-func unpackError(result []byte) (string, error) {
-	if len(result) < 4 || !bytes.Equal(result[:4], errorSig) {
-		return "<tx result not Error(string)>", errors.New("TX result not of type Error(string)")
+func (c *Contract) Execute(opts ExecOpts, result interface{}, method string, args ...interface{}) (uint64, error) {
+	defer c.meterExecutionTime(method)()
```
oneeman, line 53: Given how small the difference is between `Execute` and `Query` compared to how much is the same, I would favor undoing this duplication.  In some cases, I would agree duplication is better to make the control flow simpler, but in this case I don't think so.

There could be a single private method they both use which does all the stuff before and after the call itself, and receives an anonymous function for the call (which will in one case use Call and in the other use StaticCall.

mcortesi:
Closed in favor of #1499

PR #1469, opened 2021-03-28
Adds cached contract backend implementation

mcortesi:
### Description

Adds a Cached contract backend; that will cache evm calls

The cache has a known race condition in the scenario that more than one thread is using the same `statedb` instance will making a call.

Add an example of usage on `getExchangeRate()`. (2nd commit)

### Tested

* ci
* unit tests


contracts/cache_backend.go
```diff
@@ -0,0 +1,114 @@
...
+	cacheHits    = metrics.NewRegisteredMeter("contract_comm/caller/cache_hits", nil)
+	cacheMisses  = metrics.NewRegisteredMeter("contract_comm/caller/cache_misses", nil)
+	cacheBadCast = metrics.NewRegisteredMeter("contract_comm/caller/cache_bad_cast", nil)
+)
+
+type cacheKey struct {
+	stateRoot common.Hash
+	caller    vm.ContractRef
+	addr      common.Address
+	input     []byte
```
trianglesphere, line 22: Needs to be a non-slice type: https://play.golang.org/p/0vSKQnveUiW

oneeman: I'm likewise wondering about `caller`, above, since it's an interface.  To be safe, it would have to be something like `callerAddr common.Address` and to get it we'd call `caller.Address()`?

contract_comm/evm.go
```diff
@@ -122,14 +122,23 @@ func QueryWithAddress(backend contracts.Backend, contractAddress common.Address,
...
-	vmevm, err := createBackend(header, state)
+	vmevm, err := NewBackend(header, state)
 	if err != nil {
 		return common.ZeroAddress, err
 	}
 	return contracts.GetRegisteredAddress(vmevm, registryId)
 }
 
-func createBackend(header *types.Header, state vm.StateDB) (contracts.Backend, error) {
+func NewCachedBackend(header *types.Header, state vm.StateDB) (contracts.Backend, error) {
```
oneeman, line 132: This is `NewCachedBackend`, but the type is called `cacheBackend` rather than `cachedBackend`.  Personally, I would rename the type to `cachedBackend`.

contracts/cache_backend.go
```diff
@@ -0,0 +1,114 @@
...
+	var globalCache, err = lru.New(100)
+	if err != nil {
+		panic("error creating global cache: " + err.Error())
+	}
+	return func(backend Backend) Backend {
+		return WithCache(backend, globalCache)
+	}
+}
+
+var WithGlobalCache = createGlobalCache()
```
oneeman, line 53: I don't like the name of `createGlobalCache()`, since it suggests it returns the cache rather than a function to wrap a backend with the cache.  And, really, what makes it global isn't anything in the function itself but rather how it's used on this line to set a global var.  We could inline it, but the result isn't super-readable, so maybe just rename it to something ugly like `createWrapperWithGlobalCache()`or something ugly like that.

mcortesi:
We favored a different direction, on which a low level cache might not be needed. 

New contract refactor is on  #1499

PR #1470, opened 2021-03-29
Split out startNewSequence from core.startNewRound

trianglesphere:
### Description

This creates a function called `core.startNewSequence` that is used in place of `core.startNewRound(common.Big0)`. This helps clean up the code and simplifies the logic of starting new rounds and sequences.

### Tested

Unit tests

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -465,7 +465,8 @@ func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbu
 // startNewRound starts a new round with the desired round
 func (c *core) startNewRound(round *big.Int) error {
 	logger := c.newLogger("func", "startNewRound", "tag", "stateTransition")
-	_, headAuthor := c.backend.GetCurrentHeadBlockAndAuthor()
+	prevBlock := c.current.Sequence().Uint64() - 1
+	blockAuthor := c.backend.AuthorForBlock(prevBlock)
 
 	if round.Cmp(c.current.Round()) == 0 {
 		logger.Trace("Already in the desired round.")
@@ -490,7 +491,7 @@ func (c *core) startNewRound(round *big.Int) error {
 
 	// Calculate new propose
 	valSet := c.current.ValidatorSet()
-	nextProposer := c.selectProposer(valSet, headAuthor, newView.Round.Uint64())
+	nextProposer := c.selectProposer(valSet, blockAuthor, newView.Round.Uint64())
 
 	// Update the roundstate db
 	c.current.StartNewRound(round, valSet, nextProposer)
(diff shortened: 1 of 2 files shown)


consensus/istanbul/core/core.go
```diff
@@ -462,81 +462,102 @@ func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbu
...
+	var err error
+	request, roundChangeCertificate, err := c.getPreprepareWithRoundChangeCertificate(round)
+	if err != nil {
+		logger.Error("Unable to produce round change certificate", "err", err, "new_round", round)
+		return nil
+	}
+
+	// Calculate new propose
+	valSet := c.current.ValidatorSet()
+	nextProposer := c.selectProposer(valSet, headAuthor, newView.Round.Uint64())
```
oneeman, line 493: `headAuthor` here doesn't seem right.  If we're on sequence number 100, and in the meantime we've received block 100, shouldn't we still use the author from block 99?

oneeman: If so, I see there's a function called `AuthorForBlock()` that could be used, and it could be right here rather than at the beginning of the function.

consensus/istanbul/core/core.go
```diff
@@ -626,15 +644,8 @@ func (c *core) createRoundState() (RoundState, error) {
...
-// resetRoundState will modify the RoundState to either start a new round or a new sequence
-// based on the `roundChange` flag given
-func (c *core) resetRoundState(view *istanbul.View, validatorSet istanbul.ValidatorSet, nextProposer istanbul.Validator, roundChange bool) error {
-	// TODO(Joshua): Include desired round here.
-	if roundChange {
-		return c.current.StartNewRound(view.Round, validatorSet, nextProposer)
-	}
-
-	// sequence change
+// resetRoundState will modify the RoundState to either start a new sequence
```
oneeman, line 647:
```suggestion
// resetRoundState will modify the RoundState to start a new sequence
```

consensus/istanbul/core/core.go
trianglesphere: Yea, switch it to pulling the author based on the current sequence number.

consensus/istanbul/core/core.go
```diff
@@ -462,81 +462,103 @@ func (c *core) getPreprepareWithRoundChangeCertificate(round *big.Int) (*istanbu
 	return request, roundChangeCertificate, nil
 }
 
-// startNewRound starts a new round. if round equals to 0, it means to starts a new sequence
+// startNewRound starts a new round with the desired round
 func (c *core) startNewRound(round *big.Int) error {
-	roundChange := false
+	logger := c.newLogger("func", "startNewRound", "tag", "stateTransition")
+	prevBlock := c.current.Sequence().Uint64() - 1
+	blockAuthor := c.backend.AuthorForBlock(prevBlock)
```
oneeman, line 469: I would move these two lines to the same block as `valSet`, since they're only used there.

oneeman (approved):
LGTM!

PR #1477, opened 2021-04-02
Caches currency.Cmp

trianglesphere:
### Description

This implements a deep cache based on the contract address, transaction data, and state root. If the state db has been modified, the cache is skipped. Right now the only contract call that is cached is `medianRate` when it is being called by `currency.Cmp`

### Other changes

Refactors some of `contract_comm/evm.go`. Provides a flag `vm.checkcache` which will check the result of every cached lookup. This is a temporary flag to be used to confirm that the cache is operating correctly on test networks.

### Tested

Test with a load bot and verifying that I see the metrics that I expect (i.e. not many misses, and the same number of checks and hits when running with the check flag, also no bad casts or skips).

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/837

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -750,6 +750,10 @@ var (
 		Usage: "External EVM configuration (default = built-in interpreter)",
 		Value: "",
 	}
+	CheckStaticCallCacheFlag = cli.BoolFlag{
+		Name:  "vm.checkcache",
+		Usage: "If true, this checks the results of the static call cache. Note: this flag is experimental and will be removed",
+	}
 
 	// Istanbul settings
 
@@ -1710,6 +1714,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 
 	if ctx.GlobalIsSet(EVMInterpreterFlag.Name) {
 		cfg.EVMInterpreter = ctx.GlobalString(EVMInterpreterFlag.Name)
+	}
+	if ctx.GlobalIsSet(CheckStaticCallCacheFlag.Name) {
+		cfg.CheckStaticCallCache = ctx.GlobalBool(CheckStaticCallCacheFlag.Name)
 	}
 
 	if ctx.GlobalIsSet(RPCGlobalGasCap.Name) {
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -752,7 +752,7 @@ var (
 	}
 	CheckStaticCallCacheFlag = cli.BoolFlag{
 		Name:  "vm.checkcache",
-		Usage: "If true, this checks the results of the static call cache. Note: this flag is experimental and will be removed",
+		Usage: "If true, this checks the results of the static call cache. Note: this flag is temporary and will be removed",
 	}
 
 	// Istanbul settings
--- a/contract_comm/currency/currency.go
+++ b/contract_comm/currency/currency.go
@@ -212,7 +212,7 @@ func medianRate(currencyAddress *common.Address, memoize bool) (*exchangeRate, e
 		err         error
 	)
 	if memoize {
-		leftoverGas, err = contract_comm.MakeStaticCall(params.SortedOraclesRegistryId, medianRateFuncABI, "medianRate", []interface{}{currencyAddress}, &returnArray, params.MaxGasForMedianRate, nil, nil)
+		leftoverGas, err = contract_comm.MakeMemoizedStaticCall(params.SortedOraclesRegistryId, medianRateFuncABI, "medianRate", []interface{}{currencyAddress}, &returnArray, params.MaxGasForMedianRate, nil, nil)
 	} else {
 		leftoverGas, err = contract_comm.MakeStaticCall(params.SortedOraclesRegistryId, medianRateFuncABI, "medianRate", []interface{}{currencyAddress}, &returnArray, params.MaxGasForMedianRate, nil, nil)
 
--- a/contract_comm/currency/currency.go
+++ b/contract_comm/currency/currency.go
@@ -164,8 +164,8 @@ func Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *com
 		return val1.Cmp(val2)
 	}
 
-	exchangeRate1, err1 := getExchangeRate(currency1)
-	exchangeRate2, err2 := getExchangeRate(currency2)
+	exchangeRate1, err1 := memoizedGetexchangeRate(currency1)
+	exchangeRate2, err2 := memoizedGetexchangeRate(currency2)
 
 	if err1 != nil || err2 != nil {
 		currency1Output := "nil"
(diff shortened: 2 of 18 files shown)


contract_comm/evm.go
```diff
@@ -41,22 +41,95 @@ type InternalEVMHandler struct {
 	chain vm.ChainContext
 }
 
+// MakeStaticCallToAddress performs a static (read-only) ABI call to the smart contract address given.
```
asaj, line 44: Keeping this in the same place as the previous function will make reviewing easier

contract_comm/evm.go
```diff
@@ -41,22 +41,95 @@ type InternalEVMHandler struct {
...
+
+// MakeMemoizedStaticCall performs a static (read-only) ABI call to the smart contract address given.
+// It will attempt to memoize the call based on the contract address, transaction name, and state root.
+func MakeMemoizedStaticCall(registryId [32]byte, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, header *types.Header, state vm.StateDB) (uint64, error) {
+	// Record a metrics data point about execution time.
+	timer := metrics.GetOrRegisterTimer("contract_comm/systemcall/"+funcName, nil)
+	start := time.Now()
+	defer timer.UpdateSince(start)
+
+	scAddress, err := resolveAddressForCall(registryId, funcName, header, state)
```
asaj, line 83: Should we be memoizing this as well?

core/state/statedb.go
```diff
@@ -866,3 +866,24 @@ func (s *StateDB) Commit(deleteEmptyObjects bool) (common.Hash, error) {
 	}
 	return root, err
 }
+
+// Clean returns true if the statedb does not have any pending modifications
+func (s *StateDB) Clean() bool {
+	// It is ditry is any single of the conditions holds
```
asaj, line 872: nit: typo

core/state/statedb.go
```diff
@@ -866,3 +866,24 @@ func (s *StateDB) Commit(deleteEmptyObjects bool) (common.Hash, error) {
 	}
 	return root, err
 }
+
+// Clean returns true if the statedb does not have any pending modifications
+func (s *StateDB) Clean() bool {
```
asaj, line 871: nit: Suggest `IsClean`

contract_comm/currency/currency.go
```diff
@@ -190,33 +190,50 @@ func Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *com
...
 
 func getExchangeRate(currencyAddress *common.Address) (*exchangeRate, error) {
+	if currencyAddress == nil {
+		return &exchangeRate{cgExchangeRateNum, cgExchangeRateDen}, nil
+	} else {
+		return medianRate(currencyAddress, false)
+	}
+}
+
+func memoizedGetexchangeRate(currencyAddress *common.Address) (*exchangeRate, error) {
```
oneeman, line 200: should be `memoizedGetExchangeRate` (E instead of e)

core/vm/evm.go
```diff
@@ -602,6 +604,86 @@ func (evm *EVM) CallFromSystem(contractAddress common.Address, abi abipkg.ABI, f
...
+	cacheCheck   = metrics.NewRegisteredMeter("contract_comm/caller/cache_check", nil)
+)
+
+func (evm *EVM) MemoizedStaticCallFromSystem(contractAddress common.Address, abi abipkg.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64) (uint64, error) {
+	staticCall := func(transactionData []byte) ([]byte, uint64, error) {
+		if !evm.StateDB.IsClean() || staticCallCache == nil {
+			cacheSkipped.Mark(1)
+			return evm.StaticCall(systemCaller, contractAddress, transactionData, gas)
+		}
+		stateRoot := evm.StateDB.StateRoot()
```
asaj, line 633: Should we consider it a cache miss if the state root is dirty? Otherwise we risk returning incorrect values

contract_comm/currency/currency.go
```diff
@@ -164,8 +164,8 @@ func Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *com
 		return val1.Cmp(val2)
 	}
 
-	exchangeRate1, err1 := getExchangeRate(currency1)
-	exchangeRate2, err2 := getExchangeRate(currency2)
+	exchangeRate1, err1 := memoizedGetExchangeRate(currency1)
```
asaj, line 167: Just to confirm, the *only* place `Cmp` is used is in comparing gas prices for tx inclusion, right? Is there anywhere else this is used that is not safe to memoize?

trianglesphere: It's used in the tx pool as well to compare newer transactions. It's not use anywhere in consensus (though it should theoretically be safe to use in consensus)

core/vm/evm.go
trianglesphere: I'm calling it a cache skip (3 lines above)

cmd/utils/flags.go
```diff
@@ -750,6 +750,10 @@ var (
 		Usage: "External EVM configuration (default = built-in interpreter)",
 		Value: "",
 	}
+	CheckStaticCallCacheFlag = cli.BoolFlag{
+		Name:  "vm.checkcache",
+		Usage: "If true, this checks the results of the static call cache. Note: this flag is experimental and will be removed",
```
nategraf, line 755: I might recommend "temporary" instead of "experimental", although I am not actually sure why we would remove this flag in the future. It seems like something helpful to have for debugging should we adjust this code in the future and as we add new calls that use it.

core/state/statedb.go
```diff
@@ -866,3 +866,24 @@ func (s *StateDB) Commit(deleteEmptyObjects bool) (common.Hash, error) {
...
+func (s *StateDB) IsClean() bool {
+	// It is dirty is any single of the conditions holds
+	return !(s.txIndex > 0 ||
+		s.thash != (common.Hash{}) ||
+		s.bhash != (common.Hash{}) ||
+		len(s.stateObjectsDirty) > 0 ||
+		len(s.stateObjectsPending) > 0 ||
+		len(s.journal.dirties) > 0 ||
+		len(s.journal.entries) > 0)
+}
```
nategraf, line 880: One (simpler) alternative might be to drill one layer deeper and add a function to `trie.Trie` that will allow for checking if the trie is dirty based on the `t.unhashed` value. https://github.com/celo-org/celo-blockchain/blob/5f046188332511f2974357653a84e3f0008a58fe/trie/trie.go#L54

core/vm/evm.go
```diff
@@ -602,6 +604,86 @@ func (evm *EVM) CallFromSystem(contractAddress common.Address, abi abipkg.ABI, f
 	return evm.handleABICall(abi, funcName, args, returnObj, call)
 }
 
+type cacheKey struct {
```
nategraf, line 607: We need to include `gas` in this key as it can effect the execution. It shouldn't reduce our hit rate because we use the same gas value every time.

core/vm/evm.go
```diff
@@ -602,6 +604,86 @@ func (evm *EVM) CallFromSystem(contractAddress common.Address, abi abipkg.ABI, f
...
+}
+
+type cacheResult struct {
+	ret     []byte
+	gasLeft uint64
+}
+
+var staticCallCache, _ = lru.New(100)
+var (
+	cacheHits    = metrics.NewRegisteredMeter("contract_comm/caller/cache_hits", nil)
```
nategraf, line 620: I would suggest a namespace relevant to the `core/vm` package, but there aren't currently any others. I would suggest `contract_comm/systemcall` just to align with the ones in `contract_comm/evm` though
```suggestion
	cacheHits    = metrics.NewRegisteredMeter("contract_comm/systemcall/cache_hits", nil)
```

core/vm/evm.go
```diff
@@ -602,6 +604,86 @@ func (evm *EVM) CallFromSystem(contractAddress common.Address, abi abipkg.ABI, f
...
+				staticCallCache.Remove(key)
+				return evm.StaticCall(systemCaller, contractAddress, transactionData, gas)
+			}
+			cacheHits.Mark(1)
+			ret := make([]byte, len(cachedResult.ret))
+			copy(ret, cachedResult.ret)
+			if evm.vmConfig.CheckStaticCallCache {
+				actualRet, actualGasLeft, actualErr := evm.StaticCall(systemCaller, contractAddress, transactionData, gas)
+				if actualErr != nil {
+					log.Error("non nil error when performing evm static call that wanted to use the cached result")
```
nategraf, line 653: We should make sure these error messages have a string that makes them easy to find for error detection purposes. "cached result" is probably good enough, but "static call cached result" might be better.

cmd/utils/flags.go
trianglesphere: I'll switch it to temporary. Maybe "developer" is another option. I think we may keep this for a couple releases, but I want to be able to remove it in the future.

nategraf (approved):
🚢

PR #1478, opened 2021-04-03
Cache currency comparisons when building blocks

asaj:
### Description

This PR introduces a CurrencyComparator object that the miner uses to compare transactions when deciding which transactions to include in their next block proposal. The CurrencyComparator caches exchange rates between different fee currencies to avoid fetching them every time.

### Other changes

None.

### Tested

- Unit tests
- [TODO] Deploy to baklava and alfajores
- [TODO] Load testing

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


mcortesi:
Already included in https://github.com/celo-org/celo-blockchain/pull/1479

PR #1487, opened 2021-04-07
Celostats reliability

### LINKED ISSUE #1395: celostats reliability
Improve reliability to the connection with the celostats server

### LINKED ISSUE #1397: celostats validator true client version
Validators client version, should be available in the celostats server

Today if it's behind a proxy, some of the fields are from the proxy and not the validator

gastonponti:
### Description

Improve reliability from the ethstats module

The connectionWrapper was cherrypicked from go-ethereum upstream:
https://github.com/ethereum/go-ethereum/commit/82a9e11058623e73ace50445b41a06ccf8fd888f (https://github.com/ethereum/go-ethereum/pull/21404)

### Other changes

Validator injecting its version to the proxy stats chunk

### Tested

Manually in a local testnet

### Related issues

- Fixes #1395 
- Fixes #1397 

### Backwards compatibility

Yes

### DIFF
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -529,18 +529,18 @@ func (s *Service) fillWithValidatorInfo(message *StatsPayload) {
 			proxyInfo, ok := msg["info"].(map[string]interface{})
 			if ok {
 				infos := s.server.NodeInfo()
-				marta := proxyInfo["node"].(string)
-				proxyInfoParts := nodeNameRegex.FindStringSubmatch(marta)
-				validatorInfoParts := nodeNameRegex.FindStringSubmatch(infos.Name)
+				proxyNode := proxyInfo["node"].(string)
+				proxyNodeParts := nodeNameRegex.FindStringSubmatch(proxyNode)
+				validatorNodeParts := nodeNameRegex.FindStringSubmatch(infos.Name)
 				// if one of the regex failed, maintain the proxy node name
-				if proxyInfoParts != nil && validatorInfoParts != nil {
+				if proxyNodeParts != nil && validatorNodeParts != nil {
 					proxyInfo["node"] = fmt.Sprintf(
 						"%s/%s(val:%s)/%s/%s",
-						proxyInfoParts[1],
-						proxyInfoParts[2],
-						validatorInfoParts[2],
-						proxyInfoParts[3],
-						proxyInfoParts[4],
+						proxyNodeParts[1],
+						proxyNodeParts[2],
+						validatorNodeParts[2],
+						proxyNodeParts[3],
+						proxyNodeParts[4],
 					)
 				}
 			}
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -286,6 +286,7 @@ func (s *Service) loop(ctx context.Context) {
 			for {
 				select {
 				case delegateSignMsg := <-signCh:
+					s.fillWithValidatorInfo(&delegateSignMsg.Payload)
 					if err := s.handleDelegateSign(&delegateSignMsg.Payload, delegateSignMsg.PeerID); err != nil {
 						log.Warn("Delegate sign failed", "err", err)
 					}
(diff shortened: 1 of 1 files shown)


ethstats/ethstats.go
```diff
@@ -428,12 +498,53 @@ func (s *Service) login(conn *websocket.Conn, sendCh chan *StatsPayload) error {
...
+var nodeNameRegex = regexp.MustCompile(`(.*)/(.*)/(.*)/(.*)`)
+
+func (s *Service) fillWithValidatorInfo(message *StatsPayload) {
+	if message.Action == actionHello {
+		msg, ok := message.Stats.(map[string]interface{})
+		if ok {
+			proxyInfo, ok := msg["info"].(map[string]interface{})
+			if ok {
+				infos := s.server.NodeInfo()
+				marta := proxyInfo["node"].(string)
```
trianglesphere, line 532: Nit: I'm not sure what `marta` means

ethstats/ethstats.go
```diff
@@ -259,81 +306,104 @@ func (s *Service) loop(ctx context.Context) {
...
+				case <-errTimer.C:
+					var (
+						conn *connWrapper
+						err  error
+					)
+					// Establish a websocket connection to the server on any supported URL
+					dialer := websocket.Dialer{HandshakeTimeout: 5 * time.Second}
+					header := make(http.Header)
+					header.Set("origin", "http://localhost")
+					for _, url := range urls {
```
trianglesphere, line 325: This isn't new, but it is it correct that we only use one connection?

ethstats/ethstats.go
```diff
@@ -259,81 +306,104 @@ func (s *Service) loop(ctx context.Context) {
...
+							if err = s.reportPending(conn); err != nil {
+								log.Warn("Post-block transaction stats report failed", "err", err)
+							}
+						case <-txCh:
+							if err = s.reportPending(conn); err != nil {
+								log.Warn("Transaction stats report failed", "err", err)
+							}
+						case signedMessage := <-sendCh:
+							// if it is a ping or hello message, it shouldn't be handled here
+							if signedMessage.Action != actionNodePing && signedMessage.Action != actionHello {
```
trianglesphere, line 389: Would it make sense to split this out into several channels (1 for each message type)? It seems like we're selecting from it in multiple places, but are only looking for a specific message.

I think with the retry logic we'll be fine, but I think it may be possible to rely on it less (but you'll probably still need it b/c of the network).

trianglesphere (commented):
Generally looks good. 

I'd be curious if upstream has done something similar with the connWrapper (it seems like they'd have similar issues w/ concurrency).

Otherwise I had a couple comments on some potential future work. I think this is better than what was, but I wonder if it might be possible to make the `sendCh` less racey.

ethstats/ethstats.go
gastonponti: My bad, sorry. Changing it

gastonponti:
> I'd be curious if upstream has done something similar with the connWrapper (it seems like they'd have similar issues w/ concurrency).

Yes, that change is from upstream: https://github.com/ethereum/go-ethereum/pull/21404/files

trianglesphere:
> Yes, that change is from upstream: https://github.com/ethereum/go-ethereum/pull/21404/files

Oh nice. Could you tag it in the PR as a cherry-pick for when we eventually get around to merging that from upstream.

ethstats/ethstats.go
gastonponti: To the celostats server? I think that as is not critical data, handling only one connection keeps the other code (and even the celostats server) simpler.

ethstats/ethstats.go
gastonponti: This is actually a check that shouldn't happened because those are messages that should be consumed before, or fail due to a timeout, and that timeout should trigger a disconnect -> reconnect. But I wanted to differentiate that case for future logs if for some reason there was something under my radar.

Regardless of handling in different channels, it was something I thought about, and it's probably a thing that we could add when we change the way we send stats (to be defined in these days), because there won't be necessary for the proxy to know which message it is (if we start using it as a transport layer, instead of generator+transport as it is today)

ethstats/ethstats.go
trianglesphere: Yea. I agree it makes it simpler, but it's weird to have multiple urls but only use 1

trianglesphere (approved):
LGTM

PR #1489, opened 2021-04-08
Revert "Caches currency.Cmp"

mcortesi:
Reverts celo-org/celo-blockchain#1477

We are using https://github.com/celo-org/celo-blockchain/pull/1467 for now


oneeman (approved):
LGTM

PR #1495, opened 2021-04-09
[Exploration] Introduces notion of BlockContext to process transactions

mcortesi:
### Description

**This is a Hardfork change.** 

**Before:** For every transaction we make several EVM calls (getGasPriceMinimun, getGasForAlternativeCurrencies, IsWhitelisted(currency)). 

With this PR, instead, at the beginning of each block, we fetch all necessary EVM state and use it for all the transactions. This means that if a governance proposal or any other transaction changes the value of a parameter mid-block we won't see it until the next block. There is no problem in delaying this, and it's actually more simpler to reason about. But it does make this change a hardfork.

### Tested

* CI

### Backwards compatibility

Hardfork change


piersy:
@mcortesi What's the reason for doing this?

mcortesi:
just an exploration, served its purpose

PR #1498, opened 2021-04-10
Make --proxy.proxyenodeurlpairs optional if --proxy.proxied specified

silasbw:
### Description

Documentation implies this behavior is intended. For example,
https://docs.celo.org/validator-guide/proxy.

This facilitates running geth without initially configuring a proxy and adding
proxies later via RPCs. For example, like
https://github.com/mantalabs/proxy-informer.

### Tested

Ad hoc testing locally.

### Backwards compatibility

This change is technically a breaking one because it changes the behavior of the
(deprecated) --proxy.proxyenodeurlpair option from required to
optional. This will break uses where --proxy.proxyenodeurlpair is elided and
there's an expectation that geth will fail to start.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1508,10 +1508,6 @@ func SetProxyConfig(ctx *cli.Context, nodeCfg *node.Config, ethCfg *eth.Config) 
 			Fatalf("Option --%s must be used if option --%s is used", MiningEnabledFlag.Name, ProxiedFlag.Name)
 		}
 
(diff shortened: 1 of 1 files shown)


silasbw:
The `ci/circleci: end-to-end-governance-test` failure looks spurious, but I can investigate if that's not the case.

trianglesphere:
> The `ci/circleci: end-to-end-governance-test` failure looks spurious, but I can investigate if that's not the case.

It's a flaky test, and likely not an issue

trianglesphere (approved):
Thanks for finding this and making it consistent with the docs.

PR #1499, opened 2021-04-12
txpool: contract call caching & remove gpm check

mcortesi:
#### Contract Call Caching 

This PR adds a caching layer to the txpool. It works as follows:

Cache has 2 parts:

 * `BlockContext` which represents contract's information required to process a block. (gasPriceMinimum per currency, whitelistedCurrencies, gasForAlternativeCurrency)
 * `CurrencyManager` object which is able to make currency conversion, and maintains al local cache of exchange rates

Both objects are part of the `txpoolContext`, and they are recreated with every `reset()`. This means they are recreated everytime the txpools learns about a new block.

#### TXPool Behaviour Changes

1. txpool no longer checks for gas price minimum (see https://github.com/celo-org/celo-blockchain/pull/1499#issuecomment-819089144)
2. If `getCurrencyWhitelist()` fails, then assume no custom fee currency
is whitelisted (before: any fee currency was whitelisted)

### Other changes

* Refactors `currency.go` and cleans it up

### Tested

 * ci, unit tests

### Backwards compatibility

YES

### DIFF
--- a/consensus/istanbul/backend/pos.go
+++ b/consensus/istanbul/backend/pos.go
@@ -106,10 +106,11 @@ func (sb *Backend) distributeEpochRewards(header *types.Header, state *state.Sta
 	currencyManager := currency.NewManager(nil, nil)
 
 	// Validator rewards were paid in cUSD, convert that amount to CELO and add it to the Reserve
-	totalValidatorRewardsConvertedToCelo, err := currencyManager.ToCelo(totalValidatorRewards, &stableTokenAddress)
+	stableTokenCurrency, err := currencyManager.GetCurrency(&stableTokenAddress)
 	if err != nil {
 		return err
 	}
+	totalValidatorRewardsConvertedToCelo := stableTokenCurrency.ToCELO(totalValidatorRewards)
 
 	if err = gold_token.Mint(header, state, reserveAddress, totalValidatorRewardsConvertedToCelo); err != nil {
 		return err
--- a/consensus/istanbul/backend/pos.go
+++ b/consensus/istanbul/backend/pos.go
@@ -103,8 +103,10 @@ func (sb *Backend) distributeEpochRewards(header *types.Header, state *state.Sta
 		return err
 	}
 
+	currencyManager := currency.NewManager(nil, nil)
+
 	// Validator rewards were paid in cUSD, convert that amount to cGLD and add it to the Reserve
-	totalValidatorRewardsConvertedToGold, err := currency.Convert(totalValidatorRewards, &stableTokenAddress, nil)
+	totalValidatorRewardsConvertedToGold, err := currencyManager.ToGold(totalValidatorRewards, &stableTokenAddress)
 	if err != nil {
 		return err
 	}
--- a/consensus/istanbul/backend/pos.go
+++ b/consensus/istanbul/backend/pos.go
@@ -105,13 +105,13 @@ func (sb *Backend) distributeEpochRewards(header *types.Header, state *state.Sta
 
 	currencyManager := currency.NewManager(nil, nil)
 
-	// Validator rewards were paid in cUSD, convert that amount to cGLD and add it to the Reserve
-	totalValidatorRewardsConvertedToGold, err := currencyManager.ToGold(totalValidatorRewards, &stableTokenAddress)
+	// Validator rewards were paid in cUSD, convert that amount to CELO and add it to the Reserve
+	totalValidatorRewardsConvertedToCelo, err := currencyManager.ToCelo(totalValidatorRewards, &stableTokenAddress)
 	if err != nil {
 		return err
 	}
 
-	if err = gold_token.Mint(header, state, reserveAddress, totalValidatorRewardsConvertedToGold); err != nil {
+	if err = gold_token.Mint(header, state, reserveAddress, totalValidatorRewardsConvertedToCelo); err != nil {
 		return err
 	}
 
--- a/contract_comm/currency/currency.go
+++ b/contract_comm/currency/currency.go
@@ -130,6 +130,7 @@ func (er *ExchangeRate) FromBase(goldAmount *big.Int) *big.Int {
 	return new(big.Int).Div(new(big.Int).Mul(goldAmount, er.numerator), er.denominator)
 }
 
+// CmpValues compares two values with potentially different exchange rates
 func (er *ExchangeRate) CmpValues(amount *big.Int, anotherTokenAmount *big.Int, anotherTokenRate *ExchangeRate) int {
 	// if both rates are noop rate (CELO rate), compare values
 	if er == nil && anotherTokenRate == nil {
@@ -159,6 +160,7 @@ type CurrencyManager struct {
 	_getExchangeRate func(*common.Address, *types.Header, vm.StateDB) (*ExchangeRate, error) // function to obtain exchange rate from blockchain state
 }
 
+// NewManager creates a new CurrencyManager
 func NewManager(header *types.Header, state vm.StateDB) *CurrencyManager {
 	return newManager(GetExchangeRate, header, state)
 }
@@ -192,7 +194,8 @@ func (cc *CurrencyManager) GetExchangeRate(currency *common.Address) (*ExchangeR
 	return val, nil
 }
 
-func (cc *CurrencyManager) Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
+// CmpValues compares values of potentially differeny currencies
+func (cc *CurrencyManager) CmpValues(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
 	// Short circuit if the fee currency is the same. nil currency => native currency
 	if (currency1 == nil && currency2 == nil) || (currency1 != nil && currency2 != nil && *currency1 == *currency2) {
 		return val1.Cmp(val2)
@@ -217,6 +220,7 @@ func (cc *CurrencyManager) Cmp(val1 *big.Int, currency1 *common.Address, val2 *b
 	return exchangeRate1.CmpValues(val1, val2, exchangeRate2)
 }
 
+// ToCelo converts an amount on a given currency to CELO token
 func (cc *CurrencyManager) ToCelo(amount *big.Int, currency *common.Address) (*big.Int, error) {
 	rate, err := cc.GetExchangeRate(currency)
 	if err != nil {
@@ -225,6 +229,7 @@ func (cc *CurrencyManager) ToCelo(amount *big.Int, currency *common.Address) (*b
 	return rate.ToBase(amount), nil
 }
 
+// GetExchangeRate retrieves the exchange rate from the blockchain state
 func GetExchangeRate(currencyAddress *common.Address, header *types.Header, state vm.StateDB) (*ExchangeRate, error) {
 	if currencyAddress == nil {
 		return &NoopExchangeRate, nil
--- a/contract_comm/currency/currency.go
+++ b/contract_comm/currency/currency.go
@@ -94,89 +94,119 @@ const (
 )
 
 var (
-	cgExchangeRateNum = big.NewInt(1)
-	cgExchangeRateDen = big.NewInt(1)
-
 	medianRateFuncABI, _   = abi.JSON(strings.NewReader(medianRateABI))
 	balanceOfFuncABI, _    = abi.JSON(strings.NewReader(balanceOfABI))
 	getWhitelistFuncABI, _ = abi.JSON(strings.NewReader(getWhitelistABI))
 )
 
-type exchangeRate struct {
-	Numerator   *big.Int
-	Denominator *big.Int
+// ExchangeRate represent the exchangeRate for the pair (base, token)
+// Follows the equation: 1 base * ExchangeRate = X token
+type ExchangeRate struct {
+	numerator   *big.Int
+	denominator *big.Int
 }
 
-func ConvertToGold(val *big.Int, currencyFrom *common.Address) (*big.Int, error) {
-	if currencyFrom == nil {
-		return val, nil
-	}
+// NoopExchangeRate returns the noop rate
+// which results in a 1:1 conversion
+func NoopExchangeRate() *ExchangeRate { return nil }
 
-	celoGoldAddress, err := contract_comm.GetRegisteredAddress(params.GoldTokenRegistryId, nil, nil)
-	if err == errors.ErrSmartContractNotDeployed || err == errors.ErrRegistryContractNotDeployed {
-		log.Warn("Registry address lookup failed", "err", err)
-		return val, err
+// MustNewExchangeRate creates an exchange rate, panic on error
+func MustNewExchangeRate(numerator *big.Int, denominator *big.Int) *ExchangeRate {
+	rate, err := NewExchangeRate(numerator, denominator)
+	if err != nil {
+		panic(err)
 	}
-
-	if *currencyFrom == celoGoldAddress {
-		// This function shouldn't really be called with the token's address, but if it is the value
-		// is correct, so return nil as the error
-		return val, nil
-	} else if err != nil {
-		log.Error(err.Error())
-		return val, err
-	}
-	return Convert(val, currencyFrom, nil)
+	return rate
 }
 
-// NOTE (jarmg 4/24/19): values are rounded down which can cause
-// an estimate to be off by 1 (at most)
-func Convert(val *big.Int, currencyFrom *common.Address, currencyTo *common.Address) (*big.Int, error) {
-	exchangeRateFrom, err1 := getExchangeRate(currencyFrom)
-	exchangeRateTo, err2 := getExchangeRate(currencyTo)
-
-	if err1 != nil || err2 != nil {
-		log.Error("Convert - Error in retreiving currency exchange rates")
-		if err1 != nil {
-			return nil, err1
-		}
-		if err2 != nil {
-			return nil, err2
-		}
-	}
-
-	// Given value of val and rates n1/d1 and n2/d2 the function below does
-	// (val * d1 * n2) / (n1 * d2)
-	numerator := new(big.Int).Mul(val, new(big.Int).Mul(exchangeRateFrom.Denominator, exchangeRateTo.Numerator))
-	denominator := new(big.Int).Mul(exchangeRateFrom.Numerator, exchangeRateTo.Denominator)
-
-	if denominator.Sign() == 0 {
-		log.Error("Convert - Error in retreiving currency exchange rates")
+// NewExchangeRate creates an exchange rate.
+// Requires numerator >=0 && denominator >= 0
+func NewExchangeRate(numerator *big.Int, denominator *big.Int) (*ExchangeRate, error) {
+	if numerator == nil || common.Big0.Cmp(numerator) >= 0 {
 		return nil, errors.ErrExchangeRateZero
 	}
-
-	return new(big.Int).Div(numerator, denominator), nil
+	if denominator == nil || common.Big0.Cmp(denominator) >= 0 {
+		return nil, errors.ErrExchangeRateZero
+	}
+	return &ExchangeRate{numerator, denominator}, nil
 }
 
-type CurrencyComparator struct {
-	exchangeRates    map[common.Address]*exchangeRate
-	_getExchangeRate func(*common.Address) (*exchangeRate, error)
+// ToBase converts from token to base
+func (er *ExchangeRate) ToBase(tokenAmount *big.Int) *big.Int {
+	// check noop rate (cGLD rate)
+	if er == nil {
+		return new(big.Int).Set(tokenAmount)
+	}
+
+	return new(big.Int).Div(new(big.Int).Mul(tokenAmount, er.denominator), er.numerator)
 }
 
-func NewComparator() *CurrencyComparator {
-	return newComparator(getExchangeRate)
+// FromGold converts from base to token
+func (er *ExchangeRate) FromBase(goldAmount *big.Int) *big.Int {
+	// check noop rate (cGLD rate)
+	if er == nil {
+		return new(big.Int).Set(goldAmount)
+	}
+
+	return new(big.Int).Div(new(big.Int).Mul(goldAmount, er.numerator), er.denominator)
 }
 
-func newComparator(_getExchangeRate func(*common.Address) (*exchangeRate, error)) *CurrencyComparator {
-	return &CurrencyComparator{
-		exchangeRates:    make(map[common.Address]*exchangeRate),
+func (er *ExchangeRate) CmpValues(amount *big.Int, anotherTokenAmount *big.Int, anotherTokenRate *ExchangeRate) int {
+	// if both rates are noop rate (cGLD rate), compare values
+	if er == nil && anotherTokenRate == nil {
+		return amount.Cmp(anotherTokenAmount)
+	}
+
+	// if both exchangeRate are the same
+	if er != nil && anotherTokenRate != nil && *er == *anotherTokenRate {
+		return amount.Cmp(anotherTokenAmount)
+	}
+
+	// Below code block is basically evaluating this comparison:
+	// amount * er.denominator / er.numerator < anotherTokenAmount * anotherTokenRate.denominator / anotherTokenRate.numerator
+	// It will transform that comparison to this, to remove having to deal with fractional values.
+	// amount * er.denominator * anotherTokenRate.numerator < anotherTokenAmount * anotherTokenRate.denominator * er.numerator
+
+	// when either er or anotherTokenRate are nil, we assume rate = 1/1 (as in cGLD)
+	var leftSide, rightSide *big.Int
+	if er == nil {
+		leftSide = new(big.Int).Mul(amount, anotherTokenRate.numerator)
+		rightSide = new(big.Int).Mul(anotherTokenAmount, anotherTokenRate.denominator)
+	} else if anotherTokenRate == nil {
+		leftSide = new(big.Int).Mul(amount, er.denominator)
+		rightSide = new(big.Int).Mul(anotherTokenAmount, er.numerator)
+	} else {
+		leftSide = new(big.Int).Mul(amount, new(big.Int).Mul(er.denominator, anotherTokenRate.numerator))
+		rightSide = new(big.Int).Mul(anotherTokenAmount, new(big.Int).Mul(anotherTokenRate.denominator, er.numerator))
+	}
+
+	return leftSide.Cmp(rightSide)
+}
+
+type CurrencyManager struct {
+	header *types.Header
+	state  vm.StateDB
+
+	exchangeRates    map[common.Address]*ExchangeRate                                        // map of exchange rates of the form (cGLD, token)
+	_getExchangeRate func(*common.Address, *types.Header, vm.StateDB) (*ExchangeRate, error) // function to obtain exchange rate from blockchain state
+}
+
+func NewManager(header *types.Header, state vm.StateDB) *CurrencyManager {
+	return newManager(GetExchangeRate, header, state)
+}
+
+func newManager(_getExchangeRate func(*common.Address, *types.Header, vm.StateDB) (*ExchangeRate, error), header *types.Header, state vm.StateDB) *CurrencyManager {
+	return &CurrencyManager{
+		header:           header,
+		state:            state,
+		exchangeRates:    make(map[common.Address]*ExchangeRate),
 		_getExchangeRate: _getExchangeRate,
 	}
 }
 
-func (cc *CurrencyComparator) getExchangeRate(currency *common.Address) (*exchangeRate, error) {
+func (cc *CurrencyManager) GetExchangeRate(currency *common.Address) (*ExchangeRate, error) {
 	if currency == nil {
-		return &exchangeRate{cgExchangeRateNum, cgExchangeRateDen}, nil
+		return NoopExchangeRate(), nil
 	}
 
 	val, ok := cc.exchangeRates[*currency]
@@ -184,7 +214,7 @@ func (cc *CurrencyComparator) getExchangeRate(currency *common.Address) (*exchan
 		return val, nil
 	}
 
-	val, err := cc._getExchangeRate(currency)
+	val, err := cc._getExchangeRate(currency, cc.header, cc.state)
 	if err != nil {
 		return nil, err
 	}
@@ -194,14 +224,14 @@ func (cc *CurrencyComparator) getExchangeRate(currency *common.Address) (*exchan
 	return val, nil
 }
 
-func (cc *CurrencyComparator) Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
+func (cc *CurrencyManager) Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
 	// Short circuit if the fee currency is the same. nil currency => native currency
 	if (currency1 == nil && currency2 == nil) || (currency1 != nil && currency2 != nil && *currency1 == *currency2) {
 		return val1.Cmp(val2)
 	}
 
-	exchangeRate1, err1 := cc.getExchangeRate(currency1)
-	exchangeRate2, err2 := cc.getExchangeRate(currency2)
+	exchangeRate1, err1 := cc.GetExchangeRate(currency1)
+	exchangeRate2, err2 := cc.GetExchangeRate(currency2)
 
 	if err1 != nil || err2 != nil {
 		currency1Output := "nil"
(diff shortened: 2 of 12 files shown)


trianglesphere:
> Consider current whitelist status when doing txList.Filter().
(before: whitelisted status was only checked on txAdd() )

One callout here is that while it would be possible to modify the fee currency contract to support removing whitelisted fee currencies, it is not currently possible to do so. As such we didn't feel a need to check this after it had already been checked in `validateTx`

core/block_context.go
```diff
@@ -0,0 +1,87 @@
...
+
+// BlockContext represents contextual information about the blockchain state
+// for a give block
+type BlockContext interface {
+	// GetGoldGasPriceMinimum retrieves the gas price minimum for the CELO token
+	GetGoldGasPriceMinimum() *big.Int
+
+	// GetGasPriceMinimum retrieves the gas price minimum for any currency
+	// Also indicates if the currency is not whitelisted
+	GetGasPriceMinimum(feeCurrency *common.Address) (gpm *big.Int, isWhitelisted bool)
```
trianglesphere, line 22: I like having an explicit error condition here, but I lean towards having a separate `IsWhiteListedFeeCurrency` function

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
 
-func ConvertToGold(val *big.Int, currencyFrom *common.Address) (*big.Int, error) {
-	if currencyFrom == nil {
-		return val, nil
+// NoopExchangeRate returns the noop rate
+// which results in a 1:1 conversion
+func NoopExchangeRate() *ExchangeRate { return nil }
+
+// MustNewExchangeRate creates an exchange rate, panic on error
+func MustNewExchangeRate(numerator *big.Int, denominator *big.Int) *ExchangeRate {
```
trianglesphere, line 114: This is only used in the test file, so it feels like it should live there.

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
-		// This function shouldn't really be called with the token's address, but if it is the value
-		// is correct, so return nil as the error
-		return val, nil
-	} else if err != nil {
-		log.Error(err.Error())
-		return val, err
+// ToBase converts from token to base
+func (er *ExchangeRate) ToBase(tokenAmount *big.Int) *big.Int {
+	// check noop rate (cGLD rate)
+	if er == nil {
```
trianglesphere, line 137: Despite the fact that is compiles and works; I much rather have a `type NoOpExchangeRate struct` which implements an exchange rate interface. Better yet would be returning an exchange rate of 1.

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
 	balanceOfFuncABI, _    = abi.JSON(strings.NewReader(balanceOfABI))
 	getWhitelistFuncABI, _ = abi.JSON(strings.NewReader(getWhitelistABI))
 )
 
-type exchangeRate struct {
-	Numerator   *big.Int
-	Denominator *big.Int
+// ExchangeRate represent the exchangeRate for the pair (base, token)
+// Follows the equation: 1 base * ExchangeRate = X token
+type ExchangeRate struct {
```
trianglesphere, line 104: Not actually used outside this package.

contract_comm/currency/currency.go
```diff
@@ -216,128 +246,100 @@ func (cc *CurrencyComparator) Cmp(val1 *big.Int, currency1 *common.Address, val2
...
+	rate, err := cc.GetExchangeRate(currency)
+	if err != nil {
+		return nil, err
 	}
+	return rate.ToBase(amount), nil
+}
 
-	exchangeRate1, err1 := getExchangeRate(currency1)
-	exchangeRate2, err2 := getExchangeRate(currency2)
+func GetExchangeRate(currencyAddress *common.Address, header *types.Header, state vm.StateDB) (*ExchangeRate, error) {
```
trianglesphere, line 260: Also not used outside this module

core/tx_pool.go
```diff
@@ -1266,6 +1265,14 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	pool.pendingNonces = newTxNoncer(statedb)
 	pool.currentMaxGas = CalcGasLimit(pool.chain.CurrentBlock(), statedb)
 
+	// we don't replace currentCtx, since it's shared with all txLists
```
trianglesphere, line 1268: If `currentCtx` is shared (which means `BlockContext` and `CurrencyManager` probably are as well, do they handle concurrency properly? If they do, documenting that is important)

core/block_context.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// GetIntrinsicGasForAlternativeFeeCurrency retrieves intrisic gas to be paid for
+	// any tx with a non native fee currency
+	GetIntrinsicGasForAlternativeFeeCurrency() uint64
+}
+
+// defaultBlockContext is the default implementation of BlockContext
+type defaultBlockContext struct {
+	goldGasPriceMinimum *big.Int
+	nonGoldCurrencies   map[common.Address]*big.Int
```
trianglesphere, line 32: `nonGoldGPMs`?

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
-func NewComparator() *CurrencyComparator {
-	return newComparator(getExchangeRate)
+func NewManager(header *types.Header, state vm.StateDB) *CurrencyManager {
+	return newManager(GetExchangeRate, header, state)
 }
 
-func newComparator(_getExchangeRate func(*common.Address) (*exchangeRate, error)) *CurrencyComparator {
-	return &CurrencyComparator{
-		exchangeRates:    make(map[common.Address]*exchangeRate),
+func newManager(_getExchangeRate func(*common.Address, *types.Header, vm.StateDB) (*ExchangeRate, error), header *types.Header, state vm.StateDB) *CurrencyManager {
```
trianglesphere, line 198: this can be merged into `NewManager`

trianglesphere (commented):
I feel like this is 2 PRs in one (though the second does half depend on the first).
1. Creating a `contract_comm/currency` interface & extending the caching implementation for it.
I think this part got a little overcomplicated. Previously, we only used the following methods from the currency package
```
// Used in tx list
ConvertToGold(val *big.Int, currencyFrom *common.Address) (*big.Int, error)
// Used in POS.go
Convert(val *big.Int, currencyFrom *common.Address, currencyTo *common.Address) (*big.Int, error)
// Used in the tx pool
func Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int
// state tranistion & tx pool. Probably not worth caching
GetBalanceOf(accountOwner common.Address, contractAddress common.Address, gas uint64, header *types.Header, state vm.StateDB) (result *big.Int, gasUsed uint64, err error)
// state tranistion & tx pool. Probably worth caching
func IsWhitelisted(currencyAddress common.Address, header *types.Header, state vm.StateDB) bool
// tx pool
func CurrencyWhitelist(header *types.Header, state vm.StateDB) ([]common.Address, error)
```
I prefer creating a currency manager / currency_comm interface that combines these methods and then it's possible to write a version that either caches or delegates to a caching static call.

2. Adding the block context to the tx pool. I like the change. I've got a couple nits, and I suspect we may want to switch to a lazy load in certain cases (for a non-caching implementation vs a caching implementation), but it does what it needs to do.

contract_comm/currency/currency.go
trianglesphere: Example showing how this works: https://play.golang.org/p/tQe3kCMVAKq

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
+	if er == nil && anotherTokenRate == nil {
+		return amount.Cmp(anotherTokenAmount)
 	}
 
-	// Given value of val and rates n1/d1 and n2/d2 the function below does
-	// (val * d1 * n2) / (n1 * d2)
-	numerator := new(big.Int).Mul(val, new(big.Int).Mul(exchangeRateFrom.Denominator, exchangeRateTo.Numerator))
-	denominator := new(big.Int).Mul(exchangeRateFrom.Numerator, exchangeRateTo.Denominator)
+	// if both exchangeRate are the same
+	if er != nil && anotherTokenRate != nil && *er == *anotherTokenRate {
```
oneeman, line 161: I don't think this short-circuit works as intended, because the values in `ExchangeRate` are pointers (`*big.Int`).  We could add a `.Cmp()`.  But I also don't really like the signature of this function.  I would prefer for them to be a type that indicates an `amount,exchangeRate` pair, and then that type would have a `.Cmp()` function.

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
-		log.Error("Convert - Error in retreiving currency exchange rates")
-		return nil, errors.ErrExchangeRateZero
+	// Below code block is basically evaluating this comparison:
+	// amount * er.denominator / er.numerator < anotherTokenAmount * anotherTokenRate.denominator / anotherTokenRate.numerator
+	// It will transform that comparison to this, to remove having to deal with fractional values.
+	// amount * er.denominator * anotherTokenRate.numerator < anotherTokenAmount * anotherTokenRate.denominator * er.numerator
+
+	// when either er or anotherTokenRate are nil, we assume rate = 1/1 (as in cGLD)
+	var leftSide, rightSide *big.Int
+	if er == nil {
```
oneeman, line 172: The need for 3 cases here makes me miss the {1, 1} exchange rate

contract_comm/currency/currency_test.go
```diff
@@ -171,14 +170,61 @@ func TestCurrencyComparator(t *testing.T) {
...
+		g := NewGomegaWithT(t)
+
+		_, err := NewExchangeRate(common.Big0, common.Big1)
+		g.Expect(err).Should((HaveOccurred()))
+
+		_, err = NewExchangeRate(big.NewInt(-1), common.Big1)
+		g.Expect(err).Should((HaveOccurred()))
+	})
+
+	t.Run("can't create with denomniator <= 0", func(t *testing.T) {
```
oneeman, line 201:
```suggestion
	t.Run("can't create with denominator <= 0", func(t *testing.T) {
```

core/block_context.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/celo-org/celo-blockchain/common"
+	"github.com/celo-org/celo-blockchain/contract_comm/blockchain_parameters"
+	"github.com/celo-org/celo-blockchain/contract_comm/currency"
+	gpm "github.com/celo-org/celo-blockchain/contract_comm/gasprice_minimum"
+	"github.com/celo-org/celo-blockchain/core/types"
+	"github.com/celo-org/celo-blockchain/core/vm"
+)
+
+// BlockContext represents contextual information about the blockchain state
+// for a give block
```
oneeman, line 15:
```suggestion
// for a given block
```

contract_comm/currency/currency.go
nategraf: I personally don't have any issue with this implementation, meaningful `nil` values are a nice feature of Go. 

One option that would simplify things a bit though would be to create `ExchangeRate.Numerator()` and `ExchangeRate.Denominator()` functions that return `common.Big1` when `er == nil`. It would result in unnecessary multiplications and divisions by one, but that's not a huge issue.

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
-		return val, err
+// ToBase converts from token to base
+func (er *ExchangeRate) ToBase(tokenAmount *big.Int) *big.Int {
+	// check noop rate (cGLD rate)
+	if er == nil {
+		return new(big.Int).Set(tokenAmount)
 	}
-	return Convert(val, currencyFrom, nil)
+
+	return new(big.Int).Div(new(big.Int).Mul(tokenAmount, er.denominator), er.numerator)
```
nategraf, line 141: Note that changing the conversion logic in a way that would effect the way rounding errors are handled can be a hard fork if used consensus code. The two places I can think of that would be a risk are the cUSD validator rewards, which are converted to CELO and minted, and gas price minimum adherence where the gas value is converted to CELO before being compared to the GPM. Luckily in both cases, a non-CELO token is being converted to/from CELO, which makes this logic equivalent to the previous implementation. I'm not 100% certain there is not a place where this may cause issue. https://github.com/celo-org/celo-blockchain/blob/2dede5cce581dfe378f85c4c9d774ccd3ef99ac9/contract_comm/currency/currency.go#L148-L149

contract_comm/currency/currency.go
```diff
@@ -94,97 +94,127 @@ const (
...
-			return nil, err1
-		}
-		if err2 != nil {
-			return nil, err2
-		}
+	return new(big.Int).Div(new(big.Int).Mul(goldAmount, er.numerator), er.denominator)
+}
+
+func (er *ExchangeRate) CmpValues(amount *big.Int, anotherTokenAmount *big.Int, anotherTokenRate *ExchangeRate) int {
+	// if both rates are noop rate (cGLD rate), compare values
```
nategraf, line 155: s/cGLD/CELO
```suggestion
	// if both rates are noop rate (CELO rate), compare values
```

contract_comm/currency/currency.go
nategraf: One way to make this section simpler is to add an `ExchangeRate.Numerator()` function that returns `common.Big1` if `er == nil`. (I mentioned this in another comment as well)

contract_comm/currency/currency.go
```diff
@@ -194,14 +224,14 @@ func (cc *CurrencyComparator) getExchangeRate(currency *common.Address) (*exchan
 	return val, nil
 }
 
-func (cc *CurrencyComparator) Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
+func (cc *CurrencyManager) Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
 	// Short circuit if the fee currency is the same. nil currency => native currency
 	if (currency1 == nil && currency2 == nil) || (currency1 != nil && currency2 != nil && *currency1 == *currency2) {
 		return val1.Cmp(val2)
 	}
```
nategraf, line 231: It seems like this logic is duplicated in `ExchangeRate.CmpValues`. Consider removing the checks here as those will always be called.

contract_comm/currency/currency.go
```diff
@@ -216,128 +246,100 @@ func (cc *CurrencyComparator) Cmp(val1 *big.Int, currency1 *common.Address, val2
...
-	rightSide := new(big.Int).Mul(val2, new(big.Int).Mul(exchangeRate2.Denominator, exchangeRate1.Numerator))
-	return leftSide.Cmp(rightSide)
+	return exchangeRate1.CmpValues(val1, val2, exchangeRate2)
 }
 
-func Cmp(val1 *big.Int, currency1 *common.Address, val2 *big.Int, currency2 *common.Address) int {
-	// Short circuit if the fee currency is the same. nil currency => native currency
-	if (currency1 == nil && currency2 == nil) || (currency1 != nil && currency2 != nil && *currency1 == *currency2) {
-		return val1.Cmp(val2)
+func (cc *CurrencyManager) ToGold(amount *big.Int, currency *common.Address) (*big.Int, error) {
```
nategraf, line 252: s/Gold/Celo. (Or s/Gold/NativeToken)
```suggestion
func (cc *CurrencyManager) ToCelo(amount *big.Int, currency *common.Address) (*big.Int, error) {
```

contract_comm/currency/currency.go
```diff
@@ -216,128 +246,100 @@ func (cc *CurrencyComparator) Cmp(val1 *big.Int, currency1 *common.Address, val2
...
-		}
-		currency2Output := "nil"
-		if currency2 != nil {
-			currency2Output = currency2.Hex()
-		}
-		log.Warn("Error in retrieving exchange rate.  Will do comparison of two values without exchange rate conversion.", "currency1", currency1Output, "err1", err1, "currency2", currency2Output, "err2", err2)
-		return val1.Cmp(val2)
+	celoGoldAddress, err := contract_comm.GetRegisteredAddress(params.GoldTokenRegistryId, nil, nil)
+	if err == errors.ErrSmartContractNotDeployed || err == errors.ErrRegistryContractNotDeployed {
+		return nil, err
```
nategraf, line 267: This represents a change of behavior when comparing or converting currencies without using `ConvertToGold`, which was the only place where this was checked before. I'm concerned this could be a hard fork, and Its also very odd behavior because this will fail to get the CELO/CELO exchange rate when the registry is not deployed. We should likely put this check in the `ToGold` function and mark it as some unfortunate behavior to be removed at the next hardfork.

oneeman:
At the risk of scope creep, let me say I think this would be a great opportunity to bundle a change to not remove transactions which are below the gas price minimum.  We added it because it seemed to be part of tx pool correctness.  But, on second thought, it doesn't seem necessary and I think we'd be better off without it:

1. If the tx pool gets filled up, transactions get replaced based on gas price, so any transactions below the GPM will naturally get replaced as new transactions above the GPM come in.  As such, it doesn't raise any DoS possibility.
2. Otherwise, the transactions' being in the pool doesn't do any harm.
3. From the point of view of user experience, leaving them in is better since it allows transactions to survive temporary spikes in GPM (so long as they're not replaced due to the tx pool being full) and to get mined when the GPM goes back down
3. The miner already includes a GPM check in the state transition and breaks out of the loop when it gets to a tx below the GPM.  However, if we wanted to add redundancy, we could filter out any transactions below the GPM when getting the list of transactions to mine.

I would even ask whether we should allow adding new transactions below the GPM, but I feel less strongly about that.

Anyway, if we don't want to make such a change in this PR, we could do it as a follow-up.

contract_comm/currency/currency.go
```diff
@@ -216,128 +246,100 @@ func (cc *CurrencyComparator) Cmp(val1 *big.Int, currency1 *common.Address, val2
...
-				log.Error("medianRate invocation error", "feeCurrencyAddress", currencyAddress.Hex(), "leftoverGas", leftoverGas, "err", err)
-				return &exchangeRate{big.NewInt(1), big.NewInt(1)}, err
-			}
-		}
+	if err == errors.ErrSmartContractNotDeployed {
+		log.Warn("Registry address lookup failed", "err", err)
+		return NoopExchangeRate(), nil
+	} else if err != nil {
+		log.Error("medianRate invocation error", "feeCurrencyAddress", currencyAddress.Hex(), "leftoverGas", leftoverGas, "err", err)
+		return NoopExchangeRate(), nil
```
nategraf, line 287: Oof, this is not great default behavior. We really should not consider the exchange rate of all tokens to be equal to 1/1 when this call fails. We should instead disable any functionality that relies on this. Probably another change we should mark for the next hard fork.

nategraf (commented):
Some initial comments.

contract_comm/currency/currency.go
mcortesi: not really, this is used internally for testing. test define a mock `_getExchangeRate` function. 

That's why this function exists, and it's private

contract_comm/currency/currency.go
mcortesi: hehe... we are going to have this fight again...

To me `ExchangeRate`, `MustNewExchangeRate()` and `GetExchangeRate()` makes perfect sense to be public on a module that is about currencies and exchanges rates.

It's true, no other module is currently using it. But from a "internal module perspective", is IMO are perfect candidates for publishing.

I intentially wanted to expose ExchangeRate, and not expose `numerator` and `denmonitor` since i don't think the caller should be aware of that implementation detail

core/block_context.go
mcortesi: I did have that function first... but the usage pattern the get more cumbersome...

Let's say you have `IsWhiteListedFeeCurrency`, what signature will you use for `GetGasPriceMinimun()`?

If you you call `GetGasPriceMinimun()` on a currency that is not whitelisted. What do you do?

1. panic
2. return defaultValue
3. change signature so that `GetGasPriceMinimun()` returns also an error, and define the `NotWhiteListedError`

Neither of those sounds good to me. (1) and (2) are defintely out of the question... and 3 implies that now every caller needs to check an error that can only happen if the caller didn't first check `IsWhitelisted...`

So, i changed it to be a single function.

contract_comm/currency/currency.go
mcortesi: see my other comment

contract_comm/currency/currency.go
mcortesi: good point, i didn't consider that. definitely unfortunate behaviour, i moved a couple of layers down to simplify/normalize the logic, but can be dangerous

I suspect it's not possible not use feeCurrency == celogGoldAddress, but i'm not 100% sure... i'll see how i can leave it as before

contract_comm/currency/currency.go
mcortesi: i though of that; but i left it here because avoid doing the `getExchangeRate` call which might result on an EVM call. Maybe not that important now that they are cached

contract_comm/currency/currency.go
mcortesi: On some intermediate version of this, i did had `TokenValue{amount, rate}`; since as you said, seems cleaner.

But the caller of this function is the txpool, so, who starts with (amount, feeCurrencyAddress); so it require the txpool to first fetch the exchangeRate, then create a TokenValue, then do the same for the other tx you want to compare with, and finally call `tokenValue.Cmp(tokenValue)`.

In the err, i've err on making txpool life simpler.

Another thing i've tried is leaving the signature as it is, but implementing it with `TokenValue`; but seemed that i was just creating the type to make this comparison, and i was adding complexity just becuase, so i removed the whole thing.

mcortesi: On the comparison, you're right, but it's intented. I was trying to short circuit the case where they are actually the same instance.. and if not i do the normal comparison.

I though that if needed to do the actual Numerator and Denominator comparison; then i was not winning much. And thus i should just do the normal operation. 

But that was just an intuition, wdyt?

contract_comm/currency/currency.go
mcortesi: My intention was to leave the exact same logic here. Minus the case of one of the rate being the 1/1 on which case i elminated the `*1` & `/1` steps. Which i though (maybe wrong) would just save a few cpu cycles, but be idential in behaviour.

Where do you see i've chaning the logic?

consensus/istanbul/backend/pos.go
```diff
@@ -103,13 +103,15 @@ func (sb *Backend) distributeEpochRewards(header *types.Header, state *state.Sta
 		return err
 	}
 
-	// Validator rewards were paid in cUSD, convert that amount to cGLD and add it to the Reserve
-	totalValidatorRewardsConvertedToGold, err := currency.Convert(totalValidatorRewards, &stableTokenAddress, nil)
+	currencyManager := currency.NewManager(nil, nil)
```
nategraf, line 106: We should almost certainly pass in a header and state here (IIUC the parent's header and state), even though it was not passed in before. Currently this represents a race condition in consensus logic. (Style note, we might inline the creation of the currency manager to when it is called)

consensus/istanbul/backend/pos.go
```diff
@@ -103,13 +103,15 @@ func (sb *Backend) distributeEpochRewards(header *types.Header, state *state.Sta
...
+
+	// Validator rewards were paid in cUSD, convert that amount to CELO and add it to the Reserve
+	totalValidatorRewardsConvertedToCelo, err := currencyManager.ToCelo(totalValidatorRewards, &stableTokenAddress)
```
nategraf, line 109: Referencing https://github.com/celo-org/celo-blockchain/pull/1499#discussion_r612810987, I believe there is a hardfork here in the case that registry lookup fails for GoldToken contract

core/block_context.go
nategraf: I actually agree with this and was thinking approach 3 would be best. I reference this in another comment where the lack of an `IsWhiteListed` results in a behavior change. My main reason for wanting to reintroduce the `IsWhiteListed` is that with this interface it is currently impossible to tell when a currency _is_ whitelisted, but the GPM was not available for an unrelated reason. (e.g. no oracle reports)

core/block_context.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/celo-org/celo-blockchain/contract_comm/blockchain_parameters"
+	"github.com/celo-org/celo-blockchain/contract_comm/currency"
+	gpm "github.com/celo-org/celo-blockchain/contract_comm/gasprice_minimum"
+	"github.com/celo-org/celo-blockchain/core/types"
+	"github.com/celo-org/celo-blockchain/core/vm"
+)
+
+// BlockContext represents contextual information about the blockchain state
+// for a given block
+type BlockContext interface {
```
nategraf, line 16: Having the interface and implementation is a bit of an anti-pattern in Go. Combined with the fact that there is only one implementation, I would suggest only defining a `type BlockContext struct` here and defining an interface with the consuming code if needed. (This is the suggested approach [in Effective Go](https://golang.org/doc/effective_go#interfaces))

core/block_context.go
```diff
@@ -0,0 +1,87 @@
...
+func NewBlockContext(header *types.Header, state vm.StateDB) BlockContext {
+	gasForAlternativeCurrency := blockchain_parameters.GetIntrinsicGasForAlternativeFeeCurrency(header, state)
+
+	whitelistedCurrencies, err := currency.CurrencyWhitelist(header, state)
+	if err != nil {
+		whitelistedCurrencies = []common.Address{}
+	}
+
+	goldGasPriceMinimum, err := gpm.GetGasPriceMinimum(nil, header, state)
+	_ = err // Ignore the error since gpm.GetGasPriceMinimum returns the Fallback value on error
```
nategraf, line 49: I would suggest assigning to `_` directly.
```suggestion
	// Ignore the error since gpm.GetGasPriceMinimum returns the Fallback value on error
	goldGasPriceMinimum, _ := gpm.GetGasPriceMinimum(nil, header, state)
```

core/block_context.go
```diff
@@ -0,0 +1,87 @@
...
+
+	nonGoldGasPriceMinimums := make(map[common.Address]*big.Int, len(whitelistedCurrencies))
+	for _, currency := range whitelistedCurrencies {
+		gpm, err := gpm.GetGasPriceMinimum(&currency, header, state)
```
nategraf, line 53: It looks like this returns an error in fewer places than we might want and will result in accepting alternate currencies even if, for example, the gas price minimum contract is not deployed https://github.com/celo-org/celo-blockchain/blob/0f7f773a5dc9d34ba032f61d405011a95aa3856a/contract_comm/gasprice_minimum/gasprice_minimum.go#L125-L130

core/state_transition.go
```diff
@@ -409,7 +408,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	contractCreation := msg.To() == nil
 
 	// Calculate intrinsic gas, check clauses 5-6
-	gas, err := IntrinsicGas(st.data, contractCreation, st.evm.Header, st.state, msg.FeeCurrency(), istanbul)
+	gasForAlternativeCurrency := uint64(0)
+	if msg.FeeCurrency() != nil {
+		gasForAlternativeCurrency = blockchain_parameters.GetIntrinsicGasForAlternativeFeeCurrency(st.evm.Header, st.state)
+	}
```
nategraf, line 414: I didn't quite understand this conditional when I first read it. This comment would have helped me.
```suggestion
	// If the fee currency is nil, do not retrieve the intrinsic gas adjustment from the chain state, as it will not be used.
	if msg.FeeCurrency() != nil {
		gasForAlternativeCurrency = blockchain_parameters.GetIntrinsicGasForAlternativeFeeCurrency(st.evm.Header, st.state)
	}
```

core/tx_pool.go
```diff
@@ -622,14 +626,14 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		return ErrInvalidSender
 	}
 
-	// Ensure the fee currency is native or whitelisted.
-	if tx.FeeCurrency() != nil && !currency.IsWhitelisted(*tx.FeeCurrency(), nil, nil) {
+	gasPriceMinimum, isWhitelisted := pool.currentCtx.GetGasPriceMinimum(tx.FeeCurrency())
```
nategraf, line 629: It would be nice to differentiate between failures due to a currency not being whitelisted and the GPM for that currency not being available. I would recommend we add the `IsWhitelisted` function to block context to use here.

core/tx_pool.go
nategraf: This data sharing makes me a little nervous. In general I think its a bad idea to mutate a shared object. (In Rust, this would not be allowed at all.) It would be better is the tx lists did not hold a pointer to the context object and instead had it provided in the function arguments where needed.

light/txpool.go
```diff
@@ -396,7 +397,9 @@ func (pool *TxPool) validateTx(ctx context.Context, tx *types.Transaction) error
 
 	// Should supply enough intrinsic gas
 	header := pool.chain.GetHeaderByHash(pool.head)
-	gas, err := core.IntrinsicGas(tx.Data(), tx.To() == nil, header, currentState, tx.FeeCurrency(), pool.istanbul)
+
+	gasForAlternativeCurrency := blockchain_parameters.GetIntrinsicGasForAlternativeFeeCurrency(header, currentState)
```
nategraf, line 401: We should probably apply the optimization here where we only call `blockchain_parameters.GetIntrinsicGasForAlternativeFeeCurrency` if the fee currency is non-nil. Static calls a extra expensive from light clients due to requiring multiple network round trips.

miner/worker.go
```diff
@@ -322,10 +322,11 @@ func (w *worker) close() {
 }
 
 func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transaction) int {
-	currencyComparator := currency.NewComparator()
+	// TODO specify header & state
+	currencyManager := currency.NewManager(nil, nil)
```
nategraf, line 326: Which header and state should be used here? Id imagine the parent of `w.current` is the best candidate. If at all possible we should not merge without fixing this.

contract_comm/currency/currency.go
nategraf: What is changed is that when converting from one arbitrary currency to another (e.g. cUSD to cEUR), we would need to convert from source to the base (CELO) and then to the target currency, instead of converting from source to target in one operation. These are not guaranteed to get the same result due to a potential loss of precision in the conversion from source to base.

After having reviewed the whole PR, I realize there are indeed no places where we make a conversion from one non-CELO currency to another, so there are no places where the operation described above would need to be done. As a result, this change is fine.

core/tx_pool.go
mcortesi: yeah, agree. txpool on a few places (goroutines) requires this, also txlist and txpricedlist... so i'll see if instead of injecting the context i'll inject a context provider that has a lock

contract_comm/currency/currency.go
mcortesi: I've checked on master, and `ConvertToGold` is only used on `tx_list` to check which of two txs pays more.

I don't think goldtoken address is actually a valid feeCurrency, so i doubt this is ever needed. And in any case, this is a shortcut. So i've removed the whole thing

consensus/istanbul/backend/pos.go
mcortesi: correct... not only a race condition, it also implies that we use previous block exchange rates, not current block ones. I've already mentioned this on slack.. but changing this is a HF

consensus/istanbul/backend/pos.go
mcortesi: removed that logic

core/block_context.go
trianglesphere: I was thinking of #3 (switch the bool to error)

contract_comm/currency/currency.go
oneeman: Makes sense.  As far as when and how much to optimize, I don't know.  But if we want to short-circuit if the instances are the same, I think we could do a single check for `er == anotherTokenRate`.  I think this would be ok both as is and if we replace the `nil` rate with a {1, 1} rate.

core/tx_pool.go
mcortesi: Check https://github.com/celo-org/celo-blockchain/pull/1499/commits/e7aa53d5ed903e53e47c39ca76b050f8e10508cc where i'm using `atomic.Value` to solve the concurrency problem.

contract_comm/currency/currency.go
mcortesi: removed `nil` as a valid value for exchange rate

contract_comm/currency/currency.go
mcortesi: please unresolved if you still think we should remove this

contract_comm/currency/currency.go
mcortesi: Removed this features, since `ConvertToGold` was only called on `txpool` and using a `feeCurrency=goldTokenAddress` is not (or shoul not)  be supported

core/block_context.go
mcortesi: done

consensus/istanbul/backend/pos.go
mcortesi: I'll leave as it is, since this is currently master branch behvaior, but i've added an entry here https://www.notion.so/clabsco/Remove-quirks-list-HardFork-Changes-7ee3a0b904ea464db6af45aba9f94a4a for us to remember changing it

contract_comm/currency/currency.go
mcortesi: Added to the list https://www.notion.so/clabsco/Remove-quirks-list-HardFork-Changes-7ee3a0b904ea464db6af45aba9f94a4a

nategraf (approved):
Looks good to me. I agree with @trianglesphere's comments about the value comparison function signatures, but don't personally consider it a blocker, so I'll give my ✅  &nbsp;now.

core/tx_pool.go
mcortesi: removed gpm check; kept only isWhitelisted

core/block_context.go
mcortesi: removed gpm check; kept only isWhitelisted

PR #1500, opened 2021-04-12
Don't check GPM before mining tx (instead, handle error from state transition if it happens)

oneeman:
### Description

This is an optimization for block generation.  Previously, for each transaction we want to add, we check that its gas price meets the gas price minimum before applying the transaction.  However, this is unnecessary, because applying the transaction includes a GPM check as well (inside the state transition).  This PR removes the redundant check and instead handles the error that is returned by the state transition if the transaction's gas price is below the GPM.  If the error appears, it breaks out of the loop (which is the same as what we were previously doing if the gas price was below the GPM).

### Other Changes

* In the state transition, if the transaction's gas price doesn't meet the GPM, it will now log that at the debug level rather than the error level.

### Tested

* Benchmarking with a single validator and with 100 validators confirms this results in a noticeable improvement in block generation time
* Have not tested `core.ErrGasPriceDoesNotExceedMinimum` actually being returned.  This will not normally happened, because the transaction pool does not admit transactions below the GPM, and if the GPM increases it removes transactions that don't meet the GPM.  As such, this code path (and also the original code path breaking out of the loop due to GPM) are unlikely to ever occur (they would only happen if the gas price minimum increases beyond some pending transaction's gas price and the miner gets the list of transactions before the transaction pool has had time to remove the transaction).
* However, the code path in the state transition which rejects transactions due to being below the GPM has previously been confirmed to work.

### Backwards compatibility

No incompatibilities.  This would result in the same transactions being included in the proposed block as before.


trianglesphere (approved):
lgtm

miner/worker.go
```diff
@@ -778,6 +768,11 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, txFe
 			log.Trace("Skipping account with hight nonce", "sender", from, "nonce", tx.Nonce())
 			txs.Pop()
 
+		case core.ErrGasPriceDoesNotExceedMinimum:
+			// We are below the GPM, so we can stop (the rest of the transactions will either have
+			// even lower gas price or won't be mineable yet due to their nonce)
```
nategraf, line 773: It might be useful to have a `Trace` log here
```suggestion
			// even lower gas price or won't be mineable yet due to their nonce)
			log.Trace("Skipping remaining transaction below the gas price minimum")
```

nategraf (approved):

miner/worker.go
oneeman: Sounds good.  The gas price and GPM will have been logged as debug already (see state_transition.go change in this PR), so a message by itself here works.

oneeman:
Linter caught an error I had seen previously but missed when doing this PR: that the break is breaking out of the case and not the loop.  Will fix in an additional commit.

PR #1504, opened 2021-04-12
[DONT MERGE YET] Upstream merge up to 1.9.15

hbandura:
First big conflict commit from upstream: b4a2681
https://github.com/ethereum/go-ethereum/pull/20758
Completely rewrites the les server pool.
-- Currently accepted upstream version and DID NOT YET reimplement our changes on top of it:
-- --  The server pool entries RPC (is this still relevant?) from https://github.com/celo-org/celo-blockchain/pull/1132
-- --  other changes (unsure if they still make sense with the rewrite)

upgraded golangci-lint to 1.27 (was 1.24 upstream, 1.25 ours)

EIP-2315
https://github.com/ethereum/go-ethereum/pull/20619

Adds p2p wirev4 (we already had it)

Adds crypto/bls12381 (we already had it), but left our versions. Another commit made a change on it and since there was no conflict I left it there, needs review.


trianglesphere:
Re the mobile discovery improvements, have we checked that we no longer need our changes for valora?

crypto/bls12381/arithmetic_x86_adx.go
```diff
@@ -0,0 +1,24 @@
...
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// +build amd64,blsadx
+
+package bls12381
+
+// enableADX is true if the ADX/BMI2 instruction set was requested for the BLS
+// implementation. The system may still fall back to plain ASM if the necessary
+// instructions are unavailable on the CPU.
+const enableADX = true
```
trianglesphere, line 24: Do we use `enableADX` anywhere?

trianglesphere:
I've taken a look at everything but the LES changes. I found one thing where we may have brought in a little too much of the upstream bls changes, but I'm not really familiar with that code.

My only ask is that we follow up on making sure that the LES changes will work with valora and then I think this is good to go.

crypto/bls12381/arithmetic_x86_adx.go
hbandura: Not according to grep, no

hbandura:
> Re the mobile discovery improvements, have we checked that we no longer need our changes for valora?

I checked on the monorepo but will check in some other repos as well

core/vm/jump_table.go
```diff
@@ -62,14 +62,6 @@ var (
 // JumpTable contains the EVM opcodes supported at a given fork.
 type JumpTable [256]operation
 
-func newYoloV1InstructionSet() JumpTable {
```
gastonponti: This is because is not used and the lint failed?

accounts/keystore/keystore.go
```diff
@@ -51,6 +51,10 @@ var (
 	ErrLocked  = accounts.NewAuthNeededError("password or unlock")
 	ErrNoMatch = errors.New("no key for given address or file")
 	ErrDecrypt = errors.New("could not decrypt key with given password")
+
+	// ErrAccountAlreadyExists is returned if an account attempted to import is
+	// already present in the keystore.
+	ErrAccountAlreadyExists = errors.New("account alreaady exists")
```
gastonponti, line 57: the typo guy missed the doble `a` alreaaaaaaady

cmd/utils/flags.go
```diff
@@ -1735,13 +1735,13 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 			cfg.NetworkId = params.BaklavaNetworkId
 		}
 		cfg.Genesis = core.DefaultBaklavaGenesisBlock()
-		setDNSDiscoveryDefaults(cfg, params.KnownDNSNetworks[params.BaklavaGenesisHash])
```
gastonponti, line 1738: mmm, is this right? seems to be different things (same for the others)

gastonponti: Nevermind

core/vm/eips.go
```diff
@@ -32,6 +32,8 @@ func EnableEIP(eipNum int, jt *JumpTable) error {
 		enable1884(jt)
 	case 1344:
 		enable1344(jt)
+	case 2315:
```
gastonponti, line 35: Should we have this as an option? Does this work for us?

core/vm/jump_table.go
hbandura: exactly, plus the whole Yolo thing is a test network for ethereum

accounts/keystore/keystore.go
hbandura: haha nice catch. do you mind if I fix this in the 1.9.16 branch so We don't re run the tests for this?

core/vm/eips.go
hbandura: Do you know of anyone who con led some light into this?

accounts/keystore/keystore.go
gastonponti: Yes, of course. I just was hating the typos guy for the big PR and tried to point it out.

core/vm/eips.go
hbandura: @mcortesi

trianglesphere:
> I checked on the monorepo but will check in some other repos as well

Awesome. It seems like we're probably fine to delete this. I'd like to double check with the wallet team, but it seems like we're probably fine to go ahead with the removal.

core/vm/eips.go
mcortesi: missing the context here?
So there's an EIP 2315, but we don't have an cli option to enable it? Or is the question wether we should ahve the code for 2315?

To me, we should add all code from EIPs since probably we will have them enabled on a future HF. And they should be disable for now, which probalby is already true since they are enabled using a chainConfig flag.

hbandura: I agree, this is just the code, I believe it wasn't activated yet

gastonponti (approved):
LGTM

Issue #1350, opened 2021-02-08
Clean up announce protocol with MP and HS


trianglesphere:
## Description

1. Clean up messages sent by replicas in `proxied_validator_engine.go:ThreadRun`. This is sending old valEnodeTable messages
2. Explore (really test) disabling querying for replicas. They should get a populated `valEnodeTable` via `announce.go:updateAnnounceVersionTickerCh`

## Context

We're getting errors related to 1, and 2 might be causing problems with MP and HS with replicas being out of date of the primary. Disabling sending any messages as the replica would be nice as well.

PR #1507, opened 2021-04-13
Stop replicas from sending enode certs to proxies

trianglesphere:
### Description

Stop replicas from sending enode certs to proxies. `proxiedValidatorEngine.sendValEnodesShareMsg`
did check if the node was the primary, but `sendEnodeCerts` did not. This matches which messages
are sent as a replica in `announce.go`.

This will stop the error `invalid enode certificate message map because of old version`
from occurring on proxies with replicas.

### Tested

No

### Related issues

- Partial Fix of  #1350

### Backwards compatibility

Yes


nategraf (approved):
Functionally I think this is fine, but is also a bit deeper in the abstraction layers than feels right to me. (e.g. It is a bit surprising that a call to `sendEnodeCerts` can result in a noop) It might be more appropriate to move that check up the stack to here: https://github.com/celo-org/celo-blockchain/blob/fc461fbd9feba2949d83f493c4248ba0d0786425/consensus/istanbul/backend/announce.go#L1197-L1199

(I also see that this function is called from the following code that would need to be revised.)
https://github.com/celo-org/celo-blockchain/blob/fc461fbd9feba2949d83f493c4248ba0d0786425/consensus/istanbul/proxy/proxied_validator_engine.go#L515-L516

If you've considered this option and decided it is not right, I'm happy to have this merged. I'll approve it and let you chose whether or not to move the check.

mcortesi (approved):

PR #1508, opened 2021-04-13
Only active validator signs stats

gastonponti:
### Description

A proxy with two validators (one active) that send a stat to be signed by the validator, was receiving 2 signed messages for the same (one per validator)

Only the active validator will sign those stats.

### Other changes

The validators MUST have the celostats flag too (as we have it in the docs)

### Tested

Manually

### Backwards compatibility

Breaking change in that proxied validators need to specify the celostats flag in order to report to celostats.  However, this new behavior agrees with our documentation, so this is really more of a bugfix.


ethstats/ethstats.go
```diff
@@ -423,18 +426,18 @@ func (s *Service) login(conn *connWrapper, sendCh chan *StatsPayload) error {
...
-			return errors.New("Could not resolve NodeInfo")
+			return errors.New("could not resolve NodeInfo")
 		}
 		network = fmt.Sprintf("%d", ethInfo.Network)
 		protocol = fmt.Sprintf("%s/%d", istanbul.ProtocolName, istanbul.ProtocolVersions[0])
 	} else {
 		lesProtocol, ok := infos.Protocols["les"]
 		if !ok {
-			return errors.New("No less protocol found")
+			return errors.New("no less protocol found")
```
oneeman, line 436:
```suggestion
			return errors.New("no LES protocol found")
```
(while we're touching this line, may as well fix the "less" and capitalize it

oneeman (approved):
Looks good to me!  I made a suggested change in one of the error messages, and edited the "backwards compatibility" section to call out the breaking change also described in "Other changes".

nategraf (approved):
:shipit:

PR #1513, opened 2021-04-15
Refactors contract call layer

mcortesi:
### Description

**Companion PR** https://github.com/celo-org/celo-blockchain/pull/1522 which rewrites all contract calls to this new layer

This PR completely rewrites the low level contract API. This is done in a backwards compatible manner, since the current contract api is maintained, and rewritten so that it uses the new api. Thus, all users of the contract API are not required to migrate immediately, and we can do that migration in steps.

#### Low Level Contract API

Rewrite the low level contract API, by defining the following types

1) `EVMRunner` interface to do write/read calls to the state using an EVM

```go
// EVMRunner provides a simplified API to run EVM calls
// EVM's sender, gasPrice, txFeeRecipient and state are set by the runner on each call
// This object can be re-used many times in contrast to the EVM's single use behaviour.
type EVMRunner interface {
	// Execute performs a potentially write operation over the runner's state
	// It can be seen as a message (input,value) from sender to recipient that returns `ret`
	Execute(recipient common.Address, input []byte, gas uint64, value *big.Int) (ret []byte, leftOverGas uint64, err error)

	// Query performs a read operation over the runner's state
	// It can be seen as a message (input,value) from sender to recipient that returns `ret`
	Query(recipient common.Address, input []byte, gas uint64) (ret []byte, leftOverGas uint64, err error)

	// StopGasMetering backward compatibility method to stop gas metering
	// Deprecated. DO NOT USE
	StopGasMetering()

	// StartGasMetering backward compatibility method to start gas metering
	// Deprecated. DO NOT USE
	StartGasMetering()

	// GetStateDB retrieves current stateDB within the runner
	GetStateDB() StateDB
}
```

 * Provides a way to make multiple calls using the same object. the EVM object should not be reused!
 * It's an entry point to implement low level cache to EVM calls, since it requires only implementing a Decorator object that first checks the cache, and on miss call the decoratee.
 * `SharedEVMRunner` implementation is provided since we DO reuse the evm, and changing that would be a HF, so we need to that later
 * Exposes gas metering methods also as a backward compatible measure that we need to remove later

2) To create `EVMRunner` we expose:
```go
func NewEVMRunner(chain ExtendedChainContext, header *types.Header, state vm.StateDB) vm.EVMRunner

func GetSystemEVMRunnerFactory(chain ExtendedChainContext) func(header *types.Header, state vm.StateDB) (vm.EVMRunner, error)
```

The second one, is for backward compatibility with contract_comm and will be removed on a later PR

On https://github.com/celo-org/celo-blockchain/pull/1522 we introduce utility factories on `Blockchain` and `Lightchain` who already provide the `ExtendedChainContext` and thus only the header and state are needed.

3) On `contracts` module we expose `Method` and `BoundMethod`. The first represents a method can be called once it's binded to an address with `Bind(common.address)`.

The `BoundMethod` exposes:
* `Query(vmRunner vm.EVMRunner, result interface{}, args ...interface{}) error`: this runs the method a a query (read-only) over the given runner and parameters
* `Execute(vmRunner vm.EVMRunner, result interface{}, value *big.Int, args ...interface{}) error`: this methods runs a write operation over the given runner and parameters

Nice property about this object is that can be statically created and reused. So work as configuration for method call.

#### ChainContext and ExtendedChainContext

In the past we've used & abused `ChainContext` as an object that is available, and so we added methods to it for different objects to access it.

I would like to revert that decision, and have minimal changes to the ChainContext. I've created `ExtendedChainContext` (inherits from ChainContext) as the interface needed to create an `EVMRunner`. So that methods that are only needed for that object belong there, and not to a commonly used interface as `ChainContext`

#### Changes on Error logging

On master currently logging on error happens 3 times.

The typical call stack is:
  1. `makeCallWithContractId()`
  2. `makeCallFromSytem()`
  3. StaticCallFromSystem or CallFromSystem (that both call `handleAbiCall()`

On error we end up logging 3 TIMES!

  1. First [contracts/utils.go:handleAbiCall](https://github.com/celo-org/celo-blockchain/blob/master/contracts/utils.go#L49-L88) will log using `trance` or `error` level depending on `methodName` or type of error.
  2. On [makeCallFromSystem()](https://github.com/celo-org/celo-blockchain/blob/master/contract_comm/evm.go#L117-L120) we log with error level
  3. On [makeWallWithContractId()](https://github.com/celo-org/celo-blockchain/blob/master/contract_comm/evm.go#L117-L120) we log again with error level
  
This PR changes that. We know **only log errors** on `Call.Run()`. And it's always on `error` level.

#### Current Contract API Changes

All function on `contracts_comm/evm.go` are now deprecated. They external interface for call function was left unchanged, still they are now implemented using the new contract API.

### Tested

* CI & unit tests

### Backwards compatibility

Yes


contract_comm/evm.go
```diff
@@ -18,139 +18,76 @@ package contract_comm
...
-func MakeStaticCall(registryId [32]byte, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, header *types.Header, state vm.StateDB) (uint64, error) {
-	return makeCallWithContractId(registryId, abi, funcName, args, returnObj, gas, nil, header, state, true)
-}
-
-func MakeCall(registryId [32]byte, abi abi.ABI, funcName string, args []interface{}, returnObj interface{}, gas uint64, value *big.Int, header *types.Header, state vm.StateDB, finaliseState bool) (uint64, error) {
-	gasLeft, err := makeCallWithContractId(registryId, abi, funcName, args, returnObj, gas, value, header, state, false)
-	if err == nil && finaliseState {
-		state.Finalise(true)
+func SetEVMCallerFactory(_evmCallerFactory vm.EVMCallerFactory) {
+	if evmCallerFactory == nil {
```
oneeman, line 36: Feels funny for it to no-op silently if it's called again with a different factory.  I can think of two alternatives: (a) returning an error or panicking if it's called again, or (b) overwriting.

I guess we have no need for (b) yet, but I could imagine it maybe coming in handy during tests.  And, at first sight, I don't see a big problem with overwriting.

piersy: I'd be in favour of panicking, this is clearly not something that should be called more than once in production code. There is another problem here though which is that this lacks synchronisation and there doesn't seem to be any guarantee that this would only be called from a single go-routine.

contracts/contract_call.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/celo-org/celo-blockchain/common/hexutil"
+	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/log"
+)
+
+// Call represent a runnable call on the EVM
+type Call interface {
+	// Run will execute the call using the caller.
+	// Execution result will be populated into result
+	Run(caller vm.EVMCaller, result interface{}) (leftoverGas uint64, err error)
```
piersy, line 17: I'm not convinced about this design. Where previously we would call a method on a VM instance we now:

Construct Call object -> Call Run on call object with EVMCaller as param -> Run then calls Method on EVMCaller -> EVMCaller then calls method on VM instance.

This seems like a lot of redirection, and to me makes understanding the path of calls to the VM more difficult to follow.

The original problem that I think we were trying to solve here was that we were using VM instances more than once, looking at the history of this #938 it looks like people were well aware of the fact that they were re-using VM instances, Its not clear whether they realised they should not be doing this though.

The approach taken here is to use the EVMCaller construct which should then be used everywhere and always ensures a new instance of the VM is used for each call. This then allows us to have calls that internally make multiple calls.

I feel like this could also be solved by just making it easy to create a new VM instance and ensuring that we use a new instance per call. We could do this by funnelling our calls to the VM through a higher level function that ensures that we use a new VM instance per call (for example something like the Run method here that takes a vm factory in place of the EVMCaller) . This would reduce the hierarchy of calls to get to the evm (just, we call Run -> Run calls VM).

I prefer the approach of making VM instantiation easy, as opposed to the EVMCaller because I think, beyond reducing the call heirarchy depth to get to the VM, a factory is a more flexible component that is less likely to need to be changed as the code changes.

Both a factory and EVMCaller will need to be changed when the required parametrs to construct a VM change (hopefully this should not be too often, but we don't really have any control of this because of upstream).

Where the two designs differ is that if we ever need access to any of the 12 other public methods of the vm we need to then either expose them through the EVMCaller interface or at that point create a factory to give ourselves a vm instance. Its not even clear to me how you would correctly expose those methods on an instance of the EVMCaller, for example, for the  the `Cancelled` method we would then need to store old evm instances in the EVMCaller to be able to retrieve the result of `Cancelled` but if the EVMCaller is allowing multiple internal EVMs to be constructed in its lifetime how do we expose `Cancelled` correctly?

Right now given that it looks like our contract calling code is going to change considerably I think providing a factory and handling evm instances on a case by case basis will probably introduce less friction. Once we have clearer idea of how this code is going to end up we may then see that an EVMCaller is a useful component.

I've made these changes in another branch, see [here](https://github.com/celo-org/celo-blockchain/compare/mc/evm-caller..731425bc2f935729d1bf2d0ce0f88a4b053d1292#diff-f64965fa055835e4285f91971de4d4aec7c1a4a4c47aedb8e4c58761517115cdR69-R79) for how calling methods have changed.  Essentially there is no EVMCaller any more just a factory and provider, and we don't have the concept of "multi call" calls instead callers pass a new evm instance to each `Call.Run` invocation and `Call.Run` invocations only make one call. These changes resulted in the dropping of about 150 lines of code which I think is a significant enough change to point to the fact that this is a simplification.

contracts/contract_call.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// copy call template
+	call := rc.callTemplate
+	call.to = contractAddress
+
+	return call.Run(caller, result)
+}
+
+// contractCall represents a Call to a contract
+type contractCall struct {
```
piersy, line 119: I don't think that the introduction of `contractCall` is helping much. It seems to be a somewhat arbitrary abstraction.

If we look at the usage of `Call.Run` vs what was previously ocurring we see:
```go
	_, err := StaticCallFromSystem(evm, params.RegistrySmartContractAddress, getAddressForFuncABI, "getAddressFor", []interface{}{registryId}, &contractAddress, params.MaxGasForGetAddressFor) // old
	_, err := QueryCallFromVM(params.RegistrySmartContractAddress, params.MaxGasForGetAddressFor, NewMessage(&getAddressForFuncABI, "getAddressFor", registryId)).Run(caller, &contractAddress) // new
```

The user needs to provide the same parameters but they now need separate them between 2 function calls, the constructor and Run.

This would potentially be helpful if `contractCall` instances needed to be constructed in a separate location to where run was called or if `contractCall` instances needed to be reused. But in all cases in this PR `contractCall` instances are constructed at the same time as they are used and then discarded, like the code shown above. If all instances of `contractCall` are used immediately after construction then we don't derive any benefit from the introduction of a struct. We may as well just call a function with all the parameters as we did before.

In order to maximise the value of a struct we should try to cluster its fields such that they are likely to change together and such that a single instance of a struct can be reused.

Looking at the fields of `contractCall` we can see that they do not change together.
```go
	readOnly bool // changes with the method and abi which change on redeployments.
	from     common.Address // Always the same, the VMAddress (could be dropped entirely) 
	to       common.Address // Since this is a contract call this changes only with redeployments.
	msg      Message  // (abi, method and args) abi, and method change with redeployments, args will likely change on every call.
	value    *big.Int // Likely to change on every call.
	maxGas   uint64 // A value hardcoded in the code, could change on redeployments or code changes.
```

If we instead clustered the fields of this struct such that all struct fields are likely to change in response to redeployments or code changes then we would be able to construct a struct at node startup and reuse it for the lifetime of a node.

Such a struct would have the following fields, with the args and value being provided at runtime.
```go
        readOnly
	address common.Address
	method  string
	abi     *abi.ABI
	maxGas  uint64
```

I've had a go at this [here](https://github.com/celo-org/celo-blockchain/compare/731425bc2f935729d1bf2d0ce0f88a4b053d1292..piersy/evm-refactor-2) its also a bit leaner than its predecessor, dropping ~30 LOC, but more importantly setting up a framework where we can reuse instances for the lifetime of a node.

contracts/contract_call.go
mcortesi: I don't agree :) 

To me, sending an evm vs sending an evm caller implies that:

* **sending evm:** we rely on a social contract for developers not to missuse the evm to make contract calls
* **sending evm caller**: we rely on code to ensure the contract

I will always favor code. I don't think it adds a ton of complexity, nor it makes it unreadable. 

I'm happy to discuss the other aspects, maybe having a `Call` interface doesn't make sense. Maybe having 2 types of calls doesn't make sense... i'm not bullish about that, i was trying to make calls more "clear" (vs sending 9 positional arguments to a function); but surely other was can be found to do the same thing.

About the topic of "accessing other evm methods". Please check the `vm/runtime` package. It states 
```
// Package runtime provides a basic execution model for executing EVM code.
```

It exposes `runtime.NewEnv() runtinme.Config` and then methods `Call`, `Create`, `Execute` that receve the config and will create an EVM, do a single call, and then discard that.

What i'm doing is basically that. Maybe `EVMCaller` is a bad name, since gives you the sense that let's you access to the full EVM, and maybe instead should be called `ContractCaller` and hide the fact that an EVM is inside. To a contract you either want to do queries or exeuctions. The rest is not important.

contracts/contract_call.go
mcortesi: For someone that has been doing golang for so many years, you surely give # of LOC a lot of importance! Try haskell! generics, and all that stuff. Golang doesn't very well suited to give you a lot of expressive power in a short number of LOC! So i don't really measure it by LOC, but readability (which of course is subjective)

I like the argument of being able to reuse the struct and i like the `ContractMethod` struct. I'm not very fond of ending with a 9 positional arguments `call()` after that. That's what i was trying to avoid in the first place. Quite a well known [code smell](https://sourcemaking.com/refactoring/smells/long-parameter-list). 

On other languages, to solve it, you can try to split the function in parts, so to avoid doing ALL in one place.  Have smaller function, encode different parts in different functions. That's why i created the `Message` struct, to try to encode the abi part of it, and also because using "vargs" as opposed to a `interface{}[]` seemed more natural to me, and provides a better UX (at least in my opinion, again ux can be subjective).

Another strategy to reduce the parameter bloat, if to have property objects. You see a lot of that in javascript/typescript. Or if the language supports it "named arguments". The `ContractCall` object it's a parameter object in disguise; it provides names for all the things you need to pass, but  instead of passing a Call object to a function,  you create a call and then call `Run()`.

That also has a secondary benefit. I'm able to add new behaviour in "types of call" without modyfing the interface; nor adding conditionals to the code (and longer functions). That's ["replacing conditional for polymorphism"](https://refactoring.guru/replace-conditional-with-polymorphism); so let's the caller worry less how it works, and just do a call.

Finally, the idea of have a reusable call did cross my mind. But i didn't find much reusable parameters. Surely `(abi, method, maxGas)` seems to be reusable. `ContractAddress` comes from the registry is can change. Even `maxGas` is an argument for some calls (check `getBalance`). So i didn't end up with nothing too convincing.

In any case, even if no reuse, creating a object just to make a call and consume it; it's nothing an unusual pattern in my experience. Actually `Call` interface can be seen as an implementation of [Command design pattern](https://www.gofpatterns.com/behavioral/patterns/command-pattern.php) from the classic GOF book.

mcortesi: I've tried to get the "Best of both worlds" and create something that can be reused. Check this commit https://github.com/celo-org/celo-blockchain/pull/1513/commits/da5f69432532dadefeef92738ed53476320396e1

Basically i've tranformed `Message` into `Method` which contains (abi, method, maxGas)

And then created `BoundedMethod` which also contains the contract address (in fact the ability to resolve the address).

The benefit from this approach is that now i can initialize at a module level `BoundMethods` which can greatly simply the code of calling them.

For example:
```
var getWhitelistMethod = contracts.NewRegistryContractMethod(params.FeeCurrencyWhitelistRegistryId, &getWhitelistFuncABI, "getWhitelist", params.MaxGasForGetWhiteList)

func CurrencyWhitelist(caller vm.EVMCaller) ([]common.Address, error) {
	returnList := []common.Address{}
	_, err := getWhitelistMethod.VMQuery(caller, &returnList)
       return returnList, err
}
```


Check the demo here https://github.com/celo-org/celo-blockchain/pull/1522

piersy: Hey @mcortesi Sure go is verbose, but when I consider LOC its not really the absolute value of LOC I'm interested in, its the change. Did LOCs go up? If so the code likely got more complex, and conversely did they go down, its likely that things got simpler. Its a language agnostic measure.

You are right about `call`, it had a lot of parameters, but it was also a private function of that package, so users of the package would not need to interact with it and it was being called in only 2 places. So although I generally prefer less parameters, if usage of functions with large numbers of parameters is contained within a small space I don't see it as a big problem.

I could have made it a function on `ContractCaller` which would save 4 params, but given that I'd already written the code calling `call` it wouldn't have helped me, and I don't think that readability would be changed by doing that (If anything I think readability would be slightly degraded by converting it to a method because the reader would then have to refer back to the struct definition to know what the types of the fields are vs just looking at the parameter definition).

I don't think how the `call` function is presented (as a method or as a pure function) really has much of an impact on the code, I wouldn't have a preference either way. The only thing that would push me one way or the other would be these questions:

* Do we need the functionality of `call` in another struct? If yes then we can make it a pure function and share it between structs, or duplicate the logic between structs, I would be in favour of having a pure function.
* Do we make lots of calls to `call` within a single struct? In this case it looks like it makes sense to make it a method and save repeatedly passing the parameters.

In my branch it was too early to say which of the above situations would be the case.

Sometimes lots of parameters is a code smell but I would generally try to asses this on a case by case basis, using LOC to guide me to an optimal solution, which may end up having a function with lots of params.

Would you consider the new `run` function to still exhibit a code smell? According to the [article](https://sourcemaking.com/refactoring/smells/long-parameter-list) you linked it still exhibits a code smell because it has:
> More than three or four parameters

It has 6. I don't consider it a code smell, but I didn't consider it a code smell with 9. The fact is you need 5 parameters just to call the EVM, does that make all of our code that calls the EVM bad? I don't think so. It could be argued that the EVM is badly designed and that we should refactor it, but sometimes you are doing something complicated and you just need lots of parameters. In those cases I think large numbers of parameters can be justified, there are of course cases where functions can be nicely refactored to remove the large parameter requirements and its good for the code, but I would be wary of trying to force the parameter count down by any means necessary.

With regards to parameter objects sometimes they can help, but if they are being used to just put stuff inside and then get it back out its not really helping, I would say they are hiding the problem.

E.G.

```go
// Callsite
printAll("a", "b", "c", "d", "e", "f")

// Implementation
func printAll(s1, s2, s3, s4, s5, s6) {
    println(s1)
    println(s2)
....
}


```

So we can fix this with a parameter object, you can probably guess :) but just for the purposes of comparison.

```go

// Callsite
printAll(ss{"a", "b", "c", "d", "e", "f"})

// Implementation
func printAll(ss params) {
    println(params.s1)
    println(params.s2)
....
}
```

This solves the "code smell" we are down to one parameter, but the reality is that the function became a little bit harder to call and a little bit more complex in the implementation and we now have the additional code for the struct which is not shown.

The `ContractCall` implementation as it was originally, was like this parameter object example above. It just packed up a bunch of parameters together and then the `Run` method needed to unpack them.

Where previously the caller would write:
```go
return contract_comm.MakeStaticCall(registryId, abi, method, args, returnObj, gas, header, state)
```
The introduction of `ContractCall` means the caller would have to write this:
```go
caller, err := getCaller(header, state)
if err != nil {
	return 0, err
}

return contracts.QueryCallOnRegisteredContract(registryId, gas, contracts.NewMessage(&abi, method, args...)).Run(caller, returnObj)
```

Even if we discount the code required to get the caller, I think we could agree that it's clearly not easier than what went before, its just different. Parameters have been bundled up in a different way, technically it reduced a "code smell", but actually the underlying problem of the code smell remains, which is that the caller when making a contract call is faced with collating a large bunch of parameters, and that is a complex operation.

What's not shown above is the cost of introducing the `ContractCall` which is about 100 LOC, and this in general does increase complexity, its more lines for people to read and understand before they can use the code, its more places that bugs can hide.

If the construction of the `ContractCall` and `Message` objects were to happen elsewhere such that callers only need to call `Run` then that would likely be an improvement because it would have really reduced the burden on the caller to collate so many parameters at the point of making a call. But because the contract call object includes the `args` its seems unlikely to me that you would be able to construct the contract call elsewhere because args are likely to be unique for each call.

You mention ["replacing conditional for polymorphism"](https://refactoring.guru/replace-conditional-with-polymorphism) I'd agree that your design makes that easier, but there aren't any conditionals currently, and its not clear that there would be, so I would consider introducing `ContractCall` now for this reason would be an instance of premature abstraction.

You also mention that Call interface can be seen as an implementation of the Command design pattern. But if you look at the description in the [link](https://www.gofpatterns.com/behavioral/patterns/command-pattern.php) they say.

> The Command pattern encapsulates a request in an object, which enables you to store the command, pass the command to a method, and return the command like any other object.

`ContractCall` was not being passed, stored or returned so from the point of view of GOF it was a needless application of the pattern.

So to make `ContractCall` a useful addition I think it needs to be something that can be re-used/passed and I think this is what your latest changes are doing with `Method` and `BoundMethod`.

contracts/contract_call.go
piersy: >To me, sending an evm vs sending an evm caller implies that:
>
>sending evm: we rely on a social contract for developers not to missuse the evm to make contract calls
>sending evm caller: we rely on code to ensure the contract

Maybe I didn't explain myself very well, I too generally favour code to enforce things.

What I was suggesting is that we could enforce the use of the EVM at a different level.

So currently in this PR the EVM use is enforced at the following level.

BoundMethod.Execute -> evmCaller.Execute (vm non reuse is enforced here)

I was suggesting that we could instead enforce this in BoundMethod.Execute, by having it and its sibling methods accept a VM factory or provider in the place of the caller.

I don't think this introduces a reliance on a social contract, in both cases users will call BoundMethod.XXX and the contract call package handles enforcing correct vm usage. In either case there is still a reliance on social contract, which is that users call BoundMethod.XXX, if you are not calling BoundMethod.XXX then you can do what you like, neither approach enforces anything at that point.

The contract_comm api would hardly change, what I'm suggesting is
```go
	m.VMExecute(provider, returnObj, value, args...)
	// as opposed to 
	m.VMExecute(caller, returnObj, value, args...)
```

I'm not clear about what your point is with the runtime package. Maybe I'm missing some context but I just wonder if we can be sure that we wont need to access other VM methods?

The question of needing other methods aside, If both implementations are providing the same level of enforcement over the correct use of the VM why not choose the solution that has 150 less LOC?

trianglesphere:
Looking ahead to the external API change: I see that the external contract comm API is changed to
```
// Old
DistributeEpochRewards(header *types.Header, state vm.StateDB, groups []common.Address, maxTotalRewards *big.Int, uptimes map[common.Address][]*big.Int) (*big.Int, error)
// New
DistributeEpochRewards(caller vm.EVMCaller, groups []common.Address, maxTotalRewards *big.Int, uptimes map[common.Address][]*big.Int) (*big.Int, error)
```
Essentially we substitute `header *types.Header, state vm.StateDB` to `caller vm.EVMCaller`.

I agree with combining the `header, state` tuple into one for a couple reasons (switch provider / caller depending on the internal usage):
1. Easier to differentiate between a `current state provider` and a `specific state provider`
2. Easier to then layer on caching to a `current state provider` (could create `caching provider`)

In terms of the rewrite here, I'm more ambivalent to calling it a `EVMCaller` or `EVMProvider`. I think the interface above doesn't change much (though provider currently seems to be more what it's doing, but caller may make more sense if you cache).

In terms of the long function calls, I think they're relatively fine. The only users of the long form api are in `contract_comm/$CONTRACT_NAME`. Those functions can't quite be auto generated, but they're pretty close as they do the translation from high level go types and human names (plus some more processing in certain cases) down to the solidity ABI. I think there's probably a case to isolate the actual calls from the wrappers around them, (like `medianRate` should be it's own function, not part of `GetExchangeRate`).

contracts/contract_call.go
mcortesi: To be honest @piersy . I don't care as much. The way is see it, i've spent several days going over and over the whole codebase, trying to understand it all, and find a way to simplify the way we do call to contracts, and then migrate all the calls to the system.

Where it's better to call it provider or caller, makes no difference (but implies changing it again). I don't buy the LOC argument. A singleton surely takes a lot less LOC than passing a service all over the system to achieve the same goal. Does that make the singleton simpler or easier to understand? I don't think so. It's subjective, there are times where LOC can be a clearer indication on complexity, others where they won't be. At least that my opinion. There are several ways to measure complexity of code. So i don't appreciate being measured by just a single one.

I see this discussion as going to an annoying level of detail about a minor thing, when the much bigger thing (cleaning this up, and migrating the contracts) it's not being addressed.

If you strongly believe that provider is strictly better, then go ahead and replace everything with Provider (And please change https://github.com/celo-org/celo-blockchain/pull/1522 in the process). Just be sure that the notion of no reuse of evm doesn't leave the BoundContract methods...  Also, have in mind that `EVMProvider` implies it creates a EVM.. but what kind of EVM? In particular we create a specific evmContext `context := New(common.ZeroAddress, common.Big0, header, chain, nil)` which only makes sense for this type of calls; so i would include that in the name. `EVMProvider` seems too generic.

contracts/contract_call.go
mcortesi: As i stated on the other comment:

> Did LOCs go up? If so the code likely got more complex, and conversely did they go down, its likely that things got simpler. Its a language agnostic measure.

i don't believe this to be true. But again, i don't have to continue with this "who's right!" kinda of conversation. Too much time is wasted on it; and no real gain.

Just note that on you example, even if you don't reduce the number of parameter, using a parameter object let;s you give name to them, which is why it's help reduce the smell. You won't argue that accessing your parameter by `args[5]` is not as good as saying `args.price`. And when you read a contract call that uses 9 parameters, you'll definietly need some editor tooltip or tool to let you know which parameter is whchi, and you'll have to double check that you're passing the right parameter in the right position. Specially is parameters have the same type.

So, even if it's not good as not needing them, i do believe it makes it clearer.

But again, you seem to like the BoundMethod, so let's leave at that.

piersy: Hey @mcortesi, I wouldn't say this a a "whos right" conversation, hopefully we get to understand each other better and ideally we come to a shared understanding (not quite there yet :) maybe soon!

Totally agree that `args.price` is preferable to `args[5]` but I'd be interested to know where you would apply this? `args []interface{}` is passed around quite a lot in this PR but its just getting passed through to the vm. 

My IDE actually tells me the name and type of function parameters as I'm typing so for me its equivalent to it auto-completing a parameter object field so I don't find one approach better than the other.

I generally try to reserve the use of parameter objects for times when a large number of parameters may be nil, so rather than filling out a function call with a whole bunch of nils you can just initialise a struct with a couple of fields. I find this good for config objects.

I do like the `BoundMethod` it moves lots of config to a one-off initialisation step, and this definitely reduces the parameter burden when calls :D

contracts/contract_call.go
piersy: Hey @mcortesi I do think the provider is a slightly simpler approach that would provide more flexibility in future changes, but at this point I'm not sure its worth switching and changing everything in #1522. If we run into problems with the EVMCaller we can always re-consider the provider route.

Good point regarding the singleton, I can see how you could drop some LOC with one and it's almost never going to be an improvement.

I think at a certain point though everyone has to buy the LOC argument, anyone would be hard pressed to argue that 10,000 LOC is simpler than 1000 LOC that provides the same functionality. The question is about where you draw your thresholds.

In this case an extra layer was added the `EVMCaller` and that added LOC, and in my opinion complexity because it adds an extra layer of indirection to get to the EVM, but it also adds the feature of ensuring non reuse of the EVM, and you can't have features without LOC :)

core/vm/vmcontext/caller.go
```diff
@@ -0,0 +1,84 @@
...
+	vmContext   vm.Context
+	chainConfig *params.ChainConfig
+	state       vm.StateDB
+
+	dontMeterGas bool
+}
+
+func NewEVMCallerForCurrentBlock(chain ExtendedChainContext) (vm.EVMCaller, error) {
+	header := chain.CurrentHeader()
+	// FIXME small race condition here (Head changes between get header & get state)
```
piersy, line 23: Can we fix this?

scripts/testgenesis/main.go
```diff
@@ -0,0 +1,42 @@
+package main
```
piersy, line 1: Accidental addition?

contracts/contract_call.go
```diff
@@ -0,0 +1,120 @@
...
+	return bm.run(caller, result, true, sender, nil, args...)
+}
+
+func (bm *BoundMethod) VMExecute(caller vm.EVMCaller, result interface{}, value *big.Int, args ...interface{}) error {
+	return bm.run(caller, result, false, VMAddress, value, args...)
+}
+
+func (bm *BoundMethod) Execute(caller vm.EVMCaller, result interface{}, sender common.Address, value *big.Int, args ...interface{}) error {
+	return bm.run(caller, result, false, sender, value, args...)
+}
```
piersy, line 81: I think there might be a problem here, because you are passing `sender` but the from address has been hardcoded in the `EVMCaller` constructor as the zero address. I'd be in favoiur of getting rid of both `Query` and `Execute` as they do not seem to be used in either this or the subsequent PR.

```go
func NewEVMCaller(chain ExtendedChainContext, header *types.Header, state vm.StateDB) vm.EVMCaller {
	// The EVM Context requires a msg, but the actual field values don't really matter for this case.
	// Putting in zero values.
	context := New(common.ZeroAddress, common.Big0, header, chain, nil)

	return &evmCaller{
		vmConfig:    *chain.GetVMConfig(),
		vmContext:   context,
		chainConfig: chain.Config(),
		state:       state,
	}
}
```

contracts/contract_call.go
```diff
@@ -0,0 +1,120 @@
...
+	return am.abi.Pack(am.method, args...)
+}
+
+// decodeResult will decode the result of msg execution into the result parameter
+func (am Method) decodeResult(result interface{}, output []byte) error {
+	if result == nil {
+		return nil
+	}
+	return am.abi.Unpack(result, am.method, output)
+}
```
piersy, line 47: These are private and only used once so I would suggest in-lining them

scripts/testgenesis/main.go
```diff
@@ -0,0 +1,42 @@
...
+	db := rawdb.NewMemoryDatabase()
+
+	block, err := genesis.Commit(db)
+	exitOnError(err, "can't setup genesis")
+
+	statedb, err := state.New(block.Root(), state.NewDatabase(db), nil)
+	exitOnError(err, "can't create statedb")
+
+	_ = statedb
+	// caller := vmcontext.NewEVMCaller()
```
piersy, line 34:
```suggestion
```

contracts/contract_call.go
mcortesi: yeah, good point. I rediscoverd that yesterday. So evm "origin" is hardcoded to zeroAddress and then we do set the sender (`msg.from` in solitidy). That's nasty, we should use the same.

Only the tobin tax call changes the sender... and i'd argue it should. So i'll change it to always use the `vmaddress`

In fact, how do you feel about renaming `EVMCaller` to `SystemCaller` :D?

scripts/testgenesis/main.go
mcortesi: totally! thanks

core/vm/vmcontext/caller.go
mcortesi: I tried. problem is that `LightChain` also implements `ExtendedChainContext` and it doesn't have `StateAt()`. Maybe i can add a new method called `CurrentHeaderAndState()` to both

core/vm/vmcontext/caller.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"math/big"
+
+	"github.com/celo-org/celo-blockchain/common"
+	"github.com/celo-org/celo-blockchain/core/types"
+	"github.com/celo-org/celo-blockchain/core/vm"
+	"github.com/celo-org/celo-blockchain/params"
+)
+
+type evmCaller struct {
```
piersy, line 12: I feel like the evmCaller and its factory would be better placed in the contracts package, it doesn't seem to have much to do with vmcontext

contracts/contract_call.go
mcortesi: Totally, at some point a big difference in LOC speaks for itself.

Let's keep it as it is, and keep an eye on the future to see if needs further changes or to move to provide. In a way, this is a refactor, but not necessarily the last one; we can always refine later. What is important, is to be sure it's in the right direction (aka. "better than before")

contracts/contract_call.go
mcortesi: no worries...! we'll get there

And i agree, `BoundMethod` makes it much simpler by providing this one-off initializatiion step outside. Thanks to your suggestion of having things that can be reused in the struct.

To clarify, on my `args[5]` example. i've used an array as a parallel to positional argument. I mean when you have a function:

```go
func doSomething(a,b,c,d,e,f,g,h uint)
```

in the function implementation you have names (you use `a`,`b,` and not `args[5]`). But on the call site

```
doSomething(1,2,4,5,2,5,8)
```
It's harder to see which is which, and to detect if you have a bug by confusing the parameters. And i personally had to double check that when i see some function calls on the current codebase (on other parts of master). I use vscode tooltip, to double check. But it's not immediately evident. So that's why i prefer it.

The only bad thing about parameter objects is that someone can forget to initialize a field, and go doesn't provide a good way of checking that (unless you create a construction function with positional parameters :( ). So in the future if you add a field, you need to double check all call sites, to see it's constructed properly.

I guess golang doesn't really help. a lot in defensive programming

contracts/contract_call.go
piersy: :D I think actually `SystemEVM` would be a better name here `SystemCaller` has no reference to EVM and given the generic use of the term `call` in programming and also the connotation of `System` with OS I think it would be easy to misinterpret.

core/vm/vmcontext/caller.go
trianglesphere: I also wonder if this would fix the cyclic dependency issue when contracts are migrated to use this approach.

core/vm/vmcontext/caller.go
piersy: I think adding this extra method this is the only way to be safe, you have to get both at once.

piersy: Or we could expose the blockchain lock, but I don't think that is a sensible option.

contracts/contract_call.go
piersy: Happy to do the change, but just want to make sure we agree on the name first

core/vm/vmcontext/caller.go
piersy: Which cyclic dependency is this?

contracts/contract_call.go
mcortesi: chaging it to SystemEVM then

contracts/contract_call.go
mcortesi: i like them as they are :P

even if they add LOC, reading `encodeCall` and `decodeResult` tells a better story than just `am.abi.Unpack` `am.abi.Pack`.

core/vm/vmcontext/caller.go
mcortesi: not best pkg name for sure.

The problem is that

* `Context` depends on Contracts

contract_comm/evm.go
mcortesi: can't panic, since tests actually do override this. Acutally i don't trust tests too much, since maybe we have conflicting singleton on them :(

Luckliy next PR actually removes this singleton

contracts/contract_call.go
mcortesi: For now, i've renamed to `SystemEVM` but kept the `VMQUery` etc... instead i'm creating the vm.Context on the fly using the sender, instead of zeroAddress

core/vm/vmcontext/caller.go
mcortesi: yeah, i would wait for now, since if you take a look at the next PR the Blockchain and Lightchain objects are the one implementing the Factory interface and this factory is either not used, or only used on a single ocassion, so we can revisit there.

Either removing this altogheter, or adding the getStateAndHeader option to blockchain and lightchain

core/vm/interface.go
```diff
@@ -99,13 +98,36 @@ type ChainContext interface {
...
 	GetHeaderByNumber(uint64) *types.Header
 
-	// GetVMConfig returns the node's vm configuration
-	GetVMConfig() *Config
+	// Config returns the blockchain's chain configuration
+	Config() *params.ChainConfig
+}
 
-	CurrentHeader() *types.Header
+type SystemEVM interface {
```
piersy, line 105: I don't think this makes sense anymore because this is called 'SystemEVM' but it no longer hard codes the system address as the caller.

I see 2 options:
* Call it "SystemEVM" and always hardcode the caller to be the zero address and the gas price to zero, this means caller can be removed as a parameter to all of the methods.
* Call it "EVMCaller" and let both caller and gas price be provided as parameters, this makes it a more general use object that we can use for applying transactions as well as making system calls. I prefer this route because it should allow us to always call the evm in the same way.

contract_comm/evm.go
```diff
@@ -18,139 +18,92 @@ package contract_comm
...
-	return makeCallFromSystem(scAddress, abi, funcName, args, returnObj, gas, nil, header, state, true)
-}
-
-func GetRegisteredAddress(registryId [32]byte, header *types.Header, state vm.StateDB) (common.Address, error) {
-	vmevm, err := createEVM(header, state)
+func MustGetCaller(header *types.Header, state vm.StateDB) vm.SystemEVM {
+	caller, err := getCaller(header, state)
 	if err != nil {
-		return common.ZeroAddress, err
+		panic("failed to get caller")
```
oneeman, line 45: If we're going to panic we should at least include the error in the message, or log it before panicking.

contracts/registry.go
```diff
@@ -32,23 +32,33 @@ const (
...
 
 // GetRegisteredAddress returns the address on the registry for a given id
-func GetRegisteredAddress(evm *vm.EVM, registryId common.Hash) (common.Address, error) {
-	evm.DontMeterGas = true
-	defer func() { evm.DontMeterGas = false }()
+func GetRegisteredAddress(caller vm.SystemEVM, registryId common.Hash) (common.Address, error) {
+	caller.StopGasMetering()
+	defer caller.StartGasMetering()
 
 	// TODO(mcortesi) remove registrypoxy deployed at genesis
```
oneeman, line 55: While we're here, remove or clarify this TODO.  And maybe also the ones below (lines 63-64 on this branch.

contracts/contract_call.go
```diff
@@ -0,0 +1,120 @@
...
+
+func NewBoundMethod(contractAddress common.Address, abi *abi.ABI, methodName string, maxGas uint64) *BoundMethod {
+	return NewMethod(abi, methodName, maxGas).Bind(contractAddress)
+}
+
+func NewRegistryContractMethod(registryId common.Hash, abi *abi.ABI, methodName string, maxGas uint64) *BoundMethod {
```
oneeman, line 53: Nit: the name makes it sounds like the method is on the registry contract.  Maybe would be better as `NewRegisteredContractMethod`?  That would fit with the existing `GetRegisteredAddress`

contracts/contract_call.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (bm *BoundMethod) VMQuery(caller vm.SystemEVM, result interface{}, args ...interface{}) error {
+	return bm.Query(caller, result, VMAddress, args...)
+}
+
+func (bm *BoundMethod) Query(caller vm.SystemEVM, result interface{}, sender common.Address, args ...interface{}) error {
+	return bm.run(caller, result, true, sender, nil, args...)
+}
+
+func (bm *BoundMethod) VMExecute(caller vm.SystemEVM, result interface{}, value *big.Int, args ...interface{}) error {
```
oneeman, line 75: There's a lack of symmetry here: `VMQuery` calls `Query` but `VMExecute` calls `run` directly.  I don't think there's a big difference between the two, you I might lean a little towards calling `.Execute()`.  But either way it'd be nicer if they were consistent.

core/vm/vmcontext/system_evm.go
```diff
@@ -0,0 +1,120 @@
...
+	vm.ChainContext
+
+	// GetVMConfig returns the node's vm configuration
+	GetVMConfig() *vm.Config
+
+	CurrentHeader() *types.Header
+
+	State() (*state.StateDB, error)
+}
+
```
oneeman, line 24: I think the below is pretty confusing with the proliferation of SystemEVM-related names (system, factory, and provider).  As I see it: SystemEVMFactory generates SystemEVM, which has a provider, which generates EVMs.  I'm not sure how to make the naming clearer.  If we can't, maybe we should include some comments to lay it out for readers.  It might also help to make `NewSystemEVM` and `NewSystemEVMForCurrentBlock` private in the same way that `newEVMProvider` is private, to delineate functions which are intended as part of the API from helper functions (I'm assuming those two are in fact just helper functions and not intended to be called directly elsewhere).  And maybe even place them separately in the code.  Granted, none of this would be a huge boost to the readability, but perhaps still recommendable in the absence of a better way to simplify the things in this file.

contracts/contract_call.go
mcortesi: Remove the `VM..` variants.

core/vm/interface.go
piersy: Sounds good :+1:

scripts/testgenesis/main.go
piersy, line 1: This accidental addition is still here

core/vm/vmcontext/caller.go
piersy: I spent some time trying to refactor this, but it's quite involved, so probably best left to another PR

contracts/contract_call.go
```diff
@@ -0,0 +1,112 @@
...
+// NewMethod creates a new contract message
+func NewMethod(abi *abi.ABI, method string, maxGas uint64) Method {
+	return Method{
+		abi:    abi,
+		method: method,
+		maxGas: maxGas,
+	}
+}
+
+func (am Method) Bind(contractAddress common.Address) *BoundMethod {
```
piersy, line 29: Needs some documentation

contracts/contract_call.go
piersy: I'm not convinced about these functions telling a better story.

For me when I read the use of encodeCall I jumped to it's definition to see what it did, then I saw that it just calls abi.Pack and at that point I knew what it did, because I'm familiar with the go-ethereum codebase and I'd seen it before.

Having encodeCall adds one extra layer between the reader of the code and the underlying function Pack. Pack is much more likely to be known to anyone who is familar with the go-ethereum codebase and has more comprehensive documentation than encode call.

So I would say encodeCall actually obscures what is really happening from the reader. In my case I needed to jump to its definition to see what it did, whereas if we just called Pack directly I wouldn't have had to break my flow and figure out the meaning of encodeCall.

contracts/contract_call.go
piersy, line 29:
```suggestion
// Bind returns a BoundMethod instance which can be used to call the contract method represented by am
// and residing at contracAddress.
func (am Method) Bind(contractAddress common.Address) *BoundMethod {
```

core/vm/vmcontext/runner.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// GetVMConfig returns the node's vm configuration
+	GetVMConfig() *vm.Config
+
+	CurrentHeader() *types.Header
+
+	State() (*state.StateDB, error)
+}
+
+// SystemEVMRunnerBuilder returns a builder factory for a System's EVMRunner
```
oneeman, line 28: I don't like adding "builder" to the mix.  This is a function that returns a factory, so I would rename it to `GetSystemEVMRunnerFactory()`.  Then the name is more self-explanatory in the places it's used, e.g.

```
contract_comm.SetEVMRunnerFactory(vmcontext.GetSystemEVMRunnerFactory(blockchain))
```

contracts/contract_call.go
```diff
@@ -0,0 +1,112 @@
...
+func NewRegisteredContractMethod(registryId common.Hash, abi *abi.ABI, methodName string, maxGas uint64) *BoundMethod {
+	return &BoundMethod{
+		Method: NewMethod(abi, methodName, maxGas),
+		resolveAddress: func(vmRunner vm.EVMRunner) (common.Address, error) {
+			return resolveAddressForCall(vmRunner, registryId, methodName)
+		},
+	}
+}
+
+type BoundMethod struct {
```
piersy, line 62: I was trying to add some suggested documentation for BoundMethod but I found that I wasn't able to easily express what it is. I think that the naming is not quite right here, bound implies locked to something and makes sense in the context of the Bind method.

```go
func (am Method) Bind(contractAddress common.Address) *BoundMethod {
```

Because `Bind` is binding the Method to a specific address, but the type BoundMethod is not actually bound to a specific address in fact it has a function to resolve the address when invoked, meaning that the address can change during the lifetime of the object, which is incongruent with the concept of bound.

Also given that Method doesn't actually provide any functionality to be called I think it would be clearer to rename Method →MethodTemplate/MethodSignature and BoundMethod→Method.

contracts/contract_call.go
```diff
@@ -0,0 +1,112 @@
...
+	Method
+	resolveAddress func(vm.EVMRunner) (common.Address, error)
+}
+
+func (bm *BoundMethod) Query(vmRunner vm.EVMRunner, result interface{}, args ...interface{}) error {
```
piersy, line 67:
```suggestion
// Query executes the method as a read only action, the returned value is unpacked into result.
func (bm *BoundMethod) Query(vmRunner vm.EVMRunner, result interface{}, args ...interface{}) error {
```

core/vm/interface.go
```diff
@@ -99,13 +98,30 @@ type ChainContext interface {
...
 
-	// GetVMConfig returns the node's vm configuration
-	GetVMConfig() *Config
+	// Config returns the blockchain's chain configuration
+	Config() *params.ChainConfig
+}
 
-	CurrentHeader() *types.Header
+// EVMRunner provides a simplified API to run EVM calls
+// EVM's sender, gasPrice, txFeeRecipient and state are set by the runne on each call
```
oneeman, line 106:
```suggestion
// EVM's sender, gasPrice, txFeeRecipient and state are set by the runner on each call
```

core/vm/vmcontext/runner.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// NewSystemEVMRunnerForCurrentBlock creates an EVMRunner where calls are originated by the VMAddress (uses chain's current block)
+func NewSystemEVMRunnerForCurrentBlock(chain ExtendedChainContext) (vm.EVMRunner, error) {
+	return newEVMRunnerForCurrentBlock(chain, VMAddress, common.Big0)
+}
+
+// NewSystemEVMRunner creates an EVMRunner where calls are originated by the VMAddress
+func NewSystemEVMRunner(chain ExtendedChainContext, header *types.Header, state vm.StateDB) vm.EVMRunner {
+	return newEVMRunner(chain, VMAddress, common.Big0, header, state)
```
oneeman, line 54: Would it make sense to remove `sender` `gasPrice` as arguments from `newEVMRunnerForCurrentBlock` and `newEVMRunner`?  They are only called with the zero values, and we made them private to indicate they shouldn't be called directly, so I don't see a reason for them to take these arguments unless we envision their getting different values in the future.

core/vm/vmcontext/runner.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func newEVMRunnerForCurrentBlock(chain ExtendedChainContext, sender common.Address, gasPrice *big.Int) (vm.EVMRunner, error) {
+	header := chain.CurrentHeader()
+	// FIXME small race condition here (Head changes between get header & get state)
```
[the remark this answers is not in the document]
mcortesi: Moved inside `GetSystemEVMRunnerFactory` which is the one that dissapears after, to make this more clear

core/vm/vmcontext/runner.go
oneeman: Right, the reason I thought of this was because on `BoundMethod` you removed the non-system-address varieties of query and execute, so I thought maybe the same rationale would apply here.

oneeman (approved):
Still some pending comments, but nothing major, so I'll approve this now and you can merge it after you're done with the smaller changes (or after the migration PR is merged into this one, as you prefer).

core/vm/vmcontext/runner.go
oneeman: Done, marking as resolved.

contracts/contract_call.go
piersy: Hi @mcortesi I will try to be more flexible in future. I see these conversations as allowing us to get to understand each other's coding values better, so its not just about the specific details here but also setting the scene for us to be able to come to conclusions on similar situations much more quickly in the future. So from that perspective too I don't consider this wasted time.

contracts/contract_call.go
piersy: Interesting point about `maxGas`, if you are saying that it does not have anything to do with the signature (which I agree with) then I think it must relate to the implementation, which means that it should be a field on `BoundMethod` because that can change based on the address that a method is bound to.

PR #1514, opened 2021-04-15
Migrates minimun version check to new contract layer

mcortesi:
### Description

Migrates the minimun version check to the new contract layer

### Other changes

* Adds to `Ethereum` and `LightEthereum` objects an `evmCallerFactory` which they create (as they are the ones with blockchain objects)

* Renames version check flag to `DisableVersionCheck` since it's a disable flag, not an enable one
### Tested

 - Would be nice to do some

### Backwards compatibility

Yes


mcortesi:
closed in favor of #1522

PR #1516, opened 2021-04-15
Add block production and consensus cycle metrics

### LINKED ISSUE #1511: Add loadtest reporting flag to geth
Instead of requiring a modified version of geth to be present, add a flag that does the printing as needed.

Maybe also add in the metrics flags.

trianglesphere:
### Description

Adds better metrics for load testing.
Also adds a command line option `--metrics.loadtestcsvfile` to specify a file to write out information about the block production and consensus cycle in a CSV format. This was previously on a separate branch that was cherry-picked onto the branch that we wanted to loadtest.

### Tested

Local load test.

### Related issues

- Fixes #1511

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -1038,12 +1038,12 @@ func (sb *Backend) recordBlockProductionTimes(blockNumber, txCount, gasUsed, rou
 	sleepGauge := sb.sleepGauge
 	consensusGauge := metrics.Get("consensus/istanbul/core/consensus_commit").(metrics.Gauge)
 	verifyGauge := metrics.Get("consensus/istanbul/core/verify").(metrics.Gauge)
-	blockConstructGuage := metrics.Get("miner/worker/block_construct").(metrics.Gauge)
+	blockConstructGauge := metrics.Get("miner/worker/block_construct").(metrics.Gauge)
 	cpuSysLoadGauge := metrics.Get("system/cpu/sysload").(metrics.Gauge)
 	cpuSysWaitGauge := metrics.Get("system/cpu/syswait").(metrics.Gauge)
 	cpuProcLoadGauge := metrics.Get("system/cpu/procload").(metrics.Gauge)
 
 	sb.csvRecorder.WriteRow(blockNumber, txCount, gasUsed, round,
-		cycle.Nanoseconds(), sleepGauge.Value(), consensusGauge.Value(), verifyGauge.Value(), blockConstructGuage.Value(),
+		cycle.Nanoseconds(), sleepGauge.Value(), consensusGauge.Value(), verifyGauge.Value(), blockConstructGauge.Value(),
 		cpuSysLoadGauge.Value(), cpuSysWaitGauge.Value(), cpuProcLoadGauge.Value())
 }
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -113,7 +113,7 @@ func New(config *istanbul.Config, db ethdb.Database) consensus.Istanbul {
 		blocksFinalizedGasUsedGauge:        metrics.NewRegisteredGauge("consensus/istanbul/blocks/gasused", nil),
 		sleepGauge:                         metrics.NewRegisteredGauge("consensus/istanbul/backend/sleep", nil),
 	}
-	if metrics.EnabledLoadTest {
+	if config.LoadTestCSVFile != "" {
 		backend.csvRecorder = metrics.NewStdoutCSVRecorder("blockNumber", "txCount", "gasUsed", "round",
 			"cycle", "sleep", "consensus", "block_verify", "block_construct",
 			"sysload", "syswait", "procload")
@@ -497,7 +497,7 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		Bitmap:    aggregatedEpochValidatorSetSeal.Bitmap,
 		Signature: aggregatedEpochValidatorSetSeal.Signature,
 	})
-	if metrics.EnabledLoadTest {
+	if sb.csvRecorder != nil {
 		sb.recordBlockProductionTimes(block.Header().Number.Uint64(), uint64(len(block.Transactions())), block.GasUsed(), aggregatedSeal.Round.Uint64())
 	}
 
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -505,7 +505,7 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		Signature: aggregatedEpochValidatorSetSeal.Signature,
 	})
 	if sb.csvRecorder != nil {
-		sb.recordBlockProductionTimes(block.Header().Number.Int64(), int64(len(block.Transactions())), int64(block.GasUsed()), aggregatedSeal.Round.Int64())
+		sb.recordBlockProductionTimes(block.Header().Number.Uint64(), len(block.Transactions()), block.GasUsed(), aggregatedSeal.Round.Uint64())
 	}
 
 	sb.logger.Info("Committed", "address", sb.Address(), "round", aggregatedSeal.Round.Uint64(), "hash", proposal.Hash(), "number", proposal.Number().Uint64())
@@ -1039,7 +1039,7 @@ func (sb *Backend) UpdateReplicaState(seq *big.Int) {
 }
 
 // recordBlockProductionTimes records information about the block production cycle and reports it through the CSVRecorder
-func (sb *Backend) recordBlockProductionTimes(blockNumber, txCount, gasUsed, round int64) {
+func (sb *Backend) recordBlockProductionTimes(blockNumber uint64, txCount int, gasUsed, round uint64) {
 	cycle := time.Since(sb.cycleStart)
 	sb.cycleStart = time.Now()
 	sleepGauge := sb.sleepGauge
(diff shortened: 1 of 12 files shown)


consensus/istanbul/core/core.go
```diff
@@ -781,6 +794,8 @@ func (c *core) verifyProposal(proposal istanbul.Proposal) (time.Duration, error)
 		return 0, verificationStatus
 	} else {
 		logger.Trace("verification status cache miss")
+		start := time.Now()
+		defer func() { c.verifyGauge.Update(time.Since(start).Nanoseconds()) }()
```
oneeman, line 798: Not much of a difference, but I think this (and the others like it in this PR) could be done as:
```
defer func(start time.Time) { c.verifyGauge.Update(time.Since(start).Nanoseconds()) }(time.Now())
```

e.g. https://play.golang.org/p/XzlpNvFellY

miner/worker.go
```diff
@@ -193,30 +193,33 @@ type worker struct {
 
 	// Needed for randomness
 	db ethdb.Database
+
+	blockConstructGuage metrics.Gauge
```
oneeman, line 197: Typo: Guage -> Gauge

oneeman:
Unless there's a strong reason against it, I would prefer having to specify a filename for the csv (i.e. instead of a boolean flag to enable csv, a flag that has the filename) and outputting it there rather to stdout.  If nothing else, it would mean outputting a clean csv without any unrelated output from geth (e.g. the starting and stopping messages)

trianglesphere:
> Unless there's a strong reason against it, I would prefer having to specify a filename for the csv (i.e. instead of a boolean flag to enable csv, a flag that has the filename) and outputting it there rather to stdout. If nothing else, it would mean outputting a clean csv without any unrelated output from geth (e.g. the starting and stopping messages)

Yea, using the filename definitely makes more sense. This was copied what was already done (which at various points filtered out extraneous output). It's better to switch that than emulate the weird behavior here.

consensus/istanbul/backend/backend.go
```diff
@@ -1013,3 +1030,20 @@ func (sb *Backend) UpdateReplicaState(seq *big.Int) {
...
 }
+
+// recordBlockProductionTimes records information about the block production cycle and reports it through the CSVRecorder
+func (sb *Backend) recordBlockProductionTimes(blockNumber, txCount, gasUsed, round uint64) {
+	cycle := time.Since(sb.cycleStart)
+	sb.cycleStart = time.Now()
+	sleepGauge := sb.sleepGauge
+	consensusGauge := metrics.Get("consensus/istanbul/core/consensus_commit").(metrics.Gauge)
+	verifyGauge := metrics.Get("consensus/istanbul/core/verify").(metrics.Gauge)
+	blockConstructGuage := metrics.Get("miner/worker/block_construct").(metrics.Gauge)
```
oneeman: `guage` here too.  There are others as well, so to fix them all go through and search for `guage`.

metrics/csv_metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"sync"
+)
+
+// A CSVRecorder enables easy writing of CSV data a specified writer.
+// The header is written on creation. Writing is thread safe.
+type CSVRecorder struct {
```
oneeman, line 29: I would say to use https://golang.org/pkg/encoding/csv/ rather than do all the formatting yourself.  Generally speaking it's better to use the library rather than DIY, since CSV has many edge cases, like if one of the values has a comma in it (though that shouldn't be the case here).  Though it does require the row to consist of strings, whereas here we're getting `interface{}`, but that seems manageable.

Given the limited nature and relatively simplicity of the CSV we're generating, I don't feel super-strongly about this, but I think it's a best-practice that's always a good idea to follow (especially when the language has csv support built-in and you don't need an external dependency).

consensus/istanbul/backend/backend.go
trianglesphere: 🤦 I'll go fix them.

metrics/csv_metrics.go
trianglesphere: Yea, using the the library definitely makes sense.

What do you think about switching the interface to be []string (or varags string) and then calling out the csv library. I think I still want to wrap the library (specifically to be able to eventually close the file).

trianglesphere: Ok, I've switched to using the csv library. It makes things a little harder b/c there's no way easy way to convert primitive types (like `int64`) to string values outside the fmt package, but it works fine.

oneeman: There is `strconv.FormatInt()` for int64, so we could use that.  But, yeah, we definitely need to convert to strings one way or another.  I don't have a strong preference on whether it should be inside or outside `CSVRecorder`.

consensus/istanbul/backend/backend.go
```diff
@@ -1013,3 +1037,26 @@ func (sb *Backend) UpdateReplicaState(seq *big.Int) {
...
+	verifyGauge := metrics.Get("consensus/istanbul/core/verify").(metrics.Gauge)
+	blockConstructGauge := metrics.Get("miner/worker/block_construct").(metrics.Gauge)
+	cpuSysLoadGauge := metrics.Get("system/cpu/sysload").(metrics.Gauge)
+	cpuSysWaitGauge := metrics.Get("system/cpu/syswait").(metrics.Gauge)
+	cpuProcLoadGauge := metrics.Get("system/cpu/procload").(metrics.Gauge)
+
+	vals := [...]int64{blockNumber, txCount, gasUsed, round,
+		cycle.Nanoseconds(), sleepGauge.Value(), consensusGauge.Value(), verifyGauge.Value(), blockConstructGauge.Value(),
+		cpuSysLoadGauge.Value(), cpuSysWaitGauge.Value(), cpuProcLoadGauge.Value()}
+	var strs []string
```
oneeman, line 1056: I don't think it matters, but in such cases we would usually preallocate with `strs := make([]string, len(vals))`

consensus/istanbul/core/commit.go
```diff
@@ -137,6 +138,7 @@ func (c *core) broadcastCommit(sub *istanbul.Subject) {
 }
 
 func (c *core) handleCommit(msg *istanbul.Message) error {
+	defer func(start time.Time) { c.handleCommitTimer.UpdateSince(start) }(time.Now())
```
mcortesi, line 141: you don't need the xtra funciton
```suggestion
	defer c.handleCommitTimer.UpdateSince(time.Now())
```

consensus/istanbul/core/prepare.go
```diff
@@ -173,6 +174,7 @@ func (c *core) getViewFromVerifiedPreparedCertificate(preparedCertificate istanb
 }
 
 func (c *core) handlePrepare(msg *istanbul.Message) error {
+	defer func(start time.Time) { c.handlePrepareTimer.UpdateSince(start) }(time.Now())
```
mcortesi, line 177: same

consensus/istanbul/core/preprepare.go
```diff
@@ -50,6 +50,8 @@ func (c *core) sendPreprepare(request *istanbul.Request, roundChangeCertificate
 }
 
 func (c *core) handlePreprepare(msg *istanbul.Message) error {
+	defer func(start time.Time) { c.handlePrePrepareTimer.UpdateSince(start) }(time.Now())
```
mcortesi, line 53: same

metrics/csv_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+	writer        csv.Writer
+	backingWriter io.Writer
+	writeMu       sync.Mutex
+}
+
+// NewStdoutCSVRecorder creates a CSV recorder to standard out
+// The header is immediately written upon construction.
+func NewStdoutCSVRecorder(fields ...string) *CSVRecorder {
```
oneeman, line 36: I would remove this function since we're not using and there isn't another clear reason to keep it.

mcortesi (commented):
Something i wonder here is that we add a couple of metrics that seems to be designed for the CSV Recorder. if that's the case, we can create & report them ONLY if the csv file was given.

A way to do that is that instead of having a `backend.csvRecorder` have a `backend.performanceWatcher` that then exposes function do each type of reporting. And so if not present, no reporting happens.

If it helps you can also have a `nilWatcher` does is a noop, to avoid the `if x != nil` on every step

metrics/csv_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+	return NewCSVRecorder(os.Stdout, fields...)
+}
+
+// NewStdoutCSVRecorder creates a CSV recorder that writes to the supplied writer.
```
oneeman, line 40:
```suggestion
// NewCSVRecorder creates a CSV recorder that writes to the supplied writer.
```

metrics/csv_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+// The writer is aborbed into the recorder and the user is responsible for calling
```
oneeman, line 41: absorbed has a typo, but also I'm not sure if absorbed is the best word here.

oneeman (approved):
Left a few nits, but otherwise this looks ready.

consensus/istanbul/core/commit.go
oneeman: Nice!

trianglesphere:
> Something i wonder here is that we add a couple of metrics that seems to be designed for the CSV Recorder. if that's the case, we can create & report them ONLY if the csv file was given.

All of the metrics are actually used when load testing/re-purposed from existing metrics. There's actually a few different use cases pulled together in this PR: loadtesting w/ metrics, loadtesting with csvs, better mainnet metrics. I wouldn't really classify any of these metrics as expensive metrics either.

metrics/csv_metrics.go
trianglesphere: done

consensus/istanbul/backend/backend.go
trianglesphere: I changed the location of this, but did include this optimization

metrics/csv_metrics.go
trianglesphere: switch to "retained" and making it clear that it current only accepts a WriterCloser

PR #1522, opened 2021-04-20
Migrates all contractst calls to EVMRunner

### LINKED ISSUE #1549: Clean up contract call layer
## Description

As part of node operations, node needs to do many contract calls. To do so, it relies on `contract_comm` package, that maintains a singleton instance to be able to make such calls. Code is particularly hairy and hard to follow.

The request, is to refactor it and clean it up. Remove also the possibility of making contract calls without specifying the context (header & state).

## Context

This help maintainability. Cleaner and more strict code helps us make changes with confidence.

mcortesi:
### Description

This PR removes `contract_comm` modules, and recreates them into `contracts`. On that move, all calls are now done using `contract.BoundMethod` and they expect an`vm.EVMRunner` instance instead of a `(header,state)` pair.

Doing so, implies that contract callers now require the ability to create an `vm.EVMRunner` which in the past was done thanks to a singleton living in `contract_comm/evm.go`. On this PR, that responsibility is on `Blockchain` and `Lightchain` object. These 2 objects are easily available in most of the other packages in some form of another, so migration is direct is most of the cases.

#### random.IsRunning() quirk

This PR makes evident something that already happened:

 * Before we used `ramdom.isRunning()`, that used state in `blockchain.CurrentBlock()`; instead of the block that is actually being processed at the place of the call. 
 * This could create a weird condition if you call `random.IsRunning()` when processing a block that is not `currentHead+1` or if you've already modyfied state in the block.

For this PR, we consider that it's safe to use current context state, and not blockchain's current block. Since calls to this function only happen during block processing; at the very beginning of the block. Thus the block being processed state is equal to the last's block state.

#### Injection an EVMRunner into Istanbul

This is particularly tricky because of circular dependencies. On master, we have `vm.ChainContext` that depends on `consensus.Engine`; then we have `consensus.IstanbulEngine` that has a `SetChain` to inject a `consensus.ChainReader`.  Object that is then used to access a "blockchain interface". We need to add `EVMRunner` factory methods to `ChainReader` so that then `istanbul.Backend` can create them. This introduces a circular dependency between `consensus` and `vm`

To break such dependency several changes are included:

 1. Split consensus `ChainReader` into `ChainHeaderReader` (this comes from upstream master) and `ChainContext` (which inherits from `ChainReader` and adds `EVMRunner` factory methods)

 2. Modifies all `consensus.Engine` methods either to remove the `ChainReader` parameter or change it to ChainHeaderReader`. Potentially the parameter can be removed since Istanbul engine already receives the ChainContext on creation (with `SetChain()`)

 3. Moves `vm.ChainContext` to `core.ChainContext`. This is the original location in upstream. And helps break the dependency

 4. Duplicates `core.ChainContext` in `vmcontext.ChainContext` to break
circular dependency (core <-> vmcontext). core MUST depend on vmcontext, so we have to avoid the
inverse. For that, instead of expecting a `core.ChainContext` argument, it now expects its own local interface which is a copy of the former

Thanks to this the `consensus <-> vm` circular dependency is avoided, and code structure is closer to upstream. Also, this provides access to `EVMRunner` factory methods within `IstanbulEngine`

#### A Note on testing

Before we rely on calling `contract_comm.SetInternalEVMCallerFactory()` and initialize the singleton to be able to make contract calls through the systems. But we didn't inject do that on every call. The result is that probably a lot of tests did contract call that failed... and since our error handling strategy ignores and continue, the tests worked. I don't think that's great as a testing scenario, since it seems many things are untested.

This PR tries to do better, which means potentially some tests will fail

### Issues

* Fixes #1549


core/blockchain.go
```diff
@@ -355,6 +356,21 @@ func (bc *BlockChain) GetVMConfig() *vm.Config {
...
+// NewEVMCaller retuns the EVMCaller factory of the blockchain
+func (bc *BlockChain) NewCurrentHeadEVMCaller() (vm.EVMCaller, error) {
+	header := bc.CurrentHeader()
+	state, err := bc.StateAt(header.Root)
+	if err != nil {
+		return nil, err
+	}
+	return vmcontext.NewEVMCaller(bc, header, state), nil
+}
+
```
piersy, line 373: I would consider not adding these methods to BlockChain. Given that long term we want to get rid of the object adding methods on it will add to the work we need to do in the future. Also having these methods on blockchain may encourage people to pass blockchain when really they only need an EVMCaller thus increasing coupling ( because blockchain has so much other stuff in it)

core/blockchain.go
```diff
@@ -355,6 +356,21 @@ func (bc *BlockChain) GetVMConfig() *vm.Config {
...
 
+// NewEVMCaller retuns the EVMCaller factory of the blockchain
+func (bc *BlockChain) NewEVMCaller(header *types.Header, state vm.StateDB) vm.EVMCaller {
+	return vmcontext.NewEVMCaller(bc, header, state)
+}
+
+// NewEVMCaller retuns the EVMCaller factory of the blockchain
+func (bc *BlockChain) NewCurrentHeadEVMCaller() (vm.EVMCaller, error) {
+	header := bc.CurrentHeader()
+	state, err := bc.StateAt(header.Root)
```
piersy, line 367: This looks like the same race condition we identified in the EVMCaller PR

core/state_transition.go
```diff
@@ -190,12 +190,13 @@ func IntrinsicGas(data []byte, contractCreation bool, feeCurrency *common.Addres
 }
 
 // NewStateTransition initialises and returns a new state transition object.
-func NewStateTransition(evm *vm.EVM, msg Message, gp *GasPool) *StateTransition {
-	gasPriceMinimum, _ := gpm.GetGasPriceMinimum(msg.FeeCurrency(), evm.Header, evm.StateDB)
+func NewStateTransition(evm *vm.EVM, msg Message, gp *GasPool, caller vm.EVMCaller) *StateTransition {
```
piersy, line 193: This is confusing, why take an evm and a caller?

core/blockchain.go
mcortesi: I considered this; but i was convinced to go fro the lesser of evils. Blockchain is generally available through the code; and it's alreayd used to get other things like the `VMConfig`. You'll notice that i didn't had to do much "object passing" so that places where an evmCaller was required were able to fetch one.

Initially i put the factory on `Ethereum` & `LightEthereum` service; but that proved hard, since not many objects have reference to them, and it would imply passing a new object everywhere, while blockchain was already there.

I agree it's not the best, that changing that involves a much deeper refactor, that i would prefer to leave for a subsequent refactor.

core/blockchain.go
mcortesi: This is not, but LightChain does have it.

here you get the header, and the you get the state from the header.

the race is getting "current" header, and then getting "current" state. Where the race is that maybe the notion of "current" has changed

core/state_transition.go
mcortesi: I know, any ideas?

i can't really use the evm to construct the evmCaller; nor i can get a new evm from the caller. Since they are different, they have different vm.Context object.

core/vm/vmcontext/caller.go
```diff
@@ -18,6 +18,15 @@ type evmCaller struct {
...
 }
 
+func NewTestingEVMCaller(vmConfig vm.Config, vmContext vm.Context, chainConfig *params.ChainConfig, state vm.StateDB) vm.EVMCaller {
+	return &evmCaller{
+		vmConfig:    vmConfig,
+		vmContext:   vmContext,
+		chainConfig: chainConfig,
+		state:       state,
+	}
+}
```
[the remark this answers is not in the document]
mcortesi: won't work, since `evmCaller` is private, and the test is not in this pkg, so needs a public facing funciton

core/blockchain.go
piersy: So yeah, if you want to get to cleaner code at a certain point you have to do the object passing, so either we do it now or we do it later. But adding methods like this is backing us into a corner and will increase the chances of encountering circular dependency problems. So if we continue to make compromises like this, eventually we will get stuck, so lets schedule the refactor as soon as possible.

core/blockchain.go
piersy: Ahh, yes thanks for pointing that out, what about the light chain though?

core/state_transition.go
piersy: Whats the difference in the vm.Contexts? Can we share one instead?

core/vm/vmcontext/caller.go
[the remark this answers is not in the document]
piersy: I had a look at the test and I can see that the caller is used to lookup the gasPriceMinimum, but that fails because the registry is not deployed. As a result the fallback gas price minimum is returned. GetGasPriceMinimum also checks for a nil evmCaller and if so will return the fallback value as well.

So in this case I think it would be neater to just use a nil caller in the test and then there is no need for to add this test constructor.

core/state_transition.go
piersy: So I looked into this a bit more and I think all that differs between the evms is the sender and the gas price. So we could pass a factory here and provide a method on the factory to build an evm taking the caller, gas price, header and state. Then we would just need that one factory instance. This would change the meaning of the EVMCaller since it would no longer be an object reserved solely for calls by the system, but I think is a good change since then I think it supports all our use cases and we don't end up with 2 different ways of calling the evm.

cmd/utils/flags.go
```diff
@@ -403,10 +403,6 @@ var (
 		Usage: "Number of CPU threads to use for mining",
 		Value: 0,
 	}
-	MinerNotifyFlag = cli.StringFlag{
```
oneeman, line 406: Maybe mention these flags' removal on the PR as "Other Changes". and/or maybe on the parent PR after this one is merged.

cmd/geth/main.go
```diff
@@ -490,7 +489,23 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 		}
 	}
 	if !ctx.GlobalBool(utils.VersionCheckFlag.Name) {
-		blockchain_parameters.SpawnCheck()
+		var callerFactory func() (vm.EVMRunner, error)
```
oneeman, line 492: Rename to `runnerFactory` (here and also in `blockchain_parameters.go`)

cmd/geth/main.go
```diff
@@ -490,7 +489,23 @@ func startNode(ctx *cli.Context, stack *node.Node) {
...
+		if ctx.GlobalString(utils.SyncModeFlag.Name) == "full" || ctx.GlobalString(utils.SyncModeFlag.Name) == "fast" {
```
oneeman, line 493: I don't like this, but I don't see a way to avoid it entirely.  At first I thought to combine this with the block above that also checks full/light, but that involves its own nested checks.

However, I have a feeling that this `"full" || "fast"` may bite us in the future, in the same way that https://github.com/celo-org/celo-blockchain/blob/master/cmd/utils/flags.go#L1224 is now wrong since we added the `lightest` sync mode.  Perhaps what I would suggest is having a local var higher up like:
```
fullNode := ctx.GlobalString(utils.SyncModeFlag.Name) == "full" || ctx.GlobalString(utils.SyncModeFlag.Name) == "fast"
```

and then using that below for the two if's.  This will likely cause a conflict later on when `snap` sync mode is brought in from upstream, but at least such a conflict would prevent us from overlooking that the new condition added in this PR goes out of date.  Then again, I guess if something like that did happen, we would probably get a `Fatalf` here which would get detected by one of the end-to-end sync tests.

So I don't know, whatever you think best.

core/state_processor.go
```diff
@@ -59,20 +60,28 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
+		vmRunner = p.bc.NewSystemEVMRunner(block.Header(), statedb)
+		gp       = new(GasPool).AddGas(blockchain_parameters.GetBlockGasLimitOrDefault(vmRunner))
 	)
 	// Mutate the block and state according to any hard-fork specs
 	if p.config.DAOForkSupport && p.config.DAOForkBlock != nil && p.config.DAOForkBlock.Cmp(block.Number()) == 0 {
 		misc.ApplyDAOHardFork(statedb)
 	}
 
-	if random.IsRunning() {
+	// TODO(HF) use vmRunner instead of parentVMRunner (HF required)
```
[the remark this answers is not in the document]
mcortesi: i guess... can we assume that `state_processor` is only called with block N+1 when blockchain head is N? I was not entirely sure about that.

core/blockchain.go
```diff
@@ -355,6 +356,21 @@ func (bc *BlockChain) GetVMConfig() *vm.Config {
...
 }
 
+// NewSystemEVMRunner creates the System's EVMRunner for given header & sttate
+func (bc *BlockChain) NewSystemEVMRunner(header *types.Header, state vm.StateDB) vm.EVMRunner {
+	return vmcontext.NewSystemEVMRunner(bc, header, state)
+}
+
+// NewSystemEVMRunnerForCurrentBlock creates the System's EVMRunner for current block & state
+func (bc *BlockChain) NewSystemEVMRunnerForCurrentBlock() (vm.EVMRunner, error) {
+	header := bc.CurrentHeader()
```
oneeman, line 366: Can `CurrentHeader()` ever return `nil`?  I never thought of it until I saw this nil check in `validatorRandomnessAtBlockNumber()` (in code which is changed in this PR): https://github.com/celo-org/celo-blockchain/blob/master/consensus/istanbul/backend/backend.go#L738-L740

contracts/blockchain_parameters/blockchain_parameters.go
```diff
@@ -0,0 +1,209 @@
...
+func GetBlockGasLimitOrDefault(vmRunner vm.EVMRunner) uint64 {
+	val, err := GetBlockGasLimit(vmRunner)
+	if err != nil {
+		logError("blockGasLimit", err)
+		return params.DefaultGasLimit
+	}
+	return val
+}
+
+func GetBlockGasLimit(vmRunner vm.EVMRunner) (uint64, error) {
```
oneeman, line 154: Is this function intended to possibly be used directly anywhere?  If not, I would make it private or even inline it like it originally was (and which would make it be more similar to `GetIntrinsicGasForAlternativeFeeCurrency()` which handles the error case inline).

contracts/contract_call.go
```diff
@@ -0,0 +1,116 @@
...
+		output, leftoverGas, err = vmRunner.Execute(contractAddress, input, bm.maxGas, value)
+	}
+
+	if err != nil {
+		message, _ := unpackError(output)
+		logger.Error("Error invoking evm function: EVM call failure", "input", hexutil.Encode(input), "err", err, "message", message)
+		return err
+	}
+
+	if err := bm.decodeResult(result, output); err != nil {
```
piersy, line 109: Maybe we could return an error here if readOnly is set and result is nil

contracts/contract_call.go
```diff
@@ -0,0 +1,116 @@
...
+type BoundMethod struct {
+	Method
+	resolveAddress func(vm.EVMRunner) (common.Address, error)
+}
+
+func (bm *BoundMethod) Query(vmRunner vm.EVMRunner, result interface{}, args ...interface{}) error {
+	return bm.run(vmRunner, result, true, nil, args...)
+}
+
+func (bm *BoundMethod) Execute(vmRunner vm.EVMRunner, result interface{}, value *big.Int, args ...interface{}) error {
```
piersy, line 71: Can we remove the value parameter, its set to 0 in all calls. And in fact the same is true for the EVMRunner. I cant imagine a situation where the system would need to transfer value.

core/state_transition.go
mcortesi: i would wait on that... actually i've chatter with @oneeman before making this change since a faced two paths:

 * Add a argument and minimize changes that would conflict upstream
 * Do something more fancy like you suggest, and complicate upstream merges

And our "merge master" discouraged that approach. 

And since that would be a longer conversations i would leave this here as it is now, and we can get back at it later.

cmd/utils/flags.go
mcortesi: done

contracts/blockchain_parameters/blockchain_parameters.go
mcortesi: The intent here was to make explicit that there's an error and we are using a default. That's why there's `GetBlockGasLimitOrDefault`. I think most of the time we are not really aware of using defaults, and we should be. 

I'm not sure yet, but i was thinking maybe defaults should live elsewhere, more on the caller side, and not on the fn implementation side. So this module would have `GetBlockGasLimit` and the caller will be the one handling the default, and thus be more explicit about defaults

oneeman: I kind of like the default being together with the contract call (together as in in the same file).  But it does make sense to have separate functions to make that clearer.  If so, I would (a) make the other function private, (b) make the other methods also be the same way.

contracts/random/random.go
```diff
@@ -0,0 +1,203 @@
...
+		log.Debug("Unable to find last randomness commitment in smart contract")
+	}
+
+	return lastCommitment, nil
+}
+
+// ComputeCommitment calulcates the commitment for a given randomness.
+func ComputeCommitment(vmRunner vm.EVMRunner, randomness common.Hash) (common.Hash, error) {
+	commitment := common.Hash{}
+	// TODO(asa): Make an issue to not have to do this via StaticCall
```
oneeman, line 167: I don't know what this meant.  Ideally we'd see what this was about and take care of it and/or remove the TODO.  If not, at least update the TODO to not reference `StaticCall` since now it uses vmRunner

core/vm/vmcontext/caller.go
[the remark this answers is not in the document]
mcortesi: done!

cmd/geth/main.go
mcortesi: done!

contracts/random/random.go
```diff
@@ -0,0 +1,203 @@
...
+
+// RevealAndCommit performs an internal call to the EVM that reveals a
+// proposer's previously committed to randomness, and commits new randomness for
+// a future block.
+func RevealAndCommit(vmRunner vm.EVMRunner, randomness, newCommitment common.Hash, proposer common.Address) error {
+
+	log.Trace("Revealing and committing randomness", "randomness", randomness.Hex(), "commitment", newCommitment.Hex())
+	err := revealAndCommitMethod.Execute(vmRunner, nil, common.Big0, randomness, newCommitment, proposer)
+
+	// FIXME this should be done by the caller, remove from here after
```
oneeman, line 185: Why is this necessary when it wasn't before?  The calling code seems to not have changed in this regard.  Are there other places `Execute()` is used where Finalise would be needed?

contracts/blockchain_parameters/blockchain_parameters.go
mcortesi: I did (b), and added godocs comments.

Not sure if you were suggesting `a & b` or `a | b`. I think having them public is not bad. True, they are not being used today, but doesn't seem wrong from a "what's my public interface" perspective as a module designer.

contracts/contract_call.go
mcortesi: I tend to make a different "balance" between the "you ain't gonna need it" and "it's logical to have it".

To me, it's logical to have the `value`, even if we don't use it today. Makes sense for an EVMRunner to be able to define the value on a write call, as it's part of the evm protocol.

It's true that we don't use it today, still. I don't think it adds noise; nor it adds a level of generality that it's "too much".

To be a bit extreme, is like a TDD example of having a `sum(x,y)` function that needs to pass the test `sum(4,5) == 9`. The implementation can be `return 9` and doing `return x + y` is much more than you actually need. Still, doing the latter "knowing" that it makes sense for the `sum` function to do it, it's ok. I know this is very extreme, but just to make a point.

contracts/random/random.go
oneeman, line 185: Do you happen to know why this `MakeCall` has finalise true and all the other have it false?  Also, what is the reason for adding it here rather than at the caller?

contracts/contract_call.go
mcortesi: The system does value transfers sometimes, currently only CELO transfer, doing `state.SetBalance()` but it would't be unthinkable to do a value transfer in a contract call in the future.

Also, i see that the EVMRunner could prove to be more useful than just the system call. For example for the testing library o mycelo; you can do calls, transfer, etc, operating directly to the EVM/State layer.

contracts/random/random.go
mcortesi: Asking @asaj to see if he remember or if we can remove it

contracts/blockchain_parameters/blockchain_parameters.go
oneeman: I was suggesting (a) and (b).  My inclination is to make it private unless we have a use-case or envision a future use-case.  It helps prevent accidentally using the wrong one without realizing it's the wrong one.  And it's always possible to make it public if we have a reason to in the future, and if we do that then it's having been private until now won't cause any problems.

contracts/random/random.go
mcortesi: I don't.... @asaj ?

contracts/currency/currency.go
```diff
@@ -282,12 +298,13 @@ func GetExchangeRate(currencyAddress *common.Address, header *types.Header, stat
 }
 
 // GetBalanceOf returns an account's balance on a given ERC20 currency
-func GetBalanceOf(accountOwner common.Address, contractAddress common.Address, header *types.Header, state vm.StateDB) (result *big.Int, err error) {
+func GetBalanceOf(vmRunner vm.EVMRunner, accountOwner common.Address, contractAddress common.Address) (result *big.Int, err error) {
 	log.Trace("GetBalanceOf() Called", "accountOwner", accountOwner.Hex(), "contractAddress", contractAddress)
 
-	err = contract_comm.MakeStaticCallWithAddress(contractAddress, balanceOfFuncABI, "balanceOf", []interface{}{accountOwner}, &result, params.MaxGasToReadErc20Balance, header, state)
+	err = getBalanceMethod.Bind(contractAddress).Query(vmRunner, &result, accountOwner)
```
oneeman, line 304: This is nice!

contracts/currency/currency.go
```diff
@@ -282,12 +298,13 @@ func GetExchangeRate(currencyAddress *common.Address, header *types.Header, stat
...
+
 	if err != nil {
-		log.Error("GetBalanceOf evm invocation error", "err", err)
+		log.Error("GetBalanceOf evm invocation error", "leftoverGas", "err", err)
```
oneeman, line 307: This would give an error due to the odd number of arguments after the message.  I don't remember the exact error it gives, but I've seen it many times when adding logging to debug something.

contracts/gasprice_minimum/gasprice_minimum.go
```diff
@@ -80,27 +78,43 @@ const (
...
+	gasPriceMinimum, err := GetGasPriceMinimum(vmRunner, currency)
 	return new(big.Int).Mul(gasPriceMinimum, suggestionMultiplier), err
 }
 
-func GetGasPriceMinimum(currency *common.Address, header *types.Header, state vm.StateDB) (*big.Int, error) {
+func GetGasPriceMinimum(vmRunner vm.EVMRunner, currency *common.Address) (*big.Int, error) {
 	var currencyAddress common.Address
 	var err error
 
+	// TODO this is not ideal, we fail on vmRunner == nil
```
oneeman, line 109: Why would vmRunner be nil?  And why is this necessary here but not elsewhere?  Wouldn't all the contract methods go wrong if it were nil?  Is it because of `tracers_test.go`?  If so, I would rather create some stub struct that satisfies the `VMRunner` interface for that test so that we don't need to have special handling here for what only happens in that one test.

contracts/registry.go
```diff
@@ -49,12 +48,16 @@ func init() {
 
 // GetRegisteredAddress returns the address on the registry for a given id
 func GetRegisteredAddress(vmRunner vm.EVMRunner, registryId common.Hash) (common.Address, error) {
+	if vmRunner == nil {
```
oneeman, line 51: Can this be nil?  Is it also due to the tracer test?  If so, same comment on it as https://github.com/celo-org/celo-blockchain/pull/1522/files#r632775742

core/chain_makers.go
```diff
@@ -61,7 +62,7 @@ func (b *BlockGen) SetCoinbase(addr common.Address) {
 		panic("coinbase can only be set once")
 	}
 	b.header.Coinbase = addr
-	b.gasPool = new(GasPool).AddGas(CalcGasLimit(b.parent, b.statedb)) //TODO: fix this
+	b.gasPool = new(GasPool).AddGas(params.DefaultGasLimit)
```
oneeman, line 65: I guess this only used in tests and so it doesn't really matter whether it uses the contract or the default?  I guess here it could be possible to use some kind of testing-only stub struct which satisfies VMRunner.  Though I'm guessing here it doesn't matter so probably not worth it.

Still, a couple of questions:
1. What did that `//TODO: fix this` mean?  I don't see anything obviously wrong with the old code.  does it have to do with the parent and state not being set?
2. If we're going to just use the default gas limit, update the comment on this function to remove Note since it'd no longer be applicable.

core/chain_makers.go
```diff
@@ -89,12 +90,14 @@ func (b *BlockGen) AddTx(tx *types.Transaction) {
...
 // the block in chain will be returned.
-func (b *BlockGen) AddTxWithChain(bc vm.ChainContext, tx *types.Transaction) {
+func (b *BlockGen) AddTxWithChain(bc ChainContext, tx *types.Transaction) {
 	if b.gasPool == nil {
 		b.SetCoinbase(common.Address{})
 	}
 	b.statedb.Prepare(tx.Hash(), common.Hash{}, len(b.txs))
-	receipt, err := ApplyTransaction(b.config, bc, &b.header.Coinbase, b.gasPool, b.statedb, b.header, tx, &b.header.GasUsed, vm.Config{})
+
+	caller := &fakeEVMRunner{b.statedb}
```
oneeman, line 99: Not called 'caller' anymore

contracts/gasprice_minimum/gasprice_minimum.go
oneeman: Actually, I know see you already have a `fakeEVMRunner`.  Couldn't it be used in tracers_test.go too, and if so then would we be able to remove this here?

core/chain_makers.go
```diff
@@ -276,3 +279,23 @@ func (cr *fakeChainReader) GetHeaderByHash(hash common.Hash) *types.Header
 func (cr *fakeChainReader) GetHeader(hash common.Hash, number uint64) *types.Header { return nil }
 func (cr *fakeChainReader) GetBlock(hash common.Hash, number uint64) *types.Block   { return nil }
 func (cr *fakeChainReader) StateAt(root common.Hash) (*state.StateDB, error)        { return nil, nil }
+
+type fakeEVMRunner struct {
```
oneeman, line 283: I think this is fine, but if for any reason we need some calls to succeed with certain values, we could create a fake one which you can 'program' when you create it (e.g. make it return a certain value if the method name is `getGasPriceMinimum`, or whatever.  It's not as nice as if we could mock out the higher-level function directly, but could do in a pinch.

core/evm.go
```diff
@@ -18,12 +18,33 @@ package core
...
+type ChainContext interface {
+	// Engine retrieves the blockchain's consensus engine.
+	Engine() consensus.Engine
+
+	// GetHeader returns the hash corresponding to the given hash and number.
+	GetHeader(common.Hash, uint64) *types.Header
+
+	// GetHeaderByNumber returns the hash corresponding number.
+	// FIXME: Use of this function, as implemented, in the EVM context produces undefined behavior
+	// in the pressence of forks. A new method needs to be created to retrieve a header by number
```
oneeman, line 39:
```suggestion
	// in the presence of forks. A new method needs to be created to retrieve a header by number
```

Though, since we will be disabling reorgs, maybe we don't need this FIXME at all.

core/state_processor.go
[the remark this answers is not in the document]
oneeman: It's only used in `insertChain()`.  Now, I can think of two scenarios of `insertChain()`:

1. Adding a new block to the head of the chain
2. Inserting a block into a sidechain

Though since you reviewed the disable reorg PR you know better than me whether there are other cases.

For #1, we're fine.  For #2, well, we're going to be disabling reorgs and so it won't happen.  But, even without disabling reorgs, if it does happen, wouldn't using the chain head instead of the block being inserted mean that different nodes could use different states (while processing the same block) depending on the timing (e.g. a new block comes in)?  In other words, it's not like the current behavior results in different nodes agreeing with each other in the first place.  Though this is largely academic for the long-lived networks, since `IsRunning` just checks whether the contract has been deployed and registered (i.e. at some point it goes from false to true and then it stays true).

Still, I guess if we want to be super-safe we can do this as part of a hard fork.  But then do we have to keep the conditional forever?  Or do we try to sneakily remove it and pretend it was never part of a HF?

core/tx_pool.go
```diff
@@ -1262,12 +1257,13 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	}
 	pool.currentState = statedb
 	pool.pendingNonces = newTxNoncer(statedb)
-	pool.currentMaxGas = CalcGasLimit(pool.chain.CurrentBlock(), statedb)
 
+	pool.currentVMRunner = pool.chain.NewEVMRunner(newHead, statedb)
```
oneeman, line 1261: I notice this uses `newHead` while the `currentMaxGas` before used `CurrentBlock()`.  I guess they should be the same, but if in some crazy race condition they were somehow to differ then it seems like `newHead` would be more correct.

core/tx_pool_test.go
```diff
@@ -81,10 +81,16 @@ func (bc *testBlockChain) GetHeader(common.Hash, uint64) *types.Header {
 	return nil
 }
 
+func (bc *testBlockChain) NewEVMRunner(header *types.Header, state vm.StateDB) vm.EVMRunner {
+	return nil
```
oneeman, line 85: Maybe use the fake evm runner here?

core/tx_pool_test.go
```diff
@@ -81,10 +81,16 @@ func (bc *testBlockChain) GetHeader(common.Hash, uint64) *types.Header {
...
+}
+
 func (bc *testBlockChain) GetVMConfig() *vm.Config {
 	return nil
 }
 
+// func (bc *testBlockChain) Config() *params.ChainConfig { return nil }
```
oneeman, line 92: Was this addition a mistake?  If not, explain why we want it.

core/vm/contracts_test.go
```diff
@@ -134,17 +102,7 @@ var vmctx = Context{
...
-		// chain, so return false if this is not an Istanbul chain. As a consequence of this the
-		// seal is always verified against the canonical chain, which makes behavior undefined if
-		// this function is evaluated on a chain which does not have the highest total difficulty.
-		if chainCtx.Config().Istanbul == nil {
-			return false
-		}
-
-		// Submit the header to the engine's seal verification function.
-		return chainCtx.Engine().VerifySeal(nil, header) == nil
+		return true
```
oneeman, line 105: I kind of want to ask what the deal was with the `return chainCtx.Engine().VerifySeal(nil, header) == nil`, but then again I probably don't want to know.

core/vm/interface.go
```diff
@@ -83,27 +81,8 @@ type CallContext interface {
...
-	// in the correct fork.
-	GetHeaderByNumber(uint64) *types.Header
-
-	// Config returns the blockchain's chain configuration
-	Config() *params.ChainConfig
-}
-
 // EVMRunner provides a simplified API to run EVM calls
-// EVM's sender, gasPrice, txFeeRecipient and state are set by the runner on each call
+// EVM's sender, gasPrice, txFeeRecipient and state are set by the runne on each call
```
oneeman, line 85:
```suggestion
// EVM's sender, gasPrice, txFeeRecipient and state are set by the runner on each call
```

core/vm/vmcontext/context.go
```diff
@@ -4,16 +4,37 @@ import (
...
+type chainContext interface {
+	// Engine retrieves the blockchain's consensus engine.
+	Engine() consensus.Engine
+
+	// GetHeader returns the hash corresponding to the given hash and number.
+	GetHeader(common.Hash, uint64) *types.Header
+
+	// GetHeaderByNumber returns the hash corresponding number.
+	// FIXME: Use of this function, as implemented, in the EVM context produces undefined behavior
+	// in the pressence of forks. A new method needs to be created to retrieve a header by number
```
oneeman, line 28: Same as in `core`: either fix the typo in `pressence` or remove the FIXME since it's about forks

les/odr_test.go
```diff
@@ -135,9 +135,8 @@ func odrContractCall(ctx context.Context, db ethdb.Database, config *params.Chai
 				context := core.NewEVMContext(msg, header, bc, nil)
 				vmenv := vm.NewEVM(context, statedb, config, vm.Config{})
 
-				//vmenv := core.NewEnv(statedb, config, bc, msg, header, vm.Config{})
 				gp := new(core.GasPool).AddGas(math.MaxUint64)
-				result, _ := core.ApplyMessage(vmenv, msg, gp)
+				result, _ := core.ApplyMessage(vmenv, msg, gp, nil)
```
oneeman, line 139: More places where we use nil.  Again, I would prefer using a fake evm runner if it allows us to remove the handling for nil in GetGasPriceMinimum.  Same with the one 11 lines below this one.

light/txpool.go
```diff
@@ -389,19 +389,17 @@ func (pool *TxPool) validateTx(ctx context.Context, tx *types.Transaction) error
 		return core.ErrNegativeValue
 	}
 
+	caller := pool.chain.NewEVMRunner(pool.chain.CurrentHeader(), currentState)
```
oneeman, line 392: rename to runner

miner/worker.go
```diff
@@ -693,7 +695,7 @@ func (w *worker) updateSnapshot() {
 func (w *worker) commitTransaction(tx *types.Transaction, txFeeRecipient common.Address) ([]*types.Log, error) {
 	snap := w.current.state.Snapshot()
 
-	receipt, err := core.ApplyTransaction(w.chainConfig, w.chain, &txFeeRecipient, w.current.gasPool, w.current.state, w.current.header, tx, &w.current.header.GasUsed, *w.chain.GetVMConfig())
+	receipt, err := core.ApplyTransaction(w.chainConfig, w.chain, &txFeeRecipient, w.current.gasPool, w.current.state, w.current.header, tx, &w.current.header.GasUsed, *w.chain.GetVMConfig(), w.chain.NewEVMRunner(w.current.header, w.current.state))
```
oneeman, line 698: Nit: I would use an intermediate variable `vmRunner` here, since that's what we do in most such places and it's possibly nicer.

miner/worker.go
```diff
@@ -919,50 +921,64 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
...
-	if w.isRunning() && random.IsRunning() {
-		istanbul, ok := w.engine.(consensus.Istanbul)
-		if !ok {
-			log.Crit("Istanbul consensus engine must be in use for the randomness beacon")
+	if w.isRunning() {
+		// TODO use caller instead of parentCaller (HF required)
+		randomIsActive := false
+		header := w.chain.CurrentHeader()
+		state, err := w.chain.StateAt(header.Root)
+		// TODO(HF) make this check use w.current.header instead of blockchain's current
```
oneeman, line 929: Same comment as elsewhere as to whether this is really a HF and whether there is a reason to keep it this way.

miner/worker.go
```diff
@@ -919,50 +921,64 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
...
+		if err == nil {
+			parentVMRunner := w.chain.NewEVMRunner(parent.Header(), state)
```
oneeman, line 931: This looks strange to me.  The origin `random.IsRunning()` uses the current chain head, whatever it may be.  Here you are using the state from the current chain head with the header from `parent`.  Now, if `parent` and `header` are from the same block, why not use `header` here instead?  On the other hand, if they're from different blocks, then the runner gets a header and a state which don't correspond.

My preference would be to just use a runner like `vmRunner` below which uses `w.current`, as discussed in the other comments.  But if instead you'd rather keep the old behavior, then this should use `header` instead of `parent.Header()` to match it more closely and be consistent in and of itself.

light/lightchain.go
```diff
@@ -501,10 +502,26 @@ func (lc *LightChain) GetHeaderByNumberOdr(ctx context.Context, number uint64) (
...
 
+// NewEVMRunner creates the System's EVMRunner for given header & sttate
+func (lc *LightChain) NewEVMRunner(header *types.Header, state vm.StateDB) vm.EVMRunner {
+	return vmcontext.NewEVMRunner(lc, header, state)
+}
+
+// NewEVMRunnerForCurrentBlock creates the System's EVMRunner for current block & state
+func (lc *LightChain) NewEVMRunnerForCurrentBlock() (vm.EVMRunner, error) {
+	header := lc.CurrentHeader()
+	// FIXME small race condition here. Need to make sure state matches header
```
oneeman, line 517: I believe we can avoid the race condition by doing:
```
state, err := NewState(context.Background(), header, bc.odr), nil
```

Essentially, doing the same as what `lc.State()` does except with this header rather than have it call `lc.CurrentHeader()` again.

core/blockchain.go
oneeman: Suggested fix for lightchain at https://github.com/celo-org/celo-blockchain/pull/1522/files#r632842994 .  Marking this one as resolved.

contracts/contract_call.go
piersy: If its just coincidental that _right now_ we do not happen to use the send value functionality, then I think it's not a problem and it is logical to have the ability to pass value.

But I don't think it is coincidental. As I understand, the system makes calls from the zero address, so if the system were ever to make use of the send value functionality it would be sending celo from the zero address to another address? Looking at the zero address on the celo explorer I can see that the balance is and always has been 0. So it seems to me that unless we undertake some systemic change there will never be a situation where the system account wants to send value.

I'm confused about your point that the vm does do value transfers using `state.SetBalance()` given that the system has a zero balance?  But we can see it is not doing that through the EVMRunner otherwise there would be a call with a non zero value.

I feel that EVMRunner is sitting on the fence, on the one hand it has elements that mean that it is only suitable for system calls but on the other it still has elements that support non system calls. I don't think this is a good state to be in, it's confusing for people and opens up avenues for misuse (I.E actually passing a non zero value in any of the places where it is currently used). So I think we should decide is it a tool just for use with system calls or not?

core/state_transition.go
```diff
@@ -487,7 +488,7 @@ func (st *StateTransition) distributeTxFees() error {
 	caller := &vmcontext.SharedEVMRunner{EVM: st.evm}
```
piersy, line 488: I would really like to get rid of this. The original purpose of the EVMRunner was to enforce that we couldn't use an evm more than once, so the existence of the SharedEVMRunner undermines the existence of EVMRunner and essentially renders it useless as a tool to enforce singular use of evms. So until we get rid of this I don't see that we get any benefit from the addition of the concept of EVMRunner.

I think it is also possible that the EVMRunner abstraction may not turn out to fit particularly well once we do the work to remove SharedEVMRunner and so the sooner we can do that work the better, because less will have been built on top of it and so less will need to be re-worked.

mcortesi (commented):
zz

core/vm/contracts_test.go
mcortesi: lol... it's not that bad. This test used a mock engine where VerifySeal is implemented as `return nil`, so doing `nil == nil` is alwas true

core/blockchain.go
mcortesi: I believe not, since `NewHeaderChain()` already sets it and use it. And it should be a big bug if someone did `currentHeader.Store(nil)`

core/state_processor.go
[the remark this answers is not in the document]
mcortesi: i'll take the simplistic approach an use current state.

As you said, this only happens once over the blockchain history. And that's on migrations. And it's unlikely that you call `Process()` on a block that is not currentHead +1.

For starters, consensus engine will also use a N+1 block, then fetcher/downloader will only use N+1. Maybe a `SetHead()` or `reorg`, but i dont' think so. And even if that were the case, it is a bug with un specified behvaiour, so better to fix this here.

core/state_transition.go
mcortesi: piers.... 

1. If i use a normal EVMRunner here, that implies a hardfork
2. If i don't use, i need to do some custom thing for this particular case

On the next HF, i can then easily remove this, and use the normal VMRunner. Until then, sadly i can't.

So i ask you not to be so "absolute" about this... maybe wait a few months (next HF) until we're finally able to remove this, instead of saying that EVMRunner doesn't really do anything, because we have the SharedEVMRunner patch 😠

core/tx_pool.go
mcortesi: true, i'd say new code fixes the race condition. 

And fixing it doesn't generate a HF, so we're fine

miner/worker.go
mcortesi: Changed to use vmRunner from current header & state

miner/worker.go
mcortesi: Changed to use vmRunner from current header & state

light/txpool.go
mcortesi: done

core/tx_pool_test.go
mcortesi: Using the new Mock VMRunner

contracts/registry.go
mcortesi: removed

les/odr_test.go
mcortesi: using celomock runner now

contracts/gasprice_minimum/gasprice_minimum.go
mcortesi: using celoMock runner now

contracts/contract_call.go
mcortesi: that would change the previous API, so i wouldn't risk it

contracts/contract_call.go
mcortesi: well... EVMRunner is not necessarily a System EVMRunner, meaning that sender doesn't always need to be the system. That was my original idea.. .but then i was "suggested" to make that change, so that it's always the system.

So, now, that logic lives only on the `NewEVMRunner()` code, and could be easily changed by adding a parameter.

I just think you're being "too picky" about it... yet again my dear piers.... having all these conversations about what the EVMRunner should or should not be... it is exhausting.

core/blockchain.go
mcortesi: this PR is just too big as it is, so doing a architectural change so everything lives in another place, not the blockchian is to me work for another PR. Still, i'm not convinced that would be a great idea thinking on tradeoff valueObtained vs upstreamMergeComplexity. But we can chat about it later.

trianglesphere:
Looks like something with the flags got messed up. All the e2e tests are immediately failing, and when I ran a local mycelo instance I go the error: `Fatal: Miners and Proxies must be run as a full node`

cmd/geth/main.go
```diff
@@ -445,9 +444,11 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 			}
 		}()
 	}
+
+	isFullNode := ctx.GlobalString(utils.SyncModeFlag.Name) == "full" || ctx.GlobalString(utils.SyncModeFlag.Name) == "fast"
```
trianglesphere, line 448: Seems like `&&` go turned into `||`

trianglesphere:
Looks like some of the side changes caused some issues. I definitely don't always succeed at this, but creating smaller PRs (with less side changes) makes them at a lot easier to review.
The main migration looks fine.

cmd/geth/main.go
mcortesi: yeah, actually this is correct, but the if below as `!isFullNode`

trianglesphere (approved):
finally getting rid of the singleton

looks good to me

piersy (approved):
I still have some reservations about this design, but lets see how it goes!

PR #1523, opened 2021-04-20
Removes unnecessary arguments/returns to GetBalance

mcortesi:
### Description

Everycall used the same maxGas parameter, so instead of passing it, we
use it directly. This is what every other call in the system does.

Also, usedGas was not really used as a return parameter, so removed

### Tested

* unit tests, ci


gastonponti (approved):
LGTM

PR #1503, opened 2021-04-12
Fix homebrew error in CI (due to outdated version).

oneeman:
### Description

This fixes the error seen on the ios build in CI:

```
==> Downloading https://homebrew.bintray.com/go-1.16.3.catalina.bottle.tar.gz
##O#- #                                                                       
curl: (22) The requested URL returned error: 403 Forbidden
Trying a mirror...
==> Downloading https://ghcr.io/v2/homebrew/core/go-1.16.3.catalina.bottle.tar.g
==> Downloading from https://github.com/-/v2/packages/container/package/homebrew
#=#=#                                                                         
curl: (22) The requested URL returned error: 404 
Error: Failed to download resource "go"
Download failed: https://ghcr.io/v2/homebrew/core/go-1.16.3.catalina.bottle.tar.gz
Exited with code exit status 1
CircleCI received exit code 1
```

The fix comes recommended from https://github.com/Homebrew/brew/issues/11119.

### Tested

Confirmed as working by the CI checks.

PR #1531, opened 2021-05-05
Fix ios build in CI (again)

oneeman:
### Description

We previously added `brew update` in #1503.  However, since then `brew update` has started giving the following error:

```
Error: 
  homebrew-core is a shallow clone.
  homebrew-cask is a shallow clone.
To `brew update`, first run:
  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core fetch --unshallow
  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask fetch --unshallow
These commands may take a few minutes to run due to the large size of the repositories.
This restriction has been made on GitHub's request because updating shallow
clones is an extremely expensive operation due to the tree layout and traffic of
Homebrew/homebrew-core and Homebrew/homebrew-cask. We don't do this for you
automatically to avoid repeatedly performing an expensive unshallow operation in
CI systems (which should instead be fixed to not use shallow clones). Sorry for
the inconvenience!
```

This PR removes the `brew update`, which should work if CI now has a recent version of homebrew.

### Tested

* We'll see in CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,7 +116,6 @@ jobs:
       - run:
           name: Setup Go language
           command: |
-            brew update
             brew install go@1.16
             brew link go@1.16
             # Check that homebrew installed the expected go version


trianglesphere:
I saw this as well. TBH can we directly install golang without homebrew?

trianglesphere (approved):

oneeman:
> I saw this as well. TBH can we directly install golang without homebrew?

I'd be up for that if we run into more homebrew troubles in the future

PR #1536, opened 2021-05-06
Upstream merge up to 1.9.16

hbandura:
1.19.16 notes

ethstats/ethstats.go conflicts (we have a cherrypicked version). Still was missing a line, added it.

TxFee cap: possible issue with currencies at:
internal/ethapi/api.go SubmitTransaction (RPCTxFeeCap check): instead of the flat comparison brought by upstream, I'm using a currency manager for this

cmd/evm/internal/t8ntool/execution.go:113&114 -> commented to ensure compilation, not sure
what the impact of this is.

Changeset https://github.com/ethereum/go-ethereum/commit/ddeea1e0c68df82c1190d3ee71aae87ed3d92bfe changes comparisons on gas prices, conflicts with our own since we also use feecurrency. I kept our versions of code in some of these (tx_pool.go, tx_list.go)

Skipped new test cmd/geth/les_test.go , may need to modify the configuration of the startup? (priority client test)

Big changes in a commit for tx_list.go and tx_pool.go got reverted in this same PR (This was particularly a pain because resolving the conflicts was hard, only to then resolve the same conflicts again when they had to be undone).
- https://github.com/ethereum/go-ethereum/commit/af5c97aebe1d37486635521ef553cb8bd4bada13
- https://github.com/ethereum/go-ethereum/commit/bcb308745010675671991522ad2a9e811938d7fb


core/tx_list.go
```diff
@@ -100,7 +100,30 @@ func (m *txSortedMap) Forward(threshold uint64) types.Transactions {
...
+	for nonce := range m.items {
+		*m.index = append(*m.index, nonce)
+	}
+	heap.Init(m.index)
+	m.cache = nil
+}
+
+// filter is identical to Filter, but **does not** regenerate the heap. This method
+// should only be used if followed immediately by a call to Filter or reheap()
+func (m *txSortedMap) filter(filter func(*types.Transaction) bool) types.Transactions {
```
gastonponti: why didn't they used another name for this?????????

core/tx_list.go
```diff
@@ -659,6 +705,34 @@ func (l *txPricedList) Underpriced(tx *types.Transaction, local *accountSet) boo
...
+			for _, tx := range *items {
+				if !local.containsTx(tx) {
+					discardable++
+				}
+				if discardable >= slots {
+					break
+				}
+			}
+		}
+		// if slots > discardable {
```
gastonponti: Added from upstream or something that we did?

eth/downloader/downloader.go
```diff
@@ -369,22 +371,7 @@ func (d *Downloader) Synchronise(id string, head common.Hash, td *big.Int, mode
 		}
 		return err
 	}
-
-	switch err {
```
gastonponti, line 372: why not adding the `errInvalidChain` to this switch? it was an upstream change? (it's more readable as a switch in my opinion)

core/tx_list.go
```diff
@@ -100,30 +100,7 @@ func (m *txSortedMap) Forward(threshold uint64) types.Transactions {
 
 // Filter iterates over the list of transactions and removes all of them for which
 // the specified function evaluates to true.
-// Filter, as opposed to 'filter', re-initialises the heap after the operation is done.
-// If you want to do several consecutive filterings, it's therefore better to first
-// do a .filter(func1) followed by .Filter(func2) or reheap()
 func (m *txSortedMap) Filter(filter func(*types.Transaction) bool) types.Transactions {
-	removed := m.filter(filter)
```
gastonponti: mmmmmmm
This was added a few commits ago... they added it and revert it in the same release?

eth/config.go
```diff
@@ -50,7 +50,7 @@ var DefaultConfig = Config{
 
 	TxPool:      core.DefaultTxPoolConfig,
 	RPCGasCap:   25000000,
-	RPCTxFeeCap: 1, // 1 celo
+	RPCTxFeeCap: 100 * params.Ether, // 100 celo
```
gastonponti: Ok, I think I ask this in an old commit. You can ignore that question

internal/ethapi/api.go
```diff
@@ -1560,8 +1560,9 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
 		if err != nil {
 			return common.Hash{}, err
 		}
-		if currencyManager.CmpValues(tx.Fee(), tx.FeeCurrency(), big.NewInt(int64(b.RPCTxFeeCap())), nil) > 0 {
-			return common.Hash{}, fmt.Errorf("tx fee (%d wei) exceeds the configured cap (%d wei)", tx.Fee().Uint64(), int64(b.RPCTxFeeCap()))
+		feeCap := GetWei(b.RPCTxFeeCap())
+		if currencyManager.CmpValues(tx.Fee(), tx.FeeCurrency(), feeCap, nil) > 0 {
+			return common.Hash{}, fmt.Errorf("tx fee (%d celo) exceeds the configured cap (%d celo)", tx.Fee().Uint64(), int64(b.RPCTxFeeCap()))
```
gastonponti, line 1565: why not wei in the message? it could be a stable token, right?

cmd/geth/les_test.go
```diff
@@ -0,0 +1,149 @@
...
+	freeCli.callRPC(&peers, "admin_peers")
+	if len(peers) != 1 {
+		t.Errorf("Expected: # of client peers == 1, actual: %v", len(peers))
+		return
+	}
+
+	// Set up priority client, get its nodeID, increase its balance on the lightServer
+	prioCli := startClient(t, "prioCli")
+	defer prioCli.killAndWait()
+	// 3_000_000_000 once we move to Go 1.13
```
gastonponti, line 122: we are even better than that

cmd/geth/les_test.go
```diff
@@ -0,0 +1,149 @@
...
+	return server
+}
+
+func startClient(t *testing.T, name string) *gethrpc {
+	datadir := initGeth(t)
+	return startGethWithRpc(t, name, "--datadir", datadir, "--nodiscover", "--syncmode=light", "--nat=extip:127.0.0.1")
+}
+
+func TestPriorityClient(t *testing.T) {
+	t.Skip() // Currently not working
```
gastonponti, line 100: Do we think that we should skip this one, or maybe addapt it?

core/vm/stack.go
```diff
@@ -18,36 +18,36 @@ package vm
...
 // expected to be changed and modified. stack does not take care of adding newly
 // initialised objects.
 type Stack struct {
-	data []*big.Int
+	data []uint256.Int
 }
 
 func newstack() *Stack {
-	return &Stack{data: make([]*big.Int, 0, 1024)}
+	return &Stack{data: make([]uint256.Int, 0, 16)}
```
gastonponti, line 33: 1024 -> 16? is this right?

gastonponti (commented):
There are a few comments that were answered in different commits, but was part of the way I reviewed the PR (per commit)
Probably will be better to read every comment before answering

core/vm/stack.go
hbandura: It seems so, yes. Remember that number is the capacity of the slice, that means the preallocated size, not necessarily the max or current size

hbandura: https://github.com/ethereum/go-ethereum/blob/master/core/vm/stack.go#L28

cmd/geth/les_test.go
hbandura: Since it's an extra test, I'd say that whatever road we decide on, we could do that on a separate PR. I don't know the answer though. I think we can create an issue for that (decide & resolve)

cmd/geth/chaincmd.go
```diff
@@ -305,8 +305,10 @@ func importChain(ctx *cli.Context) error {
 	// Import the chain
 	start := time.Now()
 
+	var importErr error
 	for _, arg := range ctx.Args() {
 		if err := utils.ImportChain(chain, arg); err != nil {
```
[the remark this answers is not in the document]
hbandura: This is the commit from upstream that brought this message:
https://github.com/celo-org/celo-blockchain/commit/413358abb915dd2cfd948f2f9a58b5bff3e672fe

while we could think of a way to maybe improve on it, the change itself is an improvement over the previous version of code that would always return nil even in the face of errors

internal/ethapi/api.go
hbandura: it can be any currency, yes. What do you think the err message should look like?

core/rawdb/freezer_table.go
```diff
@@ -232,8 +232,8 @@ func (t *freezerTable) repair() error {
 	t.index.ReadAt(buffer, 0)
 	firstIndex.unmarshalBinary(buffer)
 
-	t.tailId = firstIndex.offset
-	t.itemOffset = firstIndex.filenum
+	t.tailId = firstIndex.filenum
```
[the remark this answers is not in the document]
hbandura: funny one

cmd/geth/les_test.go
hbandura: Created #1557 for this

cmd/geth/testdata/clique.json
```diff
@@ -8,14 +8,12 @@
     "byzantiumBlock": 0,
     "constantinopleBlock": 0,
     "petersburgBlock": 0,
-    "clique": {
```
gastonponti: If it's not a `clique` config anymore, we should probably rename this file

gastonponti (approved):
LGTM
(Try to rename the file it's not a clique thing anymore)

PR #1535, opened 2021-05-06
Use Go v1.16.4 and release celo-blockchain v1.3.2

oneeman:
The `golang:alpine-1.16` docker image has not been updated yet, contrary to what it says on docker hub's website.  Once it's updated, I will amend commit and force push to trigger a new build which will use 1.16.4

### Description

Within less than an hour after we released v1.3.1, a Golang security patch was announced: https://groups.google.com/g/golang-announce/c/cu9SP4eSXMc/m/vBLImzQOAgAJ .  This PR updates celo-blockchain to use the new Golang release, and set the celo-blockchain version to v1.3.2.  

### Tested

* CI
* Confirmed that the generated docker image uses go version 1.16.4:

```
❯ docker run -it us.gcr.io/celo-testnet/geth:2ea8e9c0f9f7483bb6cfb3ad8410791fb8bd6acc version
Celo
Version: 1.3.2-stable
Architecture: amd64
Protocol Versions: [66 65 64]
Go Version: go1.16.4
Operating System: linux
GOPATH=
GOROOT=/usr/local/go
```

### Backwards compatibility

No backwards compatibility issues.

PR #1537, opened 2021-05-07
Use golang v1.16.4

oneeman:
### Description

Go v1.16.4 is a security release.  We've upgraded to it in celo-blockchain v1.3.1 (#1535).  This PR also upgrades on master.  Note that, as usual, the places where we use the `golang:1.16-alpine` image don't need to be modified as the newest 1.16 image will be used automatically.

### Tested

* CI

### Backwards compatibility

No incompatibilities.


gastonponti (approved):
LGTM

Issue #1426, opened 2021-03-10
Implement pre-sharing of proposed block


nategraf:
* [x] Develop a end to end miner test
* [x] Re-structure the miner to enable this change. (in progress)
* [ ] Develop a prototype implementation of block pre-constructing
* [ ] Develop a prototype implementation of block pre-sharing
* [ ] Deploy a testnet with the change

mcortesi:
streamming block construction superseeds this

PR #1545, opened 2021-05-12
Refactor the miner

### LINKED ISSUE #1419: RPC "pending" block can be behind "latest" block
## Expected Behavior

When sending RPC requests, you can optionally specify the block number to use or "latest" or "pending" (or "earliest", but that's not relevant for this bug).  "latest" means the to use the node's current blockchain head, while "pending" means to use what the miner's worker (which runs even when the `--mine` flag isn't used, and is necessary in order to have a "pending" state) has, which should be with pending transactions from the tx pool added to the latest block.

Therefore, when you make a request using "pending", you shouldn't get state that's older than when you use "latest".

This is particularly significant because gas estimation uses the "pending" state: https://github.com/celo-org/celo-blockchain/blob/0c404fddb51b2339eda960fbe6bf8db6e1ac121b/internal/ethapi/api.go#L958

Therefore, this could be the behavior which has been observed in the wild of gas estimation sometimes failing as part of scripts or test cases when it shouldn't (which has been suspected to be due to different requests going to different Forno nodes, but could actually be due to this bug instead).  I originally discovered this bug due to some e2e tests giving errors during gas estimation when they shouldn't and it seems to be due to this.

## Current Behavior

Due to one or more race conditions, you can sometimes get an older state (from the previous block) if you use "pending" compared to "latest".  I've observed this on full nodes during e2e tests as well as using Alfajores forno (with a websocket connection so that the requests go to the same node).

What's going on is that the miner's worker (worker.go) uses the istanbul engine and doing some things which were intended for validators and include some waiting:

https://github.com/celo-org/celo-blockchain/blob/d3b153fe00407494c212902bc12b947c778b81f8/consensus/istanbul/backend/engine.go#L392

https://github.com/celo-org/celo-blockchain/blob/d3b153fe00407494c212902bc12b947c778b81f8/consensus/istanbul/backend/engine.go#L1201-L1220

I'm not clear on the exact nature of the race conditions, because the worker's `updateSnapshot()` function is called from several places.

Note also that on non-mining nodes there is another way that transactions get added to the pending state: https://github.com/celo-org/celo-blockchain/blob/8cfe3d15966eb99deb80ba82686230ca588e91dd/miner/worker.go#L460-L494

But this transactions are applied on top of the worker's current state, which will generally be the pending state from the previous block due to the sleep.

Run this script, with two private keys for Alfajores (the first one should have some funds to send a transaction with).  I believe the use of websocket in the script ensures both calls go to the same node.

```
// tslint:disable:no-console
import { newKitFromWeb3 } from '@celo/contractkit'
import { privateKeyToAddress } from '@celo/utils/lib/address'
import BigNumber from 'bignumber.js'
import Web3 from 'web3'

// Use websocket, which guarantees we are sending all requests to the same node
const forno = 'wss://alfajores-forno.celo-testnet.org/ws'
const pk1 = process.env.ALFAJORES_PK_1!
const pk2 = process.env.ALFAJORES_PK_2!

async function main() {
  if (!pk1 || !pk2) {
    console.log('Please set ALFAJORES_PK_1 (with funds) and ALFAJORES_PK_2 (empty)')
    process.exit(1)
  }

  const kit = newKitFromWeb3(new Web3(forno))
  kit.addAccount(pk1)
  const addr1 = privateKeyToAddress(pk1)
  console.log('Account 1:', addr1)
  kit.addAccount(pk2)
  const addr2 = privateKeyToAddress(pk2)
  console.log('Account 2:', addr2)
  const amount = new BigNumber('1')
  console.log('Sending 1 wei from acc 1 to 2')
  const res = await kit.sendTransaction({
    from: addr1,
    value: amount.toString(),
    to: addr2,
  })
  console.log('Awaiting first tx conf')
  const receipt1 = await res.waitReceipt()
  console.log(
    `First transaction confirmed in block ${receipt1.blockNumber} with tx hash ${receipt1.transactionHash}`
  )
  const pending = kit.web3.eth.getBalance(addr2, 'pending')
  const latest = kit.web3.eth.getBalance(addr2, 'latest')
  console.log('Latest:', await latest)
  console.log('Pending:', await pending)
  process.exit(0)
}

main()
  .then()
  .catch()

```

It doesn't happen every time, but maybe around half the time the pending balance is 1 wei less than the latest balance, indicating it uses the state from before the transaction while latest uses the state from after the transaction.

I've observed this on full nodes, but not on validators.  But given the race conditions I'm really not clear on how it happens and when it doesn't.

This could probably be worked around by adding an `updateSnapshot()` call earlier on (e.g. right after `makeCurrent()` is called), but this would not be a clean solution.  And the fact that full nodes are running the istanbul engine at all is probably not how things should be.

trianglesphere:
### Description

This refactors the miner with a couple goals
1. Simplify the block construction logic.
2. Enable pre-construciton / pre-sharing of blocks in the miner
3. Fix the pending state bug

I worked towards these goals by

- Reducing the number of go-routines passing work to each other. There is now 1 main loop goroutine.
- Make the block creation process as linear as possible (and use contexts to cancel instead of global interrupts).
    - Each step in the block production cycle is now done in `miner/block.go`
    - There is no longer a shared "current environment" for the in flight block, but a new environment is made for each request (which is the `blockState` struct).
    - Block production is cancelable with contexts instead of global interrupts
    - I have removed as much blocking as possible from `engine.Prepare`
      - It no longer sleeps until the block time - that is done in the worker
      - It does not wait until the next sequence for nodes that are not validating

### Other changes

- Removes proof of work specific miner interfaces.
- Removes miner/unconfirmed.go
- Includes some upstream changes made to `miner/miner.go` to simplify the start up logic.

### Tested

Unit tests still pass.

Tested end to end on `trianglesphere/miner_refactor_v2`
TODO: test that this fixes the pending bug.

### Related issues

- Fixes #1419
- Partial fix of #1426

### Backwards compatibility

Yes


miner/block.go
```diff
@@ -0,0 +1,400 @@
...
+				*cpy[i] = *l
+			}
+			w.pendingLogsFeed.Send(cpy)
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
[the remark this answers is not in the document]
trianglesphere: Yea, the previous implementation did not send the pending logs in this case. I'm not really sure of where we use pendingLogs / what the semantics are, but I'll fix this so it retains it's behavior.

miner/miner.go
```diff
@@ -67,132 +64,145 @@ type Miner struct {
...
+				continue
 			}
 			switch ev.Data.(type) {
 			case downloader.StartEvent:
-				atomic.StoreInt32(&miner.canStart, 0)
-				if miner.Mining() {
-					miner.Stop()
-					atomic.StoreInt32(&miner.shouldStart, 1)
+				wasMining := miner.Mining()
+				miner.worker.stop()
```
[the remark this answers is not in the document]
trianglesphere: I don't think it really matters either way. `worker.stop` is fine to call even if it's stopped.

oneeman:
Haven't reviewed the code, but I think this is important:

> * Re-introduce the transaction sub channel and apply each transaction as it comes in to full nodes (may want two different loops in that case)
> * construct multiple "pending" blocks per cycle for full nodes

Am I right in understanding that these two are being considered as alternatives to serve the same purpose?  At first glance I would have a strong preference for the first option.

The pending state is used for:

1. Gas estimation
2. In contractkit to get the nonce to use (https://github.com/celo-org/celo-monorepo/blob/4efbf24277bf5045bfe4cb3af8f2ff6be217a7a2/packages/sdk/connect/src/connection.ts#L377-L386)

In order for a script or program to be able to send multiple transactions without waiting each time for the previous one to be confirmed, you need the pending state to quickly be updated when a transaction is submitted.  Otherwise, we're introducing a different pending state bug.  And I think to make the pending state as responsive as possible (to minimize the cases of it being used before it's been updated with a just-added transaction), I think using a channel makes more sense than a periodic job.

miner/worker.go
```diff
@@ -403,177 +243,91 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
-		case <-w.exitCh:
-			interrupt()
-			return
-		}
+	if w.skipSealHook != nil && w.skipSealHook(task) {
+		return
+	}
+
+	w.pendingMu.Lock()
+	w.pendingTasks[w.engine.SealHash(task.block.Header())] = task
```
mcortesi, line 326: can we make a change to remove this concept? Maybe i'm mistaken, but to me there can only be ONE sealing task at any moment, not many. What can happen, is that i need to abort previous one, and start a new one; but no more than that.

miner/worker.go
```diff
@@ -403,177 +243,91 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+	w.pendingMu.Unlock()
+
+	if err := w.engine.Seal(w.chain, task.block, w.resultCh, w.prevTaskStopCh); err != nil {
```
mcortesi, line 329: i imagine this can be refactor to something like:

```go
ctx := // create context (save in to local field so we can cancel it )
go func() {
  sealedBlock, err := engine.Seal(ctx)
  // check ctx
  // do whatever we do on resultLoop()
}()


```

miner/block.go
```diff
@@ -0,0 +1,397 @@
...
+func (w *worker) constructAndSubmitNewBlock(ctx context.Context) {
+	start := time.Now()
+
+	// Initialize the block.
+	b, err := w.prepareBlock()
+	if err != nil {
+		log.Error("Failed to create mining context", "err", err)
+		return
+	}
+	w.updateSnapshot(b)
```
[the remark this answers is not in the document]
mcortesi: It seems it's being use to define th `PendingBlock` which is then use on the RPC. Now if that's true, i wonder that's the case of have a block that is "half-done". I can imagine the value of having a block that is not yet "sealed" but already has all the txs in it, but what's point of having one that has only passed the "prepare" state.

More generally, we could rethink the definition of `pending` block. For example, pending block should be `blockchain.currentHead` or `block with seal in progress`. So you would only set the snapshot just before launching a seal task, not before

miner/block.go
```diff
@@ -0,0 +1,397 @@
...
+
+	block, err := b.finalizeAndAssemble(w)
+	if err != nil {
+		log.Error("Failed to finalize and assemble the block", "err", err)
+		return
+	}
+	w.updateSnapshot(b)
+	if w.isRunning() {
```
mcortesi, line 88: we could change `constructAndSubmitNewBlock` to `createNewBlock() blockInfo`. So, it will be responsible to create a new block, add txs to it, and will return the create block (plus state, receipt, etc) as a result.

That way, it's less couple with the overall worker, doesn't need to know about channel, and the overall block flow.

Then caller will then get the result and submit it to the channel.

That way, the `pipeline` is defined on a single place:

```go
func blockPipeline(ctx) {
  block, receipts, state, created := createNewBlock(ctx)
  sealedBlock, err := engine.Seal(ctx, task.block)
  // result loop code
  addBlockToChain()
}
```

miner/block.go
```diff
@@ -0,0 +1,397 @@
...
+		if w.fullTaskHook != nil {
+			w.fullTaskHook()
+		}
+		w.submitTaskToEngine(&task{receipts: b.receipts, state: b.state, block: block, createdAt: time.Now()})
+
+		feesEth := totalFees(block, b.receipts)
```
mcortesi, line 94: there's probably some what to check current log level, and avoid calling `totalFees()` if log level is lower than INFO

miner/block.go
```diff
@@ -0,0 +1,397 @@
...
+		state:          state,
+		tcount:         0,
+		gasLimit:       core.CalcGasLimit(parent, state),
+		header:         header,
+		txFeeRecipient: txFeeRecipient,
+	}
+	b.gasPool = new(core.GasPool).AddGas(b.gasLimit)
+
+	// Play our part in generating the random beacon.
+	if w.isRunning() && random.IsRunning() {
```
mcortesi, line 158: not part of this PR, but i would love to see all randomenss related code in a single module, which is being called from the different places we call it today (here & blockchain)

miner/worker.go
```diff
@@ -403,177 +243,91 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
 
-// taskLoop is a standalone goroutine to fetch sealing task from the generator and
-// push them to consensus engine.
-func (w *worker) taskLoop() {
-	var (
-		stopCh chan struct{}
-		prev   common.Hash
-	)
+// interrupt aborts the in-flight sealing task.
+func (w *worker) interruptSealingTask() {
```
mcortesi, line 300: take a look at my other comments, but if we follow what i suggested, where basically we have a "mainLoop" only, that creates a go routine that runs the whole block creation process, we can greatly simplify this, by using contexts.

The main loop will create a new context for each block, and it will cancel it on stop()

mcortesi (commented):
Loving this PR, added several suggestion about how further improve the readability!

miner/block.go
trianglesphere: You need to split the pipeline into the top half and the bottom half [(creation & seal) an (result loop & insertion)] as the result that you get back from the results loop channel does not have to be associated with the task that the specific instance of the pipeline submitted. As such I don't want to mix up state by working on different tasks in the same local (function) context.

trianglesphere:
> Am I right in understanding that these two are being considered as alternatives to serve the same purpose? At first glance I would have a strong preference for the first option.

They'd be for essentially the same purpose yes.

> Otherwise, we're introducing a different pending state bug. 

After thinking through this, I think so as well. I wasn't that convinced that we were hitting the code path to actually add pending transactions, but I think that may have been a bad check and that we are probably hitting this code for full nodes.

miner/block.go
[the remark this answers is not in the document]
trianglesphere: It's actually interesting that it'd probably be possible to get rid of the pending block from the validator loop, but retain it for the full node loop.

miner/block.go
```diff
@@ -0,0 +1,397 @@
...
+		b.randomness = &types.Randomness{Revealed: lastRandomness, Committed: newCommitment}
+	} else {
+		b.randomness = &types.EmptyRandomness
+	}
+
+	return b, nil
+}
+
+// selectAndApplyTransactions selects and applies transactions to the in flight block state.
+func (b *blockState) selectAndApplyTransactions(ctx context.Context, w *worker) error {
```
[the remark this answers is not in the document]
trianglesphere: I don't like 2 as a solution (we're mutating blockState, but not the worker)
For 1: It would be possible, but a fair amount of work to pull disentangle the worker (this code relies on accessing a lot of data directly)

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
-	}
+// mainLoop is a standalone goroutine to create tasks and submit to the engine.
+func (w *worker) mainLoop() {
+	defer w.chainHeadSub.Unsubscribe()
+	defer w.txsSub.Unsubscribe()
+	// Context and cancel function for the currently executing block construction
+	// Cancel needs to be called in each exit path to make the linter happy
+	// because go struggles with analyzing lexical scoping.
+	var taskCtx context.Context
+	var cancel context.CancelFunc
```
mcortesi, line 355: suggest you add:
```go
defer func() {
			if cancel != nil {
				cancel()
			}  
}()
```

here, and remove the checks on "stop". seems simpler

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
-		}
-		// Short-circuit if the transaction requires more gas than we have in the pool.
-		// If we didn't short-circuit here, we would get core.ErrGasLimitReached below.
-		// Short-circuiting here saves us the trouble of checking the GPM and so on when the tx can't be included
-		// anyway due to the block not having enough gas left.
-		if w.current.gasPool.Gas() < tx.Gas() {
-			log.Trace("Skipping transaction which requires more gas than is left in the block", "hash", tx.Hash(), "gas", w.current.gasPool.Gas(), "txgas", tx.Gas())
-			txs.Pop()
-			continue
+	newWorkFunc := func() {
```
mcortesi, line 359: maybe calling it `generateNewBlock()` or similar is better that `newWorkFun`

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
-		limit := new(big.Int).Add(daoBlock, params.DAOForkExtraRange)
-		if header.Number.Cmp(daoBlock) >= 0 && header.Number.Cmp(limit) < 0 {
-			// Depending whether we support or oppose the fork, override differently
-			if w.chainConfig.DAOForkSupport {
-				header.Extra = common.CopyBytes(params.DAOForkBlockExtra)
-			} else if bytes.Equal(header.Extra, params.DAOForkBlockExtra) {
-				header.Extra = []byte{} // If miner opposes, don't let it use the reserved extra-data
+		if w.isRunning() {
+			if h, ok := w.engine.(consensus.Handler); ok {
+				h.NewWork()
```
mcortesi, line 367: Similarly, this could have a better name

miner/block.go
```diff
@@ -0,0 +1,348 @@
...
+
+	header         *types.Header
+	txs            []*types.Transaction
+	receipts       []*types.Receipt
+	randomness     *types.Randomness // The types.Randomness of the last block by mined by this worker.
+	txFeeRecipient common.Address
+}
+
+// prepareBlock intializes a new blockState that is ready to have transaction included to.
+func prepareBlock(w *worker) (*blockState, error) {
```
mcortesi, line 54: i would say this is cheating: so, to avoid having a worker method in another file, you make it a standalone function...

but you access worker state is if where your private state.

for example, using the Lock to access internal state. This can be nasty, since the whole idea of having an object with "private state" is that you don't need to worry about anyone else when modifying it... that's encapsulation. But we don't have any of that here.

I can look to the other side for now, if we want to get this to master, but it is worth mentioning nonetheless

miner/block.go
```diff
@@ -0,0 +1,348 @@
...
+
+		// make a copy, the state caches the logs and these logs get "upgraded" from pending to mined
+		// logs by filling in the block hash when the block was mined by the local miner. This can
+		// cause a race condition if a log was "upgraded" before the PendingLogsEvent is processed.
+		cpy := make([]*types.Log, len(coalescedLogs))
+		for i, l := range coalescedLogs {
+			cpy[i] = new(types.Log)
+			*cpy[i] = *l
+		}
+		w.pendingLogsFeed.Send(cpy)
```
mcortesi, line 291: to me it's ackward that we are accessing an internal feed object on the worker here

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
-			}
-		} else {
-			next = prev*(1-intervalAdjustRatio) + intervalAdjustRatio*(target-intervalAdjustBias)
-			// Recap if interval is less than the user specified minimum
-			if next < float64(minRecommit.Nanoseconds()) {
-				next = float64(minRecommit.Nanoseconds())
-			}
-		}
-		recommit = time.Duration(int64(next))
+	w.updatePendingBlock(b)
```
mcortesi, line 252: asking again, does it make sense to have a pending block with no transactions?

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
-
-	// Play our part in generating the random beacon.
-	if w.isRunning() && random.IsRunning() {
-		istanbul, ok := w.engine.(consensus.Istanbul)
-		if !ok {
-			log.Crit("Istanbul consensus engine must be in use for the randomness beacon")
-		}
+	for {
+		select {
+		case <-w.startCh:
```
mcortesi, line 378: if we call start() twice that means we will cancel current work and start again, feature or bug?

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
+	defer w.chainHeadSub.Unsubscribe()
+	defer w.txsSub.Unsubscribe()
+	// Context and cancel function for the currently executing block construction
+	// Cancel needs to be called in each exit path to make the linter happy
+	// because go struggles with analyzing lexical scoping.
+	var taskCtx context.Context
+	var cancel context.CancelFunc
 
-	var coalescedLogs []*types.Log
+	txsCh := make(chan core.NewTxsEvent, txChanSize)
```
mcortesi, line 357: shouldn't we try to get txpool configured size to set this size?

miner/worker.go
```diff
@@ -360,619 +238,200 @@ func (w *worker) createTxCmp() func(tx1 *types.Transaction, tx2 *types.Transacti
...
-		// Short-circuit if the transaction requires more gas than we have in the pool.
-		// If we didn't short-circuit here, we would get core.ErrGasLimitReached below.
-		// Short-circuiting here saves us the trouble of checking the GPM and so on when the tx can't be included
-		// anyway due to the block not having enough gas left.
-		if w.current.gasPool.Gas() < tx.Gas() {
-			log.Trace("Skipping transaction which requires more gas than is left in the block", "hash", tx.Hash(), "gas", w.current.gasPool.Gas(), "txgas", tx.Gas())
-			txs.Pop()
-			continue
+	newWorkFunc := func() {
+		if cancel != nil {
```
mcortesi, line 360: It's important to note that `cancel()` doesn't imply the goroutine is finished, just that it letting it know that it should finish. 

It would be nice if we wait for the routinte to finish, so we use Context as the kill signal, and a WaitGroup to wait for it to finish.

That way we know for a fact there are not two concurrent goroutines trying to build a block at the same time

miner/worker.go
trianglesphere: It would be, but golang cannot recognize that form as equivalent and the linter complains.

miner/worker.go
trianglesphere: We need to make sure that the pending block is the latest head block

consensus/istanbul/backend/test_utils.go
```diff
@@ -223,6 +223,7 @@ func makeBlockWithoutSeal(chain *core.BlockChain, engine *Backend, parent *types
 	// The worker that calls Prepare is the one filling the Coinbase
 	header.Coinbase = engine.address
 	engine.Prepare(chain, header)
+	time.Sleep(time.Until(time.Unix(int64(header.Time), 0)))
```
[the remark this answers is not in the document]
trianglesphere: The tests have issues if we don't sleep at all, but it's definitely possible to sleep for less time.

miner/worker.go
trianglesphere: That's correct. Unintended consequence that doesn't really matter. We want to cancel the pending block and start proposing blocks on `start`. It would be possible to gate sending to the start channel on the current miner state, but it doesn't seem like a big issue.

miner/worker.go
trianglesphere: We could, but this isn't regressing.

miner/block.go
trianglesphere: Yea. This is the same thing as above that the miner used to be so entangled in the block production that making it only depend on specific fields is a pain.

miner/miner.go
```diff
@@ -67,132 +64,145 @@ type Miner struct {
 	txFeeRecipient common.Address
```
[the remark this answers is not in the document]
trianglesphere: txFee can go. We definitely still need `validator` since not being able to `recoverRandomness` causes a lot of problems, (it also takes a while, so it's important that if it needs to happen, it gets done in a one-shot event).

miner/worker.go
trianglesphere: Minimal example showing go can't handle this: https://play.golang.org/p/dwrvFNabLhm

miner/worker.go
```diff
@@ -84,22 +54,6 @@ type callBackEngine interface {
 		onNewConsensusBlock func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB)) error
```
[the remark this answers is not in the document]
trianglesphere: https://github.com/celo-org/celo-blockchain/issues/1558

mcortesi (approved):
LGTM! Great work!

PR #1550, opened 2021-05-14
Fix cross compiling

### LINKED ISSUE #1544: Fix cross compiling with `make`
## Description

Fix cross compiling with `make`
Also require the release binaries to be built as part of CI.

The desired targets are:
```
linux        386, mips64, mips64le  
linux        arm 5, arm 6, arm 7  
darwin  
windows      386,  amd64
```

Note: Once the cross compile is fixed, the google cloud build hook should be re-enabled.

## Context

Cross compiling with make is broken. (Same with `go build`, but as we have dependencies that use cgo, it is hard to make go build work).

trianglesphere:
### Description

Note: requires a properly set gopath.

Current status (from my laptop darwin to ->)
```
linux        386   no
linux      amd64  yes
linux     mips64   no
linux   mips64le   no
linux      arm 5   no
linux      arm 6   no
linux      arm 7   no
linux      arm64  yes
darwin             no
windows      386   no
windows    amd64   no
```
### Related issues

- Fixes #1544


piersy:
I can cross compile without setting GOPATH, so just wondering if we really need that?

trianglesphere:
> I can cross compile without setting GOPATH, so just wondering if we really need that?

Are you sure it's not set? When I unset it (& confirmed it wasn't in my env), I got an error that there was no GOPATH being passed to xgo (note that this is all from the makefile, fixing `go build` with different GOOS/GOARCH requires dealing with the fact that we have CGO enabled).

piersy:
> > I can cross compile without setting GOPATH, so just wondering if we really need that?
> 
> Are you sure it's not set? When I unset it (& confirmed it wasn't in my env), I got an error that there was no GOPATH being passed to xgo (note that this is all from the makefile, fixing `go build` with different GOOS/GOARCH requires dealing with the fact that we have CGO enabled).

Sorry I was talking about just plain go cross compiling, not the system used in the Makefile, which I now realise is totally different to standard go cross compiling.

trianglesphere:
NP. Standard cross compiling works really well right until you have to deal with cgo (or have cgo dependencies) which we do.

PR #1556, opened 2021-05-27
Remove Miner Flags

mcortesi:
### Description

Remove flags:
 - Notify
 - NoVerify
 - GasTarget
 - GasFloor
 - GasLimit
 - Recommit

Keeps `GasPrice` as a flag, but only use to set up txpool gas price.

### DIFF
--- a/cmd/utils/flags_legacy.go
+++ b/cmd/utils/flags_legacy.go
@@ -38,8 +38,6 @@ var DeprecatedFlags = []cli.Flag{
 	LegacyLightServFlag,
 	LegacyLightPeersFlag,
 	LegacyMinerThreadsFlag,
-	LegacyMinerGasTargetFlag,
-	LegacyMinerGasPriceFlag,
 	LegacyMinerExtraDataFlag,
 }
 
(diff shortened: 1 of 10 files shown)


cmd/utils/flags_legacy.go
```diff
@@ -38,8 +38,6 @@ var DeprecatedFlags = []cli.Flag{
 	LegacyLightServFlag,
 	LegacyLightPeersFlag,
 	LegacyMinerThreadsFlag,
-	LegacyMinerGasTargetFlag,
-	LegacyMinerGasPriceFlag,
```
trianglesphere, line 42: If we're keeping the gas price flag, it seems like we may want to keep the legacy version.

eth/backend.go
```diff
@@ -90,8 +90,8 @@ type Ethereum struct {
 
 	APIBackend *EthAPIBackend
 
-	miner          *miner.Miner
-	gasPrice       *big.Int
+	miner *miner.Miner
+	// gasPrice       *big.Int
```
trianglesphere, line 94: Should be entirely removed if that is the intention.

cmd/utils/flags_legacy.go
mcortesi: moved `--miner.gasprice` to deprecated and removed the already deprecated one

PR #1564, opened 2021-06-01
Run full network tests in under 1 second (pure go)

### LINKED ISSUE #1476: Allow pure go testing of celo networks
## Description

We currently don't have any way to test whole networks in go. Testing whole network functionality currently requires multiple manual steps resulting in running multiple processes, it is complex to set up and complex to analyze the results which makes it hard and also a lot of work to encode these tests in CI. The result of this is that some of our tests are executed in an ad-hoc manner and not added to CI, which leaves us open to regressions at the next point that the code is modified.

## Proposed Solution

If we could write test like this, which starts up a network sends some celo from one account to another and then shuts down.
```go
// This test checks that we can process transactions that transfer value from
// one participant to another.
func TestSendingValue(t *testing.T) {
	participants := 3
	network, err := test.NewNetwork(participants, generateAccount)
	require.NoError(t, err)
	defer network.Shutdown()

	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
	defer cancel()
	err = network[0].SendValue(ctx, network[1].Address, 10)
	require.NoError(t, err)
}
```

Then we could reduce overhead creating tests of networks, tests would be part of CI without any extra work and it provides the opportunity to easily modify the behaviour of individual participants to simulate attacks, failures ... etc. Also problems encountered in these tests are debuggable with a debugger.

piersy:
### Description

This PR adds a package that enables writing tests of full networks within a singe go test process. It also adds an example test.

To try this out you will first need to run `make prepare-system-contracts` and then you can run the test (note you will need node v10 for the make command to succeed)

The commits for this PR are well separated and I think it would be easier to review commit by commit.

### Tested

The test works most of the time on my machine and seems to work most of the time on CI, sometimes the network hangs, I believe this is a problem with the blockchain code not the test code. I'm wondering whether it would be better to skip the test, but at least for now it's useful to know that the test is able to be run in CI.

You can run the e2e test with `go test ./e2e_test -count 1` it should take roughly a second to run.

I also noticed that having the monorepo checked out under the celo-blockchain folder seems to make `gopls` operate very slowly so I'm considering making the monorepo be checked out alongside the celo-blockchain project dir as opposed to under it, but before doing that I'd like to know if this is affecting anyone else?

### Related issues

- Fixes #1476

### Backwards compatibility

I believe so, no major changes were made to any core code.

### DIFF
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -1,0 +1,38 @@
+package e2e_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/celo-org/celo-blockchain/test"
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
+	// This statement is commented out but left here since its very useful for
+	// debugging problems and its non trivial to construct.
+	//
+	//log.Root().SetHandler(log.LvlFilterHandler(log.LvlTrace, log.StreamHandler(os.Stdout, log.TerminalFormat(true))))
+}
+
+// This test starts a network submits a transaction and waits for the whole
+// network to process the transaction.
+func TestSendCelo(t *testing.T) {
+	accounts := test.Accounts(3)
+	gc := test.GenesisConfig(accounts)
+	network, err := test.NewNetwork(accounts, gc)
+	require.NoError(t, err)
+	defer network.Shutdown()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	// Send 1 celo from the dev account attached to node 0 to the dev account
+	// attached to node 1.
+	tx, err := network[0].SendCelo(ctx, network[1].DevAddress, 1)
+	require.NoError(t, err)
+
+	// Wait for the whole network to process the transaction.
+	err = network.AwaitTransactions(ctx, tx)
+	require.NoError(t, err)
+}
--- a/test/node.go
+++ b/test/node.go
@@ -60,6 +60,11 @@ var (
 		TxPool:          core.DefaultTxPoolConfig,
 		Istanbul: istanbul.Config{
 			Validator: true,
+			// Set announce gossip period to 1 minute, if not set this results
+			// in a panic when trying to set a timer with a zero value. We
+			// don't actually rely on this mechanism in the tests because we
+			// pre share all the enode certificates.
+			AnnounceQueryEnodeGossipPeriod: 60,
 		},
 	}
 )
--- a/test/node.go
+++ b/test/node.go
@@ -388,14 +388,13 @@ func NewNetwork(accounts *env.AccountsConfig, gc *genesis.Config) (Network, erro
 	// Connect nodes to each other, although this means that nodes can reach
 	// each other nodes don't start sending consensus messages to another node
 	// until they have received an enode certificate from that node.
-	for i := range network {
-		en := enodes[i]
-		for j := range network {
+	for i, en := range enodes {
+		for j, n := range network {
 			if j == i {
 				continue
 			}
-			network[j].Server().AddPeer(en, p2p.ValidatorPurpose)
-			network[j].Server().AddTrustedPeer(en, p2p.ValidatorPurpose)
+			n.Server().AddPeer(en, p2p.ValidatorPurpose)
+			n.Server().AddTrustedPeer(en, p2p.ValidatorPurpose)
 		}
 	}
 
--- a/test/node.go
+++ b/test/node.go
@@ -1,0 +1,567 @@
+package test
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+	"net"
+	"os"
+	"strconv"
+	"time"
+
+	ethereum "github.com/celo-org/celo-blockchain"
+
+	"github.com/celo-org/celo-blockchain/accounts/keystore"
+	"github.com/celo-org/celo-blockchain/common"
+	"github.com/celo-org/celo-blockchain/consensus/istanbul"
+	"github.com/celo-org/celo-blockchain/consensus/istanbul/backend"
+	"github.com/celo-org/celo-blockchain/core"
+	"github.com/celo-org/celo-blockchain/core/types"
+	"github.com/celo-org/celo-blockchain/crypto"
+	"github.com/celo-org/celo-blockchain/eth"
+	"github.com/celo-org/celo-blockchain/eth/downloader"
+	"github.com/celo-org/celo-blockchain/ethclient"
+	"github.com/celo-org/celo-blockchain/log"
+	"github.com/celo-org/celo-blockchain/mycelo/env"
+	"github.com/celo-org/celo-blockchain/mycelo/genesis"
+	"github.com/celo-org/celo-blockchain/node"
+	"github.com/celo-org/celo-blockchain/p2p"
+	"github.com/celo-org/celo-blockchain/p2p/enode"
+	"github.com/celo-org/celo-blockchain/params"
+	"github.com/celo-org/celo-blockchain/rlp"
+)
+
+var (
+	baseNodeConfig *node.Config = &node.Config{
+		Name:    "celo",
+		Version: params.Version,
+		P2P: p2p.Config{
+			MaxPeers:    100,
+			NoDiscovery: true,
+			ListenAddr:  "0.0.0.0:0",
+		},
+		NoUSB: true,
+		// It is important that HTTPHost and WSHost remain the same. This
+		// ensures that only one server is started up on the HTTPHost. That one
+		// server can still handle both http and ws connectons.
+		HTTPHost:             "0.0.0.0",
+		WSHost:               "0.0.0.0",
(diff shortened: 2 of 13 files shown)


e2e_test/e2e_test.go
```diff
@@ -0,0 +1,38 @@
...
+	// This statement is commented out but left here since its very useful for
+	// debugging problems and its non trivial to construct.
+	//
+	//log.Root().SetHandler(log.LvlFilterHandler(log.LvlTrace, log.StreamHandler(os.Stdout, log.TerminalFormat(true))))
+}
+
+// This test starts a network submits a transaction and waits for the whole
+// network to process the transaction.
+func TestSendCelo(t *testing.T) {
+	accounts := test.Accounts(3)
```
trianglesphere, line 22: nit: this produces an account config, not the accounts.

test/node.go
```diff
@@ -0,0 +1,567 @@
...
+// others not.
+func NewNetwork(accounts *env.AccountsConfig, gc *genesis.Config) (Network, error) {
+
+	genesis, err := genesis.GenerateGenesis(accounts, gc, "../monorepo/packages/protocol/build/contracts")
+	if err != nil {
+		return nil, err
+	}
+
+	validatorAccounts := accounts.ValidatorAccounts()
+	network := make([]*Node, len(validatorAccounts))
```
trianglesphere, line 365: More a note than nit, but I usually set the capacity with make (`make(T, 0, (len(validatorAccounts))`) and then use append rather than directly setting the value.

test/node.go
```diff
@@ -0,0 +1,567 @@
...
+			return nil, err
+		}
+		en := enode.NewV4(&n.Key.PublicKey, net.ParseIP(host), portNum, portNum)
+		enodes[i] = en
+	}
+	// Connect nodes to each other, although this means that nodes can reach
+	// each other nodes don't start sending consensus messages to another node
+	// until they have received an enode certificate from that node.
+	for i := range network {
+		en := enodes[i]
```
trianglesphere, line 392: Merge this into `for i, en := rand enodes {` ?

test/node.go
```diff
@@ -0,0 +1,567 @@
...
+		for j := range network {
+			if j == i {
+				continue
+			}
+			network[j].Server().AddPeer(en, p2p.ValidatorPurpose)
```
trianglesphere, line 397: could use a  `for j, n :=` and then `n.Server()` rather than directly indexing.

test/transaction_tracker.go
```diff
@@ -0,0 +1,193 @@
...
+	if tr.sub == nil {
+		return errors.New("attempted to stop already stopped tracker")
+	}
+	tr.sub.Unsubscribe()
+	close(tr.stopCh)
+	tr.wg.Wait()
+	tr.wg = sync.WaitGroup{}
+
+	// Free fields
+	tr.client = nil
```
trianglesphere, line 189: Shouldn't these be cleaned up by the GC once the TransactionTracker can no longer be reached

test/transaction_tracker.go
```diff
@@ -0,0 +1,193 @@
...
+			if ok {
+				delete(hashmap, hash)
+			}
+		}
+		// If there are no transactions left then they have all been processed.
+		if len(hashmap) == 0 {
+			close(txsFound) // Signal to the go routine that it can exit
+			return nil
+		}
+		tr.newTxs.Wait()
```
trianglesphere, line 170: If I understand correctly we need the condition vars to deal with the fact that while we can put all the finalized transaction into the shared hashmap `tr.processed` (maybe `tr.seenTransactions`), we only want to scan through it once we get a new block.

What do you think about using a channel to signal a new block event? I think it'd play nicer with ctx.Done() and allow you to get rid of the extra go-routine. The extra channel could be a new sub or a channel  that `trackTransactions` only sends to when `AwaitTransactions` is running (requires a lock, but is doable).

test/node.go
```diff
@@ -0,0 +1,567 @@
...
+	ethConfigCopy.TxFeeRecipient = n.Address
+	ethConfigCopy.Istanbul.AnnounceQueryEnodeGossipPeriod = 60
+	ethConfigCopy.Istanbul.Validator = true
+	// ethConfigCopy.Istanbul.BlockPeriod = 0
+	// Somehow this is not being honored, its being set to 3000
+	// ethConfigCopy.Istanbul.RequestTimeout = 100
+	ethConfigCopy.Istanbul.TimeoutBackoffFactor = 10
+	ethConfigCopy.Istanbul.MinResendRoundChangeTimeout = 15 * 1000
+	ethConfigCopy.Istanbul.MaxResendRoundChangeTimeout = 2 * 60 * 1000
+	// ethConfigCopy.Istanbul.Epoch = 100
```
trianglesphere, line 165: dead code or a future todo?
ditto for the above commented out code.

trianglesphere (commented):
Looking pretty good. I had a couple small nits and 1 bigger question about the transaction tracker. I'm really excited to be able to use these tests.

test/node.go
piersy: I do that when I don't have access to an index variable, for example copying from a map to a pre-allocated slice. But when I have access to the index I think prefer this approach, just because its slightly more concise and I find it more explicit about the fact that only the pre-allocated memory is being used.

```
network[i] = n
// vs
network = append(network, n)
```

e2e_test/e2e_test.go
piersy: Even though the object type name is AccountsConfig if you look at how that variable is used I think accounts is actually an appropriate name.
```
accounts.ValidatorAccounts()
accounts.AdminAccount()
accounts.DeveloperAccounts()
```

There's actually a nice article that discusses the practice of naming things after their types
https://dave.cheney.net/2019/01/29/you-shouldnt-name-your-variables-after-their-types-for-the-same-reason-you-wouldnt-name-your-pets-dog-or-cat

test/transaction_tracker.go
piersy: Those fields are set in StartTracking so if we were to start and stop a tracker those fields would remain uncollected until start was called again. However the freeing of them definitely seems an unnecessary optimisation (if it can even be called that), I can't remember what I was thinking when I wrote that code. So I think we can safely just remove these lines.

test/transaction_tracker.go
piersy: I'm going to have a shot at this, getting rid of that go-routine would be nice :)

piersy: I've made the change, its looking much neater. Dropped a sweet 20 LOC as well :)

With regards to renaming 'processed', I chose that name because I thought it made a clear distinction between transactions that came from finished blocks versus transactions that the tracker had been asked to wait for. With 'seen' I feel like it makes the distinction less clear.

test/node.go
piersy: Just some dev cruft, I think they can be safely removed.

piersy:
> Looking pretty good. I had a couple small nits and 1 bigger question about the transaction tracker. I'm really excited to be able to use these tests.

Did you manage to run this on your local machine?

trianglesphere:
> Did you manage to run this on your local machine?

Yep. And it correctly fails when I change [this](https://github.com/celo-org/celo-blockchain/blob/4e42e7e8d683929b1317859d86431d5b3d3037c6/miner/block.go#L317-L318) as well

test/transaction_tracker.go
```diff
@@ -0,0 +1,167 @@
...
+	stopCh      chan struct{}
+	newTxs      event.Feed
+}
+
+// NewTransactionTracker creates a new transaction tracker, it subscribes to
+// new head events on the client and starts processing the events in a
+// goroutine.
+func NewTransactionTracker() *TransactionTracker {
+	return &TransactionTracker{
+		heads:       make(chan *types.Header),
```
trianglesphere, line 40: Subscriptions will block if the channels are full, so we typically use buffered channels.
```suggestion
		heads:       make(chan *types.Header, 10),
```

test/transaction_tracker.go
```diff
@@ -0,0 +1,167 @@
...
+// AwaitTransactions waits for the transactions listed in hashes to be
+// processed, it will return the ctx.Err() if ctx expires before all the
+// transactions in hashes were processed or ErrStopped if StopTracking is
+// called before all the transactions in hashes were processed.
+func (tr *TransactionTracker) AwaitTransactions(ctx context.Context, hashes []common.Hash) error {
+	hashmap := make(map[common.Hash]struct{}, len(hashes))
+	for i := range hashes {
+		hashmap[hashes[i]] = struct{}{}
+	}
+	ch := make(chan struct{})
```
trianglesphere, line 129: Ditto on using buffereed channels with subscriptions
```suggestion
	ch := make(chan struct{}, 10)
```

trianglesphere (approved):
I've got a couple small suggestions, but this looks good.
The CI was just flaky, nothing actually was wrong.

piersy:
@trianglesphere Did you notice any drop in performance of gopls after having the monorepo checked out under celo-blockchain?

trianglesphere:
> @trianglesphere Did you notice any drop in performance of gopls after having the monorepo checked out under celo-blockchain?

Not appreciably, but maybe a little (once I nuked the monorepo dir), but it's hard to tell.

PR #1565, opened 2021-06-02
Remove support for celo/64 and celo/65 protocols.

### LINKED ISSUE #1213: Remove `Primary` protocol field
~**NOTE**: We can only do this once we are sure that nodes are no longer running istanbul versions `Celo65` or older, which we could know based on minimum client version or a hard fork (we would have to be sure that the network has been upgraded to a version that supports versions higher than 65).  Because of this I'm marking the issue as blocked in its name.~

June 1, 2021: Now that Donut has been activated on Baklava, Alfajores, and mainnet, we know all full nodes which are still able to connect and sync must be running v1.3.x (and thus support Celo66), and so we can do what this issue says.

---

Once it becomes possible, we should remove the `Primary` field from `Protocol` in `p2p/protocol.go` and everything that relates to it.  It is defined at: https://github.com/celo-org/celo-blockchain/blob/4914a9768fb85eb4f0827b88721543f8cd98ee2c/p2p/protocol.go#L39-L40

Included in this is removing the `IsPrimary()` function which would then become unused: https://github.com/celo-org/celo-blockchain/blob/4914a9768fb85eb4f0827b88721543f8cd98ee2c/consensus/istanbul/protocol.go#L36-L42

Once we know that the network has switched to a newer version of istanbul, we can remove the older versions from https://github.com/celo-org/celo-blockchain/blob/4914a9768fb85eb4f0827b88721543f8cd98ee2c/consensus/istanbul/protocol.go#L32-L34

We could also clean up by removing any actual code involved in only the older versions, but it's not clear that we should, since that would likely just lead to more problems in later merges from upstream.

### Related issues

* #333, which introduced the Primary field.
* #1131, which is superseded by this issue.
* #1201, which made the Primary concept apply to specific versions of istanbul and thus will enable us to eventually remove the field entirely in a backwards compatible way.

oneeman:
### Description

Now that Donut has been activated on Alfajores, Baklava, and mainnet, all nodes which are still able to connect and sync are running v1.3.x and thus support celo/66.  Therefore, it is possible to remove support for celo/64 and celo/65 and with them the notion of a "primary" protocol.

See #1213 for more context.

### Tested

* Automated testing
* A full node successfully starts syncing to mainnet.

### Related issues

- Closes #1213

### Backwards compatibility

This breaks compatibility with nodes running v1.2.x or older.

On Alfajores, Baklava, and mainnet we have already broken compatibility with such nodes when the Donut hard fork was activated, so this PR doesn't add any additional incompatibility.

However, on private testnets that might not be the case, and so for them this PR would constitute a breaking change requiring nodes to upgrade to v1.3.0 or higher in order to be able to communicate with each other.

### DIFF
--- a/consensus/istanbul/protocol.go
+++ b/consensus/istanbul/protocol.go
@@ -22,8 +22,12 @@ import (
 
 // Constants to match up protocol versions and messages
 const (
+	// No longer supported, can be removed.
+	// The corresponding version upstream (eth/65) is removed in upstream PR #22636.
 	Celo64 = 64 // eth/63 + the istanbul messages
 	Celo65 = 65 // incorporates changes from eth/64 (EIP)
+
+	// Supported versions
 	Celo66 = 66 // incorporates changes from eth/65 (EIP-2464)
 )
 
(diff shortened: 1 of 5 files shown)


consensus/istanbul/protocol.go
```diff
@@ -22,25 +22,21 @@ import (
 
 // Constants to match up protocol versions and messages
 const (
+	// No longer supported, can be removed.
+	// The corresponding version upstream (eth/65) is removed in upstream PR #22636.
 	Celo64 = 64 // eth/63 + the istanbul messages
```
trianglesphere, line 27: Do we want to remove these now or wait for the upstream PR?

trianglesphere (approved):
lgtm

consensus/istanbul/protocol.go
oneeman: My inclination is to leave them for now, but I wouldn't be opposed to removing them either.

trianglesphere: sounds good

PR #1569, opened 2021-06-03
Fix RunTaskRepeteadly flakyness

mcortesi:
### Description

Increase internal, to cope for resource contention when running tests


gastonponti (approved):
We can extract those into constants but, as you wish, it's also valid
LGTM

PR #1573, opened 2021-06-08
Remove Registry Contract Check

### LINKED ISSUE #1542: Remove Registry Contract Check
We currently check wether RegitryProxy contract is deployed, but that contract is deployed on the genesis block and must be in a fixed address.

So, we could avoid doing this check. Which will allow us to remove the `GetStateDB()` method from the `EVMRunner`

Optionally, we could do a check on `geth init` or genesis setup, to verify that the genesis actually has that contract.

_Originally posted by @oneeman in https://github.com/celo-org/celo-blockchain/pull/1513#discussion_r629639906_

YoduYodu:
### Tested

Added tests to cover:

- DeveloperGenesisBlock, DefaultAlfajoresGenesisBlock, DefaultBaklavaGenesisBlock, MainnetGenesisBlock have registry contract.
- run `geth` and the registry contract is there.
- run `geth init` on the genesis json file with the registry, and the registry contract is there.

### Related issues

- Fixes #1542

### Backwards compatibility

Yes


core/genesis_test.go
```diff
@@ -176,3 +176,16 @@ func TestSetupGenesis(t *testing.T) {
 		}
 	}
 }
+
+// TestRegistryInGenesis tests if RegistrySmartContract sits in the state of genesis block
+func TestRegistryInGenesis(t *testing.T) {
+	db := rawdb.NewMemoryDatabase()
+	MainnetGenesisBlock().MustCommit(db)
```
trianglesphere, line 183: FYI this check also holds true for our other genesis blocks.

YoduYodu: > FYI this check also holds true for our other genesis blocks.

Added Dev, Alfajores, Baklava. Thanks for pointing it out!

core/genesis.go
```diff
@@ -374,7 +374,7 @@ func DefaultAlfajoresGenesisBlock() *Genesis {
 }
 
 // DeveloperGenesisBlock returns the 'geth --dev' genesis block.
-func DeveloperGenesisBlock(period uint64, faucet common.Address) *Genesis {
+func DeveloperGenesisBlock() *Genesis {
```
trianglesphere, line 377: nit: I'd provide these parameters (like `15, common.Address{}`) in the genesis test rather than removing them from places that did use the overrides

YoduYodu: What do you mean by **overrides**? Those deleted params are not used

trianglesphere: Oh, I didn't see that they were not being used. In this case it makes sense to either use them or not pretend to be able to override them.

alice-dfi:
Why your overriding ? Can you explain what are you trying to do ???

alice-dfi:
Or are you one of those hackers working with do kwon stealing and changing
things ?

contracts/registry.go
```diff
@@ -19,11 +19,6 @@ func GetRegisteredAddress(vmRunner vm.EVMRunner, registryId common.Hash) (common
 	vmRunner.StopGasMetering()
 	defer vmRunner.StartGasMetering()
 
-	// TODO(mcortesi) remove registrypoxy deployed at genesis
-	if vmRunner.GetStateDB().GetCodeSize(params.RegistrySmartContractAddress) == 0 {
```
piersy, line 23: Shouldn't we now be able to remove the GetStateDB method from the vm runner interface and all its implemenationns?

YoduYodu: https://github.com/celo-org/celo-blockchain/blob/c4ea1da3bcba2a1e8909efbc741e3d4c74c5d4f4/contracts/random/random.go#L72-L75

There is another call to `GetCodeSize()`

YoduYodu: @mcortesi Can you elaborate how this FIXME means?

mcortesi: the second one is still required... but i'm 100% sure why (predates me).

My comment is because IMO the caller this function should do the `Finalise()` and not the function itself. The function as all other function, shoudl just make a contract call... if you want to finalse state after,, that's the caller's reponsability.

core/genesis_test.go
```diff
@@ -176,3 +175,51 @@ func TestSetupGenesis(t *testing.T) {
 		}
 	}
 }
+
+// TestRegistryInGenesis tests if the params.RegistrySmartContract that defined in the genesis block sits in the blockchain
+func TestRegistryInGenesis(t *testing.T) {
```
mcortesi, line 180: love that we're doing a tests for this. Still, i would also ask that when we launch a node with a genesis block we should check that registry is there.

Not sure if we should do this on `geth init` or on the startup logic when we run `geth`. Maybe in both.

Basically, we want to make sure that if you've done `geth init` then `geth` you have the registry installed.

mcortesi (changes requested):
see the comment

contracts/registry.go
YoduYodu: The chunk could be safely delete, because two callers already called `stateDB.IntermediateRoot()`, which contains calling `Finalise()`.

https://github.com/celo-org/celo-blockchain/blob/c4ea1da3bcba2a1e8909efbc741e3d4c74c5d4f4/miner/block.go#L142-L147

https://github.com/celo-org/celo-blockchain/blob/c4ea1da3bcba2a1e8909efbc741e3d4c74c5d4f4/miner/block.go#L142-L147

core/genesis_test.go
YoduYodu: Added `geth` and `geth init` tests

contracts/random/random.go
```diff
@@ -69,10 +69,6 @@ func RevealAndCommit(vmRunner vm.EVMRunner, randomness, newCommitment common.Has
 	log.Trace("Revealing and committing randomness", "randomness", randomness.Hex(), "commitment", newCommitment.Hex())
 	err := revealAndCommitMethod.Execute(vmRunner, nil, common.Big0, randomness, newCommitment, proposer)
 
-	// FIXME this should be done by the caller, remove from here after
-	if err == nil {
-		vmRunner.GetStateDB().Finalise(true)
```
mcortesi, line 74: why are you removing this? It's not clear to me that this can be removed

core/genesis_test.go
mcortesi: @YoduYodu maybe i didn't express myself correctly. The idea here is not to add new tests, but to check that genesis has the contract deployed whenever we run `geth` and `geth init`. So, not a test, but a check you do during normal execution, just to be sure you're not operating with a genesis without it.

For example, imagine i create a genesis.json with an empty alloc. then i do `geth init mybogusgenesis.json` and then `geth`

With current code, you will only notice something is wrong when you reach to the place we did the check (codeSize != 0), and now we don't. So it would fail and it would be hard to debug.

Instead, what i'm asking is taht when you run geth init with a bogus genesis.json it fails

mcortesi (changes requested):

contracts/random/random.go
YoduYodu: They are already handled, check my comment:
https://github.com/celo-org/celo-blockchain/pull/1573#discussion_r656619294

cmd/geth/genesis_test.go
```diff
@@ -17,36 +17,60 @@
...
...10871565b73ffffffffffffffffffffffffffffffffffffffff163373ffffffffffffffffffffffffffffffffffffffff1614610865576040517f08c379a00000000000000000000000000000000000000000000000000000000081526004018080602001828103825260148152602001807f73656e64657220776173206e6f74206f776e657200000000000000000000000081525060200191505060405180910390fd5b61086e816108bd565b50565b600080600160405180807f656970313936372e70726f78792e61646d696e000000000000000000000000008152506013019050604051809103902060001c0360001b9050805491505090565b600073ffffffffffffffffffffffffffffffffffffffff168173ffffffffffffffffffffffffffffffffffffffff161415610960576040517f08c379a00000000000000000000000000000000000000000000000000000000081526004018080602001828103825260118152602001807f6f776e65722063616e6e6f74206265203000000000000000000000000000000081525060200191505060405180910390fd5b6000600160405180807f656970313936372e70726f78792e61646d696e000000000000000000000000008152506013019050604051809103902060001c0360001b90508181558173ffffffffffffffffffffffffffffffffffffffff167f50146d0e3c60aa1d17a70635b05494f864e86144a2201275021014fbf08bafe260405160405180910390a2505056fea165627a7a723058206808dd43e7d765afca53fe439122bc5eac16d708ce7d463451be5042426f101f0029"
```
mcortesi, line 29: You can get these variables from the orignal code.

Check https://github.com/celo-org/celo-blockchain/blob/d3b153fe00407494c212902bc12b947c778b81f8/mycelo/genesis/genesis_state.go#L22 for the proxy code

contracts/random/random.go
mcortesi: nice

mcortesi (approved):

YoduYodu:
Will merge after [this issue](https://github.com/celo-org/celo-monorepo/issues/8276) is fixed.

Issue #1384, opened 2021-02-17
Use mycelo for the e2e tests


oneeman:
Currently, the e2e take a long time (over 10 minutes) just in running the core contract migrations.  This could be reduced to be nearly instant by using mycelo to generate a genesis.json which already includes the core contracts.

It isn't clear whether we should use mycelo just to generate genesis.json, or also use it for subsequent steps such as actually running the local testnet.  That would depend on the code changes involved and the resulting code.  The big gain in terms of test running times would come from the genesis generation making subsequent migrations unnecessary.

oneeman:
Current status:

* Opened up draft PRs on celo-monorepo and celo-blockchain.
* celo-monorepo PR depends on https://github.com/celo-org/celo-monorepo/pull/8060 being merged first.
* Not converted: e2e attestations tests and e2e governance tests (more details on the monorepo PR).

PR #1579, opened 2021-06-10
Update mycelo config values to work with monorepo e2e tests

oneeman:
### Description

This is a companion PR to https://github.com/celo-org/celo-monorepo/pull/8086, with some changes which are needed to make the tests work:

1. Monorepo uses a single validator group, so we update the mycelo monorepo template to do so as well.
2. Make the DowntimeSlasher's slashable downtime small so it works with the small epoch sizes we use (if the slashable downtime is larger than the epoch size, you can't test out slashing)
3. Modify the monorepo template's config values for epoch rewards and initial reserve balance to match what is done in the monorepo for the 'testing' context (which is what the e2e tests use).  See https://github.com/celo-org/celo-monorepo/blob/c90edf1c938363e24746c90153d36263bb668aef/packages/protocol/migrationsConfig.js#L244-L281 .

### Tested

* Tested in conjunction with the monorepo tests PR above, locally and multiple times in CI.  Did not witness flakiness in the e2e tests.

### Related issues

- Necessary for #1384

### Backwards compatibility

No incompatibility

### DIFF
--- a/mycelo/genesis/base_config.go
+++ b/mycelo/genesis/base_config.go
@@ -147,7 +147,7 @@ func BaseConfig() *Config {
 		DowntimeSlasher: DowntimeSlasherParameters{
 			Reward:            bigIntStr("10000000000000000000"),  // 10 cGLD
 			Penalty:           bigIntStr("100000000000000000000"), // 100 cGLD
-			SlashableDowntime: 60,                                 // Should be overridden on public testnets
+			SlashableDowntime: 4,                                  // make it small so it works with small epoch sizes, e.g. 10
 		},
 		Governance: GovernanceParameters{
 			UseMultiSig:             true,
--- a/cmd/mycelo/templates.go
+++ b/cmd/mycelo/templates.go
@@ -164,7 +164,7 @@ func (e monorepoEnv) createEnv(workdir string) (*env.Environment, error) {
 		Accounts: env.AccountsConfig{
 			Mnemonic:             env.MustNewMnemonic(),
 			NumValidators:        3,
-			ValidatorsPerGroup:   1,
+			ValidatorsPerGroup:   5, // monorepo uses a single validator group, max group size is 5
 			NumDeveloperAccounts: 0,
 			UseValidatorAsAdmin:  true, // monorepo doesn't use the admin account type, uses first validator instead
 		},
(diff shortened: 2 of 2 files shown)


alice-dfi (commented):
Why and who approved it the changes ?? Why are you changing ?? Do not change it

oneeman:
> Why and who approved it the changes ?? Why are you changing ?? Do not change it

Could you elaborate on your concern?  Mycelo is a tool to facilitate development and testing.  It it used to generate genesis files for new ephemeral testnets.   The changes in this PR modify two things:

1. The default slashable downtime value, because the previous one doesn't make sense given the default epoch size
2. Settings for mycelo's 'monorepo' template to make it match more closely the configuration used in celo-monorepo (which is the reason that template exists)

These changes only affect the generation of new genesis files from scratch using mycelo.  They do not affect existing networks nor even the generation of a genesis file from an existing mycelo directory with a genesis config.  Plus, the genesis config as a whole can be customized by the user to suit their needs.  So if you have any concerns about this, please provide more details.

mycelo/genesis/base_config.go
```diff
@@ -147,7 +147,7 @@ func BaseConfig() *Config {
 		DowntimeSlasher: DowntimeSlasherParameters{
 			Reward:            bigIntStr("10000000000000000000"),  // 10 cGLD
 			Penalty:           bigIntStr("100000000000000000000"), // 100 cGLD
-			SlashableDowntime: 60,                                 // Should be overridden on public testnets
+			SlashableDowntime: 4,                                  // make it small so it works with small epoch sizes, e.g. 10
```
mcortesi, line 150: is this desirable for every network template or only for monorepo one?

mcortesi (approved):
looks good, left one comment

mycelo/genesis/base_config.go
oneeman: I think so.  The default epoch size (i.e. the epoch size of the default template) is 10, so the former slashable downtime of 60 doesn't make sense with that.

The only template that has a different epoch size is the load test (1000).  If we wanted to avoid slashable downtime of 4 with epoch size 1000 it would make sense to override it on the loadtest template rather than have 60 be the default.  But the low slashable downtime makes no difference if nobody does the slashing, so I don't see it as mattering for the load test template anyway.

PR #1590, opened 2021-06-14
Announce protocol code refactor

hbandura:
Extracts most of the Announce protocol behaviour into the `AnnounceManager` type. This is a partial PR, in the sense that most work is yet to be done to completely move all announce code into the announce manager, but since it's already a really big change, I decided to merge this and continue working on it on a different branch.

This PR extracts a big part of the announce code into the new announce manager type. Work yet to be done is:
- finish removing everyting from backend into announcemanager
- decide on what the public interface of the announcemanager will be, once everything is in place
- minimize the dependencies on the announce manager constructor
- improve the unit testing on it, using the advantage of it being decoupled from istanbul.Backend

Some previous PRs were written in orden to make this easier (GossipCache, the atomic.Value for wallets).


consensus/istanbul/backend/backend.go
```diff
@@ -195,6 +193,20 @@ func New(config *istanbul.Config, db ethdb.Database) consensus.Istanbul {
...
+		AWallets:           &backend.aWallets,
+		VcDbPath:           config.VersionCertificateDBPath,
+	}
+
+	backend.announceManager = NewAnnounceManager(
+		announceConfig,
+		backend,
+		backend,
+		backend,
+		backend.valEnodeTable,
```
mcortesi, line 207: thinking that maybe makes sense for the announce manager to OWN the valEnodeTable. Basically, the purpose of the announce protocol, is to:
- disclose our own enode url to other address
- maintain a table of others validators enodes (valEnodeTable)

so i can be represented as this "interface" where i go to fectch validator enodes, and that hooks into part of the system to do the announcing

consensus/istanbul/backend/backend.go
```diff
@@ -442,7 +438,7 @@ func (sb *Backend) Close() error {
 	if err := sb.valEnodeTable.Close(); err != nil {
 		errs = append(errs, err)
 	}
-	if err := sb.versionCertificateTable.Close(); err != nil {
+	if err := sb.announceManager.Close(); err != nil {
```
mcortesi, line 441: not of this PR, but i'm sure there's some library of code example on how to actually group errors. At least using %w on Errof, but doing that would imply (err1 -> err2 -> err3); when is actually (err -> [err1, err2, err3])

consensus/istanbul/backend/handler.go
```diff
@@ -381,7 +381,7 @@ func (sb *Backend) RegisterPeer(peer consensus.Peer, isProxiedPeer bool) error {
 		}
 	}
 
-	if err := sb.sendVersionCertificateTable(peer); err != nil {
+	if err := sb.announceManager.SendVersionCertificateTable(peer); err != nil {
```
mcortesi, line 384: thinking out loud on alternative "semantics". We could do:


1) announce doesn't  know about this, we do it: 
table = announceManager.getTable(); peer.send(table)

2) announce knows about this
announceManager.handleRegisterPeer(peer)

consensus/istanbul/backend/api.go
```diff
@@ -185,7 +185,7 @@ func (api *API) GetValEnodeTable() (map[string]*vet.ValEnodeEntryInfo, error) {
 }
 
 func (api *API) GetVersionCertificateTableInfo() (map[string]*vet.VersionCertificateEntryInfo, error) {
-	return api.istanbul.versionCertificateTable.Info()
+	return api.istanbul.announceManager.versionCertificateTable.Info()
```
mcortesi, line 188: i would hide `versionCertificateTable` and instead expose a getter for the `Info` on the announce manager

consensus/istanbul/backend/announce.go
```diff
@@ -70,6 +72,102 @@ const (
...
+}
+
+// NewAnnounceManager creates a new AnnounceManager using the valEnodeTable given. It is
+// the responsibility of the caller to close the valEnodeTable, the AnnounceManager will
+// not do it.
+func NewAnnounceManager(config AnnounceManagerConfig, support AnnounceSupport, proxyContext ProxyContext,
+	addrProvider AddressProvider, valEnodeTable *enodes.ValidatorEnodeDB,
+	gossipCache GossipCache) *AnnounceManager {
+	am := &AnnounceManager{
+		logger:                          log.New(),
```
mcortesi, line 142: you can add something like `log.New("module", "announceManager")`

consensus/istanbul/backend/announce.go
```diff
@@ -70,6 +72,102 @@ const (
...
+	// RetrieveValidatorConnSet returns the validator connection set
+	RetrieveValidatorConnSet() (map[common.Address]bool, error)
+	// Multicast will send the eth message (with the message's payload and msgCode field set to the params
+	// payload and ethMsgCode respectively) to the nodes with the signing address in the destAddresses param.
+	Multicast(destAddresses []common.Address, payload []byte, ethMsgCode uint64, sendToSelf bool) error
+}
+
+type AnnounceManagerConfig struct {
+	IsProxiedValidator bool
+	AWallets           *atomic.Value
```
mcortesi, line 98: it's strange to have an atomic value in the `Config` object.

consensus/istanbul/backend/announce.go
```diff
@@ -111,7 +209,7 @@ func (sb *Backend) announceThread() {
 			logger.Debug("Announce version is not newer than the existing version", "existing version", sb.announceVersion, "attempted new version", version)
 			return
 		}
-		if err := sb.setAndShareUpdatedAnnounceVersion(version); err != nil {
+		if err := sb.announceManager.setAndShareUpdatedAnnounceVersion(version); err != nil {
```
mcortesi, line 212: i would expect this whole function to be part of the announce manager.. maybe on a later PR

consensus/istanbul/backend/announce.go
```diff
@@ -510,29 +608,29 @@ func (sb *Backend) generateAndGossipQueryEnode(version uint, enforceRetryBackoff
...
+	valEnodeEntries, err := m.valEnodeTable.GetValEnodes(nil)
 	if err != nil {
 		return nil, err
 	}
 
 	var queryEnodeValEnodeEntries []*istanbul.AddressEntry
 	for address, valEnodeEntry := range valEnodeEntries {
 		// Don't generate an announce record for ourselves
-		if address == sb.Address() {
+		if address == m.wallets().Ecdsa.Address {
```
mcortesi, line 633: i realie that `wallets` is a strange name. Implies we have many wallets, as in many identities/accounts. But it's just that have an ecdsa key and a bls key. We should find a better name for this

consensus/istanbul/backend/announce.go
```diff
@@ -1190,21 +1162,21 @@ func (sb *Backend) setAndShareUpdatedAnnounceVersion(version uint) error {
 			return err
 		}
 
-		if err := sb.Multicast(destAddresses, payload, istanbul.EnodeCertificateMsg, false); err != nil {
+		if err := m.support.Multicast(destAddresses, payload, istanbul.EnodeCertificateMsg, false); err != nil {
```
mcortesi, line 1165: support seems to be "network"

consensus/istanbul/backend/announce.go
```diff
@@ -1347,16 +1323,20 @@ func (sb *Backend) handleEnodeCertificateMsg(_ consensus.Peer, payload []byte) e
...
+		sb.announceManager.proxyContext.GetProxiedValidatorEngine().SendValEnodesShareMsgToAllProxies()
 	}
 
 	return nil
 }
 
 // SetEnodeCertificateMsgMap will verify the given enode certificate message map, then update it on this struct.
 func (sb *Backend) SetEnodeCertificateMsgMap(enodeCertMsgMap map[enode.ID]*istanbul.EnodeCertMsg) error {
-	logger := sb.logger.New("func", "SetEnodeCertificateMsgMap")
+	return sb.announceManager.SetEnodeCertificateMsgMap(enodeCertMsgMap)
```
mcortesi, line 1335: i suppose this method can be removed, as only the proxy calls this and it could call the announce manager directly

consensus/istanbul/backend/announce.go
```diff
@@ -1376,22 +1356,22 @@ func (sb *Backend) SetEnodeCertificateMsgMap(enodeCertMsgMap map[enode.ID]*istan
...
 		// This function may be called with the same enode certificate.
 		// Proxied validators will periodically send the same enode certificate to it's proxies,
 		// to ensure that the proxies to eventually get their enode certificates.
 		logger.Trace("Attempting to set an enode certificate with the same version as the previous set enode certificate's")
 	} else {
 		logger.Debug("Setting enode certificate", "version", *enodeCertVersion)
-		sb.enodeCertificateMsgMap = enodeCertMsgMap
-		sb.enodeCertificateMsgVersion = *enodeCertVersion
+		m.enodeCertificateMsgMap = enodeCertMsgMap
+		m.enodeCertificateMsgVersion = *enodeCertVersion
```
mcortesi, line 1374: the two functions below should be part of the announceManager if we move ownership of the valEnodeTable to it

mcortesi (commented):
A good first step. left some comments

consensus/istanbul/backend/backend.go
mcortesi: article about wrapping errors https://blog.golang.org/go1.13-errors

consensus/istanbul/backend/backend.go
hbandura: Yes, in fact that's what I wanted at first. However it's being used in a lot of places, so I want to be more careful with it. I'd like to first have a better implementation of the AnnounceManager before venturing into fully moving this one

consensus/istanbul/backend/backend.go
hbandura: +1 We can / should do those

consensus/istanbul/backend/handler.go
hbandura: Mhm. I think I like the second one better. In the future PRs (where everyting is moved into the AnnounceManager) I'll explore actual refactors to the interface instead of just moving pieces around

consensus/istanbul/backend/api.go
hbandura: I'll change it

consensus/istanbul/backend/announce.go
hbandura: Yes, actually it has already been moved in PR # 2

consensus/istanbul/backend/announce.go
hbandura: We have a discussion about that name in the exact same object for Backend with @piersy , but we couldn't find any. Definitely open for discussion but I would change that in a separate PR since it's already 'wallets' in Backend in master

consensus/istanbul/backend/announce.go
hbandura: Not as easily, no. Because the proxy calls this via an interface method, not through the Backend struct type directly.

After the whole migration to the announce manager is done, I will consider options like maybe adding a `GetAnnounceManager` to these interfaces and remove all these forwarding methods from it.

consensus/istanbul/backend/announce.go
hbandura: Yes it's something I'll definitely explore in future PRs. Moving the valEnodeTable so soon seemed like a lot.

consensus/istanbul/backend/announce.go
hbandura: agreed. changed it

consensus/istanbul/backend/announce.go
hbandura: done

consensus/istanbul/backend/announce.go
hbandura: Mmm true. I think this will get shuffled around when we have a clear understanding of what are all the dependencies and they can be regrouped.

mcortesi (approved):

PR #1591, opened 2021-06-14
Move version_certificate from announce protocol to its own file

hbandura:


piersy:
@hbandura I think this is going to conflict with #1592, which I just raised. I think #1592 should be helpful in your goal of simplifying the announce code, because it just moves lots of encode/decode code out of announce, making the announce code a bit leaner.

hbandura:
> @hbandura I think this is going to conflict with #1592, which I just raised. I think #1592 should be helpful in your goal of simplifying the announce code, because it just moves lots of encode/decode code out of announce, making the announce code a bit leaner.

I have a bigger branch that completely changes the whole announce.go file so that will definitely be a bigger conflict :)

hbandura:
I'm putting this on hold while I'll continue working on the announce-tidy branch

PR #1592, opened 2021-06-15
Consolidate istanbul.Message encoding/decoding

piersy:
### Description

Message decoding and encoding, previously happened in 2 steps. This commit
makes decoding and encoding a message an atomic operation, either it
succeeds or it fails and the full result of decoding is now stored in the message.
It also enforces the mapping between message code and message type, reducing
burden on users who previously had to manage these mappings in multiple places.

This allows us to decode in just one place in the code, allowing us to
have a consistent approach to decode errors. By being able to respond to
decode errors at a higher level in the stack, we remove the need to pass
errors back up the stack to indicate invalid messages. It also helps
separate encoding and decoding logic from our domain logic. By storing
the result of the decoding we avoid repeating the decode operation. Previously
messages going to the backlog would be decoded 3 times, the initial decoding,
then when being stored in the backlog, then again when being processed from
the backlog. Also commit messages would be decoded up to 3 times, the initial
decoding, then once for the bls signature generation then once for the epoch
signature generation.

### Tested

CI is passing

Codecov is failing and I don't understand why, I seem to be rate-limited now,
but when I last checked to report it looked like coverage had dropped in the
eth core package and I haven't touched that. So I would appreciate any help
understanding that.

### Related issues

This PR lays the groundwork for #1434, we needed a way to mark commit messages as verified and this change allows us to cache this information on the commit message.

### Backwards compatibility

This is a mechanical change substituting a 2 step process for a single step process, so backwards compatibility should not have been changed.

### DIFF
--- a/consensus/istanbul/core/handler_test.go
+++ b/consensus/istanbul/core/handler_test.go
@@ -64,7 +64,7 @@ func TestMalformedMessageDecoding(t *testing.T) {
 		Proposal: makeBlock(1),
 	}, v0.Address())
 
-	// Prepprepare message but prepare message code
+	// Preprepare message but prepare message code
 	m.Code = istanbul.MsgPrepare
 
 	payload, err = m.Payload()
--- a/consensus/istanbul/core/handler_test.go
+++ b/consensus/istanbul/core/handler_test.go
@@ -22,6 +22,8 @@ import (
 
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/consensus/istanbul"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // notice: the normal case have been tested in integration tests.
@@ -36,90 +38,77 @@ func TestHandleMsg(t *testing.T) {
 	v0 := sys.backends[0]
 	r0 := v0.engine.(*core)
 
-	m, _ := Encode(&istanbul.Subject{
+	m := istanbul.NewMessage(&istanbul.Subject{
 		View: &istanbul.View{
 			Sequence: big.NewInt(0),
 			Round:    big.NewInt(0),
 		},
 		Digest: common.BytesToHash([]byte("1234567890")),
-	})
+	}, v0.Address())
+
 	// with a matched payload. istanbul.MsgPreprepare should match with *istanbul.Preprepare in normal case.
-	msg := &istanbul.Message{
-		Code:      istanbul.MsgPreprepare,
-		Msg:       m,
-		Address:   v0.Address(),
-		Signature: []byte{},
-	}
+	m.Code = istanbul.MsgPreprepare
 
-	_, val := v0.Validators(nil).GetByAddress(v0.Address())
-	if err := r0.handleCheckedMsg(msg, val); err != errFailedDecodePreprepare {
-		t.Errorf("error mismatch: have %v, want %v", err, errFailedDecodePreprepare)
-	}
+	payload, err := m.Payload()
+	require.NoError(t, err)
 
-	m, _ = Encode(&istanbul.Preprepare{
+	err = r0.handleMsg(payload)
+	assert.Error(t, err)
+
+	m = istanbul.NewMessage(&istanbul.Preprepare{
 		View: &istanbul.View{
 			Sequence: big.NewInt(0),
 			Round:    big.NewInt(0),
 		},
 		Proposal: makeBlock(1),
-	})
+	}, v0.Address())
+
 	// with a unmatched payload. istanbul.MsgPrepare should match with *istanbul.Subject in normal case.
-	msg = &istanbul.Message{
-		Code:      istanbul.MsgPrepare,
-		Msg:       m,
-		Address:   v0.Address(),
-		Signature: []byte{},
-	}
+	m.Code = istanbul.MsgPrepare
 
-	_, val = v0.Validators(nil).GetByAddress(v0.Address())
-	if err := r0.handleCheckedMsg(msg, val); err != errFailedDecodePrepare {
-		t.Errorf("error mismatch: have %v, want %v", err, errFailedDecodePreprepare)
-	}
+	payload, err = m.Payload()
+	require.NoError(t, err)
 
-	m, _ = Encode(&istanbul.Preprepare{
+	err = r0.handleMsg(payload)
+	assert.Error(t, err)
+
+	m = istanbul.NewMessage(&istanbul.Preprepare{
 		View: &istanbul.View{
 			Sequence: big.NewInt(0),
 			Round:    big.NewInt(0),
 		},
 		Proposal: makeBlock(2),
-	})
+	}, v0.Address())
+
 	// with a unmatched payload. istanbul.MsgCommit should match with *istanbul.Subject in normal case.
-	msg = &istanbul.Message{
-		Code:      istanbul.MsgCommit,
-		Msg:       m,
-		Address:   v0.Address(),
-		Signature: []byte{},
-	}
+	m.Code = istanbul.MsgCommit
 
-	_, val = v0.Validators(nil).GetByAddress(v0.Address())
-	if err := r0.handleCheckedMsg(msg, val); err != errFailedDecodeCommit {
-		t.Errorf("error mismatch: have %v, want %v", err, errFailedDecodeCommit)
-	}
+	payload, err = m.Payload()
+	require.NoError(t, err)
 
-	m, _ = Encode(&istanbul.Preprepare{
+	err = r0.handleMsg(payload)
+	assert.Error(t, err)
+
+	m = istanbul.NewMessage(&istanbul.Preprepare{
 		View: &istanbul.View{
 			Sequence: big.NewInt(0),
 			Round:    big.NewInt(0),
 		},
 		Proposal: makeBlock(3),
-	})
-	// invalid message code. message code is not exists in list
-	msg = &istanbul.Message{
-		Code:      uint64(99),
-		Msg:       m,
-		Address:   v0.Address(),
-		Signature: []byte{},
-	}
+	}, v0.Address())
 
-	_, val = v0.Validators(nil).GetByAddress(v0.Address())
-	if err := r0.handleCheckedMsg(msg, val); err == nil {
-		t.Errorf("error mismatch: have %v, want nil", err)
-	}
+	// invalid message code. message code is not exists in list
+	m.Code = uint64(99)
+
+	payload, err = m.Payload()
+	require.NoError(t, err)
+
+	err = r0.handleMsg(payload)
+	assert.Error(t, err)
 
 	// with malicious payload
-	if err := r0.handleMsg([]byte{1}); err == nil {
-		t.Errorf("error mismatch: have %v, want nil", err)
-	}
+	err = r0.handleMsg([]byte{1})
+	assert.Error(t, err)
 }
 
 func BenchmarkHandleMsg(b *testing.B) {
--- a/consensus/istanbul/core/handler_test.go
+++ b/consensus/istanbul/core/handler_test.go
@@ -38,7 +38,7 @@ func TestHandleMsg(t *testing.T) {
 	v0 := sys.backends[0]
 	r0 := v0.engine.(*core)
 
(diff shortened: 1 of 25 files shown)


trianglesphere (commented):
I think we need to track the message flow throughout the system a little bit more, but I think that once we decode a message from RLP into what we call `Message`, we should decode it into a struct associated with the subclass (typically needs RLP decoding as well).

From that point on I kinda feel like we should use checked (rather than panicking) typecasting to deal with figuring out how to dispatch a message and then have the handler's only take that message type.

I haven't found much on how other people do this (and if you have examples, please let me know), but what we have is the following:
```
// prev
raw bytes -> Message (via RLP decode) -> switch on message.code (and use this to send it to the handler) -> decode SubMessage in handler via RLP Decode
// yours
raw bytes -> Message (via RLP decode) -> switch on message.code -> decode SubMessage into message via RLP Decode -> get sub message in handler based on message.code
// mine
raw bytes -> Message (via RLP decode) -> switch on message.code -> decode to a new type -> switch on type to send message to handler 
```
I'd like to return a new type once we've switched on message.code and decoded the sub-message. I suppose at some level using safe typecasts is not much different that doing `ptr := Message.Type()` and then checking that the pointer isn't nil, but I do see the benefit of being able to do a `switch` on the type do message dispatch.

consensus/istanbul/types.go
```diff
@@ -343,6 +411,65 @@ type Message struct {
...
+		message.versionCertificates = t
+	case *ValEnodesShareData:
+		message.Code = ValEnodesShareMsg
+		message.valEnodeShareData = t
+	default:
+		panic(fmt.Sprintf("attempt to construct Message unrecognised inner message type %T", t))
+	}
+	bytes, err := rlp.EncodeToBytes(innerMessage)
+	if err != nil {
+		panic(fmt.Sprintf("attempt to serialise inner messgae of type %T failed", innerMessage))
```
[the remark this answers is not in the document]
piersy: Hey @trianglesphere, thanks for taking a look at this. Just to clarify, the approach I'm taking here is to panic when we encounter a code error. If the function is called with the wrong type or nil it will panic. This seems to be the approach that we use throughout the codebase, for example.

https://github.com/celo-org/celo-blockchain/blob/4e42e7e8d683929b1317859d86431d5b3d3037c6/consensus/istanbul/backend/backend.go#L427-L429
https://github.com/celo-org/celo-blockchain/blob/4e42e7e8d683929b1317859d86431d5b3d3037c6/consensus/istanbul/backend/backend.go#L431-L433
https://github.com/celo-org/celo-blockchain/blob/4e42e7e8d683929b1317859d86431d5b3d3037c6/consensus/istanbul/backend/backend.go#L686-L688

In all of these cases and many more, the approach we take is to panic if a nil parameter is passed to a function (coding error). So I'd like to understand what it is about this particular panic that makes you uncomfortable but not all the others that we have in the codebase?

piersy:
> I think we need to track the message flow throughout the system a little bit more, but I think that once we decode a message from RLP into what we call `Message`, we should decode it into a struct associated with the subclass (typically needs RLP decoding as well).
> 
> From that point on I kinda feel like we should use checked (rather than panicking) typecasting to deal with figuring out how to dispatch a message and then have the handler's only take that message type.
> 
> I haven't found much on how other people do this (and if you have examples, please let me know), but what we have is the following:
> 
> ```
> // prev
> raw bytes -> Message (via RLP decode) -> switch on message.code (and use this to send it to the handler) -> decode SubMessage in handler via RLP Decode
> // yours
> raw bytes -> Message (via RLP decode) -> switch on message.code -> decode SubMessage into message via RLP Decode -> get sub message in handler based on message.code
> // mine
> raw bytes -> Message (via RLP decode) -> switch on message.code -> decode to a new type -> switch on type to send message to handler 
> ```
> 
> I'd like to return a new type once we've switched on message.code and decoded the sub-message. I suppose at some level using safe typecasts is not much different that doing `ptr := Message.Type()` and then checking that the pointer isn't nil, but I do see the benefit of being able to do a `switch` on the type do message dispatch.

Hey I like the idea of decoding to a new type and passing that to the appropriate handler, and that would be my preferred solution if working without constraints. But the problem with doing that now is just that it's going to be a big change, round state has message sets which hold istanbul.Message instances, and istanbul.Message instances are used throughout the handler code, so doing this would end up rewriting most of the Istanbul package.

consensus/istanbul/types.go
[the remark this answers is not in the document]
trianglesphere: Sorry, wrong line: I meant it for `panic(fmt.Sprintf("attempt to construct Message unrecognised inner message type %T", t))`. For that, it's easy enough to have a specialization of the new message function for each type which would completely eliminate that panic. That would help us in the case that an incorrect type got passed into the function (which given we take `interface{}` is a logic error, but also something that we can eliminate at compile time).

For this panic, I think it's probably fine. Removing it requires changing rlp.Encode to not error, and my understanding of what would cause rlp to error is that it has an internal logic error or you're trying to encode an incorrect type. I think it'd be hard to accidentally pass a bad type to be RLP encoded here.

trianglesphere:
> But the problem with doing that now is just that it's going to be a big change, round state has message sets which hold istanbul.Message instances

I guess it regresses with what you are trying to do, but you could put the raw message type into the backlog and decode it multiple times. Then the hacky solution is that each specialized message type has a reference back to the `Message` object that it was constructed from (so it can place store itself in the round change certificate or round state db).

> istanbul.Message instances are used throughout the handler code, so doing this would end up rewriting most of the Istanbul package.

I don't think it'd do so much more than what this PR is doing (and is still a worthwhile change).

consensus/istanbul/core/roundchange.go
```diff
@@ -29,51 +29,24 @@ import (
 
 // sendRoundChange broadcasts a ROUND CHANGE message with the current desired round.
 func (c *core) sendRoundChange() {
-	logger := c.newLogger("func", "sendRoundChange")
-
-	msg, err := c.buildRoundChangeMsg(c.current.DesiredRound())
-	if err != nil {
-		logger.Error("Could not build round change message", "err", msg)
```
trianglesphere, line 36: I'm picking this as a random example, but I think my preference is to potentially hang and be unable to progress rather than panicking.

consensus/istanbul/core/roundchange.go
```diff
@@ -175,12 +145,7 @@ func (c *core) handleRoundChangeCertificate(proposal istanbul.Subject, roundChan
 func (c *core) handleRoundChange(msg *istanbul.Message) error {
 	logger := c.newLogger("func", "handleRoundChange", "tag", "handleMsg", "from", msg.Address)
 
-	// Decode ROUND CHANGE message
-	var rc *istanbul.RoundChange
```
trianglesphere, line 179: Ditto that I'd like some form of this check (either enforced through the type system or check that `rc != nil`).

consensus/istanbul/types.go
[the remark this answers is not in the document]
piersy: So just to clarify, you are saying that you are happy with this panic, but only in the case that we create a set of typesafe constructors?

So in this case:
```
NewQueryEnodeMessage(en *queryEnode, address common.Address)
```

If rlp decode fails its because *queryEnode is not rlp encodable, and in that case you are ok with the panicking.

But in this case:
```
NewMessage(inner interface{}, address common.Address)
```

If rlp decode fails its because whatever type `inner` is,  cannot be rlp encoded, and in that case you would be against panicking?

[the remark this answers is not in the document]
trianglesphere: In the second case, I don't like the panic if we hit the default case of the switch on the type. I'm actually a little split on panicking if the rlp encode fails, but that is less risky than panicking if we pass the wrong type to `interface {}`.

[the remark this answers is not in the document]
piersy: So as discussed I'm going to make this change on this PR.

piersy:
> > But the problem with doing that now is just that it's going to be a big change, round state has message sets which hold istanbul.Message instances
> 
> I guess it regresses with what you are trying to do, but you could put the raw message type into the backlog and decode it multiple times. Then the hacky solution is that each specialized message type has a reference back to the `Message` object that it was constructed from (so it can place store itself in the round change certificate or round state db).
> 
> > istanbul.Message instances are used throughout the handler code, so doing this would end up rewriting most of the Istanbul package.
> 
> I don't think it'd do so much more than what this PR is doing (and is still a worthwhile change).

As we discussed, we agree that this is a good change to make (passing specific types to handlers rather than istanbul.Message), but it probably makes sense to make this change in subsequent PRs as the results will probably be quite wide ranging.

consensus/istanbul/core/handler_test.go
```diff
@@ -36,90 +38,77 @@ func TestHandleMsg(t *testing.T) {
...
-	_, val := v0.Validators(nil).GetByAddress(v0.Address())
-	if err := r0.handleCheckedMsg(msg, val); err != errFailedDecodePreprepare {
-		t.Errorf("error mismatch: have %v, want %v", err, errFailedDecodePreprepare)
-	}
+	payload, err := m.Payload()
+	require.NoError(t, err)
 
-	m, _ = Encode(&istanbul.Preprepare{
+	err = r0.handleMsg(payload)
+	assert.Error(t, err)
```
trianglesphere, line 56: I'm a little worried that these tests are no longer testing what we want them to test.
Previously we were looking for the specific error (at least on the first 3 cases), but we send in the correct message (or we accidentally admit incorrect messages), this test will still pass (first is we need to sign, then make sure that the state is properly set up to accept the message based on it's code).

trianglesphere (commented):
two nits + one question on  some of the tests. Otherwise looks good to me.

consensus/istanbul/proxy/proxy_engine_test.go
```diff
@@ -109,6 +109,9 @@ func TestHandleValEnodeShare(t *testing.T) {
 	}
 
 	expectedVetEntry := istanbul.AddressEntry{Address: val1BE.Address(), Node: val1BE.SelfNode()}
+	spew.Dump(expectedVetEntry)
```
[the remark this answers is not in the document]
piersy: yup!

consensus/istanbul/core/handler_test.go
piersy: Hey @trianglesphere, are you saying that we need to ensure that the mock engine needs to be setup correctly, otherwise practically any input would return an error from `handleCheckedMsg`?

trianglesphere: Yea. I think that the first set of message needs signatures and then the core to be in the correct state (such that they would accept the message if the payload was correct). If the core is not in the correct state for the message type, it rejects it with err old message or err future message.

 For the later tests with totally corrupt payloads, b/c we don't report a specific error, I think the best we can do is keep checking for an error.

piersy: Looking at the previous tests I think what they were really trying to assert is that if a message is invalid then when decoded an error is returned.

Since all the decoding is now handled in `FromPayload`, rather than being spread out across the 3 message handlers ,I think we could replace these test cases with direct calls to `FromPayload`. We wouldn't be asserting that errors raised by `FromPayload` were propagated by `handleMsg`, but the previous tests called `handleCheckedMsg` directly and also didn't assert that errors raised by `handleCheckedMsg` were propagated by `handleMsg` (except for the malicious payload case, which hasn't really changed so we can keep it as it was). So I think this approach would be roughly equivalent.

piersy: @trianglesphere I've now updated the tests in line with my previous comment.

consensus/istanbul/core/handler_test.go
```diff
@@ -63,13 +64,14 @@ func TestHandleMsg(t *testing.T) {
 		Proposal: makeBlock(1),
 	}, v0.Address())
 
-	// with a unmatched payload. istanbul.MsgPrepare should match with *istanbul.Subject in normal case.
+	// Prepprepare message but prepare message code
```
trianglesphere, line 67:
```suggestion
	// Pre-prepare message but prepare message code
```

trianglesphere (approved):
1 misspell that lint didn't catch, but otherwise good to go.

PR #1596, opened 2021-06-16
Set master version to be ahead of latest release, and add comment about versioning

### LINKED ISSUE #1580: Geth version is not set correctly between releases
## Description

Getting the version from geth  on master (ab78658ae3a738d91d877f289e37fead2a0f1cf7) reports 1.0.0-unstable which is wrong.

```
$ ./celo-blockchain/build/bin/geth version 
Celo
Version: 1.0.0-unstable
Git Commit: ab78658ae3a738d91d877f289e37fead2a0f1cf7
Git Commit Date: 20210527
Architecture: amd64
Protocol Versions: [66 65 64]
Go Version: go1.16.4
Operating System: darwin
GOPATH=
GOROOT=/usr/local/go
```

## Proposed Solution

The variables in `params/version.go` need to be updated to reflect the correct version and we need to make this part of the release process.

oneeman:
### Description

We have previously been setting versions only on release branches, leaving master permanently at v1.0.0-unstable.  Besides being confusing, this wouldn't work properly if we ever increase the minimum client version on the `BlockchainParameters` contract.

Instead, this PR sets it to v1.4.0-unstable, since the latest stable release was v1.3.2, so the next release branch will be v1.4.x.  It also adds a comment describing how the versioning should go.

### Tested

Verified with `geth version`:

```
❯ geth version
Celo
Version: 1.4.0-unstable
Git Commit: 538d2f885984db4e0aff0bdaab22a562a8ad2d39
Git Commit Date: 20210616
[...]
```

### Related issues

- Fixes #1580

### Backwards compatibility

No incompatibility.

### DIFF
--- a/params/version.go
+++ b/params/version.go
@@ -20,9 +20,13 @@ import (
 	"fmt"
 )
 
+// On master, the version should be the UPCOMING one and "unstable"
+// e.g. if the latest release was v1.3.2, master should be 1.4.0-unstable
+// On release branches, it should be a beta or stable.  For example:
+// "1.3.0-beta", "1.3.0-beta.2", etc. and then "1.3.0-stable", "1.3.1-stable", etc.
 const (
 	VersionMajor = 1          // Major version component of the current release
-	VersionMinor = 0          // Minor version component of the current release
+	VersionMinor = 4          // Minor version component of the current release
 	VersionPatch = 0          // Patch version component of the current release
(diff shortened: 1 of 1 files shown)


alice-dfi:
Set master version permanently??? What ? Why is being changed do not change
anything with out my approval, that has to do with me .

trianglesphere (approved):

PR #1597, opened 2021-06-17
Move several sign&address properties from istanbul.Backend to its own type

hbandura:
Properties received in the `Authorize(...)` method of `istanbul.Backend` are guarded by a lock to avoid race conditions during  the setting of the values (golang reference assigmnent is not guaranteed to be atomic I believe).

I moved all of those into a type `AuthorizeInfo` and used an `atomic.Value` to be able to remove the lock. The `decryptFn` was not being guarded by the lock, as many other `address` fields, in what I believe it was a possible race condition bug during assignment and check. It's probably the case that it would have never been triggered in real world use though.

I also incremented a sleep in an unrelated test since the wait period for a geth instance to start was too low and it was making the test a bit flaky.

### DIFF
--- a/cmd/geth/les_test.go
+++ b/cmd/geth/les_test.go
@@ -66,7 +66,7 @@ func startGethWithRpc(t *testing.T, name string, args ...string) *gethrpc {
 	t.Logf("Starting %v with rpc: %v", name, args)
 	g.geth = runGeth(t, args...)
 	// wait before we can attach to it. TODO: probe for it properly
-	time.Sleep(1 * time.Second)
+	time.Sleep(2 * time.Second)
 	var err error
 	ipcpath := filepath.Join(g.geth.Datadir, "geth.ipc")
 	g.rpc, err = rpc.Dial(ipcpath)
--- a/cmd/geth/les_test.go
+++ b/cmd/geth/les_test.go
@@ -66,7 +66,7 @@ func startGethWithRpc(t *testing.T, name string, args ...string) *gethrpc {
 	t.Logf("Starting %v with rpc: %v", name, args)
 	g.geth = runGeth(t, args...)
 	// wait before we can attach to it. TODO: probe for it properly
-	time.Sleep(2 * time.Second)
+	time.Sleep(1 * time.Second)
 	var err error
 	ipcpath := filepath.Join(g.geth.Datadir, "geth.ipc")
 	g.rpc, err = rpc.Dial(ipcpath)
--- a/consensus/istanbul/backend/announce.go
+++ b/consensus/istanbul/backend/announce.go
@@ -591,16 +591,16 @@ func (sb *Backend) generateQueryEnodeMsg(version uint, enodeQueries []*enodeQuer
 		logger.Error("Error encoding queryEnode content", "QueryEnodeData", queryEnodeData.String(), "err", err)
 		return nil, err
 	}
-	ai := sb.wallets()
+	w := sb.wallets()
 	msg := &istanbul.Message{
 		Code:      istanbul.QueryEnodeMsg,
 		Msg:       queryEnodeBytes,
-		Address:   ai.Ecdsa.Address,
+		Address:   w.Ecdsa.Address,
 		Signature: []byte{},
 	}
 
 	// Sign the announce message
-	if err := msg.Sign(ai.Ecdsa.Sign); err != nil {
+	if err := msg.Sign(w.Ecdsa.Sign); err != nil {
 		logger.Error("Error in signing a QueryEnode Message", "QueryEnodeMsg", msg.String(), "err", err)
 		return nil, err
 	}
@@ -935,13 +935,13 @@ func (vc *versionCertificate) payloadToSign() ([]byte, error) {
 }
 
 func (sb *Backend) generateVersionCertificate(version uint) (*versionCertificate, error) {
-	ai := sb.wallets()
+	w := sb.wallets()
 	vc := &versionCertificate{
-		Address:   ai.Ecdsa.Address,
-		PublicKey: ai.Ecdsa.PublicKey,
+		Address:   w.Ecdsa.Address,
+		PublicKey: w.Ecdsa.PublicKey,
 		Version:   version,
 	}
-	err := vc.Sign(ai.Ecdsa.Sign)
+	err := vc.Sign(w.Ecdsa.Sign)
 	if err != nil {
 		return nil, err
 	}
@@ -1266,7 +1266,7 @@ func (sb *Backend) generateEnodeCertificateMsgs(version uint) (map[enode.ID]*ist
 	if err != nil {
 		return nil, err
 	}
-	ai := sb.wallets()
+	w := sb.wallets()
 	for _, externalNode := range externalEnodes {
 		enodeCertificate := &istanbul.EnodeCertificate{
 			EnodeURL: externalNode.URLv4(),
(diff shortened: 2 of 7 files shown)


cmd/geth/les_test.go
```diff
@@ -66,7 +66,7 @@ func startGethWithRpc(t *testing.T, name string, args ...string) *gethrpc {
 	t.Logf("Starting %v with rpc: %v", name, args)
 	g.geth = runGeth(t, args...)
 	// wait before we can attach to it. TODO: probe for it properly
-	time.Sleep(1 * time.Second)
+	time.Sleep(2 * time.Second)
```
piersy, line 69: What is happening here?

consensus/istanbul/backend/backend.go
```diff
@@ -57,6 +58,16 @@ var (
 	errInvalidSigningFn = errors.New("invalid signing function for istanbul messages")
 )
 
+type AuthorizeInfo struct {
```
piersy, line 61: If you look at what this does, I think Wallet is an appropriate name. It holds the adderess and provides functions to sign and decrypt.

consensus/istanbul/backend/announce.go
```diff
@@ -937,7 +937,7 @@ func (vc *versionCertificate) payloadToSign() ([]byte, error) {
 func (sb *Backend) generateVersionCertificate(version uint) (*versionCertificate, error) {
 	vc := &versionCertificate{
 		Address:   sb.Address(),
-		PublicKey: sb.publicKey,
+		PublicKey: sb.authorizeInfo.Load().(*AuthorizeInfo).PublicKey,
```
piersy, line 940: I think it would be nice to have a function that encapsulates the loading of the atomic value.
```suggestion
		PublicKey: sb.auth().PublicKey,
```

consensus/istanbul/backend/backend.go
```diff
@@ -366,22 +372,22 @@ func (sb *Backend) SendDelegateSignMsgToProxiedValidator(msg []byte) error {
 
 // Authorize implements istanbul.Backend.Authorize
 func (sb *Backend) Authorize(ecdsaAddress, blsAddress common.Address, publicKey *ecdsa.PublicKey, decryptFn istanbul.DecryptFn, signFn istanbul.SignerFn, signBLSFn istanbul.BLSSignerFn, signHashFn istanbul.HashSignerFn) {
```
piersy, line 374: It would be nice if  we could do this at contstruction time, then we wouldn't need the synchronisation at all. Its probably a fair bit more work though. It seems that there is a SetValidator method on Ethereum which is accessible through the api. That seems to be the sticking point for now.

I don't see it as a particularly useful api call for most people, because I can't imagine a validator node changing its address during execution, since validators need to be elected and different validator addresses will be owned by different people so validator addresses won't change during the lifetime of a node.

I think in fact it is likely to catch people out because setting it via a call to setEtherbase witll set the validator on the eth service and calling the corresponding ehterbase api method will then return that value, but the new validator address won't have any effect on the backend until it the backend is stopped and started  by calls to stop and start mining. There doesn't seem to be any documentation covering that or the fact that for the restart to work you need to have imported all the new keys into the node's keystore otherwise calls to start mining will fail.

But if we could remove the concept of setting the validator then we could set this once at construction time and not have to worry about synchronisation at all.

consensus/istanbul/backend/announce.go
```diff
@@ -937,7 +937,7 @@ func (vc *versionCertificate) payloadToSign() ([]byte, error) {
...
 		Version:   version,
 	}
 	err := vc.Sign(sb.Sign)
```
piersy, line 943: There are 3 race conditions here, we do an atomic load for the Address the public key and for the sign operation. Between each load Authorize could have been called. We can solve this by just loading at the top of the function and making sure we reference all values from the loaded instance.

cmd/geth/les_test.go
hbandura: "I also incremented a sleep in an unrelated test since the wait period for a geth instance to start was too low and it was making the test a bit flaky."

I could do it in a separate PR but honestly it was making me nervous

consensus/istanbul/backend/backend.go
hbandura: I 100% agree on this, and I also think that it would lead to a better codebase. However I want to keep the focus on the actual motive for this PR, which is making it easier to write a clean AnnounceManager. And with our without your proposed changes (which I agree on), we'll still need probably a way to group these fields together instead of having them hanging in the istanbul Backend. The atomic would be moot, that's for sure, though it's definitely not the focus of this PR, but the ill named `AuthorizeInfo` is.

consensus/istanbul/backend/announce.go
hbandura: You are right! Nice catch

consensus/istanbul/backend/announce.go
hbandura: I actually had one but ended up deleting it because I thought it was an overkill. If you think it looks better I'll re-add it.

consensus/istanbul/backend/backend.go
hbandura: I like this.

TBH I'm not sure either about the relationship keys and signing functions :) I know that signHash relates to the Wallet.SignHash function (that was removed in upstream but kept here), so I'm guessing it's the ECDSA but...

consensus/istanbul/backend/announce.go
piersy: I think it makes sense because you have 10 instances of this loading code, i think its nice to encapsulate the specifics of that.

consensus/istanbul/backend/backend.go
piersy: Hehe :), yeah, I'm not sure, I'd have to dig into the code to find out, but that's why it would be nice for people if it was just explicit.

consensus/istanbul/backend/backend.go
piersy: Sure, I'm definitely for the grouping. I just wanted to raise the idea, I wasn't expecting to roll it into this PR, I should have been more clear about that. But it's good to know that you concur :)

consensus/istanbul/backend/backend.go
piersy: What about changing the name to Wallet? I think its more descriptive than AuthorizeInfo, and then `auth()` would become `wallet()`

cmd/geth/les_test.go
piersy: I think since this change is totally unrelated to the rest of the PR, it would be better to not include it and instead we can submit a better fix (the probing) to upstream.

hbandura: fair, I'll remove it

consensus/istanbul/backend/backend.go
hbandura: I thought about it, but this is a grouping of different properties from some of the validator wallets, and I would like to avoid the confusion when someone sees `backend.wallet...`. I also don't like `AuthorizeInfo`, but it feels like `wallet` would make the code a little bit more confusing.

piersy: I realise now I was actually previously suggesting the name Wallets, would you consider that any better?

The reason I don't like AuthorizeInfo as a name is that the meaning of Authorize is:

>to give official permission for something to happen, or to give someone official permission to do something

So autrhorizing is an action, so it makes sense for the method to be called authorize, but what is being given in this authorize action is a set of keys and addresses, which in the blockchain space is often referred to as a wallet.

Its not a biggy, just interested to hear your thoughts on this?

hbandura: Well I hate AuthorizeInfo, so I'm all in for changing it. I think Wallets is acceptable. I'll change it.

piersy: So one little last thing, now that its called Wallets and we have wallets() `ai` as a variable name doesn't make much sense. I'd be happy with `w` or `wal`.

hbandura: agh, find & replace betrayed me

piersy (approved):
Looking good @hbandura :+1:

PR #1606, opened 2021-06-23
Piersy/update prepare-system-contracts make rule

piersy:
### Description

 This PR fixes a couple of problems with the prepare-system-contracts make rule.

* The rule previously appeared first in the Makefile, which meant that running just `make` would now run `prepare-system-contracts` instead of `geth`.
* The rule also checked out celo-monorepo under the celo-blockchain dir and this slowed down gopls massively, because of the huge amount of files in celo-monorepo.
* The rule did not handle celo-monorepo commit version changes gracefully, now when updating the version the new version will be fetched, checked out and then built.
* The default checkout path clashed with some people's setup so that has been changed to be something less likely to clash.
* Some scripts were updated to ignore the monorepo symlink at the repo root.

### Tested

I've tested it locally and it seems to work.

### Backwards compatibility

Yes, no node code was changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ MONOREPO_COMMIT=celo-core-contracts-v3.rc0
 # We checkout the monorepo as a sibling to the celo-blockchain dir because the
 # huge amount of files in the monorepo interferes with tooling such as gopls,
 # which becomes very slow.
-MONOREPO_PATH=../monorepo
+MONOREPO_PATH=../.celo-blockchain-monorepo-checkout
 
 # This either evaluates to the contract source files if they exist or NOT_FOUND
 # if celo-monorepo has not been checked out yet.
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,11 @@ endif
 
 MONOREPO_COMMIT=celo-core-contracts-v3.rc0
 
+# We checkout the monorepo as a sibling to the celo-blockchain dir because the
+# huge amount of files in the monorepo interferes with tooling such as gopls,
+# which becomes very slow.
+MONOREPO_PATH=../monorepo
+
 # example NDK values
 export NDK_VERSION ?= android-ndk-r19c
 export ANDROID_NDK ?= $(PWD)/ndk_bundle/$(NDK_VERSION)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -24,28 +24,74 @@ endif
 
 MONOREPO_COMMIT=celo-core-contracts-v3.rc0
 
+# We checkout the monorepo as a sibling to the celo-blockchain dir because the
+# huge amount of files in the monorepo interferes with tooling such as gopls,
+# which becomes very slow.
+MONOREPO_PATH=../monorepo
```
trianglesphere, line 30: My only concern is this default. For my directory structure I already have `celo-monorepo` checkout out as `monorepo` as a sibling. I'd prefer something that was less likely to clash and made the purpose of the specific checkout more clear.

trianglesphere (approved):
One comment on the defaults. I'm not sure why gopls ignores symlinks, but it does and it's performing way better. I'll still have to update some of the imports scripts, but that's easy enough to do.

Makefile
piersy: Sure @celo-org/blockchain suggestions?

trianglesphere: I used `contracts_monorepo`, but that or `e2e_monorepo` or `e2e_contracts_monorepo`. I don't have too strong a preference on which variant.

oneeman: > For my directory structure I already have celo-monorepo checkout out as monorepo as a sibling.

Same here!  I'd be fine with whatever if it's unlikely to clash. It could even start with `.` (e.g. `.celo_blockchain_e2e_monorepo`). But also it might be nice to make it overridable with an environment variable.

trianglesphere: it's further down in the docs, but if you run make with `MONOREPO_PATH=$PATH` it will do the setup correctly (and then b/c it's symlinked, you don't need the env variable)

piersy: Ok, I'm going for `.celo-blockchain_monorepo_checkout` unless anyone objects?

oneeman: I object to mixing `-` and `_`.  If we want celo-blockchain due to that being the repo name, might as well use dashes for all of them.

PR #1608, opened 2021-06-23
Exclude the monorepo from recursive operations

trianglesphere:
### Description

With #1564 we vendor the monorepo to use the contracts for network testing.
In #1606 the monorepo is symlinked which helps with gopls, but not these scripts.
This commit exclude the monorepo from recursive operations performed in the
modified scripts.


piersy:
@trianglesphere I think we should just add this commit onto #1606, since it doesn't make sense by itself

trianglesphere:
> @trianglesphere I think we should just add this commit onto #1606, since it doesn't make sense by itself. Changes look fine though

Sounds good to me. Feel free to cherry-pick this over.

piersy:
> > @trianglesphere I think we should just add this commit onto #1606, since it doesn't make sense by itself. Changes look fine though
> 
> Sounds good to me. Feel free to cherry-pick this over.

:cherries: :yum:

PR #1611, opened 2021-07-02
Go back to using the transfer's sender in Tobin tax calculation.

oneeman:
### Description

Even though the Tobin tax calculation doesn't depend on the sender, the change (in PR #1465) of the sender from the transaction's sender to the zero account led to a bad block error (i.e. a hard fork change).  This was due to a conjunction of factors including the upstream implementation of EIP-161 not being quite correct for the case of Celo (because of the existence of system EVM calls made from celo-blockchain).  This commit reverts the change, by adding and using an `ExecuteFrom()` method on the EVMRunner type.

### Tested

Confirmed that this fixes the bad block which previously appeared trying to sync to Alfajores (block 536) and that it now syncs successfully past that block.

### Backwards compatibility

This fixes a breaking change that is currently on master.  As such, it is incompatible with what is currently on master (i.e. it could potentially cause bad blocks errors trying to sync to networks which have been running the current master).  However, it makes master backwards compatible (in this regard) relative to the stable releases and v1.3.2 in particular.

### DIFF
--- a/core/vm/vmcontext/runner.go
+++ b/core/vm/vmcontext/runner.go
@@ -25,7 +25,7 @@ type evmRunnerContext interface {
 }
 
 type evmRunner struct {
-	newEVM func() *vm.EVM
+	newEVM func(from common.Address) *vm.EVM
 	state  vm.StateDB
 
 	dontMeterGas bool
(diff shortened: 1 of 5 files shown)


core/vm/vmcontext/runner.go
```diff
@@ -53,6 +53,15 @@ func (ev *evmRunner) Execute(recipient common.Address, input []byte, gas uint64,
 	return ret, err
 }
 
+func (ev *evmRunner) ExecuteFrom(sender, recipient common.Address, input []byte, gas uint64, value *big.Int) (ret []byte, err error) {
+	evm := ev.newEVM()
```
mcortesi, line 57: The elm origin is still the vmaddress

oneeman: Good catch, totally missed that.  The two options are (a) adding the `from` as an argument to `ev.newEVM()` or (b) overwriting `evm.Origin` after the evm has been created.  At first my inclination was to the second because it's less disruptive.  But I'm not so sure, because that would assume that all that would be done with the `from` is to store it as the `Origin` property.  That is true now, but I'm not sure it's something we should add as an implicit assumption.  What do you think?

oneeman: Actually, including it as an argument is a small change and pretty local, so I'll do that.

oneeman: Though the whole thing seems kind of weird.  Normally 'origin' is the EOA that initiated the transaction and 'caller' is whoever called this smart contract (could be another contract).  I guess for the core contracts it doesn't matter much since they don't use `tx.origin`, and also because the only place which uses `ExecuteFrom()` is the tobin tax calculation and that uses `sharedEvmRunner` rather than `evmRunner`...

mcortesi (approved):

PR #1614, opened 2021-07-05
Update gitignore

gastonponti:
### Description

Update .gitignore to ignore the monorepo folder from every part of the repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,4 @@ docs/_site
 docs/.jekyll-metadata
 
 # Ignore monorepo checkout
-./monorepo
+**/monorepo


hbandura (approved):

oneeman:
@gastonponti Is this necessary?  My understand was that the monorepo is now checked out stored as a sibling of celo-blockchain (defaulting to `.celo-blockchain-monorepo-checkout`) rather than inside celo-blockchain as before.  In that case we can remove it from the gitignore entirely, right?  cc @piersy

gastonponti:
@oneeman yes, but we are creating a symbolic link inside the repo
I've added that to avoid adding the symlink or even the folder, in case of a mistake. I've already talked with @piersy and he will change a few things there that will avoid that symlink. This was more of a "preventing .gitignore" in the meanwhile. I would leave this gitignore until that change to avoid pushing something by mistake

piersy:
@oneeman this was a mistake, I forgot to update the gitignore when I made the change to use a symlink

PR #1616, opened 2021-07-06
Update prepare-system-contracts make rule

piersy:
### Description

This commit makes two changes:

1. Instead of symlinking to the celo-monorepo root from the
celo-blockchain root, we now symlink just to the compiled contracts.
This should help our tooling by reducing the amount of files in scope of
searches etc.

2. The overrides for the make rule (MONOREPO_PATH & MONOREPO_COMMIT) now
are not persistent and need to be provided on all invocations of the
make rule. This change was made because the previous implementation
could end up functioning in strange ways due to the statefulness. This
approach is just more straightforward.

### Tested

Works locally and on ci?

### Backwards compatibility

Yes, no node source code has been changed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,9 @@
 version: 2.1
 parameters:
+  # Increment this to force cache rebuilding
+  system-contracts-cache-version:
+    type: integer
+    default: 1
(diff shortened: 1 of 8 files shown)


piersy:
Closing in favour of #1617

PR #1617, opened 2021-07-08
Update prepare-system-contracts make rule

piersy:
### Description
This PR makes two changes:

1. Instead of symlinking to the celo-monorepo root from the
celo-blockchain root, we now copy just to the compiled contracts from
the monorepo to compiled-system-contracts. This should help our tooling
by reducing the amount of files in scope of searches etc. Previously
certain tooling would follow the symlink and end up processing all files
in celo-monorepo resulting in significant slowdown.

2. The overrides for the make rule (MONOREPO_PATH & MONOREPO_COMMIT)
now are not persistent and need to be provided on all invocations of the
make rule. This change was made because the previous implementation
could end up functioning in strange ways due to the statefulness. This
approach is just more straightforward.

Once you've pulled this update you should delete the `monorepo` symlink at
the root of the celo-blockchain repo.

### Other changes

CI has also been updated to handle the new path for the contracts.

### Tested

Tested locally and works on CI.

### Related issues

#1564
#1476 

### Backwards compatibility

Yes. No node code was changed.


Makefile
```diff
@@ -48,44 +48,44 @@ geth:
...
-# contracts with a consistent path.
-prepare-system-contracts: ./monorepo/packages/protocol/build
-
-# If any of the source files found by the find command are more recent than the
+# MONOREPO_COMMIT and compiles the system solidty contracts. It then copies the
+# compiled contracts from the monorepo to the compiled-system-contracts, so
+# that this repo can always access the contracts at a consistent path.
+prepare-system-contracts: $(MONOREPO_PATH)/packages/protocol/build
+	@rm -rf compiled-system-contracts
+	@cp -ar $(MONOREPO_PATH)/packages/protocol/build/contracts compiled-system-contracts
```
trianglesphere, line 56: The lowercase `-r` is redundant (and in conflict with) the `-a`
```suggestion
	@cp -a $(MONOREPO_PATH)/packages/protocol/build/contracts compiled-system-contracts
```

piersy: Well spotted!

Although is the `-r` really in conflict with `-a`? For my version of cp it looks like the `-r` is just redundant, since `-a` is shorthand for `-dR --preserve=all` and `-r` is equivalent to `-R`?

trianglesphere: My version (BSD cp with mac osx) was claiming conflicts.

trianglesphere (approved):

PR #1620, opened 2021-07-12
Fix logic bug in worker, and protect against blocking.

oneeman:
### Description

Due to a logic bug in the `if`'s condition, the channel was only being sent to, never received from, causing it to hang on validators once enough transactions have been sent.  This PR fixes that and also ensures that if the channel is full we don't block the main loop.

### Tested

* Automated testing
* Added regression tests for validator and non-validator cases which send lots of transactions to the channel and fail if deadlock results.  These tests would fail if the fix in this PR is undone.


oneeman:
Added a couple of regression tests adapted from ones @trianglesphere had suggested.

trianglesphere (approved):
LGTM

PR #1622, opened 2021-07-13
Fix VersionCertificate rlp encoding/decoding

### LINKED ISSUE #1621: Breaking change made to version certificate RLP encoding
PR #1592 broke backwards compatibility of the version certificate type (formerly called `VersionCertificateEntry`, now called `VersionCertificate`) when RLP-encoded.  Definitely the order of the fields has changed, which changes the RLP encoding.  This led existing nodes to fail when trying to decode values from their database.  We will need to put the fields back in the original order and verify that no other breaking changes have been made in the methods which do the RLP encoding and decoding for this type.  And may also be a good idea to see whether similar problems may exist for other types involved in that PR.

piersy:
### Description

PR #1592 introduced a backwards incompatibilty because it changed the
rlp structure that version certificates were stored and retrieved with
in the version certificate db.

This PR reinstates the previous rlp structure for use by the version
certificate db, thereby restoring backwards compatibility.

### Tested

I've run a node prior to #1592 and then run a node with this branch with the same data dir and not witnessed any encoding or decoding errors.

### Related issues

- Fixes #1621 

### Backwards compatibility

These changes are restoring backwards compatibility.


oneeman:
Haven't reviewed this yet, and I didn't realize the serialization for the peers and for the database were different.

But I had a thought that we might add regression tests for this, which have an example for which we know the correct encoding (e.g. obtained by encoding using v1.3.2), and the test can (a) decode and verify the result is correct, (b) re-encode and verify it's the same.  And since there seem to be two different encodings we could do it for both, I guess.

oneeman:
Some context on this:

* Prior to #1592, there were two types: `VersionCertificateEntry` (used for encoding/decoding to RLP for the database) and `VersionCertificate` (used for communicating with peers), and each had their own RLP encode/decode implementation
* #1592 consolidated them into a single type called `VersionCertificate` but broke compatibility with the encoding that was used for the database

This PR in its current state does the following:

* Keep it as a single type
* Add functions `encodeVersionCertificate` and `decodeVersionCertificate`, which employ the format we had used for storing in the database, but differ in that it stores zero values for the public key and address, and when decoding computes those from the signature instead of using the values from the database.  In other words, the format is kept the same but it's interpretation is changed.

What I would recommend is:
* Have `encodeVersionCertificate` and `decodeVersionCertificate` work the way things were before (so that the content and interpretation doesn't change).  Since we are not getting rid of the old format, I don't think we should modify it's content, either, since it doesn't make things simpler and requires additional explanation in the comments.

Another possibility would be to define a new version for the version certificate database, so that the old one gets wiped, and then we could use the same format that the p2p messages use.  But we would have to confirm that this is safe (i.e. that proxies and validators would unproblematically recover when that happens.

piersy:
Hey @oneeman thanks for looking at this.

I also thought about regression tests, wasn't sure how to do that nicely though.

Tomorrow, I'll update the encode/decode functions to match the previous logic.

consensus/istanbul/types.go
```diff
@@ -914,7 +914,12 @@ func NewVersionCertificate(version uint, signingFn func([]byte) ([]byte, error))
 	if err != nil {
 		return nil, err
 	}
-	return vc, vc.recoverAddressAndPubKey()
+	err = vc.recoverAddressAndPubKey()
+	if err != nil {
+		return nil, err
+	}
+
+	return vc, nil
```
piersy, line 922: This change is not related to this PR's goal, it was made because @oneeman felt that clarity was harmed by having a return statement call a function (`recoverAddressAndPubKey`) that modifies the first returned parameter.

oneeman (approved):
Looks good to me.  May still be worth it to add a regression test using the approach I outlined in the earlier comment.  Now that I realize we have two serialization formats, I guess it should test both of them.  But can also merge as is.

piersy:
> Looks good to me. May still be worth it to add a regression test using the approach I outlined in the earlier comment. Now that I realize we have two serialization formats, I guess it should test both of them. But can also merge as is.

Hey @oneeman  I made a ticket to add the regression tests, but going to merge this now so that people are less likely to face problems using master
* #1623

Issue #1627, opened 2021-07-16
Light client panics on `eth_getBlockByHash` with an incorrect block hash


oneeman:
## Expected Behavior

RPC method calls shouldn't crash the node

## Current Behavior

When running in light client mode (sync mode "light" or "lightest"), a panic can be triggered through `eth_getBlockByHash` when the hash specified doesn't correspond to an actual block.  You can also trigger it via `eth_getBlock` this way (since `eth_getBlock` uses `eth_getBlockByHash` if given a hash).

## Steps to Reproduce Behavior

1. Run a node in light client mode (e.g. with `--syncmode lightest`)
2. Wait until it's connected to some peers and started syncing (the panic doesn't happen if you have no peers at the time of the RPC method call)
3. Call `eth_getBlockByHash` by RPC or `eth.getBlockByHash` from the `geth attach` console, with a hash for a block that doesn't exist, e.g. `0x8888888888888888888888888888888888888888888888888888888888888888`:
```
> eth.getBlockByHash("0x8888888888888888888888888888888888888888888888888888888888888888", false)
```
4. The node will panic

The panic is not recovered from, even though this is a method handler, because it happens in a different goroutine since handling the RPC request involves firing off an ODR request and the panic is from the goroutine responsible for handling the response rather than the goroutine that handles the RPC request.

## Logs

```
panic: runtime error: index out of range [-1]

goroutine 248 [running]:
github.com/celo-org/celo-blockchain/core.(*HeaderChain).ValidateHeaderChain(0xc0001c4510, 0x5e761e8, 0x0, 0x0, 0x1, 0xc009fb1600, 0x0, 0x0, 0x0)
	github.com/celo-org/celo-blockchain/core/headerchain.go:282 +0xc5c
github.com/celo-org/celo-blockchain/light.(*LightChain).InsertHeaderChain(0xc000494400, 0x5e761e8, 0x0, 0x0, 0x1, 0x2a848caa00, 0x0, 0x0, 0x0)
	github.com/celo-org/celo-blockchain/light/lightchain.go:402 +0x105
github.com/celo-org/celo-blockchain/les.(*clientHandler).handleMsg(0xc008d2d2c0, 0xc009f8e8c0, 0x0, 0x0)
	github.com/celo-org/celo-blockchain/les/client_handler.go:324 +0xba5
[...]

PR #1628, opened 2021-07-16
Add support for "system calls receipt" in eth_getTransactionReceipt

oneeman:
### Description

If systems calls made by celo-blockchain result in logs, we add a receipt including those logs, which lists the block's hash as the transaction hash field on the receipt.  This PR adds support to eth_getTransactionReceipt to return this receipt when given a block hash.

One key difference, however, is that no receipt is added in case there were no logs, whereas eth_getTransactionReceipt will return an empty receipt in such cases, so that the case of no logs can't be confused with the case of not having the block.  We may in the future wish to change it so that the receipt is also created during block processing, but that would be a hard fork change.

### Tested

Manual testing using `geth attach`:

Light client, mainnet, a block without a system receipt:
```
> eth.getTransactionReceipt(eth.getBlock(10).hash)
{
  blockHash: "0x5ffe9b7813f766e33b1e0c4991cd9e25527eb3fe76417b1b4cbdfb77a797e8c8",
  blockNumber: 10,
  contractAddress: null,
  cumulativeGasUsed: 0,
  from: "0x0000000000000000000000000000000000000000",
  gasUsed: 0,
  logs: [],
  logsBloom: "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  status: "0x1",
  to: "0x0000000000000000000000000000000000000000",
  transactionHash: "0x5ffe9b7813f766e33b1e0c4991cd9e25527eb3fe76417b1b4cbdfb77a797e8c8",
  transactionIndex: 0
}
```

Light client, mainnet, a block with a system receipt and no transactions:
```
> eth.getTransactionReceipt(eth.getBlock(10000).hash)
{
  blockHash: "0x334111831abd05a6f4d4e157596ffbb95ff9a8bc1da7173c0c7616c738395185",
  blockNumber: 10000,
  contractAddress: null,
  cumulativeGasUsed: 0,
  from: "0x0000000000000000000000000000000000000000",
  gasUsed: 0,
  logs: [{
      address: "0xdfca3a8d7699d8bafe656823ad60c17cb8270ecc",
      blockHash: "0x334111831abd05a6f4d4e157596ffbb95ff9a8bc1da7173c0c7616c738395185",
      blockNumber: 10000,
      data: "0x0000000000000000000000000000000000000000000000000000000005f5e100",
      logIndex: 0,
      removed: false,
      topics: ["0x6e53b2f8b69496c2a175588ad1326dbabe2f66df4d82f817aeca52e3474807fb"],
      transactionHash: "0x334111831abd05a6f4d4e157596ffbb95ff9a8bc1da7173c0c7616c738395185",
      transactionIndex: 0
  }],
  logsBloom: "0x00000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000200000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000000000",
  status: "0x1",
  to: "0x0000000000000000000000000000000000000000",
  transactionHash: "0x334111831abd05a6f4d4e157596ffbb95ff9a8bc1da7173c0c7616c738395185",
  transactionIndex: 0
}
```

Light client, mainnet, a block with a system receipt and regular transactions:
```
> eth.getTransactionReceipt(eth.getBlock(7773884).hash)
{
  blockHash: "0x987144a7215cc5054eab65e5a70c1414d1ec3bc0373ca64f0e41ba8bdd5fc276",
  blockNumber: 7773884,
  contractAddress: null,
  cumulativeGasUsed: 0,
  from: "0x0000000000000000000000000000000000000000",
  gasUsed: 0,
  logs: [{
      address: "0xdfca3a8d7699d8bafe656823ad60c17cb8270ecc",
      blockHash: "0x987144a7215cc5054eab65e5a70c1414d1ec3bc0373ca64f0e41ba8bdd5fc276",
      blockNumber: 7773884,
      data: "0x0000000000000000000000000000000000000000000000000000000005f5e100",
      logIndex: 3,
      removed: false,
      topics: ["0x6e53b2f8b69496c2a175588ad1326dbabe2f66df4d82f817aeca52e3474807fb"],
      transactionHash: "0x987144a7215cc5054eab65e5a70c1414d1ec3bc0373ca64f0e41ba8bdd5fc276",
      transactionIndex: 5
  }],
  logsBloom: "0x00000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000200000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000000000",
  status: "0x1",
  to: "0x0000000000000000000000000000000000000000",
  transactionHash: "0x987144a7215cc5054eab65e5a70c1414d1ec3bc0373ca64f0e41ba8bdd5fc276",
  transactionIndex: 5
}
```

### Backwards compatibility

This makes a breaking change to `eth_getTransactionReceipt`, but the change only affects cases where the hash supplied was the hash of a block rather than of a transaction.


mcortesi:
Why do we have examples of contract creation and not contract creation? I guess those are regular transactions so should not be affected by this PR?

Also, not sure if ease, but when creating an empty receipt it would good to use the txIndex that corresponds to "after the last tx". In the example was zero, but was not clear if the block was empty

oneeman:
> Why do we have examples of contract creation and not contract creation? I guess those are regular transactions so should not be affected by this PR?

Correct, they're not affected, just including them as part of the testing (to confirm they're not affected).  An earlier version I tried had a bug that broke that functionality.

> Also, not sure if ease, but when creating an empty receipt it would good to use the txIndex that corresponds to "after the last tx". In the example was zero, but was not clear if the block was empty

It's ok as is, because the method's code sets that field (and some other "derived fields") in the return value:
```
		"transactionIndex":  hexutil.Uint64(index),
```
So even though it isn't set on the `receipt` variable's object, it will be set correctly in the response.  I will see whether I can replace that example with one for a block that does have some transactions so that the example can reflect that as well.

Also, I will check the e2e transfer test which seems to be failing.

oneeman:
The failing end-to-end-transfer-test is due to #1627 (the tests include sending transactions through light clients, and it seems like web3 polls using `eth_getTransactionReceipt`).  This means that we shouldn't merge this PR until we fix #1627, because doing so would essentially break the ability to send transactions through light clients and wait for their receipts using contractkit or web3.

oneeman:
Updated the example of a block without a system receipt to use block 2937 (which contains one transaction) instead of block 10 (which doesn't contain any transactions).

oneeman:
Updated the PR to add a new method `eth_getBlockReceipt` instead of overloading `eth_getTransactionReceipt`.  Updated the PR description accordingly.  The new method currently takes in a block hash.  I initially had it take either a block number or block hash, but when you do that it also accepts "earlier", "latest", and "pending".  And, while they were working, "pending" doesn't really make sense for this method, so I changed it to take only the hash.  But this is up for discussion and can be changed back or to something else.

PR #1629, opened 2021-07-16
fix: Registry code size check

YoduYodu:
### Description

Currently the Registry Proxy check in genesis is set to be a static number 2585, which may not make sense since there may be change in Proxy code & different compiler. Hence we want to reset the check so that any non-zero code size of RegistrySmartContractAddress will pass.


oneeman (approved):
LGTM, thanks!

PR #1631, opened 2021-07-20
Fix panic on lightest mode client when requesting a block with unknown hash

### LINKED ISSUE #1627: Light client panics on `eth_getBlockByHash` with an incorrect block hash

oneeman:
### Description

Previously, if you use `eth_getBlock` on a lightest mode client, and specify a hash that doesn't correspond to a block, the entire client would crash due to a panic in client_handler.go.  This PR fixes that panic, and makes getting a block by an unknown hash return "unknown block", which is the desired behavior and also what you get when you request a block by an unknown number (a number that is higher than the last block's number).

This requires modifying the `HeaderRequest` ODR request type to properly work when receiving a response with no headers.  This PR modifies it to accept such responses if the request was by hash, but not if it was by number.  This opens the door for a malicious LES server to fool a light client into thinking there is no block with this hash, when in fact there is.  However, it's simpler and seems preferable to the alternative of querying all one's peers in case some of them are malicious this way, and only returning "unknown block" after querying all of them.  **It's also preferable to the current status quo of malicious nodes being able to crash a lightest mode client by returning an empty response.**  Note, too, that requesting the header by hash is only done through the RPC API and is not a common use-pattern by light client users.

Another possibility would be, if you get an empty result, to then ask another node or another two nodes, to reduce the probability of being fooled.  But this PR takes the more minimal approach for now, and we can open up a separate issue about making such a change.

### Tested

Manual testing using `geth attach`

### Related issues

- Fixes #1627

### Backwards compatibility

See above in the description for how the behavior changes.


les/odr_requests.go
```diff
@@ -164,7 +169,15 @@ func (r *HeaderRequest) Validate(db ethdb.Database, msg *Msg) error {
...
 	}
 	headers := msg.Obj.([]*types.Header)
-	if len(headers) != 1 {
+	if len(headers) == 0 && r.isByHash() {
+		// For requests by number, we only send to peers for which we know the block number
+		// is within the range of what they have, so if they don't send us the header we reject
+		// the response and try other peers.
+		// However, for requests by hash, we have no way of knowing ahead of time whether the peer
+		// should have it or not (e.g. what if there is no such block?).  So we need to accept
+		// 'no match' as a valid response, to avoid ODR endlessly trying to send to different peers.
```
piersy, line 178: What does ODR stand for?

les/odr_requests.go
```diff
@@ -164,7 +169,15 @@ func (r *HeaderRequest) Validate(db ethdb.Database, msg *Msg) error {
 		return errInvalidMessageType
 	}
 	headers := msg.Obj.([]*types.Header)
-	if len(headers) != 1 {
+	if len(headers) == 0 && r.isByHash() {
+		// For requests by number, we only send to peers for which we know the block number
```
piersy, line 173: How do we know what range of blocks a peer has?

light/lightchain.go
```diff
@@ -298,7 +298,9 @@ func (lc *LightChain) GetBlockByHash(ctx context.Context, hash common.Hash) (*ty
 	number := lc.hc.GetBlockNumber(hash)
 	if number == nil {
 		header, err := GetHeaderByHash(ctx, lc.odr, hash)
-		if err != nil {
+		// Header may be nil, indicating the hash doesn't match any known blocks,
+		// which is a valid response to the ODR request.
+		if err != nil || header == nil {
 			return nil, errors.New("unknown block")
```
piersy, line 304: I think it would be helpful to include the hash in this error string.

les/odr_requests.go
oneeman: On-demand retrieval: https://github.com/celo-org/celo-blockchain/blob/1188111ac106a92c6c9d89c5ff3da265b404a161/light/lightchain.go#L17

Though I've also heard "on-demand requests" as the more guessable meaning.

les/odr_requests.go
oneeman: It's based on information we maintain about the peer: https://github.com/celo-org/celo-blockchain/blob/1188111ac106a92c6c9d89c5ff3da265b404a161/les/peer.go#L600-L622 .

light/lightchain.go
oneeman: I'd hesitate about lumping this in here.  If you search for "unknown block" in the codebase we return it from several places, including as https://github.com/celo-org/celo-blockchain/blob/1188111ac106a92c6c9d89c5ff3da265b404a161/consensus/istanbul/backend/engine.go#L63 .  In the case of users making API requests, they should know the hash they requested.  And as far as having it in the logs, it may be part of a bigger conversation worth having about having more information logged from API requests vs having the logs be done by the client or nginx gateway.  Regardless I wouldn't favor adding it to just this one.

mcortesi (approved):
LGTM

Issue #1641, opened 2021-07-28
Add extra code coverage job to CI to track e2e test coverage across Istanbul package 


piersy:
Our current coverage job tracks coverage for tests only in the package in which tests reside. This is great for understanding the coverage generated by our unit tests but doesn't let us know what is all the code that is actually being run.

The goal of this ticket is to add a job to run the tests in the e2e package and track the coverage across the istanbul package and all subpackages. So that we can understand what parts of the istanbul code are being executed in a more realistic environment.

PR #1662, opened 2021-08-10
Add circleci istanbul coverage

hbandura:
#1641 

Adds a job in circleci to run the e2etests and track the coverage from our istanbul package


.circleci/config.yml
```diff
@@ -119,6 +119,18 @@ jobs:
       - run: go run build/ci.go test -coverage
       - run: bash <(curl -s https://codecov.io/bash)
 
+  coverage_istanbul:
```
YoduYodu, line 122: Can we mention `e2e` here? e.g. `coverage_istanbul_for_e2e`

.circleci/config.yml
```diff
@@ -325,6 +337,11 @@ workflows:
           requires:
             - build-geth
             - prepare-system-contracts
+      - coverage_istanbul:
```
YoduYodu, line 340: ditto

YoduYodu (approved):
LGTM.
I already smelled a 80% covered Istanbul : )

.circleci/config.yml
hbandura: I think I'll remove it from circleci since I don't think we can have both reports in the PR, so I'll just create a `make` target for it

hbandura: right

build/ci.go
```diff
@@ -355,6 +356,9 @@ func doTest(cmdline []string) {
 	gotest.Args = append(gotest.Args, "-p", "1")
 	if *coverage {
 		gotest.Args = append(gotest.Args, "-covermode=atomic", "-cover", "-coverprofile=coverage.out")
+		if *coverpkg != "" {
```
piersy, line 359: Rather than update ci.go, I think we can just use `go test ` directly, this should help reduce conflicts in ci.go.

Makefile
```diff
@@ -269,3 +269,7 @@ geth-windows-amd64:
 	$(GORUN) build/ci.go xgo -- --go=$(GO) --targets=windows/amd64 -v ./cmd/geth
 	@echo "Windows amd64 cross compilation done:"
 	@ls -ld $(GOBIN)/geth-windows-* | grep amd64
+
+
+e2e_istanbul_coverage:
```
piersy, line 274: I don't see this being run in CI, I think you need to add it to the list of jobs

piersy:
Closing this since we merged #1692

PR #1668, opened 2021-08-11
E hard fork setup.

YoduYodu:
### Description

- Add E fork info without setting activation block numbers, name undetermined.
- Switch Churrito/Donut override flags to E.

### Other changes

None

### Tested

N/A

### Related issues

N/A

### Backwards compatibility

Yes. For now the `EBlock` is set to `nil`, i.e. a no-op
Later, enabling `EBlock` with a certain block number will trigger a hard fork.


trianglesphere:
> No, this triggers a hard fork
This actually shouldn't trigger the HF, but instead enable it to later be triggered.

trianglesphere (approved):
lgtm

PR #1670, opened 2021-08-13
Fix calculation and implementation of dial ratio and the max dialed connections limit

### LINKED ISSUE #1669: Incorrect split between dialed and incoming peers when running a light server
## Expected Behavior

Connection slots for eth/istanbul peers should be divided between dialed and incoming connections according to the `DialRatio` value (which defaults to 3, meaning 1/3 of slots for dialed connections and 2/3 for incoming).

The p2p server does not explicitly divide up slots between eth peers and LES clients.  However, since LES servers don't dial LES clients (only the clients dial the servers), connection slots for LES clients should all be incoming.  Therefore the correct logic is:

```
max_dialed = max_eth_peers / DialRatio
max_incoming = max_peers - max_dialed
```

where `max_peers = max_eth_peers + max_les_clients`

## Current Behavior

If running an LES server, the calculation which splits up slots between dialed and incoming uses the max total number of peers (both eth/istanbul peers and LES clients) rather than the max eth peers number: https://github.com/celo-org/celo-blockchain/blob/f82aa14059c9b1f210077123df2700b4c912f6f4/p2p/server.go#L735-L739

This behavior is inherited from upstream, so this is also a bug in go-ethereum.

For example, with the values recommended in [our docs](https://docs.celo.org/getting-started/mainnet/running-a-full-node-in-mainnet#start-the-node) (`--light.maxpeers 1000 --maxpeers 1100`), the resulting split is 366 dialed and 734 incoming.  However, the eth protocol manager has a limit of 100 peers, so the correct split would be 33 dialed and 1067 incoming.  The resulting behavior is the following:

1. Until it gets to its limit of eth peers (100 with these flag values), it keeps trying to dial peers (since it never gets to 366 dialed peers) and also accepts incoming connections.  The resulting number of dialed vs incoming connections is determined on a "first come, first serve" basis depending on how many of the dials succeed and how many peers try to connect to it.
2. Once it's reached its limit of eth peers, the p2p server it will still keep trying to dial peers and accepting incoming connections.  However, these connections (unless they are static or trusted, of course) are immediately closed by the eth protocol manager because it's already at its limit of eth peers: https://github.com/celo-org/celo-blockchain/blob/f82aa14059c9b1f210077123df2700b4c912f6f4/eth/handler.go#L380-L382

This behavior is undesirable for two reasons:

1. The split between dialed and incoming eth peers is left up to chance rather than being determined by the dial ratio (1/3 to 2/3).  This is likely the cause for the observed behavior of new nodes having a hard time finding peers that will let them connect.  If we instead had the correct split of 1/3 to 2/3, it would help ensure that peers have open incoming slots.  Of course, some peers may be behind a firewall and not forwarding the port and so unable to have any incoming connections, so we can't know for sure how many open slots would result.  But the current behavior prevents nodes running LES servers from respecting the dial ratio and so is likely a big contributing factor to the problem.
2. The node should not be trying to dial new eth peers when it already has as many eth peers as is allowed, because it just dials them and then closes the connection immediately.

* Run an Alfajores node with the following command.  Alfajores is best to illustrate this problem, since it has less nodes and so it's easy for the node to connect to peers quickly.  The lower values for maxpeers and light.maxpeers make the problem easy to see as well, since then it shows up without needing to connect to so many nodes.
```
geth --datadir [your datadir] --light.serve 100  --alfajores --maxpeers 71 --light.maxpeers 50
```
* To see the incorrect behavior you can either add some logging to show the values of max dialed and max incoming, or let the node connect to peers and look at how many eth/istanbul peers are dialed vs incoming.  I tried this right now and I have 21 dialed connections and 0 incoming connections.  Since the eth peers limit is 21, the correct behavior would be to only allow 7 dialed connections, and reserve the rest of the slots for incoming connections.

## Notes for fixing

* Fixing this will require the p2p server config to include enough information to know (a) the max number of eth peers, and (b) the max number of total peers.  Just the total can't work because the max dialed connections needs to be based on the number of eth peers allowed.  Having just the max eth peers can't work either, because the overall peers limit (dialed + incoming) needs to allow for both eth peers and LES clients).  One way to do this would be to also include the light maxpeers value in the config (since the max eth peers can then be calculated from that and the max total).
* As mentioned above, this is also a bug upstream.  However, running a LES server is less common on the Ethereum network (it's only in go-ethereum and most nodes don't run LES servers), and they don't recommend values as large as 1000 for the max light peers number, so for this reason it hasn't caused them visible issues like it has for us.  However, it is definitely a bug and once we fix it we could open up a PR against upstream as well, or file a github issue.   I had previously asked about it in discord (https://discord.com/channels/482467812179181568/482478927374188555/786674697478733845) but didn't hear back
* There is some tangentially related behavior inherited from upstream which is confusing: the value of max light peers is added to the max peers value, unless both have been specified using the flags.  That is to say: (a) if one or both of them come from a default value, the max peers value is assumed to mean max eth peers and the value of max light peers is added to it to get the total max peers, but (b) if both have been specified then it's assumed that the max peers value instead specifies the total, and the light max peers value isn't added to it.  This is very unintuitive and likely something we should change.  However, as it is a breaking change for the flags and would have implications for the docs and node operators, I wouldn't suggest changing it as part of a PR fixing this bug.  The relevant code is: https://github.com/celo-org/celo-blockchain/blob/f82aa14059c9b1f210077123df2700b4c912f6f4/cmd/utils/flags.go#L1183-L1209

## System information

```
Celo
Version: 1.4.0-unstable
Git Commit: f82aa14059c9b1f210077123df2700b4c912f6f4
Git Commit Date: 20210809
Architecture: amd64
Protocol Versions: [66]
Go Version: go1.16.4
Operating System: darwin
GOPATH=/Users/or/go
GOROOT=go
```

oneeman:
### Description

Fixes two behaviors in `p2p/Server` which cause it to get more dialed connections than it should based on the dial ratio (which in turn impairs the ability of new nodes joining the network to find peers with open incoming connection slots):

1. The bug described in #1669, which affects LES servers.
2. The fact that the dialer attempts two dials for every dialed peer connection it's missing: https://github.com/celo-org/celo-blockchain/blob/147571e5b5aab36b6de4d839a3ee9c1f85dbd94b/p2p/dial.go#L381

Initially I was planning to just fix #1669.  But then when I tested it out and saw that it still ends up with more dialed peers than it should, I remembered this `* 2` and realized that it would have to be removed in order to get the desired behavior.

This `* 2` was introduced upstream in https://github.com/ethereum/go-ethereum/pull/20592.  Prior to that, it would attempt one dial for every additional dialed peer that it wanted: https://github.com/ethereum/go-ethereum/blob/58cf5686eab9019cc01e202e846a6bbc70a3301d/p2p/dial.go#L149-L160

The new behavior was introduced in that PR, but was not mentioned in the PR description, commented on in the code, nor commented on during PR review.  But, seeing as it impairs the functioning of the dial ratio and geth was doing fine without it up to 1.9.10, I don't see a problem with removing it.  It means that you will be slower to dial peers initially (though there is at any rate the `maxActiveDials` limit which defaults to 50) and if a dialed peer connection drops you will try to refill it one at a time instead of two at a time (if that makes sense), but that seems reasonable.

### Other changes

Updated a test to work with the new behavior (the test assumed two dial slots for every dialed peer we want).

### Tested

Verified that the dial ratio is respected by connecting a node with low limits to Alfajores and seeing that it only connects to  the right number of dials peers, whether or not it's running the LES server.

### Related issues

- Fixes #1669

### DIFF
--- a/p2p/dial_test.go
+++ b/p2p/dial_test.go
@@ -39,12 +39,12 @@ func TestDialSchedDynDial(t *testing.T) {
 	t.Parallel()
 
 	config := dialConfig{
-		maxActiveDials: 5,
+		maxActiveDials: 2,
 		maxDialPeers:   4,
 	}
 	runDialTest(t, config, []dialTestRound{
-		// 3 out of 4 peers are connected, leaving 2 dial slots.
-		// 9 nodes are discovered, but only 2 are dialed.
+		// 3 out of 4 peers are connected, leaving 1 dial slots.
+		// 9 nodes are discovered, but only 1 is dialed.
 		{
 			update: func(d *dialScheduler) {
 				d.addStatic(newNode(uintID(0x01), "127.0.0.1:30303"))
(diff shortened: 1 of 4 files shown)


p2p/dial_test.go
```diff
@@ -58,61 +58,55 @@ func TestDialSchedDynDial(t *testing.T) {
 				newNode(uintID(0x01), "127.0.0.1:30303"), // not dialed because already connected as static peer
 				newNode(uintID(0x03), "127.0.0.1:30303"), // ...
 				newNode(uintID(0x04), "127.0.0.1:30303"),
-				newNode(uintID(0x05), "127.0.0.1:30303"),
-				newNode(uintID(0x06), "127.0.0.1:30303"), // not dialed because there are only two slots
+				newNode(uintID(0x05), "127.0.0.1:30303"), // not dialed because there are only two slots
+				newNode(uintID(0x06), "127.0.0.1:30303"), // ..
```
trianglesphere, line 62:
```suggestion
				newNode(uintID(0x06), "127.0.0.1:30303"), // ...
```

p2p/dial_test.go
```diff
@@ -58,61 +58,55 @@ func TestDialSchedDynDial(t *testing.T) {
 				newNode(uintID(0x01), "127.0.0.1:30303"), // not dialed because already connected as static peer
 				newNode(uintID(0x03), "127.0.0.1:30303"), // ...
 				newNode(uintID(0x04), "127.0.0.1:30303"),
-				newNode(uintID(0x05), "127.0.0.1:30303"),
-				newNode(uintID(0x06), "127.0.0.1:30303"), // not dialed because there are only two slots
+				newNode(uintID(0x05), "127.0.0.1:30303"), // not dialed because there are only two slots
```
trianglesphere, line 61: not dialed b/c 1 slot?

trianglesphere (approved):
I've got a couple small comments on the tests, but otherwise it looks good to me.

p2p/dial_test.go
oneeman: Fixed in new commit

p2p/dial_test.go
oneeman: Fixed in new commit

oneeman:
> I've got a couple small comments on the tests, but otherwise it looks good to me.

Thanks!  Updated those comments as well as another one in a new commit, and then I rebased on master.

PR #1675, opened 2021-08-23
[CI] Update lightest-sync-test circleci workflow image

jcortejoso:
### Description

Update lightest-sync-test circleci workflow image to fix the error:
```
build/bin/geth: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by build/bin/geth)
```


jcortejoso:
Replaced by #1674

PR #1681, opened 2021-09-02
Extra e2etest + minor changes in e2etests lib

hbandura:


e2e_test/e2e_test.go
```diff
@@ -36,3 +37,44 @@ func TestSendCelo(t *testing.T) {
 	err = network.AwaitTransactions(ctx, tx)
 	require.NoError(t, err)
 }
+
+func TestSingleNodeManyTxs(t *testing.T) {
+	rounds := 100
+	accounts := test.Accounts(1)
```
piersy, line 43: With only 1 account we only create 1 node and so at that point we are not testing a network and not executing most of the consensus code. So unless we specifically want to avoid executing that code I suggest a minimum of 3 accounts.

e2e_test/e2e_test.go
```diff
@@ -36,3 +37,44 @@ func TestSendCelo(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	for r := 0; r < rounds; r++ {
+		tx, err := network[0].SendCelo(ctx, common.Address{}, 1)
+		require.NoError(t, err)
+		require.NotNil(t, tx)
+		err = network.AwaitTransactions(ctx, tx)
+		require.NoError(t, err)
+	}
```
piersy, line 58: Calling `AwaitTransactions` inside the loop means that we wait for each tx in turn so each block will have 1 tx and so we wait for at least 100 blocks. If we instead just wait at the end then we will likely end up processing more than 1 tx per block and the test should be significantly faster since we are not constantly waiting.

```suggestion
	txs := make([]*types.Transaction, rounds)
	for r := 0; r < rounds; r++ {
		tx, err := network[0].SendCelo(ctx, common.Address{}, 1)
		require.NoError(t, err)
		require.NotNil(t, tx)
		txs[r] = tx
	}

	err = network.AwaitTransactions(ctx, txs...)
	require.NoError(t, err)

```

e2e_test/e2e_test.go
```diff
@@ -36,3 +37,44 @@ func TestSendCelo(t *testing.T) {
 	err = network.AwaitTransactions(ctx, tx)
 	require.NoError(t, err)
 }
+
+func TestSingleNodeManyTxs(t *testing.T) {
+	rounds := 100
```
piersy, line 42: I think rounds has the potential to be confused with rounds in the consensus, I'd go for something like txCount.

e2e_test/e2e_test.go
```diff
@@ -36,3 +37,44 @@ func TestSendCelo(t *testing.T) {
...
+	for r := 0; r < rounds; r++ {
+		tx, err := network[0].SendCelo(ctx, common.Address{}, 1)
+		require.NoError(t, err)
+		require.NotNil(t, tx)
+		err = network.AwaitTransactions(ctx, tx)
+		require.NoError(t, err)
+	}
+}
+
+// func TestSingleNodeSurviveEpoch(t *testing.T) {
```
piersy, line 61: Whats the situation with this test?

e2e_test/e2e_test.go
hbandura: well this is actually 100 rounds of waits, not really 100txs (even if there are also 100txs, but we could send more than one per wait round)

e2e_test/e2e_test.go
hbandura: epoch sometimes hang

e2e_test/e2e_test.go
hbandura: This was actually intentional. If we want more Txs then I can fire up more txs per round

e2e_test/e2e_test.go
hbandura: yes, this is intentional. I think having just one test with a single node will probably make it easier to debug problems in the future (as in, if something fails on the tests with multiple nodes but work in this one, or any other case)

e2e_test/e2e_test.go
```diff
@@ -36,3 +38,54 @@ func TestSendCelo(t *testing.T) {
 	err = network.AwaitTransactions(ctx, tx)
 	require.NoError(t, err)
 }
+
+func TestSingleNodeManyTxs(t *testing.T) {
```
piersy, line 42:
```suggestion
// This test verifies correct behavior in a network of size one, in the case that 
// this fails we know that the problem does not lie with our network code.
func TestSingleNodeNetworkManyTxs(t *testing.T) {
```

e2e_test/e2e_test.go
piersy: What about iterations,  just not not be confused with consensus rounds?

piersy: Also I think 100 is too many. When I'm testing repeated behaviour I generally chceck for 3 iterations, the first just to make sure the thing works, the second to know that it can repeat the behaviour and the third just to make sure. So I would go for 3 or 5 if I wanted to be really really sure, but after that I don't think we get much value from the extra rounds.

e2e_test/e2e_test.go
```diff
@@ -36,3 +38,54 @@ func TestSendCelo(t *testing.T) {
...
+				tx, err := network[i].SendCelo(ctx, network[(i+1)%len(network)].Address, 1)
+				require.NoError(t, err)
+				require.NotNil(t, tx)
+				txs = append(txs, tx)
+			}
+		}
+		err = network.AwaitTransactions(ctx, txs...)
+		require.NoError(t, err)
+	}
+}
```
piersy, line 91: This test fails fairly regularly for me it was just exiting with a critical log, so I added a panic into `os.Exit` to help pinpoint where this was happening and the result was this.

```
CRIT [09-13|14:59:21.583] Failed to store updated uptime           err="leveldb: closed"
----------------------------------------- someone called exit goroutine 3345 [running]:
os.Stack(0x1de02e0, 0xa2f850, 0x1ba0cb8)
	/usr/local/go/src/os/proc.go:62 +0x9f
os.Exit(0x1)
	/usr/local/go/src/os/proc.go:76 +0x26
github.com/celo-org/celo-blockchain/log.Crit(0x13f90e8, 0x1e, 0xc00cde54f8, 0x2, 0x2)
	/home/pierspowlesland/projects/celo-blockchain/log/root.go:60 +0x7e
github.com/celo-org/celo-blockchain/core/rawdb.WriteAccumulatedEpochUptime(0x7ffab807b330, 0xc017250af8, 0x1, 0xc020cd14c0)
	/home/pierspowlesland/projects/celo-blockchain/core/rawdb/accessors_chain.go:537 +0x1b1
github.com/celo-org/celo-blockchain/consensus/istanbul/uptime/store.(*uptimeStoreImpl).WriteAccumulatedEpochUptime(0xc012684080, 0x1, 0xc020cd14c0)
	/home/pierspowlesland/projects/celo-blockchain/consensus/istanbul/uptime/store/store.go:23 +0x71
github.com/celo-org/celo-blockchain/consensus/istanbul/uptime.(*Monitor).ProcessBlock(0xc00cde5ac0, 0xc00d4ece80, 0x2, 0x2)
	/home/pierspowlesland/projects/celo-blockchain/consensus/istanbul/uptime/monitor.go:133 +0x3a3
github.com/celo-org/celo-blockchain/core.(*BlockChain).insertPreprocessedBlock(0xc013515200, 0xc00d4ece80, 0xc01e734740, 0x4, 0x4, 0x0, 0x0, 0x0, 0xc0023fc580, 0xc000b0d401, ...)
	/home/pierspowlesland/projects/celo-blockchain/core/blockchain.go:1492 +0x42f
github.com/celo-org/celo-blockchain/core.(*BlockChain).InsertPreprocessedBlock(0xc013515200, 0xc00d4ece80, 0xc01e734740, 0x4, 0x4, 0x0, 0x0, 0x0, 0xc0023fc580, 0x0, ...)
	/home/pierspowlesland/projects/celo-blockchain/core/blockchain.go:1472 +0x1e5
github.com/celo-org/celo-blockchain/miner.(*worker).start.func2(0xc00d4ece80, 0xc01e734740, 0x4, 0x4, 0x0, 0x0, 0x0, 0xc0023fc580)
	/home/pierspowlesland/projects/celo-blockchain/miner/worker.go:187 +0xb3
created by github.com/celo-org/celo-blockchain/consensus/istanbul/backend.(*Backend).Commit
	/home/pierspowlesland/projects/celo-blockchain/consensus/istanbul/backend/backend.go:539 +0x634
```

We see that the error is "leveldb: closed", my guess is that we have called `Close` on the node but that there is no synchronization between that and the goroutine writing uptime. I'm happy to dig in to this further if you would like.

e2e_test/e2e_test.go
```diff
@@ -36,3 +38,54 @@ func TestSendCelo(t *testing.T) {
...
+			require.NoError(t, err)
+			require.NotNil(t, tx)
+		}
+		err = network.AwaitTransactions(ctx, txs...)
+		require.NoError(t, err)
+	}
+}
+
+func TestManyNodes(t *testing.T) {
+	accounts := test.Accounts(4)
```
piersy, line 67: 4 seems a bit low to be classed as "many", 4 is the minimum number of nodes we could use to explore situations where we have byzantine or failed nodes and the network continues to work.

piersy: So I tried setting the number higher and the test seems to get stuck, I fiddled about with the round change params and was able to get networks of size 9 at least passing sometimes but taking 40 seconds to get there. So I think just renaming this to something like `TestNetworkManyTxs` would be fine.

piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit e06e2b86ce829b64d8a4b436d766129071586fba
<details><summary>coverage:  50.8% of statements across all listed packages</summary><br>
<pre>
coverage:  65.9% of statements in consensus/istanbul
coverage:  39.8% of statements in consensus/istanbul/announce
coverage:  54.6% of statements in consensus/istanbul/backend
coverage:  70.7% of statements in consensus/istanbul/core
</pre>
CommentID: 201574cb49
</details>

e2e_test/e2e_test.go
piersy: Hi @hbandura, I've made these changes here - https://github.com/celo-org/celo-blockchain/pull/1835/files let me know what you think?

hbandura: merged it!

piersy (approved):
Looking good !

PR #1682, opened 2021-09-12
add custom marshal function for epochsnarkdata

### LINKED ISSUE #1574: ethclient.Client can't unmarshal epoch blocks
## Description

`ethclient.Client` can't unmarshal epoch blocks.

## Current Behavior

`RPCMarshalBlock` in `internal/ethapi/api.go` marshals the epoch snark bitmap as `hexutil.Bytes`:
https://github.com/celo-org/celo-blockchain/blob/44b043780564ad33301eda8c153909d53e133f34/internal/ethapi/api.go#L1067-L1071

but `ethclient.Client.getBlock` tries to json unmarshal that field into the `*big.Int` defined in `types.EpochSnarkData`
https://github.com/celo-org/celo-blockchain/blob/44b043780564ad33301eda8c153909d53e133f34/core/types/block.go#L143-L146

## Proposed Solution

I think we need a custom unmarshal method defined on  `types.EpochSnarkData` similar to `types.Header`

This is reproducible on this branch https://github.com/celo-org/celo-blockchain/tree/piersys/epoch-block-marshaling-problem.

run
`make prepare-system-contracts` and then `go test ./e2e_test -count 1 -v`

You will see the following error in the logs, which is caused by this problem.

`trackTransactions failed with error: math/big: cannot unmarshal "\"0x06\"" into a *big.Int`

jpjamipark:
### Description

Currently, when unmarshalling `EpochSnarkData` structs from epoch blocks, the standard UnmarshalJSON function fails to convert a `hexutil.Bytes` value to `big.Int`.

This update adds a custom `UnmarshalJSON` function for `EpochSnarkData` in order to properly convert byte arrays into big.Int objects. 

I was not able to get the `gencodec` library to automatically generate an `UnmarshalJSON` function that properly handle conversions from `hexutil.Bytes` to `big.Int`, so I modified `gencodec` generated code and added it to `block.go` instead.

### Other changes

Also added a `MarshalJSON` function. Doesn't solve the bug, but added in case it was needed.

### Tested

These changes were tested via the e2e test added on the previous commit. (See the related issue for more details.)

`go test ./e2e_test -count 1 -v`

 The marshal errors are no longer thrown with the new unmarshal function.

### Related issues

- Fixes #1574 

### Backwards compatibility

AFAIK, this custom unmarshal function should not create any issues with backwards compatibility.

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -18,6 +18,7 @@
 package types
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/big"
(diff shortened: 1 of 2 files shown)


piersy (approved):
Hi @jpjamipark thank you for this! This PR is not against our master, so I'm going to merge it here and then I'll cherry pick your commit into a new PR against master.

Issue #1657, opened 2021-08-04
Fix deadlock in e2e tests


piersy:
The test in the e2e package deadlocks sometimes, ~1/15 runs for me.

In order to be able to use this test effectively we need to be able to run it many times (100s or 1000s) in order to drive out all sync errors (deadlocks, data races, race conditions) that may exist in the code, in its current state it is not useful for this since it often deadlocks when running only 10 times.

PR #1684, opened 2021-09-14
Fix failure to close in validatorPeerHandler

piersy:
### Description
The validatorPeerHandler could not be closed because threadRunning
was never set to true so calling stopThread would exit early thinking the
validatorPeerHandler was already stopped and not actually stop anything.

### Tested

CI is passing

### Related issues

Discovered while investigating #1657 

### Backwards compatibility

Yes


gastonponti (approved):
PRs this big should be split into multiple ones

LGTM

hbandura (approved):

PR #1688, opened 2021-09-14
Fix epoch block marshaling

### LINKED ISSUE #1574: ethclient.Client can't unmarshal epoch blocks

piersy:
### Description

This is a continuation of the PR from @jpjamipark:
* #1682

`EpochSnarkData` unmarshaling was broken.

`RPCMarshalBlock` was marshaling both the fields of `EpochSnarkData` as `hexutil.Bytes`.

https://github.com/celo-org/celo-blockchain/blob/43ae19064484233315e4dace800d16c19915e377/internal/ethapi/api.go#L1071-L1076

@jpjamipark Had mentioned that:
> I was not able to get the gencodec library to automatically generate an UnmarshalJSON function that properly handle conversions from hexutil.Bytes to big.Int, so I modified gencodec generated code and added it to block.go instead.

I believe that it's actually not possible to use gencodec for this task because the manual conversion performed in `ethapi/api.go` when encoding the `big.Int` Bitmap field, means that we cannot use the field type override in gencodec, because you cannot mark a field of type `big.Int` to be serialized as `hexutil.Bytes`.

[gencodec doc](https://pkg.go.dev/github.com/fjl/gencodec?utm_source=godoc#hdr-Relaxed_Field_Conversions)
> Field types in the override struct must be trivially convertible to the original field type.

Unless we change the marshaling in `internal/ethapi/api.go` I don't think we can use gencodec for this task and since, changing the marshaling in `internal/ethapi/api.go` to use `hexutil.Big` actually changes the output bytes (see below) its probably not a good idea.

So in this case we take the approach of adding a manually specified marshal and unmarhsal json methods on `EpochSnarkData`.

```go
package main

import (
	"fmt"
	"math/big"

	"github.com/celo-org/celo-blockchain/common/hexutil"
)

func main() {
	i := big.NewInt(0)
	for i.Uint64() < 50 {
		i.Add(i, big.NewInt(1))
		s, _ := hexutil.Bytes(i.Bytes()).MarshalText()
		s2, _ := hexutil.Big(*i).MarshalText()
		fmt.Printf("s %s : s2 %s\n", s, s2)
	}
}

// Output
s 0x01 : s2 0x1
s 0x02 : s2 0x2
s 0x03 : s2 0x3
s 0x04 : s2 0x4
s 0x05 : s2 0x5
s 0x06 : s2 0x6
s 0x07 : s2 0x7
s 0x08 : s2 0x8
s 0x09 : s2 0x9
s 0x0a : s2 0xa
s 0x0b : s2 0xb
s 0x0c : s2 0xc
s 0x0d : s2 0xd
s 0x0e : s2 0xe
s 0x0f : s2 0xf
s 0x10 : s2 0x10
s 0x11 : s2 0x11
s 0x12 : s2 0x12
s 0x13 : s2 0x13
s 0x14 : s2 0x14
s 0x15 : s2 0x15
```

### Other changes

Added the ability to wait for blocks in `test.Tracker`, to support the test that waits for an epoch block.

### Tested

Added a new test to check that the marshaling/unmarshaling works on epoch blocks.

### Related issues

- Fixes #1574

### Backwards compatibility

No.

This breaks backwards compatibility, because programs using the geth client to decode epoch blocks will now decode the block instead of returning an error.


piersy:
Just a test

piersy:
Just a test

piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 37ed12d6b104729d88eddc3f1c702c50d75ee732
```
coverage:  44.5% of statements across all listed packages
coverage:  45.0% of statements in consensus/istanbul/backend
coverage:  22.6% of statements in consensus/istanbul/backend/internal/replica
coverage:  63.4% of statements in consensus/istanbul/core
```

core/types/block.go
```diff
@@ -183,6 +184,37 @@ func (r *EpochSnarkData) IsEmpty() bool {
...
+func (r *EpochSnarkData) UnmarshalJSON(input []byte) error {
+	type EpochSnarkData struct {
+		Bitmap    hexutil.Bytes
+		Signature hexutil.Bytes
+	}
+	var dec EpochSnarkData
+	if err := json.Unmarshal(input, &dec); err != nil {
+		return err
+	}
+	if dec.Bitmap != nil {
```
hbandura, line 209: What should `r.UnmarshalJSON()` do if the received `bitmap` and `signature` are nil, but `r` holds some values for it? Should it:

1) keep the values or
2) overwrite the `r` values with the nil interpretation of it

I have a hunch that it should do 2, but I'm honestly asking. 

Even so, this implementation seems that it does a mix. If bitmap is nil, it won't overwrite the value, but if signature is not, it would, while not returning any error.

hbandura:
```
Unmarshaler is the interface implemented by types that can unmarshal a JSON description of themselves. The input can be assumed to be a valid encoding of a JSON value. UnmarshalJSON must copy the JSON data if it wishes to retain the data after returning.

By convention, to approximate the behavior of Unmarshal itself, Unmarshalers implement UnmarshalJSON([]byte("null")) as a no-op.
```

It is not clear what it should do, but I would veer towards something that in most cases would make (marshal(unmarshal(json)) be as idempotent as possible. 

Or maybe I'm just overthinking this and it's not that important

piersy: You are right, its not that clear.

I think there's only one valid case for calling UnmarshalJson on non empty object, that would be want to re-use a struct for unmarshaling to avoid re-allocating memory.

I think we should disallow code from setting some fields of a struct and expecting the rest to be set by unmarshaling, it seems like a fragile and confusing approach.

So to support that behavior, the unmarshal function would need to set all fields when executed. Interestingly though it looks like that implementation would violate the convention you quoted above, but I wrote a test on go playground to check this and it seems that Unmarshal will overwrite a struct with nil when decoding []byte("null") so it looks like its not a no-op.

https://play.golang.org/p/dHiWmj8X7KD

So based on that I think we can drop the nil checks.

piersy: Ok so I actually raised an issue on the golang repo because this was not clear to me https://github.com/golang/go/issues/48646

piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 3dda3404acc9aaa4180d4d1b63dcc5eac9ffd4e0
```
coverage:  44.8% of statements across all listed packages
coverage:  45.1% of statements in consensus/istanbul/backend
coverage:  22.6% of statements in consensus/istanbul/backend/internal/replica
coverage:  64.2% of statements in consensus/istanbul/core
```

hbandura (approved):

PR #1701, opened 2021-09-24
Announce Protocol Spec

hbandura:
First draft for the announce protocol spec.

Please give feedback on anything: writing style, technical ambiguities, format, lack of context, etc.


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 692815cc274c8c3596a0bc303f99bc7769275c98
```
coverage:  39.6% of statements across all listed packages
coverage:  50.1% of statements in consensus/istanbul
coverage:  37.9% of statements in consensus/istanbul/backend
coverage:  22.6% of statements in consensus/istanbul/backend/internal/replica
coverage:  58.3% of statements in consensus/istanbul/core
coverage:  58.1% of statements in consensus/istanbul/uptime
coverage:  46.1% of statements in consensus/istanbul/validator
```

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
+# Celo Announce Protocol
+
+While here referred as a protocol by itself, Announce is technically a subset of the Istanbul protocol, on the [RLPx] transport, that facilitates Validators to privately discover the eNodeURL of other Validators in the network, using the underlying p2p eth network. This is used by Istanbul to have direct connections between Validators, speeding up the consensus phase of block generation.
+
+It is important to note that a validator's `eNodeURL` is never shared among third parties, but directly published to specific peers. Therefore it is allowed for a validator to publish different external facing `eNodeURL` values to different validator peers.
```
piersy, line 5:
```suggestion
It is important to note that a validator's `eNodeURL` is never shared among third parties, but directly published to specific peers. Therefore it is allowed for a validator to publish different external facing `eNodeURL` values to different validator peers.

An `eNodeURL` has the following format:

`enode://<hex node id>@<IP/hostname>:<tcp port>?discport=<udp discovery port>`

Where `<hex node id>` is the node's serialized and hex encoded ecdsa public key. The URL parameter `discport` should only be specified if the udp discovery port differs from the tcp port.

Some example `eNodeURLs` (with partially elided hex encoded public keys):
\```
enode://517318.......607ff3@127.0.0.1:34055
enode://e5e2fdf.......348ce8@127.0.0.1:33503?discport=22042
\```
```

Note: I had to escape the code block inside the suggestion, so you will need to tweak this a bit.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
+# Celo Announce Protocol
+
+While here referred as a protocol by itself, Announce is technically a subset of the Istanbul protocol, on the [RLPx] transport, that facilitates Validators to privately discover the eNodeURL of other Validators in the network, using the underlying p2p eth network. This is used by Istanbul to have direct connections between Validators, speeding up the consensus phase of block generation.
```
piersy, line 3: Even though now it is a subset of the Istanbul protocol, I think it makes sense for it to be separated into it's own protocol. You could consider it a parallel of the discovery protocol (which is separate to the eth protocol), except that rather than simply discovering nodes it discovers validators.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+
+It is important to note that a validator's `eNodeURL` is never shared among third parties, but directly published to specific peers. Therefore it is allowed for a validator to publish different external facing `eNodeURL` values to different validator peers.
+
```
piersy, line 6:
```suggestion

An `eNodeURL` has the following format:

`enode://<hex node id>@<IP/hostname>:<tcp port>?discport=<udp discovery port>`

Where `<hex node id>` is the node's serialized and hex encoded ecdsa public key. The URL parameter `discport` should only be specified if the udp discovery port differs from the tcp port.

Some example `eNodeURLs` (with partially elided hex encoded public keys):
\```
enode://517318.......607ff3@127.0.0.1:34055
enode://e5e2fdf.......348ce8@127.0.0.1:33503?discport=22042
\```
```

Note: I had to escape the code block inside the suggestion, so you will need to tweak this a bit.

consensus/istanbul/backend/announce.md
piersy, line 5: I this needs explanation to be expanded a bit. If I understand correctly a validator will share their enode via the normal discovery mechanism. But the key used in that enode is ephemeral (or at least more ephemeral than the validator key), so although the validator's enode is shared among third parties, those third parties can't tell if the enode belongs to a validator or not.

piersy: I think a section just outlining how the two different keys come into play could be useful

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+Wether a validator opens a connection or not against another validator is not in the scope of this spec, but only how and when should the protocol messages should be sent, replied and regossipped. For example, it IS part of the spec what messages should be sent to a newly connected peer or validator peer.
+
+## Protocol Messages (RLP)
+
+// add types with the RLP nomenclature
+
+### queryEnodeMsg (0x12)
+
+`[[dest_address: B_20, encrypted_enode_url: B], version: P, timestamp: P]`
```
piersy, line 37: It would be good to add a reference explaining the meaning of B_20, P ... etc

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+
+While here referred as a protocol by itself, Announce is technically a subset of the Istanbul protocol, on the [RLPx] transport, that facilitates Validators to privately discover the eNodeURL of other Validators in the network, using the underlying p2p eth network. This is used by Istanbul to have direct connections between Validators, speeding up the consensus phase of block generation.
+
+It is important to note that a validator's `eNodeURL` is never shared among third parties, but directly published to specific peers. Therefore it is allowed for a validator to publish different external facing `eNodeURL` values to different validator peers.
+
+## Basic Operation
+
+When a validator is close to being elected ([NearlyElectedValidator]), it starts periodically sending [queryEnodeMsg] messages through the p2p network. These message are regossipped by nodes, and validators reply with a direct message to the originator with an [eNodeCertificateMsg], holding their `eNodeURL`. The initial [queryEnodeMsg] contained the origin validator's `eNodeURL` encrypted with the destination public key, therefore after the direct reply, both validators are aware of each others' `eNodeURL`.
```
piersy, line 9:
```suggestion
When a validator is close to being elected ([NearlyElectedValidator]), it starts periodically sending [queryEnodeMsg]s through the p2p network. The [queryEnodeMsg] is essentially a mapping between a validator's Celo address and it's `eNodeURL` and is re-gossiped by all nodes, but the `eNodeURL` is hidden from the rest of the network by encrypting it. The `eNodeURL` is encrypted once for each other validator with that validator's public key (question does the query always contain the enode encyrpted for all other validators, or just some, and how is that decided?). Each encryption of the `eNodeURL` is marked with the corresponding validator's Celo address, so that validators receiving the message can tell which entry is intended for them. Upon receipt of this message a validator decrypts the encrypted `eNodeURL` and replies with a direct message to the originator with an [eNodeCertificateMsg], holding their own `eNodeURL`. So after the direct reply, both validators are aware of each others' `eNodeURL`
```

consensus/istanbul/backend/announce.md
piersy: I think this message is missing the validator celo address

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+When a peer is registered, all highest known version certificates should be sent to the registered peer in a [versionCertificatesMsg].
+
+#### Handling [queryEnodeMsg]
+
+Messages received of this type should be only processed once.
+Should be regossipped as is, unless another message from the same validator origin has been regossipped in the past 5 minutes.
+
+#### Handling [enodeCertificateMsg]
+
+[enodeCertificateMsg] should be ignored by non validators.
```
piersy, line 88: I don't think non validators ever receive an enode certificate message, because they are only sent directly from validator to validator.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Message spawning
+
+Every 5 minutes it should regossip all the highest known version certificates for all the [NearlyElectedValidator]
+
+### Validator Spec
+
+#### Replying [queryEnodeMsg]
+
+Should be replied with an [enodeCertificateMsg] if the origin is a validator and this validator is the destination, and if an already existing connection to the origin validator exists.
```
piersy, line 103: I think we could leave out the bit regarding having an already existing connection. Either way the response should be respond with an enode certificate, whether the connection exists already or whether a new connection needs to be established.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Receiving [enodeCertificateMsg]
+
+Should update the highest known version to the one given, even if there's no version certificate for it known by this validator.
+
+#### Query spawning
+
+One minute (60 seconds) after a validator enters the set of [NearlyElectedValidator], it should start sending [queryEnodeMsg] messages to the network, for all other NEV validators that have a higher version `eNodeURL` than the one currently known.
+
+Messages should be spaced by at least 5 minutes (300 seconds). A query for a specific `<validator, version>` tuple has a retry back off period. The `nth` attempt should be spaced from the previous one by:
```
piersy, line 113:
```suggestion
Messages should be spaced at least 5 minutes (300 seconds) apart. A query for a specific `<validator, version>` tuple has a retry back off period. The `nth` attempt should be spaced from the previous one by:
```

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+timeoutMinutes = 1.5 ^ (min(n - 1, 5))
+```
+
+Optionally, the first ten (10) query messages can be spaced by at least 60 seconds and ignoring retry back off periods for unanswered queries. This is known as the `AggressiveQueryEnodeGossip`.
+
+If the validator is no longer in the list of [NearlyElectedValidator] then it should stop sending [queryEnodeMsg] messages.
```
piersy, line 121: I think we can drop this statement, non NEV nodes are not validators so at that point it behaves as a full node.

hbandura: This raises the question of 'what is a validator' for the purpose of this document then. I was using the definition of the istanbul backend code which is if the geth node was started with the `--mine` argument. So it can be a validator, but maybe it doesn't have enough votes to be a NEV (or much less an elected validator)

piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 98e0b34eaa8c6cb5289137ff1b87b29efebd7aa3
```
coverage:  39.6% of statements across all listed packages
coverage:  50.1% of statements in consensus/istanbul
coverage:  37.8% of statements in consensus/istanbul/backend
coverage:  22.6% of statements in consensus/istanbul/backend/internal/replica
coverage:  58.4% of statements in consensus/istanbul/core
coverage:  58.1% of statements in consensus/istanbul/uptime
coverage:  46.1% of statements in consensus/istanbul/validator
```

consensus/istanbul/backend/announce.md
hbandura: As of today, if the connection is not already there, the message is not replied with the enodeCertificateMsg. I assume that the next time the origin validator sends a query, they'll likely be connected, and then the reply will happen. 

What you make me think with this, is that, if the spec does not force to to attempt to connect when receiving a queryEnodeMsg, then this reply may very well never happen, since it's a direct message. So, should the protocol be changed, and enodeCertificateMsg be made into a message that can be safely regossiped (like the queryEnodeIs) using encryption, or should we add to the spec the obligation to 'attempt' to connect when a queryEnodeMsg is received?

consensus/istanbul/backend/announce.md
hbandura: They should never receive one, it should never happen, but there could be border cases where it happen, as in, a validator being restarted into a full node, or whatever. I'm thinking we might as well make that part explicit here.

consensus/istanbul/backend/announce.md
hbandura: Isn't `dest_address` the 20 byte celo address of the destination validator?

Agree on the types, I'll try to add a link to the RLP nomenclature

piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 731fef6f6d390bb3876fe5f5e43d0a0d8d9af6a1
```
coverage:  39.6% of statements across all listed packages
coverage:  50.1% of statements in consensus/istanbul
coverage:  37.8% of statements in consensus/istanbul/backend
coverage:  22.6% of statements in consensus/istanbul/backend/internal/replica
coverage:  58.4% of statements in consensus/istanbul/core
coverage:  58.1% of statements in consensus/istanbul/uptime
coverage:  46.1% of statements in consensus/istanbul/validator
```

consensus/istanbul/backend/announce.md
hbandura: `(question does the query always contain the enode encyrpted for all other validators, or just some, and how is that decided?)` -> that is explained in other parts of the spec. Should I link to that part at this stage? I was trying to be brief in this explanation.

I'm not sold on the sentence "The [queryEnodeMsg] is essentially a mapping between a validator's Celo address and its `eNodeURL`". Why do think it as a mapping?

consensus/istanbul/backend/announce.md
piersy: I think its fine as is, right now announce _is_ part of the istanbul protocol, I was just airing the idea that I think eventually I think it would make more sense for this protocol to be entirely separate.

consensus/istanbul/backend/announce.md
piersy: So `dest_address` is the address of the validator that the enode is encrypted for but the message is lacking the source address (which is contained in the istanbul message), which the receiver will require to be able to map validator address -> enode.

consensus/istanbul/backend/announce.md
piersy: >`(question does the query always contain the enode encyrpted for all other validators, or just some, and how is that decided?)` -> that is explained in other parts of the spec. Should I link to that part at this stage? I was trying to be brief in this explanation.

I've seen it now, thanks but I think we could just say something like "The `eNodeURL` is encrypted once for each other validator that this validator needs to discover and it's encrypted with this validator's public key."

Its a bit of a mouthful, but I think it would help with the understanding, the previous description was a just a bit too light I thought to really understand the protocol. And trying to piece it together by looking at the individual behaviours is quite hard. Its nice to have a cohesive natural language description of what the protocol is doing.

piersy: >I'm not sold on the sentence "The [queryEnodeMsg] is essentially a mapping between a validator's Celo address and its eNodeURL". Why do think it as a mapping?

As a receiver of the queryEnodeMsg the knowledge it provides to me is Validator address X has enode Y. Prior to receiving it the reciever simply knows that there is a validator with address X, the query enode message lets them complete the mapping. Does that make sense?

consensus/istanbul/backend/announce.md
piersy: Ahh right, that seems reasonable but, would be good to just have a section that clarifies the terminology.

consensus/istanbul/backend/announce.md
piersy: I think the connection is explicitly made:

If you look here, we can see that if validating the node "multicasts" their enode certificate back to the sender, and that will do nothing if they are not connected, but then at the end of the function it calls `UpsertVersionAndEnode`.

https://github.com/celo-org/celo-blockchain/blob/731fef6f6d390bb3876fe5f5e43d0a0d8d9af6a1/consensus/istanbul/backend/announce.go#L741-L780

At the end of `UpsertVersionAndEnode` `AddValidatorPeer` is called. `AddValidatorPeer` adds a peer to the p2p server with validator purpose, and this will cause the p2p server to try and connect to that peer and when it does because the peer is marked with validator purpose the enode certificate will be sent as part of the handshake.

https://github.com/celo-org/celo-blockchain/blob/731fef6f6d390bb3876fe5f5e43d0a0d8d9af6a1/consensus/istanbul/backend/internal/enodes/val_enode_db.go#L326

Does that look correct to you?

consensus/istanbul/backend/announce.md
hbandura: Ah, I misunderstood you before. Ok, you mean the origin address. You are correct. I will clarify that.

consensus/istanbul/backend/announce.md
hbandura: Yes, right now when this happens the node attempts to connect to the one that sends the query. But I was talking about the spec, not the implementation. Should the spec explicitly say that a connection should be attempted when receiving the query? If we don't do that, we fall into the problem I've mentioned.

piersy: I don't think the protocol spec should dictate whether a connection is made or not because that then ties it into a specific implementation. What the spec should make clear is that a message is sent, irrespective of any particular state that the node may be in or the mechanism by which that message is sent.

consensus/istanbul/backend/announce.md
mcortesi: i guess you could say that:

* any node has a enodeURL, that enodeURL is public and is shared through discovery.
* a validator node, has also an pk/address pair that identifies itself
* announce is about revealing the mapping `valdiatorAddress => enodeURL` so nodes can easily identify/connect which nodes are validators.

I would speak to lengthy about this. enodeURL are already explained on the devp2p protocol, just a refresher. the concept that a node has an etherbase and enodeURL is confusing for most, but because they haven't read the specs ;)

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+* Nearly Elected Validator
+* eNodeURL
+
+### Full Node
+
+A Full Node is a Celo node fully operating in the Celo p2p network.
+
+### Validator
+
+A Validator is a [FullNode] that has been started with the ability to produce blocks, even if it has not yet been elected. In the geth program,
```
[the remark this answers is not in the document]
hbandura: Yeap. The concept of `Validator` is useless here. `NearlyElectedValidator` is the only one important

consensus/istanbul/backend/announce.md
piersy: @hbandura @mcortesi As this stands I think this section is misleading.

The following phrase to me seems incorrect
>It is important to note that a validator's `eNodeURL` is never shared among third parties

Validator's `eNodeURLs` are shared with everyone.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+
+```
+enode://517318.......607ff3@127.0.0.1:34055
+
+enode://e5e2fdf.......348ce8@127.0.0.1:33503?discport=22042
+```
+
+## Objective
+
+The Announce protocol objective is to allow all [NearlyElectedValidator] to maintain an updated table of `validator -> eNodeURL` for all other [NearlyElectedValidator], while somewhat concealing these values from the rest of the nodes in the network, and also avoiding an unnecessary high message traffic to achieve it. [NearlyElectedValidator] nodes can use the information from this table to open direct connections between themselves, speeding up the consensus phase.
```
piersy, line 52: This seems to be missing a reference to the currently elected validators. Also I think maybe changing `validator -> eNodeURL` to `validatorAddress -> eNodeURL` makes things a little clearer.

piersy, line 52: This seems to be missing a reference to the currently elected validators. Also I think maybe changing `validator -> eNodeURL` to `validatorAddress -> eNodeURL` makes things a little clearer.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,151 @@
...
+### queryEnodeMsg (0x12)
+
+`[[dest_address: B_20, encrypted_enode_url: B], version: P, timestamp: P]`
+
+- `dest_address`: the destination validator address this message is querying to.
+- `encrypted_enode_url`: the origin's `eNodeURL` encrypted with the destination public key.
+- `version`: the current announce version for the origin's `eNodeURL`.
+- `timestamp`: a message generation timestamp, used to bypass hash caches for messages.
+
+Max amount of encrypted node urls is 2 * size(set of [NearlyElectedValidator]).
```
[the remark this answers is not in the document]
piersy: I'm wondering why is the max amount (2 * NEV) instead of just NEV?

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Message spawning
+
+Every 5 minutes it should multicast its version certificate table to all peers.
+
+### Validator Spec
+
+#### Peer handshake
+
+When a [NearlyElectedValidator] connects to another [NearlyElectedValidator], the inbound peer can send an [enodeCertificateMsg] to identify itself as a [NearlyElectedValidator]. This allows for preferential treatment for the p2p connection.
```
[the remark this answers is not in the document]
piersy: I actually prefer inbound as opposed to initiator since it then affords the use of outbound to describe the other side of the connection which is useful. Otherwise I guess you could use initiator and non-initiator but it seems a bit clunky. Also inbound and outbound are the terms currently used in the codebase to describe this property of connections.

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+## Protocol Messages (RLP)
+
+All messages presented here are Istanbul messages, following the same format as the consensus messages (see `istanbul.Message`), which are serialized in the `payload` field for standard p2p messages.
+
+`[code: P, msg: B, sender_address: B_20, signature: B]`
+
+- `code`: the message code (0x12, 0x16, or 0x17).
+- `msg`: rlp encoded data, depending on the message type (`code`), and detailed in the following subsections.
+- `sender_address`: the originator of the message.
+- `signature`: the message signature.
```
piersy, line 89: Currently there doesn't seem to be any reference to the `sender_address` in the rest of the spec, and I think the sender address is critical to specifying the protocol, because it is the sender address that lets nodes construct the mapping between address and eNodeURL.

Further down in the message handling descriptions talk about upserting the eNodeURL, but they don't mention the key used to upsert, which is the sender_address.

consensus/istanbul/backend/announce.md
hbandura: The announce protocol doesn't really care about elected validators, only Nearly Elected Validators

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+
+[NearlyElectedValidator] can also use this table to understand which [eNodeURL] entries are stale, if the [eNodeURL] table entry has a version older than the version table entry.
+
+## Basic Operation
+
+When a validator is close to being elected ([NearlyElectedValidator]), it starts periodically sending [queryEnodeMsg] messages through the p2p network. These message are regossipped by nodes, and validators reply with a direct message to the originator with an [eNodeCertificateMsg], holding their `eNodeURL`. The initial [queryEnodeMsg] contained the origin validator's `eNodeURL` encrypted with the destination public key, therefore after the direct reply, both validators are aware of each others' `eNodeURL`.
+
+### p2p connection management
+
+Wether a validator opens a connection or not against another validator is not in the scope of this spec, but only how and when should the protocol messages should be sent, replied and regossipped. For example, it IS part of the spec what messages should be sent to a newly connected peer or validator peer.
```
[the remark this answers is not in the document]
hbandura: Added a line clarifying that

consensus/istanbul/backend/announce.md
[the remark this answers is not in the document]
hbandura: I found it puzzling too tbh

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+When a peer is registered, the whole version table should be sent to the peer in a [versionCertificatesMsg].
+
+#### Handling [queryEnodeMsg]
+
+Messages received of this type should be only processed once.
+Should be regossipped as is, unless another message from the same validator origin has been regossipped in the past 5 minutes.
+
+#### Handling [enodeCertificateMsg]
+
+[enodeCertificateMsg] should be ignored by [FullNode] instances, since they should never receive one if all other participants are behaving properly. We make
```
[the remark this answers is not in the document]
hbandura: Done

consensus/istanbul/backend/announce.md
piersy: I'm wondering though how do elected validators ensure they are connected to other elected validators?

consensus/istanbul/backend/announce.md
```diff
@@ -0,0 +1,217 @@
...
+this explicit in the spec since the regossipping of this particular message would make the [eNodeURL] of the sender public, which defeats the purpose of the
+protocol.
+
+#### Handling [versionCertificatesMsg]
+
+Messages received of this type should be only processed once.
+When received, it should upsert its version table, and then multicast to its peers a new filtered [versionCertificatesMsg] without the certificates that were not updated in the table, and also filtering out certificates if another one for the same [NearlyElectedValidator] has been gossiped in the past 5 minutes.
```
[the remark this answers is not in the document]
mcortesi: As discussed outside, this is more a future docs, not part of the spec

consensus/istanbul/backend/announce.md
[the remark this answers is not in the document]
mcortesi: i guess, what confuses me is that inbound/output referes to a connection, not to the to the actor. The actor is either the initiator or the receiving part of it.

consensus/istanbul/backend/announce.md
hbandura: sender address is a parameter from the 'parent' type istanbul.Message. That's why it's only being mentioned as to where it's coming from, and it's assumed the istanbul message format is known

mcortesi (approved):

consensus/istanbul/backend/announce.md
piersy: @hernan if the announce protocol only cares about validators that  have a "good chance of becoming an elected validator in the following epoch" how are existing validators supposed to connect to each other?

PR #1714, opened 2021-10-12
Fix flakiness in TestRegistryInGenesis

37ng:
### Description

Observed `cmd/geth TestRegistryInGenesis` test failed intermittently on CI while running fine on my laptop.
Given that the failures ends with a timeout from what I observed, I'm suspecting CI's cpu cannot spinning up `geth` within current 5s timeout.
So I set timeout to 15s, and managed run it 1000 times on CI without failure.

### Other changes

- Deletes redundant `datadir` setup in the test.

### Tested

Run `TestRegistryInGenesis` 1000 times under CircleCI job `test-TestRegistryInGenesis-1000-times`, no single failure.
https://app.circleci.com/pipelines/github/celo-org/celo-blockchain/7107/workflows/342b4446-0a25-4c77-835c-d89dc5c29e84/jobs/70510

### Backwards compatibility

Yes, only modifies CircleCI config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -426,6 +426,32 @@ jobs:
             export PATH=${PATH}:~/repos/golang/go/bin
             ./ci_test_validator_order.sh local ~/repos/geth
 
+  end-to-end-cip35-eth-compatibility-test:
+    executor: e2e
+    resource_class: xlarge
+    steps:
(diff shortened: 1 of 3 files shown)


piersy (approved):
Hi @Tong, this is looking good to me now 👍

PR #1720, opened 2021-10-15
Remove coreMu read lock uses when it's only for coreStarted

YoduYodu:
### Description

Remove coreMu read lock uses when it's only for coreStarted
A use case fixed in #1719 caused deadlock, removing others to avoid potential bugs.

### Tested

CI

### Backwards compatibility

Yes

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -645,8 +645,6 @@ func (sb *Backend) SetCallBacks(hasBadBlock func(common.Hash) bool,
 	processBlock func(*types.Block, *state.StateDB) (types.Receipts, []*types.Log, uint64, error),
 	validateState func(*types.Block, *state.StateDB, types.Receipts, uint64) error,
 	onNewConsensusBlock func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB)) error {
-	sb.coreMu.RLock()
-	defer sb.coreMu.RUnlock()
 	if sb.isCoreStarted() {
 		return istanbul.ErrStartedEngine
 	}
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -201,8 +201,6 @@ func (sb *Backend) SetP2PServer(p2pserver consensus.P2PServer) {
 func (sb *Backend) NewWork() error {
 	sb.logger.Debug("NewWork called, acquiring core lock", "func", "NewWork")
 
-	sb.coreMu.RLock()
-	defer sb.coreMu.RUnlock()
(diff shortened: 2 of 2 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 548738da360745c9ade46fa811ca64f910bf1cb9
<details><summary>coverage:  44.4% of statements across all listed packages</summary><br>
<pre>
coverage:  44.9% of statements in consensus/istanbul/backend
coverage:  63.2% of statements in consensus/istanbul/core
</pre>
CommentID: ca267d1a86
</details>

trianglesphere:
> other core engine related fields

that's the problem. An atomic means that accessing the variable is can be done across threads without issue, but it does not stop threads from interleaving modifications to the core engine fields.

YoduYodu:
@trianglesphere 
Doesn't that contradict with #1719 ? And if `coreMu` is supposed to cover `coreStarted` and other fields, why is an atomic `coreStarted` needed?
But I agree we don't need make changes if no known issue.

consensus/istanbul/backend/engine.go
```diff
@@ -645,8 +645,6 @@ func (sb *Backend) SetCallBacks(hasBadBlock func(common.Hash) bool,
 	processBlock func(*types.Block, *state.StateDB) (types.Receipts, []*types.Log, uint64, error),
 	validateState func(*types.Block, *state.StateDB, types.Receipts, uint64) error,
 	onNewConsensusBlock func(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB)) error {
-	sb.coreMu.RLock()
-	defer sb.coreMu.RUnlock()
```
piersy, line 649: I think this usage of `coreMu` needs to remain, as it ensures that there is no data race between the values set in `SetCallbacks` and when they are read in `StartValidating`.

Currently `SetCallbacks`  & `StartValidating` are called by the same thread so it would be safe, but there is nothing that enforces that.

I think a better solution would be to pass the callbacks to StartValidating, that avoids any possibility of a data race over those fields.

consensus/istanbul/backend/handler.go
```diff
@@ -201,8 +201,6 @@ func (sb *Backend) SetP2PServer(p2pserver consensus.P2PServer) {
 func (sb *Backend) NewWork() error {
 	sb.logger.Debug("NewWork called, acquiring core lock", "func", "NewWork")
 
-	sb.coreMu.RLock()
-	defer sb.coreMu.RUnlock()
```
piersy, line 205: This removal looks safe to me, since it just locks over access of `coreStarted` and a logging statement.

PR #1726, opened 2021-10-26
Istanbul spec

piersy:
### Description

Specification for the current implementation of the istanbul protocol.

Currently not 100% complete, the resend round change timeout and the future proposal timeout have not been added to the spec yet. But its enough to get an idea of the protocol.

There are currently 3 versions of the document, with different notation in each, please look at them all and let me know your preference.


consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+Only messages from participants are considered.
+
+For messages containing a value only messages with valid values are considered.
+
+Participants are able to broadcast messages to all other participants. In the
+case that a participant is off-line or somehow inaccessible they will not
+receive broadcast messages and there is no mechanism for these messages to be
+re-sent.
+
+We refer to a consensus instance to mean consensus for a specific height,
```
hbandura, line 37: I don't quite understand this sentence. What would shared state be in this context?

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+`Hc - current height`\
+`Rc - current round`\
+`Rd - desired round`\
+`Vc - currently proposed value`\
+`PCc - current prepared certificate`\
+
+### Algorithm
+See supporting [functions](#Functions) and [notation](#Appendix-1-Notation).
+```
+upon: FC_E
```
hbandura, line 51: I don't quite get this section. Are these pseudocodes for something like an event-driven transition in the node state?

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+E.g. `m<Am, Bm, Cm>`
+
+If all instances of a composite object element share the same value for one
+of its variables then that value can be used in the definition. E.g. `<A, B, Hc>`
+represents a composite element with variables `A` `B` and current height.
+
+If a composite object element has variables for which the value is not
+important then `*` is used in the place of that variable.
+
+### Variable names
```
hbandura, line 255: While correct, I find myself going back and forth too often to this section to understand what these variables are. I feel like we can probably make them a little bit more self explanatory in many cases. Not just variable names. E.g "bc" could probably be completely renamed to 'broadcast' and an explanation line for it wouldn't be needed.

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+  Sc ← AcceptRequest
+  Vc ← nil
+  schedule onRoundChangeTimeout(Hc, 0) after roundChangeTimeout(0)
+
+upon: <R_E, Hc, V> && Sc = AcceptRequest
+  if Rc = 0 && isProposer(Hc, Rd) {
+    bc(<PP_T, Hc, 0, V, nil>)
+  }
+
+upon: <PP_T, Hc, Rd, V, RCC> from proposer(Hc, Rd) && Sc = AcceptRequest
```
hbandura, line 64: Giving names to each of these transitions would probably help to make it easier to understand

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+
+`proposer(H,R)`\
+Returns the proposer for the given height and round.
+
+`isProposer(H, R)`\
+Returns true if the current participant is the proposer for the given height
+and round. 
+
+`deliverValue(V)`\
+Delivers the given value to the application.
```
hbandura, line 113: What does this mean? What is 'deliver' here?

consensus/istanbul/spec.md
```diff
@@ -0,0 +1,329 @@
...
+### High level overview
+
+The Celo IBFT protocol is a BFT (Byzantine Fault Tolerant) protocol that allows
+a group of participants to agree on the ordering of values by exchanging voting
+messages across a network. As long as less than 1/3rd of participants follow
+the protocol correctly then the protocol should ensure that there is only one
+ordering of values that participants agree on and also that and that they can
+continue to agree on new values, i.e. they don't get stuck.
+
+Ensuring that only one ordering is agreed upon is referred to as 'safety' and
```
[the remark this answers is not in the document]
piersy: Reading this again, do you mean something like
>Ensuring that only one ordering is agreed upon is commonly referred to as 'safety' in the field of distributed systems.

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+
+### High level overview
+
+The Celo IBFT protocol is a BFT (Byzantine Fault Tolerant) protocol that allows
+a group of participants to agree on the ordering of values by exchanging voting
+messages across a network. As long as less than 1/3rd of participants deviate
+from the protocol then the protocol should ensure that there is only one
+ordering of values that participants agree on and also that and that they can
+continue to agree on new values, i.e. they don't get stuck.
+
```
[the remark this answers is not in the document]
piersy: From the perspective of the consensus algorithm it doesn't need to know what it's agreeing on, so I thought if we specify block explicitly here then we limit the application of the consensus algorithm to just blocks and I didn't see any reason to do that.

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+### System model
+
+We consider a group of `3f+1` participating computers (participants) that are
+able to communicate across a network. The group should be able to tolerate `f`
+failures without losing liveness or safety.
+
+All participants know the other participants in the system and it is assumed
+that all messages are signed in some way such that participants in the protocol
+know which participant a message came from and that messages cannot be forged.
+
```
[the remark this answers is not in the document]
piersy: Similar to the reasoning for not mentioning block explicitly, from the perspective of consensus algorithm and if you think about it the [implementation](https://github.com/celo-org/celo-blockchain/blob/58804ca7236f60b19ed7fd3ec9d0fc3284621b00/consensus/istanbul/core/handler.go#L163-L182), when it receives a message it does not know, or need to know that the message was sent over the celo p2p network. Changing the network implementation has no effect on the consensus algorithm, and explicitly coupling the consensus to the p2p implementation means that if we were ever to change the network implementation we would need to go and modify the consensus spec to represent that, which is work that is avoidable.

I'm wondering if its not clear what I'm trying to specify here? If I were trying to specify the working of the entire collection of packages under `consensus/istanbul` then it would be pertinent to specify these details, but I'm only trying to specify at a conceptual level the algorithm used for achieving consensus.

consensus/istanbul/celo-ibft-spec.md
piersy: Did you see the section describing upon conditions?
https://github.com/celo-org/celo-blockchain/blob/58804ca7236f60b19ed7fd3ec9d0fc3284621b00/consensus/istanbul/celo-ibft-spec.md?plain=1#L328-L339

If that is not making sense let me know.

consensus/istanbul/celo-ibft-spec.md
piersy: There are 3 versions of this document, this one you have commented on with short names and then 2 others with long names. Can you check the others to see what you think?

consensus/istanbul/celo-ibft-spec.md
piersy: https://github.com/celo-org/celo-blockchain/pull/1726#discussion_r738256951

consensus/istanbul/celo-ibft-spec.md
piersy: I'm wondering if it was obvious that variables mostly start with the first letter of the full name (e.g. Height -> H, Round -> R ...). And then there is the subscript notation for global variables either 'c' for current or 'd' desired.
With those two rules in my head I find it easy to remember all the notation, but of course I wrote it :D

I'm wondering if it would be better to put the notation up front and explain this naming approach. I can see how if you just dive into the algorithm and look up variables as you need, the naming approach may not be so clear.

consensus/istanbul/celo-ibft-spec.md
piersy: What I'm trying to say is that no values are carried over from one instance to the next, for example the messages from the previous height have no bearing on the outcome of the current height.

consensus/istanbul/celo-ibft-spec.md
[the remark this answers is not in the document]
hbandura: Ok, makes sense. Maybe add a paragraph clarifying this at the start of the doc?

consensus/istanbul/celo-ibft-spec.md
hbandura: Ok, so it's a way of saying that this value is now accepted by the consensus? The name was misleading to me

hbandura (commented):
I feel like the comments for each event improved it by a lot

consensus/istanbul/celo-ibft-spec.md
piersy: Not really, by the time `deliverValue` is called the value has already been committed by consensus. In other BFT consensus algorithms I have seen at that point they move to the next height and begin agreeing on the next value. But in this one the consensus algorithm passes the value to the application and waits for a notification from the application. `FinalComittedEvent` before continuing to the next height. I don't know why it does this.

In fact now that I'm thinking about it seems like this could be a source of problems in the protocol. If the final comitted event is delayed, then the network I think could move to the next round and agree on another value. Then you would have a fork.

consensus/istanbul/celo-ibft-spec.md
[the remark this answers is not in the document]
piersy: I've added a description up front for this.

https://github.com/celo-org/celo-blockchain/blob/5afaaa5bb877cb106d9dec1567df41dd8eee7668/consensus/istanbul/celo-ibft-spec.md?plain=1#L5-L15

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,394 @@
...
+a group of participants to agree on the ordering of values by exchanging voting
+messages across a network. As long as less than 1/3rd of participants deviate
+from the protocol then the protocol should ensure that there is only one
+ordering of values that participants agree on and also that and that they can
+continue to agree on new values, i.e. they don't get stuck.
+
+Ensuring that only one ordering is agreed upon is referred to as 'safety' and
+ensuring that the participants can continue to agree on new values is referred
+to as 'liveness'.
+
```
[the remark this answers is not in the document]
piersy: Added - https://github.com/celo-org/celo-blockchain/blob/5afaaa5bb877cb106d9dec1567df41dd8eee7668/consensus/istanbul/celo-ibft-spec.md?plain=1#L21-L23

consensus/istanbul/celo-ibft-spec.md
piersy: Done

trianglesphere:
I think I have a preference to the single variable letter version, but it's nice to have multiple versions that suit different people. The spec itself looks correct to me.

piersy:
> I think I have a preference to the single variable letter version, but it's nice to have multiple versions that suit different people. The spec itself looks correct to me.

Good to know, when I added 3 versions to this PR I was not intending that we would keep all three. Having multiple versions increases the maintenance cost and also introduces the possibility of divergence between the specifications. I don't believe the value provided by the niceness of multiple versions offsets the costs.

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,486 @@
...
+  Hc ← Hc+1
+  Rc ← 0
+  Rd ← 0
+  Sc ← AcceptRequest
+  Vc ← nil
+  schedule onRoundChangeTimeout(Hc, 0) after roundChangeTimeout(0)
+
+// A request event is a request to reach agreement on the provided value, if this parcicipant
+// is the proposer it will propose that value by sending a preprepared message.
+upon: <RequestEvent, Hc, V> && Sc = AcceptRequest
```
hbandura, line 96: This is a bit confusing. Who is sending the 'request event'? Is it the application to consensus layer?

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,486 @@
...
+  }
+}
+```
+
+### Supporting Functions
+
+These functions can read global state but cannot modify it.
+
+#### Application provided functions
+No pseudocode is provided for these functions since their implementation is
```
hbandura, line 175: This is true for many, but for example, what `proposer(H, R)` should return is actually part of the protocol, isn't it?

hbandura (commented):
While re-reading the functions, I have some thoughts:

1) It is sometimes hard to understand what/who the actor triggering the event is (underlying application, a different node, or a timed-event (timeout)).

2) I also feel like I lose some time trying to remember if a variable name refers to stored state from the current round, or if it is an argument provided in the triggered event.  Do you think this could be made a bit clearer? Not sure how, but any kind of prepended name to make it more explicit

consensus/istanbul/celo-ibft-spec.md
piersy: As you guessed it is sent by the application to the consensus layer, will update. Thanks!

consensus/istanbul/celo-ibft-spec.md
piersy: I don't think it is, we actually have 3 different implementations currently, sticky, round robin, and shuffled round robin where the proposer at any point is defined by the implementation of the randomness contract.

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,486 @@
...
+  }
+
+// When a participant sees at least 2f+1 prepare or commit messages for a value
+// it will send a commit message for that value.
+upon: M ← { <T, Hc, Rd, Vc> : T ∈ {Prepare, Commit} } && |M| >= 2f+1 && Sc ∈ {AcceptRequest, Preprepared} 
+  Sc ← Prepared
+  PCc ← <PreparedCertificate, M, Vc>
+  bc(<Commit, Hc, Rd, Vc>)
+
+// When a participant sees at least 2f+1 commit messages for a vlaue, they
```
trianglesphere, line 118:
```suggestion
// When a participant sees at least 2f+1 commit messages for a value, they
```

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,486 @@
...
+
+```
+validRCC(V, RCC) {
+  M ← { m<RoundChange, Hm , Rm, PC> ∈ RCC : Hm = Hc && Rm >= Rd && (PC = nil || validPC(PC)) }
+  N ← { m<RoundChange, * , *, PC> ∈ M : PC != nil }
+  if |N| > 0 {
+    return 2f+1 <= |M| <= 3f+1 &&
+    ∃ m<RoundChange, *, *, Pcm> ∈ N : validPC(PCm) && PCValue(PCm) = V && ∀ n<RoundChange, * , *, PCn> ∈ N != m : PCRound(PCm) >= PCRound(PCn)
+  }
+  return 2f+1 <= |M| <= 3f+1 &&
```
trianglesphere, line 244: Dangling `&&`. I don't think there's a condition there that needs to be checked.

```suggestion
  return 2f+1 <= |M| <= 3f+1
```

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,486 @@
...
+Say all nodes are at round 0 they all time out and send round changes for round 1 so they all move to round one.
+
+Now f+1 nodes timeout and send round changes for round 2 they have desired round 2 current round 1. The remaining 2f
+nodes receive the f+1 round changes and set desired round to 2 and schedule
+roundChangeTimeout with a target round of 2.
+
+Now if those 2f nodes have their scheduled roundChangeTimeout execute before
+they have updated their current round to 2 roundChangeTimeout will be a no-op.
+if the round changes sent by these 2f nodes are lost then they will never
+schedule any further round changes without outside intervention.
```
trianglesphere, line 469: I'm not quite following this sequence of events. I agree that you could stall the network if you got into a condition where a node would not schedule a round change timeout & nodes got stuck on different rounds, but I'm not sure that there is a sequence of events that results in roundChangeTimeouts being dropped.

trianglesphere (approved):
A couple small nits, but otherwise this looks good

consensus/istanbul/celo-ibft-spec.md
piersy: So this was more of a thought experiment/working notes. I think in these cases nodes could get stuck in the case that we didn't have the resendRoundChangeTimeout functionality.


Looking at this again I think there is actually a much more straightforward sequence of events that can lead to 
nodes can getting into a situation where they have killed their local chain of round change timeouts.

```
onRoundChangeTimeout(H, R) {
  if H = Hc && R = Rc {
    Rd ← Rc+1
    Sc ← WaitingForNewRound
    schedule onRoundChangeTimeout(Hc, Rd) after roundChangeTimeout(Rd)
    bc<RoundChange, Hc, Rd, PCc>
  }
}
```

So we can see the `onRoundChangeTimeout` function will only schedule another timeout if the internal condition is met. So to kill the local chain of timeouts we just need to not meet that condition once.

So if we imagine that all nodes time out on round 0, the round change timeout condition will be met because when starting a sequence it is scheduled a such `schedule onRoundChangeTimeout(Hc, 0) after roundChangeTimeout(0)`

The next round change will be scheduled with the desired round == 1 but current round is still 0. If all the subsequently broadcast round change messages are lost, then when the scheduled round change timeout runs the expression will not be met  and all nodes will be stuck and never progress from there on.

piersy:
@hbandura I think I've addressed all your current comments

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,488 @@
...
+  if (Rd > 0 && validRCC(V, RCC)) || (Rd = 0 && RCC = nil)  {
+    Rc ← Rd
+    Vc ← V
+    Sc ← Preprepared
+    bc(<Prepare, Hc, Rd, Vc>)
+  }
+
+// When a participant sees at least 2f+1 prepare or commit messages for a value
+// it will send a commit message for that value.
+upon: M ← { <T, Hc, Rd, Vc> : T ∈ {Prepare, Commit} } && |M| >= 2f+1 && Sc ∈ {AcceptRequest, Preprepared}
```
mcortesi, line 115: shouldn't we be already in `Preprepared`?

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,488 @@
...
+// request event is sent by the application, if this parcicipant is the proposer
+// it will propose that value by sending a preprepared message.
+upon: <RequestEvent, Hc, V> && Sc = AcceptRequest
+  if Rc = 0 && isProposer(Hc, Rd) {
+    bc(<Preprepare, Hc, 0, V, nil>)
+  }
+
+// When receiving a preprepare from a proposer participants will vote for the
+// value (if valid) by sending a prepare message.
+upon: <Preprepare, Hc, Rd, V, RCC> from proposer(Hc, Rd) && Sc = AcceptRequest
```
mcortesi, line 105: I found notation a bit confusing here:
Is this pattern matching: `<Preprepare, Hc, Rd, V, RCC>`, meaning that the height in the msg must be equal to `Hc` and the round in the msg should be equal `Rd`?

It would seem like that's the interpretation, but i'm not so sure

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,488 @@
...
+  if (Rd > 0 && validRCC(V, RCC)) || (Rd = 0 && RCC = nil)  {
+    Rc ← Rd
+    Vc ← V
+    Sc ← Preprepared
+    bc(<Prepare, Hc, Rd, Vc>)
+  }
+
+// When a participant sees at least 2f+1 prepare or commit messages for a value
```
mcortesi, line 113: what's not clear from the spec, is that message set contains a single message per participant. So having the prepare and the commit from the same particpant doesn't count as `|M| = 2`

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,488 @@
...
+upon: M ← { <T, Hc, Rd, Vc> : T ∈ {Prepare, Commit} } && |M| >= 2f+1 && Sc ∈ {AcceptRequest, Preprepared} 
+  Sc ← Prepared
+  PCc ← <PreparedCertificate, M, Vc>
+  bc(<Commit, Hc, Rd, Vc>)
+
+// When a participant sees at least 2f+1 commit messages for a value, they
+// consider that value committed (agreed) and pass the value to the application,
+// which will in turn issue a final committed event if the value is considered
+// valid by the application.
+upon: M ← { <Commit, Hc, Rd, Vc> } && |M| >= 2f+1 && Sc ∈ {AcceptRequest, Preprepared, Prepared}
```
mcortesi, line 124: Is the same message set? for commits and prepares? I think we need a better explanation of what the message set are. 

2. Also, why can Sc be in all these states?
3. Why `Rd` and not `Rc`, assuming they can be different, right? I would expect this to be `Rc` since that's the current round we are trying to accumulate messages for

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,488 @@
...
+// change messages sharing the same round then switch to it. Otherwise if there are at least
+// f+1 round change messages switch to the higest round that is less than or equal to the top
+// f+1 rounds.
+upon: m<RoundChange, Hc , R, PC> && (PC = nil || validPC(PC)) 
+  if R < Rd {
+    send(<RoundChange, Hc, Rd, PCc>, sender(m))
+  } else if quorumRound() > Rd {
+    Rd ← quorumRound()
+    Rc ← quorumRound()
+    Sc ← AcceptRequest
```
mcortesi, line 139: shouldn't we clear the Message sets?

consensus/istanbul/celo-ibft-spec.md
```diff
@@ -0,0 +1,488 @@
...
+    schedule onRoundChangeTimeout(Hc, Rd) after roundChangeTimeout(Rd)
```
mcortesi, line 140: semantically i'd prefer we use Rc in these cases. They have the same value, but one is the current round, while the other is the "desired" one. We are scheduling the timeout for the current one (not the desired) semantically. The desired one in fact only make sense when we are to change rounds, rest of the time is equal to Rc.

mcortesi (commented):
This tiem it was really easy to read the spec. I guess moving everything to appendinx helped a lot, since i was able to grasp everything without accumulating all appendix knowledge before entering the spec.

So, i really like the spec. Wha i think it's missing is:

* Message Set are not defined, and it's unclear. how the work. To me they are part of the state of a participant, and should be treated as such, explaining the semnatincs of them
* upon clauses don't cover all the cases. And i think some are worth noting, for example, what happens if I receives a prepare AFTER i send the commit? Messages from previous rounds? Future Rounds?

piersy:
Hi @mcortesi
Just responding to your questions here:
> * Message Set are not defined, and it's unclear. how the work. To me they are part of the state of a participant, and should be treated as such, explaining the semnatincs of them

Yeah that bit is a bit hazy, its kind of alluded to that nodes store all the messages/events for the current height but not explicitly defined. The only part that hints at this is here:
https://github.com/celo-org/celo-blockchain/blob/be8e779fb879e63cfae8f0d60f9b4f9ff3163fd3/consensus/istanbul/celo-ibft-spec.md?plain=1#L395-L399

The standard approach to handling this in consensus specifications seems to be to define a variable that contains all messages received by a node or instance and then the upon conditions depend on the contents of that variable. See [IBFT 2.0](https://arxiv.org/pdf/1909.10194.pdf).

> * upon clauses don't cover all the cases. And i think some are worth noting, for example, what happens if I receives a prepare AFTER i send the commit? Messages from previous rounds? Future Rounds?

For your first example any prepares received after sending a commit are ignored, because sending a commit implies the node has already seen 2f+1 prepares or commits.

Messages from previous or future rounds are for the most part ignored (excepting round changes), since messages for old rounds are irrelevant, and messages for future rounds are irrelevant to the node until they reach that round. Something that is not quite right here is the way I described the upon condition as being triggered upon receipt of a message or event, instead upon conditions should simply trigger whenever the condition evaluates to true.

So in the case of future messages, they will be added to "received messages" but only become relevant when the node state changes such that the height and round match the future messages in "received messages" at which point an upon condition may be triggered.

So I think all of your points can be addressed by introducing the "received messages" variable and tweaking the description of how upon conditions are triggered.

Please let me know if there are some other un-catered for cases that you can come up with.

consensus/istanbul/celo-ibft-spec.md
piersy: Yes I think so, thanks!

consensus/istanbul/celo-ibft-spec.md
piersy: You could call it pattern matching, the approach is described here - https://github.com/celo-org/celo-blockchain/blob/be8e779fb879e63cfae8f0d60f9b4f9ff3163fd3/consensus/istanbul/celo-ibft-spec.md?plain=1#L323-L325

Let me know how you would change it if that is not clear.

consensus/istanbul/celo-ibft-spec.md
piersy: Good point, I thought I had addressed this in the system model section, but I had not.

consensus/istanbul/celo-ibft-spec.md
piersy: >Is the same message set? for commits and prepares? I think we need a better explanation of what the message set are.

I'm going to add a variable to clarify the message set.

>2. Also, why can Sc be in all these states?

Its a mistake, it should only be Preprepared and Prepared

>3. Why Rd and not Rc, assuming they can be different, right? I would expect this to be Rc since that's the current round we are trying to accumulate messages for

I don't know, thats what is in the code: https://github.com/celo-org/celo-blockchain/blob/be8e779fb879e63cfae8f0d60f9b4f9ff3163fd3/consensus/istanbul/core/backlog.go#L63-L71

I don't actually think we need to have desired round and current round, we could just do with one variable, you can see this relatively easily by looking at the spec and seeing where they would differ. We can go over this together if you're interested. I think removing desired round would make the protocol much easier to understand.

consensus/istanbul/celo-ibft-spec.md
piersy: Given that they are not defined in this version there is nothing to clear.

consensus/istanbul/celo-ibft-spec.md
piersy: The current implementation uses desired round:
https://github.com/celo-org/celo-blockchain/blob/be8e779fb879e63cfae8f0d60f9b4f9ff3163fd3/consensus/istanbul/core/core.go#L743-L748


As I mentioned above I think we can do away with the concept of desired round, which would make things much simpler to follow. But I didn't want to start switching variables around in the spec because that would make it much more confusing to map the spec to the implementation and vice versa.

consensus/istanbul/celo-ibft-spec.md
mcortesi: I'd actually define what message set M is. There's also a N set that i haven't seen used. 

In any case. If we do this by the operaor. I wouldn't use || as that since people understand the symbol as size in a set. So i wouldn't override that notion. Instead use another function

consensus/istanbul/celo-ibft-spec.md
mcortesi: I remember desired round be a corner case. When the val set are in diff rounds. Desired round helped to speed up the consensus on which is the current round

consensus/istanbul/celo-ibft-spec.md
mcortesi: I don't understand. Are they not part of the state? They seem to be since we add messages to the sets during the protocol execution

mcortesi:
> Hi @mcortesi
> 
> Just responding to your questions here:
> 
> > * Message Set are not defined, and it's unclear. how the work. To me they are part of the state of a participant, and should be treated as such, explaining the semnatincs of them
> 
> 
> 
> Yeah that bit is a bit hazy, its kind of alluded to that nodes store all the messages/events for the current height but not explicitly defined. The only part that hints at this is here:
> 
> https://github.com/celo-org/celo-blockchain/blob/be8e779fb879e63cfae8f0d60f9b4f9ff3163fd3/consensus/istanbul/celo-ibft-spec.md?plain=1#L395-L399
> 
> 
> 
> The standard approach to handling this in consensus specifications seems to be to define a variable that contains all messages received by a node or instance and then the upon conditions depend on the contents of that variable. See [IBFT 2.0](https://arxiv.org/pdf/1909.10194.pdf).
> 
 I was imagining that. Define a state variable (or 2 since we use 2 sets) and treat it as state. On the variable definition explain that is a set and how it behaves (whats the set condition), and set of "what". 

Then on the upon conditions I'm unsure it as is a stare variable do the appends in the body of the upon instead on the first line (declaration). Both ways work for me. I like the expresivness of appendind on the declaration line. But it's not consistent with how we manage other state variables 


> 
> > * upon clauses don't cover all the cases. And i think some are worth noting, for example, what happens if I receives a prepare AFTER i send the commit? Messages from previous rounds? Future Rounds?
> 
> 
> 
> For your first example any prepares received after sending a commit are ignored, because sending a commit implies the node has already seen 2f+1 prepares or commits.
> 
> 
> 
> Messages from previous or future rounds are for the most part ignored (excepting round changes), since messages for old rounds are irrelevant, and messages for future rounds are irrelevant to the node until they reach that round. Something that is not quite right here is the way I described the upon condition as being triggered upon receipt of a message or event, instead upon conditions should simply trigger whenever the condition evaluates to true.
> 
> 
> 
> So in the case of future messages, they will be added to "received messages" but only become relevant when the node state changes such that the height and round match the future messages in "received messages" at which point an upon condition may be triggered.
> 
> 
> 
> So I think both of your points can be addressed by introducing the "received messages" variable and tweaking the description of how upon conditions are triggered.
> 


Well. I guess old messages and future messages handling. Is that part od the spec. It's surely part lf the code. So i would expect to see it

Same with the parent commit set
> 
> Please let me know if there are some other un-catered for cases that you can come up with.

consensus/istanbul/celo-ibft-spec.md
piersy: So M is just a variable that represents a set of messages, what is in M is defined by whatever statement is on the right of the assignment operator, N is used in the case that we need to represent more than one set in a single scope.

I've updated the appendix to make this more clear, let me know if it is not.

https://github.com/celo-org/celo-blockchain/blob/ec196aa1e40ef304d49142d172265ea151735444/consensus/istanbul/celo-ibft-spec.md?plain=1#L311-L312

consensus/istanbul/celo-ibft-spec.md
piersy: I've scheduled some time tomorrow to go over this, seems like its quite nuanced

consensus/istanbul/celo-ibft-spec.md
piersy: I'm not sure what message sets you are referring to here?

consensus/istanbul/celo-ibft-spec.md
mcortesi: Agree. The protocol assumes there is a proposer and that it can depend on the heigh and on the round, but it doesn't specify the semantics of the selection, that's outside of the protocol

mcortesi (approved):
LGTM. Now that i read the spec by understanding the upon as "this condition is met" instead of "when this happens", i fully understand the spec. 

Maybe there's a way to make that more evident.

We want to say: "when there is message in the msg set such that...." or "when there is a set of messages in the msg that set such that.."

Instead of saying "when we receive a msg that matches this pattern".

I initially was reading it as the latter case. Which to me was more natural, as i imagined a state machine that only respond to external messages and you specify what happens when a give msg is sent.

piersy:
Hi @mcortesi 

>Maybe there's a way to make that more evident.

>We want to say: "when there is message in the msg set such that...." or "when there is a set of messages in the msg that set such that.."

>Instead of saying "when we receive a msg that matches this pattern".

I've updated the explanation for upon it now reads:
```
upon: UponCondition - Pseudocode directly following upon statements is executed
when the associated UponCondition evaluates to true
```

I'm not really sure how to make the comments proceeding each upon condition any clearer, if you have some specific suggestions please add them.

PR #1750, opened 2021-11-30
Avoid calling NextBlockValidators for preparedCert

piersy:
NextBlockValidators does processing to update the validator set in the
event that validators have been added or removed. In doing this it makes
contract calls.

Given that prepared certs are only used in the context of round changes,
the current validator set will never be different from the set used in
round 0 so the call to NextBlockValidators can be avoided.

### DIFF
--- a/consensus/istanbul/core/prepare.go
+++ b/consensus/istanbul/core/prepare.go
@@ -84,11 +84,7 @@ func (c *core) verifyPreparedCertificate(preparedCertificate istanbul.PreparedCe
 				return nil, err
 			}
 
-			newValSet, err := c.backend.NextBlockValidators(preparedCertificate.Proposal)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit d43e69646f7fc0e74f258dc126a2d70dc30ea0d5
<details><summary>coverage:  44.0% of statements across all listed packages</summary><br>
<pre>
coverage:  43.8% of statements in consensus/istanbul/backend
coverage:  63.4% of statements in consensus/istanbul/core
</pre>
CommentID: ee5aab29ec
</details>

piersy:
I had incorrectly assumed that we could use the current block validtors but in the case that we need to validate a commit with an epoch seal the next block validators are required.

PR #1770, opened 2021-12-07
Piersy/consolidate ibft logic

piersy:
### Description

This PR is a step towards consolidating the IBFT logic and separating it from all the other concerns (serialization/signatures/persistence/concurrency/synchronisation) that are currently also handled inside the istanbul package, I do not consider the work to consolidate IBFT logic as finished, there is quite a lot more that could be done.

The intention of this PR is to simply re-factor without affecting the behaviour of the code, except for this removal https://github.com/celo-org/celo-blockchain/pull/1770/commits/3d26a7f9378b567f5fd885ee9c08c06ff12372b3 which if you look at the code just can't happen currently and also seems to be totally misplaced, in that it is checked for every message processed but it's not inspecting any fields of the message.

Currently some of the logic exists within the `algorithm.go` implementation but some still exists outside (for example state manipulations, logic to handle round changes) ideally we can get to a point where all ibft logic is contained with the `algorithm.go` implementation.

This PR introduces a new message type `algorithm.Msg` that allows us to treat consensus messages homogeneously, this allows us to consolidate message processing flows into a single flow. Currently its use is rather limited and so the benefits it provides in consolidating processing logic and reducing the amount of code are quite small, but further use should yield more benefits.

This PR also moves message validation into a single place which helps with understanding the flow of message processing. The handle message function can be read linearly to see what happens to all types of messages. Having the message verification laid out like that also helps spot and potentially resolve problems in our processing. For example, we currently check whether a message is from a validator before we check if its a future message. If we wish to change this we can now by simply moving the check further down the function, previously to achieve the same we would have had to have duplicated that check in all 4 message handlers (preprepare, prepare, commit, round change).

In its current form I don't think this PR provides great benefit, notably it has not reduced the amount of code significantly. I think this is because we have added some new structures and types which where they are used help to reduce the amount of code needed, but they have not been extended them into the system enough to remove the old ones as such we need glue code bridge the gap between the new and old. I think if we extended `algorithm.Msg` type to be used almost everywhere except for when we deserialise from the network and serialise to the network this would then allow us to remove old logic and replace it with more homogeneous logic that would then significantly reduce code size.


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit cdf7ab8f852a71e1284817fb27bd1dbad155b91b
<pre>
coverage:  54.4% of statements in consensus/istanbul
coverage:  77.8% of statements in consensus/istanbul/algorithm
coverage:  44.9% of statements in consensus/istanbul/backend
coverage:  63.3% of statements in consensus/istanbul/core
</pre>
CommentID: d723a9f796
</details>

consensus/istanbul/algorithm/algorithm.go
```diff
@@ -0,0 +1,169 @@
...
+// certain logic that is really part of the consensus algorithm is abstracted
+// away by the Oracle, for example managing the round and state of the
+// algorithm and the logic for processing round change messages. In time we
+// should be able to move this logic here and slim down the oracle's interface.
+type Oracle interface {
+	CurrentState() State
+	QuorumCommit(height, round uint64, val Value) bool
+	QuorumPrepare(height, round uint64, val Value) bool
+	DesiredRound() uint64
+	QuorumRoundChange() *uint64
```
trianglesphere, line 95: I'd prefer an interface more like `bool, uint64` rather than a pointer. The best interface would be `Some(u64)` (using rust syntax), but absent that, the pair makes slightly more sense to me. (I was able to figure out the intention from reading the definition, but it still feels tricky to me)

consensus/istanbul/algorithm/algorithm.go
```diff
@@ -0,0 +1,169 @@
...
+	// We see a quorum of commits and we are not yet committed, then move to
+	// committed state.
+	if t == Commit && s < Committed && oracle.QuorumCommit(h, r, v) {
+		var round uint64 = 0
+		return nil, &round, nil
+	}
+
+	// We are not yet prepared and see a quorum of prepares (where a commit
+	// also counts as a prepare), send a commit.
+	if t.In(Prepare, Commit) && s < Prepared && oracle.QuorumPrepare(h, r, v) {
```
trianglesphere, line 136: Structurally, is it correct that the message that is being processed must be previously be provided to oracle to add to the message set?

consensus/istanbul/core/handler.go
```diff
@@ -172,43 +246,226 @@ func (c *core) handleMsg(payload []byte) error {
...
+		if proposerForMsgRound.Address() != msg.Address {
+			logger.Warn("Ignore preprepare message from non-proposer", "actual_proposer", proposerForMsgRound.Address())
+			return errNotFromProposer
+		}
+
+		defer c.handlePrePrepareTimer.UpdateSince(time.Now())
+
+		if m.Round > 0 {
+			logger.Trace("Trying to move to round change certificate's round", "target round", m.Round)
+			err := c.startNewRound(new(big.Int).SetUint64(m.Round))
```
trianglesphere, line 332: Is this not an algorithm step?

trianglesphere (commented):
First pass. I left a couple smaller comments, but I also have a couple structural questions.

It seems like the flow is
1. Verify message
2. Add message to the appropriate set
3. Run algorithm handle message
4. Perform next action (output of handle message)

It looks like 1 & 2 are merged together, 3 is in it's own function with returning information to perform 4.

I'm curious to see how you'd like to see it structured.

consensus/istanbul/core/handler.go
piersy: It definitely looks like it should be! Will look into this more, thanks

piersy:
> First pass. I left a couple smaller comments, but I also have a couple structural questions.
> 
> It seems like the flow is
> 
> 1. Verify message
> 2. Add message to the appropriate set
> 3. Run algorithm handle message
> 4. Perform next action (output of handle message)
> 
> It looks like 1 & 2 are merged together, 3 is in it's own function with returning information to perform 4.
> 
> I'm curious to see how you'd like to see it structured.

Hi @trianglesphere thanks for taking a look here:

This is the result of a 2 week time-box on refactoring, so its definitely not the final structure that I would like to see. However at a high level that structure is about what i was aiming at. The thing I would like to change most is to be able to have more processing inside the algorithm logic, theres some processing that it would be nice to move out of the verification, and some that would be nice to move out of the next action processing.

So for example such as determining the validity of a prepared certificate based on whether the all the individual messages have the same proposal, whether it contains enough messages and whether the messages have matching views. Basically verification that is to do with consensus as opposed to things like verifying signatures or checking that the encoding is correct.

And then for the next action processing it would be great if the algorithm state and state transitions could be managed inside the algorithm, leaving only message sending and committing as actions that happen outside the algorithm as part of the next action step.

Theres also a lot of "bespoke" processing that goes on in switch statements in the handler now and I'd like to slim that down, for example the need to add each message to the state in a different way. Now that there is a single type that can represent all consensus messages we should be able to add messages to the state in one place with a single method. So my goal would be to reduce those switch statements to the bare minimum code required to handle the differences in the messages.

consensus/istanbul/algorithm/algorithm.go
piersy: Hey @trianglesphere I'm not sure what you mean by structurally correct here?

If your are pointing out that the quorum calculation is taking into account the currently processed message then I think that is correct. Bear in mind that the `QuorumPrepare` method is enforcing that the quorum is with the same height, round and value and with the appropriate message types.

consensus/istanbul/algorithm/algorithm.go
piersy: Definitely these methods could do with some documentation describing the meaning of what they return.

I'm wondering if you could describe a bit more what seems "tricky" to you about this approach?

In this specific case I think using a bool could result in slightly neater use of these methods where they are currently used, so I'm not averse to it. In general I think representing non existence with nil makes a lot of sense, nothing else in the language indicates theres nothing there more than nil!

consensus/istanbul/core/handler.go
```diff
@@ -162,6 +168,74 @@ func (c *core) sendEvent(ev interface{}) {
...
+// then the round change certificate verification logic could be moved to the
+// algorithm, but this will be a bit of work.
+func buildAlgorithmMsg(msg *istanbul.Message, o *RoundStateOracle) (*algorithm.Msg, error) {
+	m := &algorithm.Msg{
+		MsgType: algorithm.Type(msg.Code),
+	}
+	switch msg.Code {
+	case istanbul.MsgPreprepare:
+		p := msg.Preprepare()
+		rccID, err := o.SetRoundChangeCertificate(p)
```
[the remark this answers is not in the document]
mcortesi: In fact, it's looks fishy, since `handleRoundChangeCertificate` assumes uses `current` to obtain state data and it looks to me that it assumes is a msg from the current view, not from a future or old one

consensus/istanbul/core/handler.go
```diff
@@ -162,6 +168,74 @@ func (c *core) sendEvent(ev interface{}) {
...
+	return false
+}
+
+// buildAlgorithmMsg builds an algorithm message from an istanbul message. It
+// takes the oracle because it needs to validate round change certificates and
+// generate an associated id. This is pretty clunky, ideally we could represent
+// the round change certificate in more detail in the algorithm message and
+// then the round change certificate verification logic could be moved to the
+// algorithm, but this will be a bit of work.
+func buildAlgorithmMsg(msg *istanbul.Message, o *RoundStateOracle) (*algorithm.Msg, error) {
```
mcortesi, line 197: Not sure i see a lot of value on this. I see value in having a method that no matter the msg is able to give me the msg's round and sequence, and if there's another thing that is shared among messages. But besides that, it does not provid emuch

mcortesi (commented):
not finished, slowly reading prev version and current version to assert they are equal, found some things that are not.

In general, i like remvoing `checkMessage()` and `catchFutureMsg()`; i prefer handling all of that at handeMsg(). The algorithm msg doesn't make sense to me yet. Some parts were "hard to find" in the new code.

mcortesi (changes requested):
by checking the code, i'm convinced this is currently broken. And also a bit confusing.

In general i like how you've consolidated the check for future and old msg in the handleMsg, instead of having a call to checkMessage in every `handleXXX` method. And i can buy the switch on handle that consolidates all `handleXXX` in a single method.

But i'm not buying the algorithmMessage, nor the Algorithm. I see as an unnecessary complexity that is being added. Expanding the surface for potential errors.

We are also introducing new cached state, which makes reasoning about correcteness more complex. And by processing message before even knowing if they are future/old we are wasting processor cycles, and potentially introducing a DoS vector. 

Right now we have these steps:

  1. buildAlgoMsg (which has sideffects)
  2. check old&future msg
  3. switch that does "some" processing by mesg
  4. call to algorithm
  5. execute action specified by algorithm

In the current state, i see that having 3, 4 and 5  as separate steps makes reasoning more complicated, not simpler. I'd prefer to be able to follow the "preprepare" mgs (or any other msg) in a single thread, instead of needing to inspect the swtich and the algorithm (and the buildAlgoMsg) to grasp all of it.

consensus/istanbul/core/handler.go
```diff
@@ -162,6 +168,74 @@ func (c *core) sendEvent(ev interface{}) {
...
 }
 
+// Returns true if this message is considered a future message.
+func IsFutureMsg(currentSequence, desiredRound *big.Int, currentState State, msgView *istanbul.View, msgCode uint64) bool {
+
+	// Future sequences indicate a future message.
+	if msgView.Sequence.Cmp(currentSequence) > 0 {
+		return true
+	}
+	// Future rounds inidcate a future message except for round changes
```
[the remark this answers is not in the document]
piersy:
```suggestion
	// Future rounds inidicate a future message except for round changes
```

consensus/istanbul/core/oracle.go
```diff
@@ -0,0 +1,149 @@
...
+	return nil
+}
+
+// SetRoundChangeCertificate maps a round change certificate to an id and
+// returns the id. This allows the algorithm to reference the round change
+// certificate without having a compile time dependency on
+// istanbul.RoundChangeCertificate.
+func (o *RoundStateOracle) SetRoundChangeCertificate(preprepare *istanbul.Preprepare) (*algorithm.Value, error) {
+	// Unset old values
+	o.roundChangeCertId = nil
```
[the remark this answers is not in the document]
piersy: It's not really a cache, at most `ValidRoundChangeCert` will be called once for each call to `SetRoundChangeCertificate`.

So the flow would be:

Receive a preprepare -> Set the round change cert -> send the message to the algo -> inspect the round change cert.

All this is happening in a single go-routine, so any time another preprepare is received it will overwrite the last value and the new value will be used just for processing that new preprepare message.

consensus/istanbul/core/handler.go
```diff
@@ -174,43 +248,226 @@ func (c *core) handleMsg(payload []byte) error {
...
-func (c *core) handleCheckedMsg(msg *istanbul.Message, src istanbul.Validator) error {
-	logger := c.newLogger("func", "handleCheckedMsg", "from", msg.Address)
-
-	// Store the message if it's a future message
-	catchFutureMessages := func(err error) error {
-		if err == errFutureMessage {
-			// Store in backlog (if it's not from self)
-			if msg.Address != c.address {
-				c.backlog.store(msg)
+	m, err := buildAlgorithmMsg(msg, c.oracle)
```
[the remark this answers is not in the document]
piersy: It is effectively stateless, this is because although the oracle is long lived the results of calling `SetRoundChangeCertificate` are only ever referenced when processing the preprepare message that resulted in the call to `SetRoundChangeCertificate`. This is easily verifiable because we can see that `ValidRoundChangeCert` is only called for preprepare messages (this is the only reference in the code).
https://github.com/celo-org/celo-blockchain/blob/baa55bd4837d3690c03324429d1858164b45a7de/consensus/istanbul/algorithm/algorithm.go#L147

And any subsequent preprepare message will overwrite the prior results.
https://github.com/celo-org/celo-blockchain/blob/baa55bd4837d3690c03324429d1858164b45a7de/consensus/istanbul/core/handler.go#L202-L204

It's not my ideal approach, but given the time constraints this seemed like the most pragmatic option. I don't see it being too much work to properly represent the round change cert in an `algorithm.Msg` which would then make this process obviously stateless.

piersy:
Hi @mcortesi I'll respond to your points here.

> We are also introducing new cached state, which makes reasoning about correcteness more complex.
Explained [here](https://github.com/celo-org/celo-blockchain/pull/1770#discussion_r783117094) its not really a cached state.

>And by processing message before even knowing if they are future/old we are wasting processor cycles, and potentially introducing a DoS vector.

Good point, I think I originally wanted to do the processing when setting rather than getting so that an error could be returned ( the algorithm doesn't consider errors), but given that in the end the error is just logged we could just move that logic to the get step (`ValidRoundChangeCert`)  which should solve these issues.

> But i'm not buying the algorithmMessage, nor the Algorithm. I see as an unnecessary complexity that is being added. Expanding the surface for potential errors.
> 
> Right now we have these steps:
> 
> 1. buildAlgoMsg (which has sideffects)
> 2. check old&future msg
> 3. switch that does "some" processing by mesg
> 4. call to algorithm
> 5. execute action specified by algorithm
> 
> In the current state, i see that having 3, 4 and 5 as separate steps makes reasoning more complicated, not simpler. I'd prefer to be able to follow the "preprepare" mgs (or any other msg) in a single thread, instead of needing to inspect the swtich and the algorithm (and the buildAlgoMsg) to grasp all of it.

In their current state, I agree it is more complicated, as I mentioned in the PR description:

>Currently some of the logic exists within the algorithm.go implementation but some still exists outside (for example state manipulations, logic to handle round changes) ideally we can get to a point where all ibft logic is contained with the algorithm.go implementation.

But I believe that if we could get to the point that all consensus logic and state is represented in the algorithm then things would be much simpler. So I see this as a first step, when re-structuring things often get messier before they get cleaner and I'd say that this is somewhere between 25% and %50 of the way there. As I mentioned in the description I'm not sure this PR constitutes an improvement in its current state, some things became a bit nicer, some became worse.

piersy:
Given the high probability that we will leverage narwhal's consensus in place of Istanbul, there is no reason to continue here.

PR #1731, opened 2021-11-02
Switch to node12 for CI

trianglesphere:
### Description

Switches from node 10 to 12 to fix an issue with yarn using an unauthenticated git
protocol in the `monorepo-checkout` step

PR #1776, opened 2021-12-15
Use node10 for prepare-system-contracts build step

piersy:
This was previously changed accidentally in #1731.

Also updated the cache key for the prepared system contracts to include
the executor image, so that any image change will cause
prepare-system-contracts to run, which should alert us to any problems
before we merge any changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,14 +11,19 @@ parameters:
   system-contracts-path:
     type: string
     default: "compiled-system-contracts"
+  system-contracts-executor-image:
+    type: string
+      # The system contracts executor needs node 10 to function, this should
+      # only be changed when system-contracts-monorepo-version is changed.
+    default: "celohq/node10-gcloud:v3"
 executors:
   golang:
     docker:
       - image: circleci/golang:1.16
     working_directory: ~/repos/geth
-  node-v12:
+  system-contracts-executor:
     docker:
-      - image: us.gcr.io/celo-testnet/circleci-node12:1.0.0
+      - image: <<pipeline.parameters.system-contracts-executor-image>>
     working_directory: ~/repos/geth
   e2e:
     docker:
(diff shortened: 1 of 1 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit e4277cccc7bff921cb7a8a28b9bce6511005bdbc
<details><summary>coverage:  45.4% of statements across all listed packages</summary><br>
<pre>
coverage:  54.9% of statements in consensus/istanbul
coverage:  45.3% of statements in consensus/istanbul/backend
coverage:  66.1% of statements in consensus/istanbul/core
</pre>
CommentID: 2f5f93b46a
</details>

.circleci/config.yml
```diff
@@ -11,14 +11,19 @@ parameters:
...
 executors:
   golang:
     docker:
       - image: circleci/golang:1.16
     working_directory: ~/repos/geth
-  node-v12:
+  system-contracts-executor:
     docker:
-      - image: us.gcr.io/celo-testnet/circleci-node12:1.0.0
+      - image: <<pipeline.parameters.system-contracts-executor-image>>
```
gastonponti, line 26: ok, so default 10, unless we set this parameter in the ci config for the specific PR, right?

gastonponti (approved):
LGTM

.circleci/config.yml
YoduYodu: node-v12(now system-contracts-executor) is only used for jobs.prepare-system-contracts

YoduYodu (approved):

.circleci/config.yml
piersy: @gastonponti that parameter is defined in the `.circleci/config.yml` at the top, so it's like a global variable of the config file.

https://github.com/celo-org/celo-blockchain/blob/e4277cccc7bff921cb7a8a28b9bce6511005bdbc/.circleci/config.yml#L14-L18

It's defined there rather than hardcoded so that it can be used in the cache key, so that if we change the image it invalidates the cache, and hence forces `prepare-system-contracts` to run fully

PR #1790, opened 2022-01-05
Fix JSON RPC return + gas estimation

### LINKED ISSUE #1789: Clean up get_Transaction JSON
## Description

Specify what JSON fields should be returned for each transaction type and make sure that those fields are being returned.

## Context

This is a non-consensus change for Espresso, but this is important for users to have a consistent schema.

Note from slack: "the fields maxFeePerGas and maxPriorityFeePerGas seem to be included in the rlp encoding for the new celo transaction type 0x7c"

trianglesphere:
### Description

The first change is passing parameters through to gas estimation. This is
important, specifically that if the fee currency parameter is not provided,
gas estimation will fail when it should not if the EOA does not have any
native currency, but does have a whitelisted fee currency and is trying to pay
for fees in that.

The second change includes the CeloDynamicFeeType transactions when
constructing responses over the JSON RPC.

### Tested

Manually. Working

### Related issues

- Fixes #1789


gastonponti (approved):
LGTM!

YoduYodu (approved):

PR #1811, opened 2022-01-25
Set default genesis in mobile client from network ID

nategraf:
### Description

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/mobile/geth.go
+++ b/mobile/geth.go
@@ -220,6 +220,7 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		return nil, err
 	}
 
+	// Add the root node object to memsize for debugging memory allocations.
 	debug.Memsize.Add("node", rawStack)
 
 	var genesis *core.Genesis
@@ -231,7 +232,10 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 		if err := json.Unmarshal([]byte(config.EthereumGenesis), genesis); err != nil {
 			return nil, fmt.Errorf("invalid genesis spec: %v", err)
 		}
+
 		// If we have the testnet, hard code the chain configs too
+		// NOTE(victor): It's unclear if this behavior is a sensible way to establish defaults. We
+		// may want to remove these statements.
 		if config.EthereumGenesis == AlfajoresGenesis() {
 			genesis.Config = params.AlfajoresChainConfig
 			if config.EthereumNetworkID == 1 {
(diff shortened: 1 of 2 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 4493a3e7a9e028368c023e5ca3fc5e52bff3db1b
<details><summary>coverage:  51.0% of statements across all listed packages</summary><br>
<pre>
coverage:  66.1% of statements in consensus/istanbul
coverage:  41.3% of statements in consensus/istanbul/announce
coverage:  54.6% of statements in consensus/istanbul/backend
coverage:  70.4% of statements in consensus/istanbul/core
</pre>
CommentID: 901e9401c1
</details>

mobile/geth.go
```diff
@@ -243,7 +247,12 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
 				config.EthereumNetworkID = int64(params.BaklavaNetworkId)
 			}
 		}
+	} else if config.EthereumNetworkID == int64(params.AlfajoresNetworkId) {
+		genesis = core.DefaultAlfajoresGenesisBlock()
+	} else if config.EthereumNetworkID == int64(params.BaklavaNetworkId) {
+		genesis = core.DefaultBaklavaGenesisBlock()
 	}
```
jeanregisser, line 254: Great! Just to be sure, we don't need a case for the `MainnetNetworkId` because it's already the default?

jeanregisser:
And one more question, what about the backward compatibility for existing Valora users who had been initializing the node by passing the genesis data?
Is there some persisted data somewhere down the line that would create a problem by switching from passing the genesis data to not passing it anymore? Especially related to the Espresso hardfork.

nategraf:
> And one more question, what about the backward compatibility for existing Valora users who had been initializing the node by passing the genesis data? Is there some persisted data somewhere down the line that would create a problem by switching from passing the genesis data to not passing it anymore? Especially related to the Espresso hardfork.

If we additionally change Valora to not pass in the chain config parameters, I don't think there should be an issue. This is certainly something we should test.

mobile/geth.go
nategraf: Default was to use mainnet, but I added a case to make it more explicit.

nategraf:
Following the process for making releases, this PR should be merged to `master` first, then cherry picked back to `release/1.5.x`

PR #1816, opened 2022-01-25
Ensure txPricedList reheaps regularly (avoid tx memory leak)

hbandura:
This PR fixes a memory leak found while running v1.5.0 nodes before the Espresso fork.

`txPricedList` keeps both a list of all txs in the pool, and also a heap for fast access to the cheapest txs. These heaps are not always synchronized with the txs list: *they increment at the same time as the list, but they do not decrement at the same time*. To decrement, the txpool marks how many txs are 'stale', and after a critical point, a `Reheap()` is called, where the whole memory of the heaps is released and they are recreated.

The issue in the code was that the critical point was never reached, and the heaps grew indefinitely. Strangely, the behavior is different after Expresso HF activation, since then `Reheap()` is called periodically on changes for the base fee. But before the HF nothing forces this so the heaps grow forever. One of the reasons for the critical point never being reached is the tx pool not respecting the contract of calling `Removed` every time it should.

The issue seems to be similar to https://github.com/ethereum/go-ethereum/issues/23690, which was workarounded by reheaping after every pool reorg. Here we also add a couple of safeguards based on the size of the tx pool to ensure that, even if the node is stuck (due to a bad block or any other reason), proper reheaps are triggered if necessary.

A follow-up investigation should be made to have an exhaustive list of all the reasons why the de-synchronization occurs. An example of those is the method `demoteUnexecutables`, where many txs are removed from the pool but not from the pricedlist.

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -336,7 +336,8 @@ func NewTxPool(config TxPoolConfig, chainconfig *params.ChainConfig, chain block
 	pool.reset(nil, chain.CurrentBlock().Header())
 	// TODO: Does this reordering cause a problem?
 	// priced list depends on the current ctx which is set in reset
-	pool.priced = newTxPricedList(pool.all, &pool.currentCtx)
+	// Use the global slots as the max amount of stale transactions in the priced heap before a re-heap.
+	pool.priced = newTxPricedList(pool.all, &pool.currentCtx, int64(pool.config.GlobalSlots))
 
 	// Start the reorg loop early so it can handle requests generated during journal loading.
 	pool.wg.Add(1)
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -1255,6 +1255,9 @@ func (pool *TxPool) runReorg(done chan struct{}, reset *txpoolResetRequest, dirt
 		pool.demoteUnexecutables()
 		if reset.newHead != nil && pool.chainconfig.IsEspresso(new(big.Int).Add(reset.newHead.Number, big.NewInt(1))) {
 			pool.priced.SetBaseFee(pool.ctx())
+		} else {
+			// Prevent the price heap from growing indefinitely
+			pool.priced.Reheap()
 		}
 	}
 	// Ensure pool.queue and pool.pending sizes stay within the configured limits.
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -779,7 +779,7 @@ func (l *txPricedList) Put(tx *types.Transaction, local bool) {
 // from the pool. The list will just keep a counter of stale objects and update
 // the heap if a large enough ratio of transactions go stale.
 func (l *txPricedList) Removed(count int) {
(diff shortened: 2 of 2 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 4d8ff2ed245e03839be984277dba013f96332564
<details><summary>coverage:  51.3% of statements across all listed packages</summary><br>
<pre>
coverage:  66.3% of statements in consensus/istanbul
coverage:  41.7% of statements in consensus/istanbul/announce
coverage:  69.8% of statements in consensus/istanbul/core
</pre>
CommentID: a350172000
</details>

diwu1989:
the ratio:
```
// Reheap if the ratio of stales is more than 25% of the heaps sizes
overStalesRatio := int(stales) > (urgentSize+floatingSize)/4
```

is it possible for the 4 to be taken in as a geth argument / flag, with default being 4?
in case the 25% threshold is not ideal, we can easily tune that without recompiling

core/tx_pool.go
```diff
@@ -336,7 +336,8 @@ func NewTxPool(config TxPoolConfig, chainconfig *params.ChainConfig, chain block
 	pool.reset(nil, chain.CurrentBlock().Header())
 	// TODO: Does this reordering cause a problem?
 	// priced list depends on the current ctx which is set in reset
-	pool.priced = newTxPricedList(pool.all, &pool.currentCtx)
+	// Use the global slots as the max amount of stale transactions in the priced heap before a re-heap.
+	pool.priced = newTxPricedList(pool.all, &pool.currentCtx, int64(pool.config.GlobalSlots))
```
mcortesi, line 340: i guess logic here is that we can have as many stales a tx have in the pool, if we consider that a tx usually is 1 slot, which i'm not sure if true in average.

Still, it's works a default number, just a bit ackward

mcortesi (approved):
LGTM

hbandura:
> the ratio:
> 
> ```
> // Reheap if the ratio of stales is more than 25% of the heaps sizes
> overStalesRatio := int(stales) > (urgentSize+floatingSize)/4
> ```
> 
> is it possible for the 4 to be taken in as a geth argument / flag, with default being 4? in case the 25% threshold is not ideal, we can easily tune that without recompiling
> 
> or even make `maxStales` a flag of its own that can be passed in, whichever one is the cleanest way to have some manual control over the frequency of these intentional reheaps.

While adding them as arguments is possible, it probable adds more noise than anything, since now the Reheap will be run on every pool reorg with a reset, which happens on every ChainHeadEvent i.e every new block. We could argue that it's not even needed, except for extreme cases where the blockchain is not advancing.

core/tx_pool.go
hbandura: if stales > as globalslots and a reheap hadn't been done before, it means that the heaps sizes are at a minimum 4x the size (due to the 25% rule) so something is off. A lower number could be set, but I feel like as a ceiling it should be ok.

PR #1818, opened 2022-01-26
Eth api using baseFee only when needed

gastonponti:
### The Problem

The `eth` RPC API after expreso required some amending.

On ethereum, `baseFee` ( called `gasPriceMinimun` on Celo) is present on the block header. For Celo, it's present on the `GasPriceMinimum` contract, and it also needs to be denominated in all different fee currencies. 

This difference poses a problem on the RPC, since whenever a `Transaction` or a `TransactionReceipt` is constructed, the `gasPrice` field is computed for new transaction. To do so, it uses the `baseFee`. Obtaining the `baseFee` for block X is easy on Ethereum and hard on Celo; since it's a smart contract call, the node needs to have the `state` available but that's not true for historical blocks on any node that it's not an archive node.

### The Solution

To solve this issue, this PR takes the following approach:

 * If node has access to the block's state, then `gasPrice` is computed as usual
 * If node doesn't have access to block's state, then `gasPrice` is `nil`

Additionally, we have the same problem with the new `AccessList` rpc endpoint, that computes the access list for a transaction on a given block. If the block is too old, and we don't have the state, then the rpc will fail.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1253,13 +1253,13 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 		result.GasTipCap = (*hexutil.Big)(tx.GasTipCap())
 		// if the transaction has been mined, compute the effective gas price
 		if blockHash != (common.Hash{}) {
-			baseFee, _ := baseFeeFn(tx.FeeCurrency())
-			if baseFee != nil {
+			baseFee, err := baseFeeFn(tx.FeeCurrency())
+			if err != nil {
 				// price = min(tip, gasFeeCap - baseFee) + baseFee
 				price := math.BigMin(new(big.Int).Add(tx.GasTipCap(), baseFee), tx.GasFeeCap())
 				result.GasPrice = (*hexutil.Big)(price)
 			} else {
-				// baseFee nil for DinamicFees implies that there is no state to retrieve the baseFee for that block
+				// error != nil for DinamicFees implies that there is no state to retrieve the baseFee for that block
 				result.GasPrice = nil
 			}
 		} else {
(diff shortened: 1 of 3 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 85d5dc227c0dfcc868a87db4fe39fbbba4dc5322
<details><summary>coverage:  47.3% of statements across all listed packages</summary><br>
<pre>
coverage:  69.9% of statements in consensus/istanbul/core
</pre>
CommentID: 8c45cd6630
</details>

internal/ethapi/api.go
```diff
@@ -1253,10 +1252,16 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 		result.GasFeeCap = (*hexutil.Big)(tx.GasFeeCap())
 		result.GasTipCap = (*hexutil.Big)(tx.GasTipCap())
 		// if the transaction has been mined, compute the effective gas price
-		if baseFee != nil && blockHash != (common.Hash{}) {
-			// price = min(tip, gasFeeCap - baseFee) + baseFee
-			price := math.BigMin(new(big.Int).Add(tx.GasTipCap(), baseFee), tx.GasFeeCap())
-			result.GasPrice = (*hexutil.Big)(price)
+		if blockHash != (common.Hash{}) {
+			baseFee, _ := baseFeeFn(tx.FeeCurrency())
```
mcortesi, line 1256: it would be better if we use the error and then check `if err != nil || baseFee == nil`. Seems more robust

internal/ethapi/api.go
```diff
@@ -1793,8 +1804,10 @@ func (s *PublicTransactionPoolAPI) PendingTransactions() ([]*RPCTransaction, err
 	for _, tx := range pending {
 		from, _ := types.Sender(s.signer, tx)
 		if _, exists := accounts[from]; exists {
-			baseFee, _ := s.b.CurrentGasPriceMinimum(context.Background(), tx.FeeCurrency())
-			transactions = append(transactions, newRPCPendingTransaction(tx, curHeader, s.b.ChainConfig(), baseFee))
+			baseFeeFn := func(feeCurrency *common.Address) (*big.Int, error) {
```
mcortesi, line 1807: this could be outside of the `for`

mcortesi (approved):
Looks goods, left 2 nits

PR #1829, opened 2022-02-02
Enforce proper check of ParentAgregatedSeal during preprepare phase

hbandura:
## Problem

During consensus, when validator has to verify that a proposed block is valid, the validators not checking the `ParentAggregatedSeal`. Thus, they can potentially accept an invalid block as valid.

Full Nodes during regular block syncing don't have the same problem, and thus reject the block. In a scenario with validators depending on proxies (full nodes) this also implies that after such a block, network connections between validators and proxies will break, thus the network will stall.

## How?

To verify a block, validators have to check the header, but ignore the `AggreatedSeal` since this field is the result of consensus and will only be added as the final step of it.

To achieve this, `backend.VerifyProposal()` calls `engine.VerifyHeader` and it will mark the block as invalid if this call fails, unless the error is related to the AggregatedSeal. This is incorrect, since if `engine.VerifyHeader()` fails due to the missing AggregatedSeal it does not imply that the rest of the header is valid, just that the first error on it found is that.

In particular, `engine.VerifyHeader()` will not verify the `ParentAggregatedSeal` if `AggreatedSeal` is missing.

## Fix

In this PR, instead of calling `VerifyHeader` we call a new function `verifyHeaderFromProposal` that checks that the `AggregatedSeal`'s signature is empty, and checks all others fields. As expected from a consensus proposal.

This PR is applied on top of the 1.5.x version branch


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 730192c6d7bb4986d712da47dab4f83703571f46
<details><summary>coverage:  46.1% of statements across all listed packages</summary><br>
<pre>
coverage:  65.3% of statements in consensus/istanbul/core
</pre>
CommentID: 198b544752
</details>

consensus/istanbul/backend/engine.go
```diff
@@ -103,14 +105,22 @@ func (sb *Backend) Author(header *types.Header) (common.Address, error) {
 // VerifyHeader checks whether a header conforms to the consensus rules of a
 // given engine. Verifies the seal regardless of given "seal" argument.
```
piersy, line 106: This comment is now now confusing. It talks about 'the' seal but there are 2 seals, so it would be good to make it explicit what seals it checks.

consensus/istanbul/backend/engine.go
```diff
@@ -103,14 +105,22 @@ func (sb *Backend) Author(header *types.Header) (common.Address, error) {
...
+	return sb.verifyHeader(chain, header, true, nil)
 }
 
 // verifyHeader checks whether a header conforms to the consensus rules.The
 // caller may optionally pass in a batch of parents (ascending order) to avoid
 // looking those up from the database. This is useful for concurrently verifying
 // a batch of new headers.
-func (sb *Backend) verifyHeader(chain consensus.ChainHeaderReader, header *types.Header, parents []*types.Header) error {
+// If emptyAggregatedSeal is set, the aggregatedSeal will be checked to be completely empty. Otherwise
+// it will be checked as a normal aggregated seal.
```
piersy, line 122: I think it would be good to point out here that the `parentAggregatedSeal` is always checked, otherwise the terminology of `aggregatedSeal` is a bit ambiguous.

consensus/istanbul/backend/engine.go
```diff
@@ -160,7 +170,9 @@ func (sb *Backend) checkEpochBlockExists(chain consensus.ChainHeaderReader, head
 // rather depend on a batch of previous headers. The caller may optionally pass
 // in a batch of parents (ascending order) to avoid looking those up from the
 // database. This is useful for concurrently verifying a batch of new headers.
-func (sb *Backend) verifyCascadingFields(chain consensus.ChainHeaderReader, header *types.Header, parents []*types.Header) error {
+// If emptyAggregatedSeal is set, the aggregatedSeal will be checked to be completely empty. Otherwise
+// it will be checked as a normal aggregated seal.
```
piersy, line 174: Same as above.

piersy (changes requested):
I think we should add a test to verify that this will actually reject blocks that are missing the parentSeals.

consensus/istanbul/backend/engine.go
hbandura: Since this is the same diff as the one applied in the release branches, I'd like for it to stay the same, and maybe we can update changes in a different PR?

mcortesi (approved):

consensus/istanbul/backend/engine.go
piersy: Ok, lets keep this the same.

piersy (approved):

PR #1830, opened 2022-02-02
Enforce proper check of ParentAgregatedSeal during preprepare phase (1.5.x release)

hbandura:
## Problem

During consensus, when validator has to verify that a proposed block is valid, the validators not checking the `ParentAggregatedSeal`. Thus, they can potentially accept an invalid block as valid.

Full Nodes during regular block syncing don't have the same problem, and thus reject the block. In a scenario with validators depending on proxies (full nodes) this also implies that after such a block, network connections between validators and proxies will break, thus the network will stall.

## How?

To verify a block, validators have to check the header, but ignore the `AggreatedSeal` since this field is the result of consensus and will only be added as the final step of it.

To achieve this, `backend.VerifyProposal()` calls `engine.VerifyHeader` and it will mark the block as invalid if this call fails, unless the error is related to the AggregatedSeal. This is incorrect, since if `engine.VerifyHeader()` fails due to the missing AggregatedSeal it does not imply that the rest of the header is valid, just that the first error on it found is that.

In particular, `engine.VerifyHeader()` will not verify the `ParentAggregatedSeal` if `AggreatedSeal` is missing.

## Fix

In this PR, instead of calling `VerifyHeader` we call a new function `verifyHeaderFromProposal` that checks that the `AggregatedSeal`'s signature is empty, and checks all others fields. As expected from a consensus proposal.

This PR is applied on top of the 1.5.x version branch

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -592,10 +592,9 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (*istanbulCore.StateProces
 		}
 	}
 
-	err := sb.VerifyHeader(sb.chain, block.Header(), false)
+	err := sb.verifyHeaderFromProposal(sb.chain, block.Header())
 
-	// ignore errEmptyAggregatedSeal error because we don't have the committed seals yet
-	if err != nil && err != errEmptyAggregatedSeal {
+	if err != nil {
 		if err == consensus.ErrFutureBlock {
 			return nil, time.Unix(int64(block.Header().Time), 0).Sub(now()), consensus.ErrFutureBlock
 		} else {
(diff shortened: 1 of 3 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 1bd6b18823dea3495d5f6464d8446b64dd549d7c
<pre>
coverage:  45.0% of statements in consensus/istanbul/backend
coverage:  63.9% of statements in consensus/istanbul/core
</pre>
CommentID: 076f09b18d
</details>

mcortesi (approved):

mcortesi:
This PR has been merged manually (`git rebase on v1.5.x`) to mantain the commit SHA that were already using during the emergency patch generation

PR #1834, opened 2022-02-07
Alternative docker build that uses libmimalloc for better runtime memory allocation & heap fragmentation

diwu1989:
### Description

Alpine's musl memory allocator is known to be not performant for concurrent workload and not handle heap fragmentation well over time.

This is alternative build to see if a different memory allocator will have better steady state memory consumption.
Jemalloc doesn't play well with Alpine, so I chose to experiment with libmimalloc instead.

### Tested

I'm switching all of my own full nodes over to this build to see its impact over time.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Change is fully backwards compatible. the LD_PRELOAD env variable can be set to empty to disable libmimalloc as needed.


Dockerfile.mimalloc
```diff
@@ -0,0 +1,57 @@
...
+RUN cd /go-ethereum && make geth-musl
+
+# Pull Geth into a second stage deploy alpine container
+FROM alpine:latest
+ARG COMMIT_SHA
+
+# Copy over compiled libmimalloc
+COPY --from=builder /mimalloc/build/*.so.* /lib
+RUN ln -s /lib/libmimalloc.so.* /lib/libmimalloc.so
+ENV LD_PRELOAD=/lib/libmimalloc.so
```
diwu1989, line 41: if this `LD_PRELOAD` env variable is not set, then geth will still run with the default musl memory allocator

carterqw2:
@diwu1989 Could you share any observations on any patterns you observe now when more than a month since the switch has passed?

PR #1837, opened 2022-02-08
Fix dynamicTxs gasPrice for rpc call

### LINKED ISSUE #1836: newRPCTransaction wrongly sets gasPrice to nil for CeloDynamicFeeTxType
On discord, jennyhys reported:

```
jennyhys: Hello we have a node with --syncmode full and --gcmode archive setting and it's fully caught up, yet when we called the eth_getTransactionByHash endpoint on some TX the gasPrice is showed as  null (other field showed the correct value). It doesn’t seem to be effecting all TXs but some, would it be related to the above pruning settings?
```

```
curl -s -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "eth_getTransactionByHash", "params": ["0x7707c458e89295d2127d30df3cfa33539dae734b9b1bb809eca28d7452d94df7"]}' localhost:8545
{"jsonrpc":"2.0","id":1,"result":{"blockHash":"0xd958198c8203cac59e8bac8b4a6e5e4ad25a6dab5a413ab431524872dbc0e6f8","blockNumber":"0x92238a","from":"0xa97a2bdb364ee277d12f8d90708da904a00ca64a","gas":"0x18204","gasPrice":null,"maxFeePerGas":"0x1dcd6500","maxPriorityFeePerGas":"0x1dcd6500","feeCurrency":null,"gatewayFeeRecipient":null,"gatewayFee":"0x0","hash":"0x7707c458e89295d2127d30df3cfa33539dae734b9b1bb809eca28d7452d94df7","input":"0x095ea7b30000000000000000000000007e20f0d08b9ba26b6a609dcc6b97e7b7e3754e46ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff","nonce":"0x8","to":"0xeecba7ab0a29dbf1869d97cff8e2a0c90905a6b7","transactionIndex":"0x1","value":"0x0","type":"0x2","accessList":[],"chainId":"0xaef3","v":"0x0","r":"0xa66c539d23657e8b698e57e0e9101de0e8257b3cca02dd2fb6a1d14c748a82c","s":"0x4c183a2e5719abaf44c02d9a5e1fc04be110130514d8748a3934206da72907ed","ethCompatible":false}}
```

A brief look into `internal/ethapi/api.go` points to a merging issue in the code:

```
			if err != nil {
				// price = min(tip, gasFeeCap - baseFee) + baseFee
				price := math.BigMin(new(big.Int).Add(tx.GasTipCap(), baseFee), tx.GasFeeCap())
				result.GasPrice = (*hexutil.Big)(price)
			} else {
				// error != nil for DinamicFees implies that there is no state to retrieve the baseFee for that block
				result.GasPrice = nil
			}
```

The if clause is probably inverted, but may need confirming. Seems to be present in 1.5.x release branch and also master.

gastonponti:
### Description

Fix gas price calculation for dynamic txs. If clause inverted

### Related issues

- Fixes #1836 

### Backwards compatibility

Yes

### DIFF
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -239,7 +239,7 @@ func TestRPCDynamicTxGasPriceWithoutState(t *testing.T) {
 	network, shutdown, err := test.NewNetwork(ac, gc, ec)
 	require.NoError(t, err)
 	defer shutdown()
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*300)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
 	defer cancel()
 
 	accounts := test.Accounts(ac.DeveloperAccounts(), gc.ChainConfig())
@@ -261,9 +261,11 @@ func TestRPCDynamicTxGasPriceWithoutState(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, json.BlockNumber)
 
-	// Wait until the state is prunned
-	err = network.AwaitBlock(ctx, json.BlockNumber.ToInt().Uint64()+3000)
-	require.NoError(t, err)
+	// Wait until the state is prunned. For this we create blocks with at least 1 tx
+	for i := 0; i < 200; i++ {
+		_, err := accounts[0].SendCeloTracked(ctx, accounts[1].Address, 1, network[0])
+		require.NoError(t, err)
+	}
 
 	var json2 *rpcCustomTransaction
 	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &json2, "eth_getTransactionByHash", tx.Hash())
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -200,6 +200,7 @@ type rpcCustomTransaction struct {
 func TestRPCDynamicTxGasPriceWithState(t *testing.T) {
 	ac := test.AccountConfig(3, 2)
 	gc, ec, err := test.BuildConfig(ac)
+	ec.TrieDirtyCache = 5
 	require.NoError(t, err)
 	network, shutdown, err := test.NewNetwork(ac, gc, ec)
 	require.NoError(t, err)
@@ -233,8 +234,9 @@ func TestRPCDynamicTxGasPriceWithState(t *testing.T) {
 }
 
 func TestRPCDynamicTxGasPriceWithoutState(t *testing.T) {
-	ac := test.AccountConfig(3, 2)
+	ac := test.AccountConfig(3, 3)
 	gc, ec, err := test.BuildConfig(ac)
+	ec.TrieDirtyCache = 5
 	require.NoError(t, err)
 	network, shutdown, err := test.NewNetwork(ac, gc, ec)
 	require.NoError(t, err)
@@ -270,6 +272,7 @@ func TestRPCDynamicTxGasPriceWithoutState(t *testing.T) {
 	var json2 *rpcCustomTransaction
 	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &json2, "eth_getTransactionByHash", tx.Hash())
 	require.NoError(t, err)
+	require.NotNil(t, json2)
 	require.NotNil(t, json2.BlockNumber)
 
 	require.Nil(t, json2.GasPrice)
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -197,10 +197,12 @@ type rpcCustomTransaction struct {
 	GasPrice    *hexutil.Big `json:"gasPrice"`
 }
 
-func TestRPCDynamicTxGasPriceWithState(t *testing.T) {
+// TestRPCDynamicTxGasPriceWithBigFeeCap test that after a dynamic tx
+// was added to a block, the rpc sends in the gasPrice the actual consumed
+// price by the tx, which could be less than the feeCap (as in this example)
+func TestRPCDynamicTxGasPriceWithBigFeeCap(t *testing.T) {
 	ac := test.AccountConfig(3, 2)
 	gc, ec, err := test.BuildConfig(ac)
-	ec.TrieDirtyCache = 5
 	require.NoError(t, err)
 	network, shutdown, err := test.NewNetwork(ac, gc, ec)
 	require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 97731e2e82026bcb6d10e4e97285f08b388f8688
<details><summary>coverage:  47.0% of statements across all listed packages</summary><br>
<pre>
coverage:  68.9% of statements in consensus/istanbul/core
</pre>
CommentID: e104c95eb7
</details>

internal/ethapi/api.go
```diff
@@ -1254,12 +1254,14 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 		// if the transaction has been mined, compute the effective gas price
 		if blockHash != (common.Hash{}) {
 			baseFee, err := baseFeeFn(tx.FeeCurrency())
-			if err != nil {
+			baseFeeFn(tx.FeeCurrency())
```
hbandura, line 1257: probably delete this line?

internal/ethapi/api_test.go
```diff
@@ -0,0 +1,54 @@
...
+	// Test unminned transaction. GasPrice == GasFeeCap
+	rpcTx3 := newRPCTransaction(types.NewTx(&types.CeloDynamicFeeTx{
+		FeeCurrency: &currency,
+		ChainID:     big.NewInt(1234567),
+
+		GasFeeCap: big.NewInt(1999001),
+		GasTipCap: big.NewInt(888400),
+	}), common.Hash{}, 0, 0, baseFeeFn)
+	assert.Equal(t, (*hexutil.Big)(big.NewInt(1999001)), rpcTx3.GasPrice)
+}
```
piersy, line 54: I think it would be good to add a test case where the baseFeeFn returns an error

internal/ethapi/api_test.go
```diff
@@ -0,0 +1,54 @@
...
+	// Test GasPrice == GasFeeCap
+	rpcTx := newRPCTransaction(types.NewTx(&types.CeloDynamicFeeTx{
+		FeeCurrency: &currency,
+		ChainID:     big.NewInt(1234567),
+
+		GasFeeCap: big.NewInt(111222),
+		GasTipCap: big.NewInt(888400),
+	}), blockHash, blockNumber, index, baseFeeFn)
+	assert.Equal(t, (*hexutil.Big)(big.NewInt(111222)), rpcTx.GasPrice)
+	// Test GasPrice == GasTipCap + baseFee
```
piersy, line 36: Maybe a bit nitpicky but I would find this much easier to read if there was a blank line between each test case

internal/ethapi/api_test.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		return nil, errors.New("unexpected")
+	}
+	// Test GasPrice == GasFeeCap
+	rpcTx := newRPCTransaction(types.NewTx(&types.CeloDynamicFeeTx{
+		FeeCurrency: &currency,
+		ChainID:     big.NewInt(1234567),
+
+		GasFeeCap: big.NewInt(111222),
+		GasTipCap: big.NewInt(888400),
```
YoduYodu, line 33: Maybe use variables/consts instead of indicate them in each test cases repeatedly?

internal/ethapi/api_test.go
gastonponti: two e2e tests added to check this 😉

nategraf (approved):
Looks good to me. Hopefully @piersy can also take a look to see if his concerns are addressed. Thanks for adding tests

hbandura (approved):

e2e_test/e2e_test.go
```diff
@@ -189,3 +191,89 @@ func TestStartStopValidators(t *testing.T) {
...
+	tx, err := accounts[0].SendCeloWithDynamicFee(ctx, accounts[1].Address, 1, gasFeeCap, gasTipCap, network[0])
+	require.NoError(t, err)
+
+	// Wait for the whole network to process the transaction.
+	err = network.AwaitTransactions(ctx, tx)
+	require.NoError(t, err)
+	var json *rpcCustomTransaction
+	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &json, "eth_getTransactionByHash", tx.Hash())
+	require.NoError(t, err)
+	require.NotNil(t, json.BlockNumber)
```
piersy, line 264: This seems unnecessary, once the network has processed the transaction you know its been executed, no need to fetch it from the node and look at the block number.

e2e_test/e2e_test.go
```diff
@@ -189,3 +191,89 @@ func TestStartStopValidators(t *testing.T) {
...
+	gasTipCap := big.NewInt(0).Mul(suggestedGasPrice, big.NewInt(2))
+
+	// Send one celo from external account 0 to 1 via node 0.
+	tx, err := accounts[0].SendCeloWithDynamicFee(ctx, accounts[1].Address, 1, gasFeeCap, gasTipCap, network[0])
+	require.NoError(t, err)
+
+	// Wait for the whole network to process the transaction.
+	err = network.AwaitTransactions(ctx, tx)
+	require.NoError(t, err)
+	var json *rpcCustomTransaction
```
piersy, line 261:
```suggestion
	// Check that the transaction can be retrieved via the rpc api
	var json *rpcCustomTransaction
```

e2e_test/e2e_test.go
```diff
@@ -189,3 +191,89 @@ func TestStartStopValidators(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotNil(t, json.BlockNumber)
+
+	// Wait until the state is prunned. For this we create blocks with at least 1 tx
+	for i := 0; i < 200; i++ {
+		_, err := accounts[0].SendCeloTracked(ctx, accounts[1].Address, 1, network[0])
+		require.NoError(t, err)
+	}
+
+	var json2 *rpcCustomTransaction
```
piersy, line 272:
```suggestion
	// Check that the transaction can still be retrieved via the rpc api
	var json2 *rpcCustomTransaction
```

e2e_test/e2e_test.go
piersy, line 264: I'm wondering why you are checking block number here? Given the name of the test I would expect it to check that the gas price is non nil and maybe also not zero.

e2e_test/e2e_test.go
```diff
@@ -189,3 +191,89 @@ func TestStartStopValidators(t *testing.T) {
...
+	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &json2, "eth_getTransactionByHash", tx.Hash())
+	require.NoError(t, err)
+	require.NotNil(t, json2)
+	require.NotNil(t, json2.BlockNumber)
```
piersy, line 276: Again wondering why you are checking block number?

e2e_test/e2e_test.go
```diff
@@ -189,3 +191,89 @@ func TestStartStopValidators(t *testing.T) {
 	require.NoError(t, err)
 
 }
+
+type rpcCustomTransaction struct {
+	BlockNumber *hexutil.Big `json:"blockNumber"`
+	GasPrice    *hexutil.Big `json:"gasPrice"`
+}
+
+func TestRPCDynamicTxGasPriceWithState(t *testing.T) {
```
piersy, line 200: I think we can drop this test because the the other test is practically already performing this test (except for checking the gasprice with respect to the tip and cap)

e2e_test/e2e_test.go
```diff
@@ -189,3 +191,89 @@ func TestStartStopValidators(t *testing.T) {
...
+	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &json, "eth_getTransactionByHash", tx.Hash())
+	require.NoError(t, err)
+	require.NotNil(t, json.BlockNumber)
+	gasPrice := json.GasPrice.ToInt()
+	require.NotNil(t, json.GasPrice)
+	require.Greater(t, gasPrice.Int64(), gasTipCap.Int64())
+	require.Less(t, gasPrice.Int64(), gasFeeCap.Int64())
+}
+
+func TestRPCDynamicTxGasPriceWithoutState(t *testing.T) {
```
piersy, line 236: I think it would be helpful to give an overview here of what the test is trying to check. Without the context of the bug this is trying to fix its not particularly clear what is being checked.

e2e_test/e2e_test.go
gastonponti: Blocknumber is my way to stay that it was added to a block, remember that I'm using a custom way because the ethclient is not serializing things as we expect. The actual `TransactionByHash` from the go ethclient, returns a `isPending` that's just a `blockNumber == nil`

e2e_test/e2e_test.go
gastonponti: I will keep it, but change the logic to assume it is an archive node, asking both, at the beginning and after N blocks. That way we will cover that case too

e2e_test/e2e_test.go
piersy: The call to AwaitTransactions ensures that all the requested transactions were added to blocks, so you shouldn't need to do this.

mcortesi (approved):

PR #1840, opened 2022-02-12
Improve miner to recover from randomness cache miss on initialization

nategraf:
### Description

In the course of getting Baklava into condition to restart block production, I noticed that the
miner would fail to initialize on nodes that had replaced there chain data from backup. The observed
error was that the randomness commitment cache would fail to be populated, which is caused by
replacing the chain data on that node. This error is supposed to be recoverable, but no code
existed on the code path that was failing to do so. Recovery would only occur after syncing to the
network, which these nodes did not do.

This PR adds a conditional to try recovering the randomness cache once if there is a cache miss
during miner initialization.

### Other changes

None

### Tested

Ran these changes on cLabs Baklava validators and got to a correctly initialized state

### DIFF
--- a/miner/block.go
+++ b/miner/block.go
@@ -128,7 +128,19 @@ func prepareBlock(w *worker) (*blockState, error) {
 		if (lastCommitment != common.Hash{}) {
 			lastRandomnessParentHash := rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
 			if (lastRandomnessParentHash == common.Hash{}) {
-				return b, errors.New("Failed to get last randomness cache entry")
+				log.Warn("Randomness cache miss while building a block. Attempting to recover.")
+
+				// We missed on the cache which should have been populated, attempt to repopulate the cache.
+				err := w.chain.RecoverRandomnessCache(lastCommitment, b.header.ParentHash)
+				if err != nil {
+					log.Error("Error in recovering randomness cache", "error", err)
+					return b, errors.New("Failed to to recover the randomness cache after miss")
+				}
+				lastRandomnessParentHash = rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
+			}
+			if (lastRandomnessParentHash == common.Hash{}) {
+				// Recover failed to fix the issue. Bail.
+				return b, errors.New("Failed to get last randomness cache entry and failed to recover")
 			}
 
 			var err error
--- a/miner/block.go
+++ b/miner/block.go
@@ -134,12 +134,12 @@ func prepareBlock(w *worker) (*blockState, error) {
 				err := w.chain.RecoverRandomnessCache(lastCommitment, b.header.ParentHash)
 				if err != nil {
 					log.Error("Error in recovering randomness cache", "error", err, "number", header.Number.Uint64())
-					return b, errors.New("Failed to to recover the randomness cache after miss")
+					return b, errors.New("failed to to recover the randomness cache after miss")
 				}
 				lastRandomnessParentHash = rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
 				if (lastRandomnessParentHash == common.Hash{}) {
 					// Recover failed to fix the issue. Bail.
-					return b, errors.New("Failed to get last randomness cache entry and failed to recover")
+					return b, errors.New("failed to get last randomness cache entry and failed to recover")
 				}
 			}
 
--- a/miner/block.go
+++ b/miner/block.go
@@ -128,12 +128,12 @@ func prepareBlock(w *worker) (*blockState, error) {
 		if (lastCommitment != common.Hash{}) {
 			lastRandomnessParentHash := rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
 			if (lastRandomnessParentHash == common.Hash{}) {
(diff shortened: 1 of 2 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit aac396bef0a864f6025218f9fb870192ccc9faa0
<pre>
coverage:  45.1% of statements in consensus/istanbul/backend
coverage:  64.0% of statements in consensus/istanbul/core
</pre>
CommentID: 633e7e20f1
</details>

hbandura (dismissed):

miner/block.go
```diff
@@ -128,7 +128,19 @@ func prepareBlock(w *worker) (*blockState, error) {
...
 			lastRandomnessParentHash := rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
 			if (lastRandomnessParentHash == common.Hash{}) {
-				return b, errors.New("Failed to get last randomness cache entry")
+				log.Warn("Randomness cache miss while building a block. Attempting to recover.")
+
+				// We missed on the cache which should have been populated, attempt to repopulate the cache.
+				err := w.chain.RecoverRandomnessCache(lastCommitment, b.header.ParentHash)
+				if err != nil {
+					log.Error("Error in recovering randomness cache", "error", err)
+					return b, errors.New("Failed to to recover the randomness cache after miss")
```
piersy, line 137: Callers of `prepareBlock` will log the error returned, so I think it would be less noisy to just return the error here.

Also it would be nice to add a bit of context to the error such as the block number that was being prepared.
```suggestion
					return b, fmt.Errorf("Failed to to recover the randomness cache for block %d after miss: %w", header.Number.Uint64(), err)
```

miner/block.go
```diff
@@ -128,7 +128,19 @@ func prepareBlock(w *worker) (*blockState, error) {
 		if (lastCommitment != common.Hash{}) {
 			lastRandomnessParentHash := rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
 			if (lastRandomnessParentHash == common.Hash{}) {
-				return b, errors.New("Failed to get last randomness cache entry")
+				log.Warn("Randomness cache miss while building a block. Attempting to recover.")
```
piersy, line 131:
```suggestion
				log.Warn(fmt.Sprintf("Randomness cache miss while building block %d. Attempting to recover.", header.Number.Uint64()))
```

YoduYodu (approved):

miner/block.go
nategraf: I think we should keep the log line because that failure in and of itself is a notable error. It's also OK for it to me more noisy because this is an error condition that prevents the miner from producing new blocks, which is quite severe.

miner/block.go
nategraf: I've added the block number to this log line in 262bf2cc3508cde195fd8585ba7c6542b2b6d72d. I used the variable length args method that is supported by this logging package.

miner/block.go
```diff
@@ -128,7 +128,19 @@ func prepareBlock(w *worker) (*blockState, error) {
...
+				// We missed on the cache which should have been populated, attempt to repopulate the cache.
+				err := w.chain.RecoverRandomnessCache(lastCommitment, b.header.ParentHash)
+				if err != nil {
+					log.Error("Error in recovering randomness cache", "error", err)
+					return b, errors.New("Failed to to recover the randomness cache after miss")
+				}
+				lastRandomnessParentHash = rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
+				if (lastRandomnessParentHash == common.Hash{}) {
+					// Recover failed to fix the issue. Bail.
+					return b, errors.New("Failed to get last randomness cache entry and failed to recover")
```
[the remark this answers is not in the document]
nategraf: Right, that's intentional. error messages in the Geth codebase are generally static. Instead that info is included in the log message

miner/block.go
```diff
@@ -128,7 +128,19 @@ func prepareBlock(w *worker) (*blockState, error) {
 		if (lastCommitment != common.Hash{}) {
 			lastRandomnessParentHash := rawdb.ReadRandomCommitmentCache(w.db, lastCommitment)
 			if (lastRandomnessParentHash == common.Hash{}) {
-				return b, errors.New("Failed to get last randomness cache entry")
+				log.Warn("Randomness cache miss while building a block. Attempting to recover.", "number", header.Number.Uint64())
+
+				// We missed on the cache which should have been populated, attempt to repopulate the cache.
+				err := w.chain.RecoverRandomnessCache(lastCommitment, b.header.ParentHash)
+				if err != nil {
+					log.Error("Error in recovering randomness cache", "error", err, "number", header.Number.Uint64())
```
mcortesi, line 136: mind that error messages are to be lowercase by golang convention. we are probably not following too much anyway

mcortesi (approved):

miner/block.go
nategraf: Made an edit to that 9e21aa8cc0a6e2a54dbc039da44a85ed1b4efa97

mcortesi (approved):

miner/block.go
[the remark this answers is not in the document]
piersy: There isn't a log message here though. If this path gets taken you will just see the error message.

Issue #1856, opened 2022-02-24
Gas estimation is unreliable in release v1.5.3 with espresso enabled


piersy:
## Description

Gas estimation is unreliable in release v1.5.3 with espresso enabled.

Calling estimate gas on a non validating node can sometimes return a gas estimate that is too low. When run against alfajores [these tests](https://github.com/dacadeorg/celo-nft-minter/blob/master/test/nft-test.js) are often failing due to the gas estimates being too low. This only seems to happen when calling estimate gas on non validating nodes and is a problem that manifests with the Espresso hard fork.

The specific case where the gas estimate seems to be too low is when a contract transaction is estimated against a pending block that contains the contract deployment. In such cases the estimation comes out slightly low.

This issue is exacerbated by another problem which is that the pending block is not updated in a timely fashion. Pending blocks should represent the next block to insert, but currently the pending block is update is delayed by the length of time of a block, so that the pending block for the most part shares the same number as the most recently inserted block.

PR #1857, opened 2022-02-24
Only sleep during prepare if node is validating

piersy:
### Description

This change mitigates the problem of low gas estimates #1856, by ensuring that non validating nodes do not wait till the next block is due when engine.Prepare is called. This results in non validating nodes having an up to date pending block( I.E a pending block with a number 1 greater than the last inserted block). For some reason validating nodes produce the correct gas estimate even when they have an old pending block (I.E a pending block with the same number as the last inserted block).

### Tested

I have tested this locally using mycelo to start a network with 1 validator and then starting up a full node that connects to that network. Without the fix the tests [here](https://github.com/piersy/celo-nft-minter/blob/ft/update-celo-depency/test/nft-test.js) repeatedly show a low gas estimate for the first few gas estimation calls made against the pending block. With the fix I have not been able to elicit an incorrect gas estimate (although it is definitely possible because block insertion and pending block creation are not part of a single atomic operation).  Prior to the fix I observed from the logs that gas estimates using the latest and pending block would execute against a block with the same number, after the fix I observed from the logs that gas estimates with the pending block executed against a block with number one greater than gas estimates with the latest block.

### Related issues

-Mitigates #1856

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -413,12 +413,24 @@ func (sb *Backend) Prepare(chain consensus.ChainHeaderReader, header *types.Head
 	// TODO(victor): Sleep here was previously removed and added to the miner instead, that change
 	// has been temporarily reverted until it can be reimplemented without causing fewer signatures
 	// to be included by the block producer.
-	delay := time.Until(time.Unix(int64(header.Time), 0))
-	if delay < 0 {
-		sb.sleepGauge.Update(0)
-	} else {
-		sb.sleepGauge.Update(delay.Nanoseconds())
-		time.Sleep(delay)
+	// (piers): Victor's re-addition of the sleep here caused the problem that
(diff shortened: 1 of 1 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 3ded159bc7cc6f2cea9bee28265a3631b833ec62
<details><summary>coverage:  51.1% of statements across all listed packages</summary><br>
<pre>
coverage:  64.1% of statements in consensus/istanbul
coverage:  51.3% of statements in consensus/istanbul/backend
coverage:  45.0% of statements in consensus/istanbul/backend/internal/db
coverage:  42.8% of statements in consensus/istanbul/backend/internal/enodes
coverage:  71.4% of statements in consensus/istanbul/core
</pre>
CommentID: 91e403c12a
</details>

piersy:
I think I have a fix for this problem now

PR #1858, opened 2022-02-25
Ensure pending state is ready for tx execution

### LINKED ISSUE #1856: Gas estimation is unreliable in release v1.5.3 with espresso enabled

piersy:
What this PR does is better described [here](https://github.com/celo-org/celo-blockchain/pull/1858#issuecomment-1054159493)

Fixes #1856 

Before introducing this change gas estimations against the pending block on non validating nodes with the espresso fork enabled would sometimes be lower than they should be.

Its still not clear to me exactly how the lower gas estimates were being calculated but ensuring that the pending block is finalized and assembled seems to resolve this. Validating nodes were already finalizing and assembling their pending blocks and looking back it seems that prior to #1545 the pending block was always finalized and assembled (see permalink below). So this seems like this change is correcting a long running problem. https://github.com/celo-org/celo-blockchain/blob/ab78658ae3a738d91d877f289e37fead2a0f1cf7/miner/worker.go#L969

This change also changes the behaviour of the debug tracing apis when used run against the pending block.

Since #1545 caused the pending block to not be finalized, the state root in the pending block was zero. Strangely, looking up the zero root in  the state trie doesn't fail but the state returned is essentially empty, meaning that all transactions would be traced as a simple send ( because that is how a contract transaction executes if the contract does not exist).

With this change, calls to the tracing apis with the pending block will return the error `required historical state unavailable`.

This is because the pending block will have been finalized and will therefore have a state root, and the tracing apis first look up the block and then try to get the state for it using `Ethereum.StateAtBlock` which doesn't take into account the pending state and therefore the state cannot be found.

### Tested

I have verified that the tests linked in #1856 produce reliable gas estimates running against a local node connected to a 1 validator network run with mycelo and also on a fullnode deployed in alfajores.

### Backwards compatibility

The results of the rpc apis are chainging.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -171,7 +171,19 @@ func (w *worker) pending() (*types.Block, *state.StateDB) {
 	stateCopy := w.snapshotState.Copy()
 	// Call Prepare to ensure that any access logs from the last executed
 	// transaction have been erased.
-	// See https://github.com/celo-org/celo-blockchain/pull/1858#issuecomment-1054159493
+	//
+	// Prior to the upstream PR
+	// https://github.com/ethereum/go-ethereum/pull/21509 the state returned
+	// from pending was ready to use for transaction execution, that PR
+	// essentially changed the contract of the pendng method, in that the
+	// returned state was not ready for transaction execution and required
+	// Prepare to be called on it first, but notably the PR did not update any
+	// of the callers of pending to ensure that Prepare was called. I think
+	// this broke some of the eth rpc apis.  Calling Prepare here essentially
+	// restores the previous contract for this method which was that the
+	// returned state is ready to use for transaction execution.
+	//
+	// See https://github.com/celo-org/celo-blockchain/pull/1858#issuecomment-1054159493 for more details.
 	stateCopy.Prepare(common.Hash{}, 0)
 	return w.snapshotBlock, stateCopy
 }
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -168,7 +168,12 @@ func (w *worker) pending() (*types.Block, *state.StateDB) {
 	if w.snapshotState == nil {
 		return nil, nil
 	}
-	return w.snapshotBlock, w.snapshotState.Copy()
+	stateCopy := w.snapshotState.Copy()
+	// Call Prepare to ensure that any access logs from the last executed
+	// transaction have been erased.
+	// See https://github.com/celo-org/celo-blockchain/pull/1858#issuecomment-1054159493
+	stateCopy.Prepare(common.Hash{}, 0)
+	return w.snapshotBlock, stateCopy
 }
 
 // pendingBlock returns pending block.


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 6ad68f2566df101b726c022d0d2d966b98b5ebd1
<details><summary>coverage:  52.4% of statements across all listed packages</summary><br>
<pre>
coverage:  64.9% of statements in consensus/istanbul
coverage:  53.0% of statements in consensus/istanbul/backend
coverage:  50.0% of statements in consensus/istanbul/backend/internal/db
coverage:  46.3% of statements in consensus/istanbul/backend/internal/enodes
coverage:  72.5% of statements in consensus/istanbul/core
</pre>
CommentID: cc27fd229a
</details>

nategraf:
Nice detective work so far!

I don't think calling `finalizeAndAssemble` before the last transaction is applied results in well-defined behavior. It ends up mutating the state DB held by the working in a number of ways, such as distributing epoch rewards, and those transactions are intended to be the last step in assembling a block. Applying additional transactions on top of that will result in behavior that is divergent from what would happen during on-chain execution of a block with the same contents (e.g. a transaction running after the GPM has changed by be valid while the real execution will end up rejecting the transaction because the GPM changes at the _end_ of the block). More critically, this code seems to result in calling finalization multiple times on the same state DB, which is not intended.

https://github.com/celo-org/celo-blockchain/blob/6ad68f2566df101b726c022d0d2d966b98b5ebd1/consensus/istanbul/backend/engine.go#L491

I'm guessing that the key in this actually the following line:

https://github.com/celo-org/celo-blockchain/blob/6ad68f2566df101b726c022d0d2d966b98b5ebd1/consensus/istanbul/backend/engine.go#L529

Or possibly the call to `state.Prepare`:

https://github.com/celo-org/celo-blockchain/blob/6ad68f2566df101b726c022d0d2d966b98b5ebd1/consensus/istanbul/backend/engine.go#L503

Both of these mutate the state DB, and the first of the two modifies the header as well.

piersy:
Thank you @nategraf.

I decided to look at the TraceBlock api, because trace block is tracing each tx in the block and has to prepare some intermediate state for each tx such that the tx traces correctly. And that is I think what we need to achieve here.

Below is the code block that prepares a state for tracing a tx.
https://github.com/celo-org/celo-blockchain/blob/6ad68f2566df101b726c022d0d2d966b98b5ebd1/eth/tracers/api.go#L561-L576

There seem to be 2 things that are done to the state in preparation for tracing.
* A call to `statedb.Prepare(txHash, txIndex)`
* A call to `statedb.Finalise(isEIP158)`

The prepare implementation is below, and we can see that it creates an `accessLIst` which was added as part of [EIP-2929 (gas cost increases for state access opcodes)](https://github.com/ethereum/go-ethereum/pull/21509). The summary for this eip is:
>Increases gas cost for SLOAD, *CALL, BALANCE, EXT* and SELFDESTRUCT when used for the first time in a transaction.

So if `Prepare` is not called between transaction executions you would expect to see the second transaction pay less gas than it should need to, if it accessed some of the same state as the prior transaction. This seems very much like the problem we are facing.
https://github.com/celo-org/celo-blockchain/blob/6ad68f2566df101b726c022d0d2d966b98b5ebd1/core/state/statedb.go#L884-L888

The `Finalise` implementation is below and it is significantly more complicated than `Prepare` it seems to be doing some sort of internal maintenance on the state representation, but the last line seems significant, the comment states:
>Invalidate journal because reverting across transactions is not allowed.

Based on that it would seem that failing to call `Finalise` between transactions could result in a revert reverting the changes from prior transactions, so something to bear in mind if executing multiple transactions across a state instance.

https://github.com/celo-org/celo-blockchain/blob/6ad68f2566df101b726c022d0d2d966b98b5ebd1/core/state/statedb.go#L775-L819

Looking into the miner code that manages the pending block I can see that both `Prepare` and `Finalise` are being called, `Prepare` before executing a tx and `Finalise` after. So I think al that is required for us is to ensure that we call prepare on the state returned from the worker's `pending` method before beginning gas estimation. We shouldn't need to worry about calling finalise since the worker returns a copy of the state and estimate gas executes just one transaction against it.

nategraf (approved):

PR #1865, opened 2022-03-10
Gas Inflation Rate flag

gastonponti:
### Description

Adding a gas inflation rate flag, to multiply the gas estimations from the rpc call.
This aims to mitigate the gas estimation bug until we found the actual fix for it

### Tested

Run tests

### Backwards compatibility

Yes

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -141,6 +141,10 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	if err := pruner.RecoverPruning(stack.ResolvePath(""), chainDb, stack.ResolvePath(config.TrieCleanCacheJournal)); err != nil {
 		log.Error("Failed to recover state", "error", err)
 	}
+	if config.RPCGasInflationRate == 0 {
+		// if it was not set, default it as 1
+		config.RPCGasInflationRate = 1
+	}
 	eth := &Ethereum{
 		config:            config,
 		chainDb:           chainDb,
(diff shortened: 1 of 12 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 5d52d99dc03a3bbaae6e2e674d50afa20381479e
<details><summary>coverage:  50.9% of statements across all listed packages</summary><br>
<pre>
coverage:  64.9% of statements in consensus/istanbul
coverage:  40.5% of statements in consensus/istanbul/announce
coverage:  54.8% of statements in consensus/istanbul/backend
coverage:  70.3% of statements in consensus/istanbul/core
</pre>
CommentID: 6ab7ba288e
</details>

eth/backend.go
```diff
@@ -141,6 +141,10 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	if err := pruner.RecoverPruning(stack.ResolvePath(""), chainDb, stack.ResolvePath(config.TrieCleanCacheJournal)); err != nil {
 		log.Error("Failed to recover state", "error", err)
 	}
+	if config.RPCGasInflationRate == 0 {
+		// if it was not set, default it as 1
+		config.RPCGasInflationRate = 1
```
mcortesi, line 146: does this mean that it's 1 by default or 1.3 as below?

mcortesi (approved):

eth/backend.go
gastonponti: it's 1.3 by default because we set all the defaults with flags. This is more a net just in case someone builds a node object by their own. In that case, as the defaults are only set as part of reading flags, the rate will be in 0 which is not good at all.
I didn't want to change how we handle the defaults, just made the same as others, and found this possible scenario

PR #1876, opened 2022-03-24
Don't use log file in sync_test.sh

piersy:
Instead of putting the node logs into a file, the lightest sync test now prints them to stdout, this makes it possible to see the logs on CI


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 244c90f7a9187b077f1166ad1bf729e077e5b4d8
<details><summary>coverage:  50.5% of statements across all listed packages</summary><br>
<pre>
coverage:  64.7% of statements in consensus/istanbul
coverage:  40.4% of statements in consensus/istanbul/announce
coverage:  54.8% of statements in consensus/istanbul/backend
coverage:  69.2% of statements in consensus/istanbul/core
</pre>
CommentID: b8b3dd4df1
</details>

piersy:
> Isn't the log file ever useful? Shouldn't we just tee the output to both stdout and the file?

In CI it's not useful because the log file is lost, thats the only usage I can see for this script.

In general if users are interested in the output then printing it to stdout seems to be a better option than putting it in a log, because it saves them having to read the script to find out where the output went and then having to open the output. If users really want a log they can easily redirect from the script (e.g. `> log`) which I think is nicer for them because they are now explicitly aware of where the log is.

Maybe I'm missing some context, but I don't see any reason to make a log the default here.

piersy:
@hbandura Any chance of getting your review on this

palango (approved):
Looks good to me.

PR #1900, opened 2022-05-18
Ensure copy of state used to generate block context

piersy:
Instead of providing an `EVMRunner` to `core.NewSysContractCallCtx` we now provide the header, state and an evm runner factory. This allows us to ensure that the state is copied before passing it to the evm runner, thus ensuring that the given state is not modified. This also encapsulates that responsibility so that callers do not have to be aware of it.

This solves our gas estimation problem, that was being caused by executing multiple transactions/call with the same access list, specifically in this case executing all the calls for `core.NewSysContractCallCtx` and then estimating gas for a transaction without first clearing the access list.

We could have instead ensured that we called `StateDB.Prepare` on the state after calling `core.NewSysContractCallCtx` but copying the state seemed neater because: 

1. `core.NewSysContractCallCtx` is intended to be read only, so making it copy state aligns well with it conceptually.
2. `StateDB.Prepare` should be called just prior to executing a transaction and additionally to clearing the access list also contains functionality to allow retrieval of transaction logs for a specific transaction. So if we wanted to use this correctly we would need to call it not immediately after calling, or inside `core.NewSysContractCallCtx`, but instead before each tx executed with that system contract call context. This means we cannot neatly encapsulate calls to `StateDB.Prepare` in `core.NewSysContractCallCtx`.

Renamed `VmRunnerAtHeader` to `NewEVMRunner` since both methods did the same thing and in some cases both methods had been implemented on structs.

Updated godoc for NewSysContractCallCtx.

### Tested

Manually checked that the the state root of the state used to execute `core.NewSysContractCallCtx` is not modified by `core.NewSysContractCallCtx` meaning we are safe to treat `core.NewSysContractCallCtx` as a read only operation and therefore it is safe to let it execute on a copy of the state.

Manually checked that gas estimation was correct for tx 0x5fb4082c0c92330c38ccd0fc64f92384c7161f8670b27972b7895a46e5710063 after making this change. Used a lightest client connected to an archive node to verify this.

Also checked that the same results were obtained for 0x5fb4082c0c92330c38ccd0fc64f92384c7161f8670b27972b7895a46e5710063 if calling `StateDB.Prepare` as opposed to copying the state.

### Backwards compatibility

There were two approaches that we could have used to solve this.
1. Calling `StateDB.Prepare` before tx execution.
2. Using a state copy for building the system contract call context.

Both approaches were being used prior to this PR in the critical path of the blockchain, one by the miner for creating a block and the other during block processing when applying the block. Given that we have not seen any problems so far, it seems reasonable to assume that both approaches result in the same chain state, so this should not result in a hardfork.

It is possible that this change would cause a difference in execution in all the other places that `StateDB.Prepare` was not being called before tx execution, the only place I can see that was not calling `StateDB.Prepare` before tx execution is in the state prefetcher, so we could expect the prefetched state to be more accurate than it previously was but the outcome of transaction processing should remain the same.

### DIFF
--- a/core/sys_context.go
+++ b/core/sys_context.go
@@ -7,6 +7,8 @@ import (
 	"github.com/celo-org/celo-blockchain/contracts/blockchain_parameters"
 	"github.com/celo-org/celo-blockchain/contracts/currency"
 	"github.com/celo-org/celo-blockchain/contracts/gasprice_minimum"
+	"github.com/celo-org/celo-blockchain/core/state"
+	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/core/vm"
 )
 
@@ -46,6 +48,16 @@ func NewSysContractCallCtx(vmRunner vm.EVMRunner) (sc *SysContractCallCtx) {
 	}
 
 	return
+}
+
+type runnerFactory interface {
+	NewEVMRunner(*types.Header, vm.StateDB) vm.EVMRunner
+}
+
+// NewSysContractCallCtx2 creates the SysContractCallCtx object and makes the contract calls.
+func NewSysContractCallCtx2(header *types.Header, state *state.StateDB, factory runnerFactory) (sc *SysContractCallCtx) {
+	vmRunner := factory.NewEVMRunner(header, state.Copy())
+	return NewSysContractCallCtx(vmRunner)
 }
 
 // GetIntrinsicGasForAlternativeFeeCurrency retrieves intrinsic gas for non-native fee currencies.
--- a/core/sys_context.go
+++ b/core/sys_context.go
@@ -22,8 +22,17 @@ type SysContractCallCtx struct {
 	gasPriceMinimums GasPriceMinimums
 }
 
-// NewSysContractCallCtx creates the SysContractCallCtx object and makes the contract calls.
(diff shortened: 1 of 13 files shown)


core/sys_context.go
```diff
@@ -7,27 +7,46 @@ import (
...
 	"github.com/celo-org/celo-blockchain/contracts/gasprice_minimum"
+	"github.com/celo-org/celo-blockchain/core/state"
+	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/core/vm"
 )
 
-// SysContractCallCtx represents a system contract call context for a given block (represented by vm.EVMRunner).
-// It MUST sit on the header.Root of a block, which is parent of the block we intend to deal with.
+// SysContractCallCtx acts as a cache holding information obtained through
+// system contract calls to be used during block processing.
```
YoduYodu, line 16: Maybe also worth mentioning the system contract calls are done based on starting of the block(or end of parent block)?

piersy: Hey @YoduYodu, I think I covered that with the constructor comment. See here - https://github.com/celo-org/celo-blockchain/pull/1900/files#diff-d591f9889dcb3c67b0f48fd0671a1a40ebaf6edf021c05b50f2a5698ae79fe24R34-R36

If not, can you elaborate a bit?

gastonponti (approved):
LGTM!
Great work around for the runner ;)

PR #1908, opened 2022-06-01
Upstream: "core/rawdb: freezer batch write"

hbandura:
Merging commit https://github.com/ethereum/go-ethereum/commit/794c6133efa2c7e8376d9d141c900ea541790bce#diff-d78b498ba524953c810b53b2f1e866245d052a063863ed820c9b4852f8762d69 from upstream into the 1.10.9 merge PR

It took me a while to make this one work. Had lots of issues with the total difficulty storage. I hacked it by forcing it to write number+1 in the batch write functions.

Opening this PR against the partial 1.10.9 PR since that one hasn't been merged yet.


core/blockchain.go
```diff
@@ -1204,120 +1183,116 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
...
-					size += rawdb.WriteAncientBlock(bc.db, b, rawdb.ReadReceipts(bc.db, h, 0, bc.chainConfig), rawdb.ReadTd(bc.db, h, 0))
-					log.Info("Wrote genesis to ancients")
+		first := blockChain[0]
+		last := blockChain[len(blockChain)-1]
+
+		// Ensure genesis is in ancients.
+		if first.NumberU64() == 1 {
+			if frozen, _ := bc.db.Ancients(); frozen == 0 {
+				b := bc.genesisBlock
+				writeSize, err := rawdb.WriteAncientBlocks(bc.db, []*types.Block{b}, []types.Receipts{nil}, big.NewInt(1))
```
piersy, line 1199: Given that `WriteAncientBlocks` is ignoring the td parameter I think we should just remove it from the parameter list, otherwise providing values here is quite confusing

core/blockchain.go
```diff
@@ -1204,120 +1183,116 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 		}
 		return false
 	}
+
```
piersy, line 1186: This PR makes significant changes to `InsertReceiptChain` but it is also missing a test that was included in the upstream PR `TestInsertReceiptChainRollback`. Looking at the results of coverage I can see that none of the error cases of `insertReceiptChain` are executed so I think it would be good to include a test here that exercises some of the error cases for `InsertReceiptChain`.

piersy:
Hey @hbandura, I left 2 comments. I think we could leave the removal of the redundant parameter to another PR, because making that change here would complicate understanding this merge.

Further testing of insert receipt chain could however be added without risk of conflict so I think that change should be made here, unless you don't think it should be tested.

Aside from that all the e2e tests on ci are failing.

Also why did you merge the original branch back into this branch? That's going to make for an overly complicated history?

core/blockchain.go
hbandura: The new test uses sidechains, which is why we can't run it. But the old test uses a test field/test mechanism "testInsert" to provoke the tests to be tested, which was removed in this refactor. The errors are tested using sidechains, which we don't have. 

Re-adding the mechanism to make it testeable was probably a change that should be made in a different PR for the same reason you mentioned the redundant parameter one

hbandura: https://github.com/celo-org/celo-blockchain/issues/1917

core/blockchain.go
piersy: @hbandura Is there an issue for this?

PR #1909, opened 2022-06-01
Faster CI

piersy:
This update reduces overall build time from about 22m to 15m, a 7m
saving or about a 32% improvement.

The time taken for the unit-tests job to complete has been dropped from
around 19m to about 4m, this should be beneficial to developers by
providing a much shorter feedback loop.

These changes were made possible by taking steps to minimise the amount
of data downloaded and shared between jobs, and most pertinently
removing usages of workspaces to persist and retrieve intermediate build
artefacts. And in place of that relying more on caches and modifying
jobs so that they are able to rebuild those caches. Greater use of
caches was enabled by better understanding of how to reuse portions of
config via yaml anchors.

### Other changes

Updated a couple of tests to prevent flakey failures.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,14 +1,111 @@
+# A note on performance.
+#
+# TL;DR: Storing and retrieving from workspaces is slow so use it sparingly.
+#
+# Workspaces are useful in sharing data between jobs where the overhead
+# involved in using a cache and providing functionality to rebuild the cache in
+# each job is too high.
+#
+# Long version:
+#
+# Storing and retrieving data from workspaces seems to be slow. And
+# caches/checkouts/plain downloads appear to be faster. Some numbers ...
+#
+####
+#
+# Restoring repo dir (no binaries) (65M) + gocache (243M) + gomodcache (1.5G)
+# from a workspace takes approx 121s
+#
+# Restoring repo dir (no binaries) (65M) + gocache (243M) from a workspace
+# takes approx 52s and then restoring the gomodcache (1.5G) from a cache takes
+# approx 12s, so 64s total, so a reduction of 57s which is almost half the
+# time!
+#
+####
+#
+# Restoring monorepo dir + celo-blockchain  + go install (684M) from workspace
+# approx 51s.
+#
+# Restoring monorepo dir + go install (659M) from workspace approx 37s +
+# shallow ceckout celo-monorepo 2s, so a total of 39s a 12 second improvement.
+#
+####
+#
+# Restoring monorepo dir + golang install from workspace (659M) takes
+# approx 35s.
+#
+# Downloading and installing go (109M) takes approx 4s and restoring monorepo
+# dir from (414M) takes approx 28s, so a total of 32s, so a 3s improvement.
+#
+#######
+#
+# As can be seen from the above numbers restoring from workspaces are
+# significantly slower than caches, a fair bit slower than github checkouts and
+# a little bit slower than downloading and installing go.
+#
+# Since workspaces are scoped to a workflow you need to persist to a workspace
+# in your workflow before you can read from it, which further slows the build
+# time and persisting seems to take about twice as long as retrieving. And on
+# top of that remember that before persisting you will need to run the task
+# that you would otherwise run directly in a downstream job.
+#
+# Workspaces work best for small amounts of data that take a long time to
+# generate that we want to share between many jobs without having to duplicate
+# everywhere the logic to generate the data.
+#
+# How are we using them?
+#
+# We are currently using workspaces to store the compiled-system-contracts
+# (4.7M) which take about 5 minutes to generate 2s to persist to the workspace
+# and 0s to restore from the workspace, we also cache them so that most of the
+# time we can skip the build step and simply persist the cache to the
+# workspace. Having this functionality contained in one job makes the config
+# simpler than if we tried to add compiled system contract generation to all of
+# the downstream jobs and since this job is quite fast and not in the critical
+# path for the workflow it's ok to be a bit slower here.
+#
+########
+#
+# The structure of this workflow
+#
+# We have a central folder ~/repos under which there are paths which are used
+# throughout the workflow:
+#
+# - geth: The celo-blockchain checkout
+# - celo-monorepo: The celo-monorepo checkout
+# - golang: A golang installation
+# - gocache: A folder containing intermediate compilation results for golang
+# - gomodcache: A folder containing downloaded modules for golang
 version: 2.1
 parameters:
-  # Increment this to force cache rebuilding
+  # Increment these to force cache rebuilding
   system-contracts-cache-version:
     type: integer
-    default: 1
+    default: 2
+  go-mod-cache-version:
+    type: integer
+    default: 3
+  go-cache-version:
+    type: integer
+    default: 3
+  checkout-monorepo-cache-version:
+    type: integer
+    default: 3
+  # Location where compiled system contracts are stored under the root of this
+  # repo.
   system-contracts-path:
     type: string
     default: "compiled-system-contracts"
   system-contracts-executor-image:
-    type: string
+    type:
+      string
+      # The system contracts executor currently needs node 12 to function,
+      # this should only be changed when the version in `monorepo_commit` is
+      # changed.
+    default: "us.gcr.io/celo-testnet/circleci-node12:1.0.0"
+  e2e-executor-image:
+    type:
+      string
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -1,14 +1,111 @@
...
+# approx 12s, so 64s total, so a reduction of 57s which is almost half the
+# time!
+#
+####
+#
+# Restoring monorepo dir + celo-blockchain  + go install (684M) from workspace
+# approx 51s.
+#
+# Restoring monorepo dir + go install (659M) from workspace approx 37s +
+# shallow ceckout celo-monorepo 2s, so a total of 39s a 12 second improvement.
```
YoduYodu, line 30: type "ceckout"

YoduYodu: Why are both "Restoring monorepo dir" and "shallow "checkout celo-monorepo" needed?

.circleci/config.yml
```diff
@@ -1,14 +1,111 @@
...
+# We are currently using workspaces to store the compiled-system-contracts
+# (4.7M) which take about 5 minutes to generate 2s to persist to the workspace
+# and 0s to restore from the workspace, we also cache them so that most of the
+# time we can skip the build step and simply persist the cache to the
+# workspace. Having this functionality contained in one job makes the config
+# simpler than if we tried to add compiled system contract generation to all of
+# the downstream jobs and since this job is quite fast and not in the critical
+# path for the workflow it's ok to be a bit slower here.
+#
+########
```
YoduYodu, line 67: I suggest putting this section `The structure of this workflow` at the top, making it easier to understand above performance note.

.circleci/config.yml
```diff
@@ -53,26 +146,122 @@ unit-tests-defaults: &unit-tests-defaults
...
+  run: date +%m > ~/repos/month
+
+end-to-end-test: &end-to-end-test
+  executor: e2e
+  resource_class: large
+  working_directory: ~/repos
+  parameters:
+    cache-key:
+      type: string
+      default: checkout-monorepo-{{ checksum "~/repos/geth/monorepo_commit" }}-v<<pipeline.parameters.checkout-monorepo-cache-version>>-<<pipeline.parameters.e2e-executor-image>>
```
gastonponti, line 205: As we talked, this `monorepo_commit` could bring some issues if for example, someone uses master. That will save a cache for the `master` branch, and with new commits, will continue to access the same cached master from before. Same with all the others branches, but probably as the contract versions are tags and not branches, will be something that could happen to master

.circleci/config.yml
```diff
@@ -1,14 +1,111 @@
...
+# that you would otherwise run directly in a downstream job.
+#
+# Workspaces work best for small amounts of data that take a long time to
+# generate that we want to share between many jobs without having to duplicate
+# everywhere the logic to generate the data.
+#
+# How are we using them?
+#
+# We are currently using workspaces to store the compiled-system-contracts
+# (4.7M) which take about 5 minutes to generate 2s to persist to the workspace
```
YoduYodu, line 59:
```suggestion
# (4.7M) which take about 5 minutes to generate, 2s to persist to the workspace
```

.circleci/config.yml
```diff
@@ -1,14 +1,111 @@
...
+      # The system contracts executor currently needs node 12 to function,
+      # this should only be changed when the version in `monorepo_commit` is
+      # changed.
+    default: "us.gcr.io/celo-testnet/circleci-node12:1.0.0"
+  e2e-executor-image:
+    type:
+      string
       # The system contracts executor currently needs node 12 to function,
       # this should only be changed when the version in `monorepo_commit` is
       # changed.
```
YoduYodu, line 111: Modify this?

.circleci/config.yml
```diff
@@ -53,26 +146,122 @@ unit-tests-defaults: &unit-tests-defaults
...
+    key: go-mod-<<pipeline.parameters.go-cache-version>>-{{ checksum "~/repos/geth/go.sum" }}
+    paths:
+      - ~/repos/gocache
+
+restore-go-mod-cache: &restore-go-mod-cache
+  restore_cache:
+    keys:
+      - go-mod-<<pipeline.parameters.go-cache-version>>-{{ checksum "~/repos/geth/go.sum" }}
+
+store-month-in-file: &store-month-in-file
```
YoduYodu, line 190: https://circleci.com/docs/2.0/caching/#introduction
max caching storage duration is 15 days

.circleci/config.yml
```diff
@@ -53,26 +146,122 @@ unit-tests-defaults: &unit-tests-defaults
...
+    key: go-cache-<<pipeline.parameters.go-cache-version>>-{{ checksum "~/repos/geth/go.sum" }}-{{ checksum "~/repos/month" }}
+    paths:
+      - ~/repos/gocache
+
+restore-go-cache: &restore-go-cache
+  restore_cache:
+    keys:
+      - go-cache-<<pipeline.parameters.go-cache-version>>-{{ checksum "~/repos/geth/go.sum" }}-{{ checksum "~/repos/month" }}
+
+go-mod-cache: &save-go-mod-cache
```
YoduYodu, line 179:
```suggestion
save-go-mod-cache: &save-go-mod-cache
```

.circleci/config.yml
```diff
@@ -53,26 +146,122 @@ unit-tests-defaults: &unit-tests-defaults
...
+    - *restore-go-cache
+    - run:
+        name: Run e2e test
+        no_output_timeout: 15m
+        command: |
+          export PATH=${PATH}:~/repos/golang/go/bin
+          cd celo-monorepo/packages/celotool
+          ./${TEST_NAME} local ~/repos/geth
+    # Note, all e2e tests call 'make all' in ~/repos/geth, this causes most code
+    # to be built and so results in a faifly well populated go cache.
```
YoduYodu, line 253:
```suggestion
    # to be built and so results in a fairly well populated go cache.
```

.circleci/config.yml
piersy: Im not sure what you are suggesting here @YoduYodu ?

.circleci/config.yml
piersy: Hmmm, I read that as meaning that caches would persist for 15 days after their last use. There doesn't seem to be any sense in deleting a cache if it's in constant use, but their documentation seems to suggest otherwise. I've opened an [issue](https://github.com/circleci/circleci-docs/issues/6909) to get some clarification on that, in the meantime I think we can keep this as it will ensure periodic cache changes and it's not a problem if the cache changes slightly more often.

common/task/task.go
```diff
@@ -4,19 +4,44 @@ import "time"
...
+	d.t.Stop()
+}
+
+func RunTaskRepeateadly(task func(), t ticker) StopFn {
 	// Setup the ticker and the channel to signal
 	// the ending of the interval
-	ticker := time.NewTicker(period)
 	stop := make(chan struct{})
 
+	println("run task started")
```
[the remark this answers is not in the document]
piersy: left over debug, thanks!

common/task/task_test.go
```diff
@@ -5,16 +5,34 @@ import (
...
+func (t *testTicker) sendTick() {
+	t.tc <- time.Now()
+}
+
 func TestRunTaskRepeateadly(t *testing.T) {
 	t.Parallel()
 
 	counter := 0
-	ping := func() { counter++ }
+	ping := func() { println("tick"); counter++ }
```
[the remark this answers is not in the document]
piersy: More left over debug!

.circleci/config.yml
piersy: I've addressed this in dbbe58e472df3882939be59e9e3a894e6e882790 `make prepare-system-contracts` will fail if `monorepo_commit` contains a branch.

piersy: > I've addressed this in [dbbe58e](https://github.com/celo-org/celo-blockchain/commit/dbbe58e472df3882939be59e9e3a894e6e882790) `make prepare-system-contracts` will fail if `monorepo_commit` contains a branch.

Let me know what you think @gastonponti ?

gastonponti (approved):
LGTM

.circleci/config.yml
piersy: Ok I discovered that we have some old caches [this job restores a cache that is over 90 days old](https://app.circleci.com/pipelines/github/celo-org/celo-blockchain/7979/workflows/a76c3667-d19b-428d-979a-c2c6d6a3b35c/jobs/82940) so I think we need to keep the rotation functionlaity.

PR #1922, opened 2022-07-14
Increase max message size for p2p peers

hbandura:


gakonst:
Is this PR related to the recent network halt due to validators rejecting / not committing large messages?

PR #1924, opened 2022-07-14
Reduce Preprepare size

mcortesi:
### Description

This PR modifies `RoundChangeCertificate` encoding so that we don't serialize `N` times the proposal, instead we do that only once.

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Change is not backward compatible, requires 2/3 of validators to use the new version at that same time

### DIFF
--- a/consensus/istanbul/types.go
+++ b/consensus/istanbul/types.go
@@ -28,6 +28,7 @@ import (
 	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/crypto"
 	blscrypto "github.com/celo-org/celo-blockchain/crypto/bls"
+	"github.com/celo-org/celo-blockchain/log"
 	"github.com/celo-org/celo-blockchain/p2p/enode"
 	"github.com/celo-org/celo-blockchain/rlp"
 )
@@ -145,6 +146,7 @@ func (c *RoundChangeCertificate) EncodeRLP(w io.Writer) error {
 	if err != nil {
 		return err
 	}
+	log.Debug("Proposals count on the round change certificate", len(proposals))
 	return rlp.Encode(w, []interface{}{proposals, messages})
 }
 
--- a/consensus/istanbul/types.go
+++ b/consensus/istanbul/types.go
@@ -211,11 +211,6 @@ type IndexedRoundChangeMessage struct {
 func NewIndexedRoundChangeMessage(message *Message) *IndexedRoundChangeMessage {
 	roundChange := getRoundChange(message)
 	pc := roundChange.PreparedCertificate
-	if pc == nil {
-		return &IndexedRoundChangeMessage{
-			Message: *message,
-		}
-	}
 
 	// Assume message.Code = MsgRoundChange
 	indexedMsg := IndexedRoundChangeMessage{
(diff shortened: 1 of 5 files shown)


consensus/istanbul/types.go
```diff
@@ -135,6 +139,144 @@ func (b *RoundChangeCertificate) IsEmpty() bool {
...
+	// create a Proposal index from the list
+	propIndex := make(map[common.Hash]Proposal)
+	for _, prop := range props {
+		propIndex[prop.Hash()] = prop
+	}
+	// Recreate Messages one by one
+	mess := make([]Message, len(iMess))
+	for i, im := range iMess {
+		mess[i] = Message{
+			Code:      im.Message.Code,
```
timmoreton, line 177: could im.Message be nil here?

consensus/istanbul/types.go
```diff
@@ -135,6 +139,144 @@ func (b *RoundChangeCertificate) IsEmpty() bool {
...
+			Address:   im.Message.Address,
+			Signature: im.Message.Signature,
+		}
+
+		// Add the proposal to the message if it had one
+		roundChange, err := getRoundChange(&im.Message)
+		if err != nil {
+			return err
+		}
+		if proposal, ok := propIndex[im.ProposalHash]; ok {
```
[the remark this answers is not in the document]
gastonponti: Only if someone is sending a malicious message, I think. But in this case, as the propIndex is generated with valid proposals, it shouldn't find anything and in that case

consensus/istanbul/types.go
```diff
@@ -135,6 +139,144 @@ func (b *RoundChangeCertificate) IsEmpty() bool {
...
+		}
+
+		if proposal != nil {
+			// we don't use the height since we know they MUST be the same
+			proposalsMap[proposal.Hash()] = proposal
+		}
+	}
+
+	// Iterate values. RLP does not support maps
+	proposals := make([]*types.Block, len(proposalsMap))
```
marekolszewski, line 225: Would it be helpful to log the number of proposals we're seeing here (e.g. with trace logging)? 

If this does not fix the issue, it would be nice to get some fresh logging data to see what the composition of the round change certificate looks like.

consensus/istanbul/types.go
```diff
@@ -135,6 +139,144 @@ func (b *RoundChangeCertificate) IsEmpty() bool {
...
+	for i, message := range c.RoundChangeMessages {
+		var proposal *types.Block
+		proposal, messages[i], err = extractProposal(&message)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if proposal != nil {
+			// we don't use the height since we know they MUST be the same
+			proposalsMap[proposal.Hash()] = proposal
```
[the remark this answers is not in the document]
gastonponti: We are using that map to build the proposal list. As the same proposal will be set in the same hash, that result array will be already a set

consensus/istanbul/types.go
hbandura: Mmm. As much as before, if an element of the list had been nil. Not sure if it's great not to check, but it's not worse than before

hbandura (approved):

PR #1937, opened 2022-08-18
Consensus Fork: Preprepare and RoundChange V2 messages

hbandura:
This PR implements a consensus fork that is determined by a specific block number. From that sequence on, the messages `Preprepare` and `RoundChange` will be no longer used, and instead `PreprepareV2` and `RoundChangeV2` will be used instead.

The main purpose is to reduce the preprepare message size. `Preprepare` can contain a `RoundChangeCertificate`, which contains a quorum of `RoundChange` messages received previously. Since a `RoundChange` message can potentially include a full block (due to a `PreparedCertificate`), the `Preprepare` message can, in the worst case, contains `f+1` different blocks (as per IBFT protocol design capping `PreparedCertificate` amount to `f+1`).

The `PreprepareV2` uses a new `RoundChangeCertificateV2` format, that instead of holding the whole `RoundChange` messages, it uses `RoundChangeRequest` objects, which are a smaller (also signed, verified) version, that do not have a full block but only a reference (hash) to it. The `PreprepareV2` should propose a block whose hash is the one referenced in the highest round `PrepareCertificateV2` found in the `RoundChangeCertificateV2`.

The `RoundChangeV2` is a new round change message that holds an inner signed `RoundChangeRequest` (to be shared in the `RoundChangeCertificateV2`) and a block proposal in the case that there is a non-empty `PreparedCertificateV2` in the request (matching the proposal's hash to the one referenced in the `PreparedCertificateV2`.

The following command is useful for locally reviewing some of the changes which are largely duplicated code:
```
for x in preprepare roundchange roundchangeset; do git diff --no-index consensus/istanbul/core/${x}.go  consensus/istanbul/core/${x}_v2.go; done
```


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 5772ea5e489b546f26f0ca662ffb6050100d6997
<details><summary>coverage:  46.5% of statements across all listed packages</summary><br>
<pre>
coverage:  54.6% of statements in consensus/istanbul
coverage:  40.1% of statements in consensus/istanbul/announce
coverage:  54.5% of statements in consensus/istanbul/backend
coverage:  53.9% of statements in consensus/istanbul/core
</pre>
CommentID: ec119ab652
</details>

piersy:
Hey @hbandura, when I was thinking of what we needed to do here I was imagining just transforming the messages that we send over the network without actually changing much of consensus (if any). So for example


Old system sending round change from one peer to another:
```
|-----------------peer 1-----------------|---network----|-------------------peer 2-----------------|
OutputRoundChange ---serialised---> bytes --------------> bytes ---deserialised---> InputRoundChange
```

System I was imagining for the v2:
```
|----------------------------------peer 1---------------------------------|---network----|-----------------------------------peer 2------------------------------------|
OutputRoundChange --transform--> OutputRoundChangeV2 --serialised--> bytes --------------> bytes ---deserialised---> OutputRoundChangeV2 --transform--> InputRoundChange
```

I was imaging this would make things simple for us because we wouldn't need to actually modify any core consensus code and as long as we could show that the transformation between v2 and v1 was correct then we could be fairly sure that everything else would work (assuming we hadn't touched it).

I'm wondering why you didn't follow this approach?

hbandura:
> Hey @hbandura, when I was thinking of what we needed to do here I was imagining just transforming the messages that we send over the network without actually changing much of consensus (if any). So for example
> 
> Old system sending round change from one peer to another:
> 
> ```
> |-----------------peer 1-----------------|---network----|-------------------peer 2-----------------|
> OutputRoundChange ---serialised---> bytes --------------> bytes ---deserialised---> InputRoundChange
> ```
> 
> System I was imagining for the v2:
> 
> ```
> |----------------------------------peer 1---------------------------------|---network----|-----------------------------------peer 2------------------------------------|
> OutputRoundChange --transform--> OutputRoundChangeV2 --serialised--> bytes --------------> bytes ---deserialised---> OutputRoundChangeV2 --transform--> InputRoundChange
> ```
> 
> I was imaging this would make things simple for us because we wouldn't need to actually modify any core consensus code and as long as we could show that the transformation between v2 and v1 was correct then we could be fairly sure that everything else would work (assuming we hadn't touched it).
> 
> I'm wondering why you didn't follow this approach?

This is what we did before during the stall, on the serialization of the Preprepare V1 message. This was not possible here. The format change on the RoundChangeV2 and PreprepareV2 messages are not just cosmetic. They force a different way of validating them. Say if we were to do the approach you suggest, then the signatures would be invalid, so we would have to modify or ignore signatures in the code if the message came from the v2 approach. And just like that, a whole lot of other validations suddenly have to be handled differently. In the end it made more sense to create new validating and state transition functions that are directly designed for the new format, but I tried to extract common code every time I was able to.

e2e_test/e2e_test.go
```diff
@@ -39,7 +39,32 @@ func TestSendCelo(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Wait for the whole network to process the transaction.
+	err = network.AwaitTransactions(ctx, tx)
+	require.NoError(t, err)
+}
+
+// TestSendCeloV2 starts a network submits a transaction and waits for the whole
+// network to process the transaction.
+func TestSendCeloV2(t *testing.T) {
```
piersy, line 58: Rather than duplicating all the test code, I think we should be able to just run each test twice with slightly different config which I would prefer because it will obviously make things simpler if we end up needing to change any tests.

piersy: I've managed to set this up, take a look at
* #1950

consensus/istanbul/core/commit_test.go
```diff
@@ -266,6 +266,243 @@ OUTER:
 	}
 }
 
+func TestHandleCommitV2(t *testing.T) {
```
piersy, line 269: If we put all the v2 tests in their own file then removing the original tests becomes as simple as deleting a file, it's also going to make any conflicts easier to resolve, in the case that changes were made to the tests.

consensus/istanbul/types.go
```diff
@@ -837,6 +855,11 @@ func (m *Message) RoundChange() *RoundChange {
 	return m.roundChange
 }
 
+// Prepare returns round change if this is a round change message.
```
piersy, line 858:
```suggestion
// RoundChangeV2 returns round change if this is a round change message.
```

consensus/istanbul/types.go
```diff
@@ -813,6 +826,11 @@ func (m *Message) Preprepare() *Preprepare {
 	return m.prePrepare
 }
 
+// Preprepare returns preprepare if this is a preprepare message.
```
piersy, line 829:
```suggestion
// PreprepareV2 returns preprepare if this is a preprepare message.
```

consensus/istanbul/core/commit_test.go
piersy: And it would make reviewing a lot easier because then I could just diff the two files.

piersy: Ok I've made this change in the below PR:
* #1954

params/config.go
```diff
@@ -71,6 +71,7 @@ var (
 			BlockPeriod:    5,
 			RequestTimeout: 3000,
 			LookbackWindow: 12,
+			V2Block:        big.NewInt(15287159),
```
piersy, line 74: These activation blocks will need to be set again, the mainnet is only 4 days from now but I see that baklava has already happened. (didn't check alfajores)

consensus/istanbul/core/errors.go
```diff
@@ -49,8 +49,14 @@ var (
...
 	errInvalidRoundChangeViewMismatch = errors.New("View for PREPARED certificate is greater than the view in the round change message")
+	// errInvalidRoundChangeRequestSignature is returned when a RoundChangeRequest is improperly signed
+	errInvalidRoundChangeRequestSignature = errors.New("Round Change Request improperly signed")
+	// errRoundChangeRequestAddressMismatch is returned when a RoundChangeRequest address does not match the message address
+	errRoundChangeRequestAddressMismatch = errors.New("Round Change Request address does not match message address")
 	// errInvalidPreparedCertificateInconsistentViews is returned when the PREPARED certificate view is inconsistent among it's messages
 	errInvalidPreparedCertificateInconsistentViews = errors.New("View is inconsistent among the PREPARED certificate messages")
+	// errRoundChangeProposalHashMismatch is returned when the proposal hash of a round change prepared certificate does not match the proposal's hash.
+	errRoundChangeProposalHashMismatch = errors.New("Proposal hash mismatch in round change message")
```
piersy, line 59: Rather than adding these errors with static messages I would suggest instead just creating an error at the point where you are using them. This would allow for much richer error descriptions that are able to provide contextual information such as what was the proposal hash and what was the mismatching hash and who sent the mismatched hash.

piersy:
@hbandura I've left a few comments and suggestions. I couldn't see any major problems with the implementation, but one thing that does seem to be missing and I would class as important is a test of the old and new consensus showing that the new version is able to handle blocks that  are so big they cause the old implementation to halt.

consensus/istanbul/types_v2.go
```diff
@@ -0,0 +1,329 @@
...
+
+// HighestRoundWithPreparedCertificate returns the highest round request with a PC. Be mindful
+// that there could be many different PCs with the same highest round.
+func (rcc *RoundChangeCertificateV2) HighestRoundWithPreparedCertificate() *big.Int {
+	var maxRound *big.Int
+	for _, req := range rcc.Requests {
+		if !req.HasPreparedCertificate() {
+			continue
+		}
+		if req.View.Round == nil {
```
gastonponti, line 264: We always have a view set, right?

consensus/istanbul/core/roundchange_v2.go
```diff
@@ -0,0 +1,247 @@
...
+			return err
+		}
+
+		// Check for duplicate ROUND CHANGE messages
+		if seen[request.Address] {
+			return errInvalidRoundChangeCertificateDuplicate
+		}
+		seen[request.Address] = true
+
+		if request.View.Sequence == nil || request.View.Round == nil {
```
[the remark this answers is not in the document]
piersy: Hey @gastonponti `request.View` is not a pointer, and there are no `istanbul.Message`s here so there are no code to check. The previous structure had a slice of `istanbul.Message` so any message could have been added there but this structure only holds `istanbul.RoundChangeRequests`.

e2e_test/e2e_test.go
hbandura: No reason, just didn't want to waste time understanding this to perfection when as soon as this version is out the V2 ones could be deleted and the V2Block would be added as default. But your solution is nice if you are 100% sure this triggers them with the v2 on.

consensus/istanbul/types_v2.go
hbandura: we should, yes

consensus/istanbul/core/core.go
```diff
@@ -515,20 +556,33 @@ func (c *core) startNewRound(round *big.Int) error {
...
 	prevBlock := c.current.Sequence().Uint64() - 1
 	blockAuthor := c.backend.AuthorForBlock(prevBlock)
 	valSet := c.current.ValidatorSet()
 	nextProposer := c.selectProposer(valSet, blockAuthor, newView.Round.Uint64())
 
+	var err error
+	var request *istanbul.Request
+	var roundChangeCertificate istanbul.RoundChangeCertificate
+	var roundChangeCertificateV2 istanbul.RoundChangeCertificateV2
+	if c.address == nextProposer.Address() && propose {
```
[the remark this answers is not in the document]
hbandura: This is a funny one:
`startNewRound` is called from two different places: `handleRoundChange` and `handleRoundChangeCertificate`. The first occurs when receiving a `RoundChange(V1 or V2)` message, and the second when receiving a `Preprepare(V1 or V2)` message (round >= 1).

So, in the second case, during a preprepare handling, this function is creating a preprepare and a roundchangecertificate with the round change messages that it received in the roundchangecertificate. So it's pointless, but this is how it was written before since it's shared code called from two different sides. 

However, with our V2 change of the RoundChangeCertificate, we may not have the round change messages, but still received the preprepare with a RCC. And the RoundChangeSet stores roundchangeMessages since it needs the full block to decide which is the one that goes with the highest round preparedCertificate. So we don't have the data to create it properly. The solution was to modify completely how the roundChangeSet works, but since this was a change to fix something that shouldn't be done anyways, I just added the propose flag. A greater rework may imply bigger changes (calling two different functions) but I wanted to keep changes at minimum.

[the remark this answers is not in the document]
gastonponti: Ok, makes sense. Could you add a comment for this to a void a future mistake?

consensus/istanbul/core/roundchange_v2.go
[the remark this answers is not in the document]
gastonponti: Ohhhh, totally missed that. Thanks

e2e_test/e2e_test.go
hbandura: Merged it

consensus/istanbul/core/commit_test.go
hbandura: Merged

consensus/istanbul/types.go
hbandura: Done

consensus/istanbul/types.go
hbandura: Done

params/config.go
hbandura: These were placeholder values I set a long time ago, yes, we need to think what the proper date will be

consensus/istanbul/core/errors.go
hbandura: I don't like much this setup as well, but wanted to be consistent with other consensus error types

consensus/istanbul/core/roundchange_v2.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	logger := c.newLogger("func", "handleRoundChangeCertificateV2", "proposal_round", subject.View.Round, "proposal_seq", subject.View.Sequence, "proposal_digest", subject.Digest.String())
+
+	if len(roundChangeCertificateV2.Requests) > c.current.ValidatorSet().Size() || len(roundChangeCertificateV2.Requests) < c.current.ValidatorSet().MinQuorumSize() {
+		return errInvalidRoundChangeCertificateNumMsgs
+	}
+
+	seen := make(map[common.Address]bool)
+	for i := range roundChangeCertificateV2.Requests {
```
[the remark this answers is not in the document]
hbandura: This is something that I don't have a proper answer for. I remember checking the paper for specifically this question, and we discussed this with @gastonponti , about if it makes sense or not to do it. In theory only the prepared certificate for the highest round should be used to match the proposal. So that's the only one that I'm verifying (below, at `preparedView, err := c.verifyPCV2WithProposal(*pc, proposal)` ).

consensus/istanbul/core/core.go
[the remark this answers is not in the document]
hbandura: Done

consensus/istanbul/core/roundchange_v2.go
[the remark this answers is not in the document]
piersy: Having thought more about this, i think this is safe.

A malicious proposer could include f+1 round change requests that are not valid (fake proposals, no prepare or commit messages), but they cannot forge a valid round change request with prepared certificate because 2f+1 prepare or commit messages would be needed for that.

So if a malicious proposer included a fake round change request with prepared certificate it will not pass verification, nodes will not vote for the proposal and the effect would be as if the proposer had not proposed.

consensus/istanbul/core/errors.go
piersy: I think it's ok to diverge here, its not like this package is particularly clean or consistent.

Using hardcoded strings like this makes troubleshooting harder, especially when you consider that we don't run validators so the best we can do is ask for the logs after a failure, so we really need to try and cram as much useful information into those logs as possible.

piersy (approved):
The test for the message size looks good! I'd prefer if we could make the errors better but I don't think it's worth blocking this PR over.

consensus/istanbul/core/roundchange_v2.go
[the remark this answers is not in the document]
hbandura: The proposer cannot actually do that, because we are checking the signatures of all the requests. So what is not being checked here is if the prepared certificate signed by the emitter of the roundChangeRequest is valid. But we know that the request is properly signed

consensus/istanbul/core/backlog.go
```diff
@@ -73,12 +74,13 @@ func (c *core) checkMessage(msgCode uint64, msgView *istanbul.View) error {
...
 	// RoundChange messages are accepted in all states and for current or future rounds.
-	if msgCode == istanbul.MsgRoundChange {
+	if msgCode == istanbul.MsgRoundChange || msgCode == istanbul.MsgRoundChangeV2 {
 		return nil
 	}
 
 	// WaitingForNewRound and StateAcceptRequest: accepts Preprepare (including for rounds >= desiredRound), other messages are future.
-	if (c.current.State() == StateWaitingForNewRound || c.current.State() == StateAcceptRequest) && msgCode != istanbul.MsgPreprepare {
+	if (c.current.State() == StateWaitingForNewRound || c.current.State() == StateAcceptRequest) &&
+		(msgCode != istanbul.MsgPreprepare && msgCode != istanbul.MsgPreprepareV2) {
```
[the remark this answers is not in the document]
hbandura: Done

Issue #1945, opened 2022-09-12
Celo RPC API is incompatible with ethers.js


piersy:
## Description

Celo's block type differs from the block used in ethereum, this causes our RPC API to be incompatible with ethers.js (and most likely many other web3 clients)

See these issues where people have encountered problems
https://github.com/ethers-io/ethers.js/issues/1735
https://github.com/ethers-io/ethers.js/discussions/2919
https://github.com/ethers-io/ethers.js/issues/2558
https://github.com/celo-org/celo-blockchain/issues/1738

Celo's block header lacks the following fields that exist in ethereum.

```
UncleHash
Difficulty
GasLimit
MixDigest
Nonce
BaseFee
```

So in order to provide an API that is compatible with ethers.js we should seek to somehow re-instate these missing fields. We could do this by actually adding the fields to the block or simply by filling them with fake data in RPC responses.

Although BaseFee and GasLimit are not currently part of the Celo block they are parameters of the system and are managed by smart contracts. So these can be added to the block in a meaningful way.

It's not so clear however what to do with UncleHash, Difficulty, MixDigest and Nonce, since these fields really do not exist in the Celo system. We could add them to the block or RPC responses, but we would need to fill them with dummy data. This could be more problematic than just not having them since people may then try to make important decisions based off the dummy data. Or may find that the dummy data causes other problems within their system.

sashaaldrick:
This is preventing us from deploying on Celo/Alfajores atm as we require hardhat forking and ethers compatibility, see https://github.com/NomicFoundation/hardhat/issues/1966

jaredcassidyluna:
ethers.js expects even length hex number for gasLimit:

We have been using ethers.js with an override to the formatter function to allow it to be used without gasLimit being present in the rpc responce.

About 2 days ago we started picking up errors on alfajores, I have removed our override and the error is still coming up

```
ethersProvider.getBlockWithTransactions(currentBlockHeight)

[...]

PR #1957, opened 2022-09-21
Improve RPC API compatibility for Ethers.js

piersy:
### Description

This PR aims to make our RPC API usable in the common case with Ethers.js, it tackles the following problems:

* Prior to this change attempts to get a block with Ethers.js would raise an exception, because the `gasLimit` field was missing. The `gasLimit` field is now filled out on the block returned by the RPC API, note the field does not actually exist on the internal representation of the block.
* Prior to this change users of Ethers.js were unable to submit EIP-1559 style transactions with the `maxFeePerGas` and `maxPriorityFeePerGas` fields, because by the RPC API did not contain the `baseFeePerGas` field which Ethers.js uses to determine whether  EIP-1559 style transactions are supported. The `baseFeePerGas` field is now filled out on blocks returned by the RPC API, note the field dos not actually exist on the internal representation of the block.

### Discussion points
The `gasLimit` field is populated on the RPC block by getting the value from the contract state. If the state doesn't exist a default value (20M) is returned (this is pre-existing functionality for getting the gas limit) this means that blocks can exhibit inconsistent gas limit values depending on whether the node they are being retrieved from has the state for that block. Should we instead return no gas limit field in the case that there is no state?

Similarly the `baseFeePerGas` is populated on the RPC block by getting the value from the contract state but in contrast there is no default value so in the case that there is no state for the requested block the block returned from the RPC API will not have the `baseFeePerGas` field populated. This actually works ok for Ethers.js with respect to sending EIP-1559 style transactions because Ethers.js always requests the latest block to determine the `baseFeePerGas`, however this could also lead to confusion for developers.

Adding a flag to disable this change, currently there is no way to disable this change, this change doesn't seem to be a problem for our tooling, because all the monorepo e2e tests have passed in circleci, however it could cause issues for other users, it's hard to know how many may be affected. Should we add the flag?

### Other changes

Updated the readme to specify a minimum go version rather than a current go version, this should relieve us from having to try and keep the readme in sync whenever we update go.

### Tested

There's an e2e test that runs Ethers.js against our blockchain node to check that it works.

### Related issues

- Related to  #1945

### Backwards compatibility

This is changing the RPC API response so Its not backwards compatible.

### Release Notes

An additional two fields have been added to the block representation returned by the RPC api, they are `gasLimit` and `baseFeePerGas`, both fields previously existed as parameters of the system held in contract state. The reason for this change is to improve compatibility with existing ethereum tooling specifically `EthersJS` which crashes if the `gasLimit` is not present on a block and is unable to send EIP-1559 transactions if `baseFeePerGas` is not present. Note that the internal block representation has not been modified so these new fields should not be taken into account in block hashing or signing operations. A new flag has been added which allows disabling the addition of the new fields as a measure to help ease the transition to the new RPC block representation, the flag will likely be removed in a future release.
TODO: talk about the inconsistent return values. due to state pruning.


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit a0174f4e2ae1069c5d4a121829e9311cc0fe9c0c
<details><summary>coverage:  50.8% of statements across all listed packages</summary><br>
<pre>
coverage:  64.0% of statements in consensus/istanbul
coverage:  42.7% of statements in consensus/istanbul/announce
coverage:  66.5% of statements in consensus/istanbul/core
</pre>
CommentID: 59959d96fe
</details>

Makefile
```diff
@@ -45,6 +45,13 @@ geth:
 	@echo "Done building."
 	@echo "Run \"$(GOBIN)/geth\" to launch geth."
 
+prepare: prepare-system-contracts prepare-ethersjs-project
+
+prepare-ethersjs-project: ./ethersjs-api-check/node_modules
+
+./ethersjs-api-check/node_modules: ./ethersjs-api-check/package.json ./ethersjs-api-check/package-lock.json
+	@cd ./ethersjs-api-check && npm ci
```
palango, line 53: Nice, I didn't know about `npm ci`

.circleci/config.yml
```diff
@@ -351,11 +353,12 @@ jobs:
       - attach_workspace:
           at: ~/repos
       - *restore-go-mod-cache
+      - run: make prepare-ethersjs-project
         # Run the tests with coverage parse the coverage and output the summary
       - run:
           name: Run tests and print coverage summary
           command: |
-            go test -coverprofile cov.out -coverpkg ./consensus/istanbul/... ./e2e_test 
+            go test -v -coverprofile cov.out -coverpkg ./consensus/istanbul/... ./e2e_test
```
palango, line 361: Do you want to keep the added `-v`?

internal/ethapi/api.go
```diff
@@ -724,11 +728,42 @@ func (s *PublicBlockChainAPI) GetBlockByNumber(ctx context.Context, number rpc.B
...
+func addEthCompatibilityFields(ctx context.Context, block map[string]interface{}, b Backend, header *types.Header) {
+	hash := header.Hash()
+	numhash := rpc.BlockNumberOrHash{
+		BlockHash: &hash,
+	}
+	gasLimit := b.GetBlockGasLimit(ctx, numhash)
+	block["gasLimit"] = hexutil.Uint64(gasLimit)
+
+	// Providing nil as the currency address gets the gas price minimum for the native celo asset.
+	baseFee, err := b.GasPriceMinimumForHeader(ctx, nil, header)
```
palango, line 754: How does this return 20M in the not found case?
If I follow the code, I find this line, which seems to be preset to 0.
https://github.com/celo-org/celo-blockchain/blob/f4b3f22a082eeb9a965d85b72c07c5d5a54e0296/contracts/gasprice_minimum/gasprice_minimum.go#L70

palango (commented):
Sorry, I mostly added some questions I had while reading the code for now.

palango:
> The gasLimit field is populated on the RPC block by getting the value from the contract state. If the state doesn't exist a default value (20M) is returned (this is pre-existing functionality for getting the gas limit) this means that blocks can exhibit inconsistent gas limit values depending on whether the node they are being retrieved from has the state for that block. Should we instead return no gas limit field in the case that thee is no state?

It will be inconsistent in both variants, right? Either it first returns 20M and then another value once the state is available, or it is first not returned and then suddenly is. How was the gas limit historically set? And how would people use it?

.circleci/config.yml
piersy: Yes that was intentional, we want to have the output in CI so we can more easily analyse failures.

internal/ethapi/api.go
gastonponti, line 754: Should we also mock the gasPriceMinimum for blocks that we don't have the state?
The same as we are doing with the `GetBlockGasLimit` where we are returning `params.DefaultGasLimit` which is set to 20M.
Both have different natures and the blockGasLimit will fluctuate only by governance (which is not that common), and the price is tied to usage. But we are already sending a possible "wrong" answer. Should we do the same with the `baseFee`? Or maybe we should put a nil in the gasLimit too if we don't have the state

internal/ethapi/api.go
```diff
@@ -724,11 +728,42 @@ func (s *PublicBlockChainAPI) GetBlockByNumber(ctx context.Context, number rpc.B
...
+	// GasPriceMinimumForHeader can return an error + non nil baseFee, an error
+	// and nil base fee or a base fee and a nil error. In all cases where a
+	// base fee is returned we want to add it as a field and if there was an
+	// error and nil base fee we log the failure.
+	if err != nil && baseFee == nil {
+		log.Debug("Not adding baseFeePerGas to RPC response, failed to retrieve gas price minimum", "block", header.Number.Uint64(), "err", err)
+		return
+	}
+	block["baseFeePerGas"] = (*hexutil.Big)(baseFee)
+}
```
gastonponti, line 765: Probably is not breaking ethers, but why are we not adding the `difficulty` too? It's a low hang fruit if we use the block number

ethersjs-api-check/test/test.ts
```diff
@@ -0,0 +1,44 @@
...
+		assert.notEqual(feeData.maxPriorityFeePerGas, null);
+		// We check the other 2 fields for completeness, they should also be set.
+		assert.notEqual(feeData.gasPrice, null);
+		assert.notEqual(feeData.lastBaseFeePerGas, null);
+	});
+
+	// Our blockchain client implementation returns a default gas limit when the
+	// actual gas limit cannot be retrieved.  We cannot check fee data against a
+	// pruned block because getFeeData always requests the latest block.
+	it('block with pruned state still reports gas limit', async () => {
```
gastonponti, line 35: Trying to understand something.
This test will be triggered probably with state.
But you are reusing this test in the e2e one, to execute with state, prune it, and execute it again right? So that is the actual test that is checking the prunning, and this one executed by circle, will always return what we have in the state

gastonponti: For the pruned one, you could also add a test checking that the baseFee is null using the same logic (if we keep the baseFee as it is, and not adding a default as I asked in another comment)

internal/ethapi/api.go
gastonponti: Same as the other comments, sorry for didn't check that before.
I agree that we should be consistent. Or we return null in both cases, or we return a mocked things

ethersjs-api-check/test/test.ts
piersy: @ponti, I don't understand why you think this test would be executed with state?

internal/ethapi/api.go
piersy: I'd rather keep this change to the minimum required for Ethers.js, because its kinda hacky (we now have 2 different block representations ) and then spend some more time figuring out how to handle the other fields properly.

internal/ethapi/api.go
piersy: I'm leaning towards no mocking, because if we switch to actually putting these fields on the block then there will be a point when the fields don't exist. Otherwise we have to maintain this shim forever. But I'm really not sure.

ethersjs-api-check/test/test.ts
piersy: Ok so just to clarify, the e2e test was controlling which test to run from the set of tests and would only run the tests for the the pruned state when it had pruned the state.

piersy:
@gastonponti I've updated this PR to not return the fields on the block when we can't retrieve them. I've also split the tests up under 2 describe headings one for tests with state and one for tests without. And finally added a few more tests.

Please take a look

e2e_test/ethersjs-api-check/test/test.ts
```diff
@@ -0,0 +1,94 @@
...
+			[ethers.utils.hexValue(process.env.npm_config_blocknum as string), true]
+		)
+		assert.isFalse(fullBlock.hasOwnProperty('baseFeePerGas'))
+	});
+
+
+});
+
+
+// const fullBlock = await provider.send(
```
gastonponti, line 91: nit: clean this?

gastonponti (approved):
I think you left some code used for debugging. Everything else, LGTM

PR #1958, opened 2022-09-22
Refactor downloader

gastonponti:
### Description

Refactor in downloader removing the logic required due to the old way we calculated the uptimeScore

### Other changes

Restore some tests from upstream

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 072606bf170b0db7587a7442b9ebf4b9f1f0e7fb
<details><summary>coverage:  51.0% of statements across all listed packages</summary><br>
<pre>
coverage:  64.0% of statements in consensus/istanbul
coverage:  55.8% of statements in consensus/istanbul/backend
coverage:  67.0% of statements in consensus/istanbul/core
</pre>
CommentID: 0ec3d51ccc
</details>

eth/downloader/downloader.go
```diff
@@ -1309,24 +1287,18 @@ func (d *Downloader) fetchHeaders(p *peerConnection, from uint64, height uint64)
...
 					// If we're still skeleton filling fast sync, check pivot staleness
 					// before continuing to the next skeleton filling
-
-					// TODO(ponti): Re add the "moving" pivot, after changing the way we calculate the uptimeScore
-					// if skeleton && pivot > 0 {
-					// 	getNextPivot()
-					// } else {
-					getHeaders(from)
-					// }
+					if skeleton && pivot > 0 {
```
piersy, line 1290: My understanding of this check is that `pivot` is set from `pivotHeader` and `pivotHeader` is onlyset when performing fast sync, so this check is essentially checking whether we are still skeleton filling and performing a fast sync. But I'd just like to double check that with your understanding.

piersy: Also why do we only update the pivot if we're still skeleton filling, I'd expect to want to update the pivot continuously regardless of where we are with respect to downloading skeleton headers or just headers?

eth/downloader/downloader_test.go
```diff
@@ -921,6 +921,58 @@ func testMultiProtoSync(t *testing.T, protocol uint, mode SyncMode) {
 	}
 }
 
+// Tests that if a block is empty (e.g. header only), no body request should be
+// made, and instead the header should be assembled into a whole block in itself.
+func TestEmptyShortCircuit67Full(t *testing.T)  { testEmptyShortCircuit(t, istanbul.Celo67, FullSync) }
+func TestEmptyShortCircuit67Fast(t *testing.T)  { testEmptyShortCircuit(t, istanbul.Celo67, FastSync) }
+func TestEmptyShortCircuit67Light(t *testing.T) { testEmptyShortCircuit(t, istanbul.Celo67, LightSync) }
```
piersy, line 928: Where did these tests come from?

eth/downloader/downloader.go
```diff
@@ -1987,17 +1923,22 @@ func (d *Downloader) processFastSyncContent() error {
...
 			// late and will drop peers due to unavailable state!!!
-			if height := latest.Number.Uint64(); height > pivot.Number.Uint64()+2*max(d.epoch, fsMinFullBlocks)-uint64(reorgProtHeaderDelay) {
-				newPivot := d.calcPivot(height)
-				log.Warn("Pivot became stale, moving", "old", pivot, "new", newPivot)
+			// BUT we are not expecting chain reorgs in IBFT (so we could assume it as zero)
+			headerDelay := reorgProtHeaderDelay
+			if d.ibftConsensus {
+				headerDelay = 0
+			}
+			if height := latest.Number.Uint64(); height > pivot.Number.Uint64()+2*uint64(fsMinFullBlocks)-uint64(headerDelay) {
```
piersy, line 1931: So the `reorgProtHeaderDelay` was added in 6ee3b26f447459d3f3a316dbb572e461a273e193 and IIUC its intended use at the time was to prevent syncing headers at the very tip of the chain in the case that the sync head was not close to the tip, in order to avoid a situation where a node syncs a header that then gets re-orged. But here it appears that the `reorgProtHeaderDelay` is being used in a different way, looking at the above comment it seems that they are using `reorgProtHeaderDelay` to avoid reaching a point where the pivot point is stale and requests for state start failing. It seems that we would want this.

piersy:
Hey @gastonponti, I've finished an initial pass on this, left a few questions for you!

eth/downloader/downloader_test.go
gastonponti: Upstream. We removed those before. I restore them

eth/downloader/downloader.go
gastonponti: Yes, agree on that, but in our case, check this line:
https://github.com/celo-org/celo-blockchain/pull/1958/files#diff-80cc409bf7f1e4a3d19174ed878e4dcb3d40b19b6e0580bd3816a174c57e0fb2R1227
We already removed that delay because we are IBFT.
So in this case it's maintaining the same logic, as there are no reorg, we removed that delay, so we shouldn' take that delay for this scenario

eth/downloader/downloader.go
piersy: Good to know that you'll be in the fill skeleton stage until you reach the pivot, but I don't understand your point about the ton of problems. The pivot definitely gets changed after starting to download state, because state only exists for the 128 most recent blocks and there's no way you can download the whole state in that time, so the pivot has to be updated probably roughly every 10 minutes (because 128 blocks take 10.66 minutes)

eth/downloader/downloader.go
piersy: Ok I've got it. Because eth was delaying pulling in `headerDelay` headers the sync head is `headerDelay` behind the real head and because state is pruned at 128 blocks after the real head we need to adjust our calculation of the prune point to take into account that the sync head is behind the real head.

eth/downloader/downloader.go
piersy: So from our chat, snap and fast sync are the only syncs that use a pivot, so `pivot > 0` is equivalent to `syncmode == FAST|SNAP`.

And we agreed that the pivot is constantly being updated, this is happening in `processFastSyncContent`

piersy: So my current understanding is that in `fetchHeaders` the pivot is updated to be 64 blocks deep whenever it approaches 128 blocks deep, that pivot header is stored in `Downloader.pivotHeader` which is used by `processFastSyncContent, but for some reason `processFastSyncContent` also does some adjustment of the pivot to prevent staleness of the pivot.

piersy (approved):
I still don't quite understand the pivot management, but it's upstream code.

PR #1961, opened 2022-09-27
Fix data race in downloader

palango:
### Description

The race detector found a problem when running with `go test -race -count=1 ./eth/downloader`.

It turns out that
https://github.com/celo-org/celo-blockchain/blob/67b6239118b04abf329a8cafea533727603d0260/eth/downloader/queue.go#L505 emits a pointer to a value owned by the `resultstore`.

This value is later changed here (added in the [latest upstream pull](https://github.com/celo-org/celo-blockchain/pull/1593)):
https://github.com/celo-org/celo-blockchain/blob/620b5fca2b8eeec1a00c9246c9141264991cfda1/eth/downloader/queue.go#L531-L534 but also accessed here in `AllDone`:
https://github.com/celo-org/celo-blockchain/blob/67b6239118b04abf329a8cafea533727603d0260/eth/downloader/resultstore.go#L154

I guess this patch is the easy solution to avoid the race, but there might be a better way to avoid this in the first place. It seems ugly that `pending` is accessed in the first place there. I'm open to suggestions.

### Other changes

Some typos.

### Tested

This triggered the race detector when run with `go test -race -count=1 ./eth/downloader`. With these changes it didn't trigger locally with a count of 20.

### Related issues
/
### Backwards compatibility

/

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -55,8 +55,9 @@ type Header struct {
 	Extra       []byte         `json:"extraData"        gencodec:"required"`
 
 	// Used to cache deserialized istanbul extra data
-	extraLock         sync.Mutex
-	deserializedExtra *IstanbulExtra
+	extraLock  sync.Mutex
+	extraValue *IstanbulExtra
+	extraError error
 }
 
 // field type overrides for gencodec
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"math/big"
 	"reflect"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -52,6 +53,10 @@ type Header struct {
 	GasUsed     uint64         `json:"gasUsed"          gencodec:"required"`
 	Time        uint64         `json:"timestamp"        gencodec:"required"`
 	Extra       []byte         `json:"extraData"        gencodec:"required"`
+
+	// Used to cache deserialized istanbul extra data
+	extraLock         sync.Mutex
+	deserializedExtra *IstanbulExtra
 }
 
 // field type overrides for gencodec
@@ -100,6 +105,10 @@ func (h *Header) SanityCheck() error {
 // EmptyBody returns true if there is no additional 'body' to complete the header
 // that is: no transactions.
 func (h *Header) EmptyBody() bool {
+	if _, err := h.IstanbulExtra(); err == nil {
+		return false
+	}
+
 	return h.TxHash == EmptyRootHash
 }
 
@@ -271,8 +280,19 @@ func NewBlockWithHeader(header *Header) *Block {
 // CopyHeader creates a deep copy of a block header to prevent side effects from
 // modifying a header variable.
 func CopyHeader(h *Header) *Header {
-	cpy := *h
-	if cpy.Number = new(big.Int); h.Number != nil {
+	cpy := Header{
+		ParentHash:  h.ParentHash,
+		Coinbase:    h.Coinbase,
+		Root:        h.Root,
+		TxHash:      h.TxHash,
+		ReceiptHash: h.ReceiptHash,
+		Bloom:       h.Bloom,
+		Number:      new(big.Int),
+		GasUsed:     h.GasUsed,
+		Time:        h.Time,
+	}
+
+	if h.Number != nil {
 		cpy.Number.Set(h.Number)
 	}
 	if len(h.Extra) > 0 {
@@ -366,10 +386,10 @@ func (c *writeCounter) Write(b []byte) (int, error) {
 // WithHeader returns a new block with the data from b but the header replaced with
 // the sealed one.
 func (b *Block) WithHeader(header *Header) *Block {
-	cpy := *header
+	cpy := CopyHeader(header)
 
 	return &Block{
-		header:         &cpy,
+		header:         cpy,
 		transactions:   b.transactions,
 		randomness:     b.randomness,
 		epochSnarkData: b.epochSnarkData,
--- a/core/types/celo_additions.go
+++ b/core/types/celo_additions.go
@@ -1,18 +1,18 @@
 package types
 
+// IstanbulExtra returns the 'Extra' field of the header deserialized into an
+// IstanbulExtra struct, if there is an error deserializing the 'Extra' field
+// it will be returned.
 func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
 	h.extraLock.Lock()
 	defer h.extraLock.Unlock()
 
-	extraData := h.deserializedExtra
-
-	if extraData == nil {
+	if h.extraValue == nil && h.extraError == nil {
 		extra, err := extractIstanbulExtra(h)
-		if err != nil {
-			return nil, err
-		}
-		h.deserializedExtra = extra
+
+		h.extraValue = extra
+		h.extraError = err
 	}
 
-	return h.deserializedExtra, nil
+	return h.extraValue, h.extraError
 }
--- a/core/types/celo_additions.go
+++ b/core/types/celo_additions.go
@@ -1,0 +1,18 @@
+package types
+
+func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
+	h.extraLock.Lock()
+	defer h.extraLock.Unlock()
+
+	extraData := h.deserializedExtra
+
+	if extraData == nil {
+		extra, err := extractIstanbulExtra(h)
+		if err != nil {
+			return nil, err
+		}
+		h.deserializedExtra = extra
+	}
+
+	return h.deserializedExtra, nil
+}
--- a/eth/downloader/queue.go
+++ b/eth/downloader/queue.go
@@ -527,11 +527,6 @@ func (q *queue) reserveHeaders(p *peerConnection, count int, taskPool map[common
 			// There are no resultslots available. Leave it in the task queue
 			break
 		}
-		// Only required if the reserve is for a body type
-		if kind == bodyType {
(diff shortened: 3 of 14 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 67b6239118b04abf329a8cafea533727603d0260
<details><summary>coverage:  47.7% of statements across all listed packages</summary><br>
<pre>
coverage:  52.9% of statements in consensus/istanbul
coverage:  56.1% of statements in consensus/istanbul/backend
coverage:  55.4% of statements in consensus/istanbul/core
</pre>
CommentID: a3fcedc7ea
</details>

palango:
An possible alternative would be to add something like:
```go
// SetBodyUndone flags the body as not finished.
func (f *fetchResult) SetBodyUndone() {
	if v := atomic.LoadInt32(&f.pending); (v & (1 << bodyType)) == 0 {
		atomic.AddInt32(&f.pending, 1)
	}
}
```

eth/downloader/resultstore.go
```diff
@@ -141,7 +141,7 @@ func (r *resultStore) HasCompletedItems() bool {
 // countCompleted returns the number of items ready for delivery, stopping at
 // the first non-complete item.
 //
-// The mthod assumes (at least) rlock is held.
+// The method assumes (at least) r.lock is held.
```
piersy, line 144: I think this comment was originally intending to say that the read lock was held for `countCompleted` because it says "at least". So now you've changed it to say that the lock is held, was that intentional?

eth/downloader/queue.go
```diff
@@ -529,8 +529,10 @@ func (q *queue) reserveHeaders(p *peerConnection, count int, taskPool map[common
 		}
 		// Only required if the reserve is for a body type
 		if kind == bodyType {
+			q.resultCache.lock.Lock()
```
piersy, line 532: Hi @palango, while this does fix the data race that we were seeing, I don't think it's a good approach because I think it's a misuse of `resultCache.lock`. I believe `resultCache.lock` is intended to protect fields of result cache from concurrent use, but here it is being used to protect a `fetchResult` that just happens to be accessed in the resultCache in a function also protected by that lock. `fetchResult` is a type that is built for concurrent access and so I think a more appropriate solution would be what you suggested in your comment.

However, I think that maybe we don't even need to add a `SetBodyUndone` because the celo chain always needs to download the block body for any header it downloads because the block contains some extra fields added by us that IIUC are required for both fast sync and lightest sync (and of course full sync).

https://github.com/celo-org/celo-blockchain/blob/2e37ebe19fcc83a43b067f21cbacaf385156ab13/core/types/block.go#L207-L208

So it may be sufficient to simply ensure that all `fetchResults` are created with the body set as pending, by removing the check in `newFetchResult`. The check for empty body there just checks to see if there are txs in the block, it doesn't take into account randomness or epochSnarkData. Alternatively we could just have `Header.EmptyBody` always return false.

https://github.com/celo-org/celo-blockchain/blob/2e37ebe19fcc83a43b067f21cbacaf385156ab13/eth/downloader/queue.go#L78-L80

I'm not sure why @gastonponti chose the approach that he did here, so lets wait for his input before choosing an approach.

palango: > Hi @palango, while this does fix the data race that we were seeing, I don't think it's a good approach because I think it's a misuse of `resultCache.lock`. I believe `resultCache.lock` is intended to protect fields of result cache from concurrent use, but here it is being used to protect a `fetchResult` that just happens to be accessed in the resultCache in a function also protected by that lock.

I agree that the current addition of locks isn't nice, but I think the above is incorrect. The `resultStore` own the `fetchResult`s until they are pulled via `GetCompleted`. In this case, `AddFetch` is used to create a new item in the store, so access should use the same lock.

> `fetchResult` is a type that is built for concurrent access and so I think a more appropriate solution would be what you suggested in your comment.

Wouldn't it need it's own lock in that case?


Anyhow, I like the idea you propose. Let's what @gastonponti thinks.

eth/downloader/resultstore.go
piersy: Hey @palango Are you saying the change is intentional or not?

eth/downloader/queue.go
gastonponti: Ok, thanks @piersy to point me that I was checking another line, haha.
Yes, I added that line to force to download every block because we need it for the randomness (probably a bad call) 
The actual problem that we have is that `header.EmptyBody`, is returning true. We should change that. Maybe use the Extra field from the Header to send a valid response

eth/downloader/resultstore.go
palango: I thought this referred to the field in the struct, but your intepretation makes more sense, so I'm gonna revert this.

eth/downloader/queue.go
piersy: So what I can think of here is that we add a method to `types.Header` to return the deserialized extra data. And then we can call that in `EmptyBody` to see if the header is an `istanbul` header,  if its an istanbul header then we should always return `false` for `EmptyBody`.

But this could turn out to be a bit of a performance bottleneck and so I'd suggest actually caching the deserialized results on the header in a private field, and at that point we should probably move to always using that method when we want to access the extra data.

core/types/celo_additions.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+func (h *Header) IsIstanbulHeader() bool {
```
piersy, line 3: I'd suggest instead just returning the deserialized extra data, so something like. Then we can use that function wherever we were previously calling `ExtractIstanbulExtra`
```
func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
```

palango: Done in [377d8bf](https://github.com/celo-org/celo-blockchain/pull/1961/commits/377d8bfa57c5d7a9b6753b1c8629cc753a218c52)

core/types/celo_additions.go
```diff
@@ -0,0 +1,13 @@
...
+func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
+	if h.deserializedExtra == nil {
+		extra, err := ExtractIstanbulExtra(h)
+		if err != nil {
+			return nil, err
+		}
+		h.deserializedExtra = extra
+	}
+
+	return h.deserializedExtra, nil
```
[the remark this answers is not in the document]
palango: Mh, in the worst case it get's decoded twice. Do you think the lock it worth the price for avoiding this?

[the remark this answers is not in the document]
piersy: Without a lock its undefined when a go-routine will see changes made from other go routines, so in the worst case it will get decoded on every call. Also this code would still trigger the race detector.

core/types/celo_additions.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
+	if h.deserializedExtra == nil {
+		extra, err := ExtractIstanbulExtra(h)
```
piersy, line 5: Since `ExtractIstanbulExtra` is now only used in `types` we could make it unexported which would force code to only use this function on the header.

core/types/celo_additions.go
```diff
@@ -1,13 +1,21 @@
...
 			return nil, err
 		}
+		h.extraLock.Lock()
 		h.deserializedExtra = extra
+		h.extraLock.Unlock()
 	}
 
+	h.extraLock.RLock()
+	defer h.extraLock.RUnlock()
 	return h.deserializedExtra, nil
```
piersy, line 20: I think just one lock over the whole func here would be clearer, as it stands this code could still allow the deserialization to happen multiple times.

palango: True, the idea was that there could be multiple readers later, but then I'll change it to use a `Mutex` instead.

piersy: Yeah, I thought about having multiple readers too, but I don't think this is something we are going to see a lot of contention on.

core/types/celo_additions.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
```
piersy, line 3:
```suggestion
// IstanbulExtra returns the 'Extra' field of the header deserialized into an
// IstanbulExtra struct, if there is an error deserializing the 'Extra' field
// it will be returned.
func (h *Header) IstanbulExtra() (*IstanbulExtra, error) {
```

core/types/celo_additions.go
```diff
@@ -0,0 +1,18 @@
...
+	if extraData == nil {
+		extra, err := extractIstanbulExtra(h)
+		if err != nil {
+			return nil, err
+		}
+		h.deserializedExtra = extra
+	}
+
+	return h.deserializedExtra, nil
+}
```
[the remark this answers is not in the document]
palango: Good point, added this.

core/types/block.go
```diff
@@ -99,8 +105,18 @@ func (h *Header) SanityCheck() error {
...
+func (h *Header) EmptyBody() (bool, error) {
+	extra, err := h.IstanbulExtra()
+
+	if err != nil {
+		return false, err
+	}
+
+	if extra != nil {
+		return false, nil
+	}
```
palango, line 117: @piersy I'm nor sure about this part. `h.IstanbulExtra` will always return either an `extra` or an `err`, so we now can never reach the statement in line 119, which doesn't make much sense.

palango: Done

piersy (approved):
Great!

PR #1964, opened 2022-09-28
Add PreprepareV2 RLP encoded size test

hbandura:


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit 3390d0f9a0ce4992f2566ff224bdb61013218bee
<details><summary>coverage:  46.6% of statements across all listed packages</summary><br>
<pre>
coverage:  52.8% of statements in consensus/istanbul
coverage:  52.2% of statements in consensus/istanbul/core
</pre>
CommentID: da573b479e
</details>

consensus/istanbul/types_v2_test.go
```diff
@@ -183,3 +187,187 @@ func TestRoundChangeV2RLPEncoding(t *testing.T) {
...
+	return header
+}
+
+func bigTxs(gasLimit int) []*types.Transaction {
+	curr := common.BytesToAddress(getF(20))
+	gateFeeRec := common.BytesToAddress(getF(20))
+	tx := types.NewTransaction(999, common.BytesToAddress(getF(20)),
+		big.NewInt(329274), 2942729, big.NewInt(294279),
+		&curr,
+		&gateFeeRec, big.NewInt(32027), getF(gasLimit/4))
```
gastonponti, line 266: extract that 4 as a different variable, and add a comment that that's the cost of an empty byte

gastonponti: Even better. Use the `TxDataZeroGas` constant in the `params/protocol_params.go` which is actually setting that scenario gas price

consensus/istanbul/types_v2_test.go
```diff
@@ -183,3 +188,188 @@ func TestRoundChangeV2RLPEncoding(t *testing.T) {
...
+func bigRCC(validators int) *RoundChangeCertificateV2 {
+	rcc := &RoundChangeCertificateV2{}
+	quorum := int(math.Ceil(float64(2*validators) / 3))
+	rcc.Requests = make([]RoundChangeRequest, quorum)
+	for i := 0; i < quorum; i++ {
+		rcc.Requests[i] = *bigRequest(quorum)
+	}
+	return rcc
+}
+
```
piersy, line 340:
```suggestion
// Constructs the largest possible consensus message in order to ensure that it
// does not exceed the built in size limits. Note that byte arrays are filled
// with the byte 0xFF, because that takes more space in RLP encoding than byte
// values < 128.
```

consensus/istanbul/types_v2_test.go
```diff
@@ -183,3 +188,188 @@ func TestRoundChangeV2RLPEncoding(t *testing.T) {
...
+func TestBigPreprepareV2Size(t *testing.T) {
+	gasLimit := 34_600_000
+	validators := 110
+	gasPerByte := params.TxDataZeroGas
+	sizeLimit := int(9.5 * 1024 * 1024)
```
piersy, line 345: What about referencing the actual message size limit here (`eth.maxMessageSize`)? It's currently not exported, but making it exported would be a very small change.

consensus/istanbul/types_v2_test.go
hbandura: Byte arrays are serialized with the 'string' logic, not the 'list' logic of RLP. So the value actually doesn't matter I think in most cases, but I used FF since that made it much more clear to the reader.

If you test serialization comparison of an array fill of zero and a different byte array filled with 0xFF, both will take the exact same space except for arrays of length one.

hbandura: (I'm still adding most of the comment)

consensus/istanbul/types_v2_test.go
hbandura: I'd rather keep it separated, but I agree that I don't have strong reasons for it

gastonponti (approved):
LGTM

PR #1969, opened 2022-10-13
Add job that build and stores executables

palango:
This might be helpful for external people who want to try out the client on a certain PR.

### Description

Adds a job to CI that build all executables and stores them as artifacts.

### Backwards compatibility

Only CI


piersy (approved):
Looks good to me!

PR #1860, opened 2022-02-28
Fix concurrent map access error when tracing blocks (#1822)

piersy:
Fixes  https://github.com/celo-org/celo-blockchain/issues/1799

This commit ensures that state.StateDB instances are not
accessed concurrently when tracing blocks and also adds
an e2e test to verify the fix.

It also disables logging from the e2e tests, because the logs
slowed the tests down and cluttered the output of test runs.

It also updates test node instances to ensure that all rpc api
modules are enabled.

PR #1987, opened 2022-12-02
Fix tracer block context

### LINKED ISSUE #1983: Cannot trace block #16292612 using callTracer
## Description

The need here is to trace block #16292612 while calling on method `debug_traceBlockByNumber` using the Tracer `callTracer`

## Current Behavior

➜  ~ curl https://RPC_PROVIDER/ \ 
 -X POST \                                                                       
 -H "Content-Type: application/json" \
 --data '{"method":"debug_traceBlockByNumber","params":["0xf89b04", {"tracer": "callTracer"}],"id":1,"jsonrpc":"2.0"}'  | jq '.result[93]'

{
  "error": "tracing failed: max fee per gas less than block base fee: address 0xadd7c542667728e6B91Dc091a6def2489f2dE35C, maxFeePerGas: 227247290 baseFee: 229157277"
}
## Desired Behavior

To see the trace results and not the failure I pasted above.

## Context

Operating System: "Ubuntu 20.04.5 LTS (Focal Fossa)"
Client Version: celo/v1.7.0-stable-5285ca48/linux-amd64/go1.13.8

## Proposed Solution

"Pending"

gastonponti:
### Description

The tracer was not using the same block context for every tx (it was generating a new context for every tx)

### Related issues

- Fixes #1983 

### Backwards compatibility

Yes

### DIFF
--- a/core/sys_context.go
+++ b/core/sys_context.go
@@ -14,6 +14,7 @@ import (
 
 // SysContractCallCtx acts as a cache holding information obtained through
 // system contract calls to be used during block processing.
+// Note: This struct should be a read only one to be safe for concurrent use
 type SysContractCallCtx struct {
 	whitelistedCurrencies map[common.Address]struct{}
 	// The gas required for a non celo (cUSD, cEUR, ...) transfer.
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -599,16 +599,16 @@ func (api *API) traceBlock(ctx context.Context, block *types.Block, config *Trac
 	}
 	blockCtx := core.NewEVMBlockContext(block.Header(), api.chainContext(ctx), nil)
 	blockHash := block.Hash()
+	var sysCtx *core.SysContractCallCtx
+	if api.backend.ChainConfig().IsEspresso(block.Number()) {
+		sysCtx = core.NewSysContractCallCtx(block.Header(), statedb, api.backend)
+	}
 	for th := 0; th < threads; th++ {
 		pend.Add(1)
 		go func() {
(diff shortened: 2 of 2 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit d60bdd21e48f1f41cd655cde237e52df54f10482
<details><summary>coverage:  47.4% of statements across all listed packages</summary><br>
<pre>
coverage:  52.4% of statements in consensus/istanbul
coverage:  56.0% of statements in consensus/istanbul/backend
coverage:  54.6% of statements in consensus/istanbul/core
</pre>
CommentID: 442e2e816d
</details>

piersy (changes requested):
Hey @gastonponti I think this change may re-introduce a problem that I previously fixed.

* #1860 

StateDB instances are not safe for concurrent use, so you have to use a copy, or ensure that all interaction is done in a single thread.

-------
Update:

@gastonponti Following our conversation I believe that this actually is safe. In #1860 I was incorrectly building the block context using the intermediate state after each transaction application, where in fact the block context is intended to be constructed with the state at the beginning of the block and re-used for all transactions in a block. This was causing the results of block tracing to diverge from execution in the case that any transaction affected the results given by the block context. 

Looking at the `SysContractCallCtx` constructor I can see that the provided `StateDB` is only used in the constructor and not retained by the struct, this makes the `SysContractCallCtx` safe for concurrent use.

hbandura (approved):

piersy (approved):
Looks good

PR #1991, opened 2023-01-09
Switch pr comment user to be celo-org

piersy:
Previously it was piersy but now we are using an org based access token.


piersy:
This didn't work

PR #1996, opened 2023-01-13
Ensure RoundStateDB is closed when core is stopped

piersy:
Prior to this change the RoundStateDB was never closed. This could result in leaked goroutines from its internal leveldb instance.

This change caused the TestStartStopValidators e2e test to fail because changing the rsdb open to be in the startup code, slowed down the start method for core this caused the gossiping of the manually added peers (done in the test after starting each previously stopped node )to happen after the validator peer handler had removed all the peers resulting in gossiping certs to no one. Thus leaving the network in a disconnected state.

The fix was really hacky, but just to wait 250ms after starting a node and hoping that the validator peer handler will have removed nodes by that point and then add the peers and then wait a little for them to be registered and then finally gossip to them.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -240,7 +240,10 @@ func (w *worker) stop() {
 	atomic.StoreInt32(&w.running, 0)
 
 	if istanbul, ok := w.engine.(consensus.Istanbul); ok {
-		istanbul.StopValidating()
+		err := istanbul.StopValidating()
+		if err != nil {
+			log.Error("Error while calling engine.StopValidating", "err", err)
+		}
 	}
 }
 
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -152,10 +152,6 @@ type core struct {
 
 // New creates an Istanbul consensus core
 func New(backend CoreBackend, config *istanbul.Config) Engine {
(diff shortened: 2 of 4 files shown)


piersy:
Coverage from tests in `./e2e_test/...` for `./consensus/istanbul/...` at commit fd62a28feb03ea84d3de66c5f6f461c6d563bfc6
<details><summary>coverage:  46.9% of statements across all listed packages</summary><br>
<pre>
coverage:  52.9% of statements in consensus/istanbul
coverage:  56.0% of statements in consensus/istanbul/backend
coverage:  52.6% of statements in consensus/istanbul/core
</pre>
CommentID: 9aa11c9703
</details>

miner/worker.go
```diff
@@ -240,7 +240,10 @@ func (w *worker) stop() {
 	atomic.StoreInt32(&w.running, 0)
 
 	if istanbul, ok := w.engine.(consensus.Istanbul); ok {
-		istanbul.StopValidating()
+		err := istanbul.StopValidating()
```
palango, line 243: Is this related to the shutdown of the db?

piersy: Partially, `StopValidating` could have also returned an error before this change, but this change added another condition where `StopValidating` could return an error.

palango (approved):
Nice find, makes sense to me.

The wait in the e2e is unfortunate, but I don't see a better way.

PR #2004, opened 2023-01-30
BLS verification in background thread

piersy:
Added functions to CommittedSubject to perform bls verification in background thread. In order to maintain the result of the verification moved to passing around deserialized istanbul.Message instances instead of the payload bytes. Ran into problems when handling proxy fwd messages (there is compile error).

This is proving overly complex because of the tangled nature of the istanbul code.

Suggestion would be to clean the code (refactor)and then attempt this change.


palango:
Closing this as we don't plan to include this in the short term.

PR #2009, opened 2023-02-13
Update error msg when contracts are not made yet

karlb:
There is no `compiled-system-contracts` make target, but `make prepare-system-contracts` creates the necessary files.


palango (approved):
Thanks!

Issue #2005, opened 2023-02-05
Upgrade dependency "github.com/celo-org/celo-bls-go"


Ben131-Go:
## Background
Repo `github.com/celo-org/celo-blockchain` depends on `github.com/celo-org/celo-bls-go@v0.3.3`.<br>
<https://github.com/celo-org/celo-blockchain/blob/master/go.mod#L19><br>
However, comparing version `v0.3.3` of `github.com/celo-org/celo-bls-go` from proxy.golang.org and github, there are inconsistencies.<br>
> [commit time of the copy on github.com](https://api.github.com/repos/celo-org/celo-bls-go/commits/d06a312)
```txt
"committer": {
      "name": "Michael Straka",
      "email": "mstraka100@gmail.com",
      "date": "2022-09-19T14:44:27Z"
    }
```
> [commit time of the copy on proxy.golang.org](https://proxy.golang.org/github.com/celo-org/celo-bls-go/@v/v0.3.3.info)
```txt
{"Version":"v0.3.3","Time":"2022-09-16T15:11:29Z","Origin":{"VCS":"git","URL":"https://github.com/celo-org/celo-bls-go","Ref":"refs/tags/v0.3.3","Hash":"c5db50e963c35e5f87033fe2d76c3eae6e344c6c"}}
```
So the checksum from the code in github does not match the checksum saved in sum.golang.org. The `v0.3.3` tag of `github.com/celo-org/celo-bls-go` might have been retagged after a minor edition on github. I guess you use proxy.golang.org to get dependencies, but that also shows that your project is depending on the copy of `github.com/celo-org/celo-bls-go@v0.3.3` before its edition. Depending upon such inconsistent tag version may also result in some unexpected errors as well as build errors due to different proxy settings.<br>
For example, when someone who does not use proxy.golang.org, say `GOPROXY=direct`, attempts to get `github.com/celo-org/celo-bls-go@v0.3.3`, the following error occurs.
```Go
go: downloading github.com/celo-org/celo-bls-go v0.3.3
go: github.com/celo-org/celo-bls-go@v0.3.3: verifying module: checksum mismatch
        downloaded: h1:m1Z51zo7z63zQhvfo3IUmEDb3S2OcKMAzheT5QeZ8J4=
        sum.golang.org: h1:kOkm/BeM0YwHWUTwSLprxhk1IFq/ZjaWVtW6jTZ0+ys=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
[...]

PR #2012, opened 2023-02-15
Update celo-bls-go version to v0.3.4

mstraka100:
### Description

Updates celo-bls-go dependency to v0.3.4, which is identical to v0.3.4 except it is in sync with the commit in https://proxy.golang.org/github.com/celo-org/celo-bls-go/@v/v0.3.4.info. This is to fix https://github.com/celo-org/celo-blockchain/issues/2005.

### Tested

No functional code changes were made. CI tests should be sufficient.


go.sum
```diff
@@ -108,18 +108,32 @@ github.com/buraksezer/consistent v0.0.0-20191006190839-693edf70fd72/go.mod h1:OE
 github.com/c-bata/go-prompt v0.2.2/go.mod h1:VzqtzE2ksDBcdln8G7mk2RX9QyGjH+OVqOCSiVIqS34=
 github.com/celo-org/celo-bls-go v0.3.3 h1:kOkm/BeM0YwHWUTwSLprxhk1IFq/ZjaWVtW6jTZ0+ys=
 github.com/celo-org/celo-bls-go v0.3.3/go.mod h1:eoMAORYWgZ5HOo3Z0bIAv/nbPtj/eArIO0nh7XFx7rk=
+github.com/celo-org/celo-bls-go v0.3.4 h1:slNePT/gVjgUi7f8M4KTwBz/YYgv3JWU6XqyY0xKN84=
```
carterqw2, line 111: Hmm, that's weird that we have dependencies for two different versions for all `celo-bls-go-*` packages now.

palango:
Did you create a new `celo-bls-go` release for this or did that already exist?

mstraka100:
> Did you create a new `celo-bls-go` release for this or did that already exist?

I made a new version (v0.3.4) which was functionally identical to the last (v0.3.3). The celo-bls-go subrepos are still at v0.3.3 since they didn't need to be changed and updating them would just have introduced unnecessary complexity

palango:
> > Did you create a new `celo-bls-go` release for this or did that already exist?
> 
> I made a new version (v0.3.4) which was functionally identical to the last (v0.3.3). The celo-bls-go subrepos are still at v0.3.3 since they didn't need to be changed and updating them would just have introduced unnecessary complexity

Will that tag be merged into master as well?

palango (approved):
👍

mstraka100:
> > > > Did you create a new `celo-bls-go` release for this or did that already exist?
> > > 
> > > 
> > > I made a new version (v0.3.4) which was functionally identical to the last (v0.3.3). The celo-bls-go subrepos are still at v0.3.3 since they didn't need to be changed and updating them would just have introduced unnecessary complexity
> > 
> > 
> > Will that tag be merged into master as well?
> 
> What do you mean by this? The current PR just updates the `celo-bls-go` dependency to v0.3.4, keeping the other `celo-bls-go-*` repos as v0.3.3.

If you're asking will it be merged into master of celo-bls-go, that's a good question. The branch of v0.3.* is not intended to be merged into master. It is a stopgap solution to avoid using unaudited cryptography code (BLS batch verification) as a dependency in celo-blockchain, which is in master on celo-bls-go as v0.6.*. It may be worth discussing whether we want to simply use this version in celo-blockchain anyways to point to master of celo-bls-go since it might be a while before we get this code audited with its deprioritization. We could just make sure to be explicit in the celo-bls-go repo description that this segment is unaudited.

PR #2013, opened 2023-02-15
testing akeyless workflow

pputman-clabs:
### Description

Changes the PR_COMMENT_TOKEN used to authenticate to github, to be a github organization app token instead of a personal access token, but generated by akeyless for each run.

Now it should use the akeyless context in circle ci to access the accessid variable, which is a nonsensitive id to know which akeyless auth method to use for circle ci.  This should use oauth2 to authenticate to akeyless by the circle CI organization and project ID, and have access to generate a temporary secret that gives access to write to PRs for the the celo-blockchain  github repository.

Then using the akeyless command line utility to generate the new key, it saves it and persists it via workspaces to share with the necessary jobs.

### Tested

Testing with this PR

### Backwards compatibility

Should be fully backwards compatible, leaving piersy's PR_COMMENT_TOKEN for now so it won't break anything while testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -279,6 +279,34 @@ jobs:
       - run: go mod download
       - *save-go-mod-cache
 
+  akeyless-get-token:
+    docker:
+      - image: akeyless/ci_base
+    environment:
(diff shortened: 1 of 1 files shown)


carterqw2 (approved):
Awesome work, thanks @pputman-clabs!

Issue #2015, opened 2023-02-19
GraphQL returns incorrect gasPrice


kamikazechaser:
## Description

For EIP 1559 transactions, the GraphQL endpoint returns an incorrect `gasPrice`.

## Current Behavior

 It seems to return the `maxFeePerGas` value, which is incorrect.

## Context

Bot the RPC `eth_getTransactionByHash` and GraphQL should return the same `gasPrice` value.

## Example

On Alfajores tx with hash: `0xe662ea49a30baf976e6372a5bf3392bd2d632f7f171225deb31d1c11f05fe761` which was submitted with the opts:

- MaxFee: 35 Gwei
- Tip: 1 wei

The `gasPrice` should be ~ 5 gwei

RPC:

```bash
╭─ [kamikaze:argon] in ~ 
╰─➜  curl --data '{"method":"eth_getTransactionByHash","params":["0xe662ea49a30baf976e6372a5bf3392bd2d632f7f171225deb31d1c11f05fe761"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST $RPC_ENDPOINT | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   916    0   776  100   140    636    114  0:00:01  0:00:01 --:--:--   751
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "blockHash": "0xe19b0df06e75d515af25964b405f46e495a2bab477a6ab981d2741ed18f75dd9",
    "blockNumber": "0xf88143",
    "from": "0x194cf4b3027e17a1bdedd87c52ade2c5c5c19026",
    "gas": "0x6aa4",
    "gasPrice": "0x12a05f201",
    "maxFeePerGas": "0x826299e00",
    "maxPriorityFeePerGas": "0x1",
    "feeCurrency": null,
    "gatewayFeeRecipient": null,
    "gatewayFee": "0x0",
    "hash": "0xe662ea49a30baf976e6372a5bf3392bd2d632f7f171225deb31d1c11f05fe761",
    "input": "0x",
    "nonce": "0x10",
    "to": "0x0577ae5a3547ac753ac499e282557ce0275a4942",
    "transactionIndex": "0x1",
    "value": "0x0",
    "type": "0x2",
    "accessList": [],
    "chainId": "0xaef3",
    "v": "0x0",
[...]

PR #2020, opened 2023-02-24
graphql: add EIP-1559 and celo gas handling

karlb:
### Description

The GraphQL-Endpoint did not properly handle post-EIP1559 gas fees. To get the correct results based on the respective baseFee, we have to take a look at the used feeCurrency.

### Tested

To get at least some test coverage for this, I added one EIP-1559 transaction to the test data. This is still far from a thorough coverage of all relevant cases (different feeCurrency, the base fee is 0 in the test, so bugs for proper values might be hidden).
No testing outside of the unit tests, yet.

### Related issues

Fixes https://github.com/celo-org/celo-blockchain/issues/2015

### Backwards compatibility

The GraphQL endpoint returns different values than before. Considering that values which changes were wrong (effectiveGasFee) or useless (baseFee was always null), the changes seem reasonable.

For information in the difference between gasPrice and effectiveGasPrive, see also https://github.com/ethereum/go-ethereum/issues/23124#issuecomment-872052444


palango (approved):
Looks good to me, let's get this merged. 

Great that you added a test!

PR #2024, opened 2023-02-27
Add fork diff config and actions job

palango:
This will create a website that lists our changesto upstream geth. This is similar to https://op-geth.optimism.io.

For now I didn't add many categories, this should be done later.

### Todos
- [x] Create a subdomain to host this
  - Pushed to https://celo-org.github.io/celo-blockchain/


fork.yaml
```diff
@@ -0,0 +1,30 @@
+title: "celo - go-ethereum fork diff overview"
```
karlb, line 1: Do we want this file in the root directory? We can put it somewhere also to avoid clutter in the root dir.

karlb (commented):
Do we have any way to test this before merging to master? We could temporarily relax the `on` condition and change it back before merging if there is no more elegant way.

palango:
once approved I need to change the branch back to `master`

.github/fork.yaml
```diff
@@ -0,0 +1,30 @@
...
+  description: |
+    This is an overview of the changes in [`celo`](https://github.com/celo-org/celo-blockchain),
+    a fork of [`go-ethereum`](https://github.com/ethereum/go-ethereum).
+  sub:
+    - title: "Core modifications"
+      sub:
+        - title: "Consensus modifications"
+          description: ""
+          globs:
+            - "consensus/instanbul/*"
```
karlb, line 23: For some reason this glob does not match, although it looks good to me.

karlb: Ah, I found the problem: It should be istanbul, not i**n**stanbul!

karlb (approved):
Looks great! We just need some more categories to make it more readable.

Do we want to ignore the docs, or just have them in a separate category?

karlb:
And an ignore pattern for `go-ethereum` -> `celo-blockchain` should also greatly reduce the diff size.

palango:
> Do we want to ignore the docs, or just have them in a separate category?

I think ignoring would be fine.

palango:
> > And an ignore pattern for `go-ethereum` -> `celo-blockchain` should also greatly reduce the diff size.
> 
> I wonder if we really need this difference at all. https://op-geth.optimism.io/ just keeps the go-ethereum name and even if we ignore the pattern here, it will still cause merge conflicts when we rebase on a newer geth version.

Yeah, fixing the diff in CI doesn't work as the tool uses the git ref. But I'm not sure why this is done at all.

palango:
I added more categories, I think the next step would be to filter out the import changes.

eelanagaraj:
Closing since these changes were merged as part of #2045

PR #2025, opened 2023-02-27
Migrate to Github Actions

carterqw2:
### Description

 - Migrates CI from CircleCI to Github Actions.
 - Increases e2e benchmark timeouts as using bigger runners will take some effort.
 - Puts the Android test behind a flag so that it's ignored if run in CI.
 - Removes akeyless github token dependency.

### Future improvements
  - move e2e tests to a reusable workflow to avoid repetition
  - figure out if we need our own codecov token

### Example
[Successful run](https://github.com/celo-org/celo-blockchain/actions/runs/4324501369/jobs/7549715922)


eelanagaraj (approved):
Looks good! I checked that all the old tests in CI seem to run in the new setup (numbers of individual test cases) with roughly the expected time.

.github/workflows/config.yml
```diff
@@ -0,0 +1,735 @@
...
+        ./build/cache/golangci-lint-1.42.0-linux-amd64/golangci-lint run --config .golangci.yml --out-format junit-xml ./... | tee /tmp/test-results/go-lint-report.xml
+    - name: Upload linter report
+      uses: actions/upload-artifact@v3
+      with:
+        name: linter-report
+        path: /tmp/test-results
+
+  unit-tests:
+    name: Unit tests
+    runs-on: ubuntu-latest
```
pputman-clabs, line 142: Defining permissions for the default github token will allow the job to write to the PR, (for codecoverage step).  This can be defined under unit-tests in line with name and runs-on

  permissions: 
      pull-requests: write

pputman-clabs: nevermind it's commenting already, closing this.

pputman-clabs:
Quick question, do you think it would be helpful to turn these into reusable workflows?  I have one for go tests here for my projects: https://github.com/celo-org/reusable-workflows/blob/main/.github/workflows/go-tests.yaml but I could use some feedback on improving it. Here's an example of what it looks like when called: https://github.com/celo-org/akeyless/blob/main/.github/workflows/akeyless-gcp-producer.yaml

  If we can make a standardized workflow that's usable in any situation with variables, it might be easier to maintain and set standards/best practices for testing across multiple repos and projects for the whole organization.

.github/workflows/config.yml
```diff
@@ -0,0 +1,735 @@
...
+        check-latest: false
+        cache: true
+    - name: Run linter
+      run: |
+        # so that the tee call at the end doesn't hide the real exit code
+        set -o pipefail
+        mkdir -p /tmp/test-results
+        go run build/ci.go ensure-linter
+        ./build/cache/golangci-lint-1.42.0-linux-amd64/golangci-lint run --config .golangci.yml --out-format junit-xml ./... | tee /tmp/test-results/go-lint-report.xml
+    - name: Upload linter report
```
palango, line 134: Can we show this as a comment instead of putting it into some artifact?

.github/workflows/config.yml
```diff
@@ -0,0 +1,735 @@
...
+    # TODO: figure out if we need a token
+    - name: Upload coverage to codecov
+      uses: codecov/codecov-action@v3
+    - name: Upload test report
+      uses: actions/upload-artifact@v3
+      with:
+        name: unit-test-report
+        path: /tmp/test-results
+
+  publish-test-results:
```
palango, line 179: nit: does it make sense to have an own job for this?

palango (commented):
Looks great in general, left some comments, but they're mostly for my understanding.

- Great that we can get rid of the custom docker image!
- Have you tested this with a PR coming from a fork? I know we always had problems with this, so it would be nice to see if this works now.
- Having a template for the e2e tests seems nice, be we can improve this later.
- Did you receive any answer regarding the use of bigger VMs for the heavy tests?

carterqw2:
> * Have you tested this with a PR coming from a fork? I know we always had problems with this, so it would be nice to see if this works now.

The problem with forks was that it couldn't obtain the akeyless token because of security constraints. I changed the implementation that has a way to solve this issue [with a proxy](https://github.com/mshick/add-pr-comment#proxy-for-fork-based-prs). Although this PR doesn't solve it, I was planning to add it later.

> * Having a template for the e2e tests seems nice, be we can improve this later.

Yes, I wanted to have this merged first.

> * Did you receive any answer regarding the use of bigger VMs for the heavy tests?

Yes, @pputman-clabs found a nice way to use self-hosted runners, so I'm optimistic that we can have a nice setup in the end.

.github/workflows/config.yml
```diff
@@ -0,0 +1,735 @@
...
+  # Location where compiled system contracts are stored under the root of this
+  # repo.
+  SYSTEM_CONTRACTS_PATH: "compiled-system-contracts"
+
+jobs:
+  prepare-system-contracts:
+    name: Build celo-monorepo
+    runs-on: ubuntu-latest
+    env:
+      NODE_VERSION: 12
```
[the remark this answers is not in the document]
carterqw2: No special reason.

.github/workflows/config.yml
```diff
@@ -0,0 +1,735 @@
...
+    - name: Conditionally build monorepo
+      if: steps.celo-monorepo-cache.outputs.cache-hit != 'true'
+      run: |
+        cd celo-monorepo
+        yarn
+        yarn build
+        cd packages/protocol
+        yarn run build:sol
+    - name: Upload built system contracts
+      if: always()
```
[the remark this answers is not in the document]
carterqw2: It would skip this step in the "best, everything is cached" scenario otherwise.

.github/workflows/config.yml
```diff
@@ -0,0 +1,735 @@
...
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+env:
+  # Increment these to force cache rebuilding
+  SYSTEM_CONTRACTS_CACHE_VERSION: 2
+  CHECKOUT_MONOREPO_CACHE_VERSION: 3
```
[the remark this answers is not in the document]
carterqw2: True, although I thought we might not need them at all since we can just manually delete the caches [here](https://github.com/celo-org/celo-blockchain/actions/caches) if needed. Left it as is just to replicate what we had.

.github/workflows/config.yml
carterqw2: Good point, changed it to an action, it should show errors right in the PR now.

.github/workflows/config.yml
carterqw2: Agreed, removed.

carterqw2:
> Quick question, do you think it would be helpful to turn these into reusable workflows? I have one for go tests here for my projects: https://github.com/celo-org/reusable-workflows/blob/main/.github/workflows/go-tests.yaml but I could use some feedback on improving it. Here's an example of what it looks like when called: https://github.com/celo-org/akeyless/blob/main/.github/workflows/akeyless-gcp-producer.yaml
> 
> If we can make a standardized workflow that's usable in any situation with variables, it might be easier to maintain and set standards/best practices for testing across multiple repos and projects for the whole organization.

Although I like the idea of reusable workflows in general, in this particular case it might be hard to generalize since different projects might need different golang versions or various additional steps, which would require extensive parameterization of the workflow and would kinda defeat the purpose. I think it works great for workflows that are the same for everybody, like the akeyless one for example.

.github/workflows/config.yml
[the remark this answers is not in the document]
palango: Ah nice, didn't know about that interface.

.github/workflows/config.yml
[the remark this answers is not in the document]
palango: Makes sense.

palango (approved):

PR #2029, opened 2023-03-02
CIP 55: Remove Tobin Tax

palango:
### Description

Remove tobin tax deduction with next hardfork.

### Tested

I tested this manually by running the `TestTransferCELO` test first without changes. This showed deduction of tobin tax. Then I changed the hardfork flag to always return true and made sure that the tobin tax deduction code is not run anymore.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain-planning/issues/18

### Backwards compatibility

Changes consensus rules and will therefor be enabled with the next hardfork.

### DIFF
--- a/core/vm/vmcontext/context.go
+++ b/core/vm/vmcontext/context.go
@@ -148,16 +148,16 @@ func VerifySealFn(ref *types.Header, chain chainContext) func(*types.Header) boo
 // If the calculation or transfer of the tax amount fails for any reason, the regular transfer goes ahead.
 // NB: Gas is not charged or accounted for this calculation.
 func TobinTransfer(evm *vm.EVM, sender, recipient common.Address, amount *big.Int) {
+	// Run only primary evm.Call() with tracer
+	if evm.GetDebug() {
+		evm.SetDebug(false)
+		defer func() { evm.SetDebug(true) }()
+	}
+
 	// Only deduct tobin tax before the g hardfork
 	if evm.ChainConfig().IsGFork(evm.Context.BlockNumber) {
 		Transfer(evm.StateDB, sender, recipient, amount)
 	} else {
-		// Run only primary evm.Call() with tracer
-		if evm.GetDebug() {
-			evm.SetDebug(false)
-			defer func() { evm.SetDebug(true) }()
-		}
-
 		if amount.Cmp(big.NewInt(0)) != 0 {
 			caller := &SharedEVMRunner{evm}
 			tax, taxRecipient, err := reserve.ComputeTobinTax(caller, sender, amount)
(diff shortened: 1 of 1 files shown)


core/vm/vmcontext/context.go
```diff
@@ -142,25 +142,30 @@ func VerifySealFn(ref *types.Header, chain chainContext) func(*types.Header) boo
...
+	if evm.ChainConfig().IsGFork(evm.Context.BlockNumber) {
+		Transfer(evm.StateDB, sender, recipient, amount)
+	} else {
+		// Run only primary evm.Call() with tracer
+		if evm.GetDebug() {
+			evm.SetDebug(false)
+			defer func() { evm.SetDebug(true) }()
+		}
 
-	if amount.Cmp(big.NewInt(0)) != 0 {
```
karlb, line 151: You could reduce this change to a one line diff with
```
	if amount.Cmp(big.NewInt(0)) != 0 && !evm.ChainConfig().IsGFork(evm.Context.BlockNumber) {
```
I'm not sure if that is clearer, though (and it would keep flipping the debugging off and on without need).

palango: I like the clear separation between logic and hardfork, so I'd prefer to keep it as is.

core/vm/vmcontext/context.go
```diff
@@ -142,25 +142,30 @@ func VerifySealFn(ref *types.Header, chain chainContext) func(*types.Header) boo
...
-	if evm.GetDebug() {
-		evm.SetDebug(false)
-		defer func() { evm.SetDebug(true) }()
-	}
+	// Only deduct tobin tax before the g hardfork
+	if evm.ChainConfig().IsGFork(evm.Context.BlockNumber) {
+		Transfer(evm.StateDB, sender, recipient, amount)
+	} else {
+		// Run only primary evm.Call() with tracer
+		if evm.GetDebug() {
```
palango, line 150: Mh, looking at this again, didn't this code before disable tracing for the expected part of the transfer?

eelanagaraj: It did disable it for the whole transfer but I think this is practically the same as what's happening here ^ since `Transfer` doesn't actually make any EVM/interpreter calls. It may be easier to reason about if you keep this on the outside of the hardfork logic anyways (should be functionally the same); also, this may be moved directly into the `SharedEVMRunner` anyways as part of this ticket https://github.com/celo-org/celo-blockchain/issues/2078

eelanagaraj (commented):
LG! Can you add a second test case, i.e. (`TestTransferCELOPreG`, `TestTransferCELOPostG`) which transfer before/after the G hardfork and checks expected behavior? I think it'd be really useful to have this going forwards to make sure we don't inadvertently mess with the old logic in the future. Wdyt? Happy to approve after that!

As far as your question/comment, I think maybe slightly preferable to move the debug setting to the top of the func (for easier modification later when redoing all the SharedEVMRunner trace fixes) but I think totally up to you!

palango:
> I think it'd be really useful to codify your testing (from the description) & have this going forwards to make sure we don't inadvertently mess with the old logic in the future. Wdyt? Happy to approve after that!

I agree it would be super useful, but the only way I see to do it involves a full e2e test at the moment, because we don't have the necessary setup logic for units/integration tests to setup the whole celo contract infrastructure. And as this is pretty simple, I'm not sure if the overhead of two e2e tests is worth it. What do you think?

Issue #1920, opened 2022-07-09
Segfault when doing fast forward receipt download


diwu1989:
Using latest stable version 1.5.5, hitting this segfault:
```
ERROR[07-09|06:27:16.927] Invalid block total difficulty RLP       hash=64c1b1..9023d4 err="rlp: value size exceeds available input length"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5275a5]

goroutine 4480 [running]:
math/big.(*Int).Cmp(0xa8f8b6642b0a06f1?, 0xa1e5406221dd63db?)
        math/big/int.go:328 +0x25
github.com/celo-org/celo-blockchain/core.(*BlockChain).InsertReceiptChain.func1(0xc0426d1a80)
        github.com/celo-org/celo-blockchain/core/blockchain.go:1198 +0x4b7
github.com/celo-org/celo-blockchain/core.(*BlockChain).InsertReceiptChain.func3({0xc043932000, 0x8d, 0x84b67016ef55ac58?}, {0xc043938000, 0x8d, 0x0?})
        github.com/celo-org/celo-blockchain/core/blockchain.go:1372 +0x812
github.com/celo-org/celo-blockchain/core.(*BlockChain).InsertReceiptChain(0xc0003e9000, {0xc023e65b00, 0x8d, 0x4?}, {0xc0426a8000, 0x8d, 0xa?}, 0x0)
        github.com/celo-org/celo-blockchain/core/blockchain.go:1398 +0x1405
github.com/celo-org/celo-blockchain/eth/downloader.(*Downloader).commitFastSyncData(0xc0005e81e0, {0xc023e65680, 0x8d, 0x8a35f30fd616db9c?}, 0xc031373220)
        github.com/celo-org/celo-blockchain/eth/downloader/downloader.go:2088 +0x4e8
github.com/celo-org/celo-blockchain/eth/downloader.(*Downloader).processFastSyncContent(0xc0005e81e0)
        github.com/celo-org/celo-blockchain/eth/downloader/downloader.go:2004 +0x905
github.com/celo-org/celo-blockchain/eth/downloader.(*Downloader).syncWithPeer.func7()
        github.com/celo-org/celo-blockchain/eth/downloader/downloader.go:593 +0x1d
github.com/celo-org/celo-blockchain/eth/downloader.(*Downloader).spawnSync.func1()
[...]

PR #2035, opened 2023-03-07
More nil checking in InsertReceiptChain

karlb:
Fixes https://github.com/celo-org/celo-blockchain/issues/1920

This just applies the change suggested in the issue. It seems reasonable and passed the tests, but I don't understand which situation would cause nil return values, yet.


karlb:
Closed in favour of https://github.com/celo-org/celo-blockchain/pull/2036 due to CI handling of fork repos.

PR #2036, opened 2023-03-07
More nil checking in InsertReceiptChain

karlb:
Fixes https://github.com/celo-org/celo-blockchain/issues/1920

This just applies the change suggested in the issue. It seems reasonable and passed the tests, but I don't understand which situation would cause nil return values, yet.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1174,7 +1174,21 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 		// Rewind may have occurred, skip in that case.
 		if bc.CurrentHeader().Number.Cmp(head.Number()) >= 0 {
 			currentFastBlock, td := bc.CurrentFastBlock(), bc.GetTd(head.Hash(), head.NumberU64())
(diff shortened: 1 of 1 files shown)


karlb:
> @karlb The upstream geth issues doesn't have a clear solution, did you check what solved it for them?

I don't think they know what solved the issue, the issue is just closed as stale. That part of the code has been changed significantly, so there is no trivial way to cherry-pick a change in either direction.

palango:
> I don't think they know what solved the issue, the issue is just closed as stale. That part of the code has been changed significantly, so there is no trivial way to cherry-pick a change in either direction.

So the idea is to add the check and see if it happens again? Maybe we should add some log message then...

karlb:
> So the idea is to add the check and see if it happens again?

Unless anyone has a good idea on how to reproduce or explain this, yes. Anything is better than a segfault. Well, anything but a corrupted state, but since we exit the function at a pretty sane point, I think we are safe here.

> Maybe we should add some log message then...

Yes, good idea.

palango (approved):

Issue #615, opened 2019-11-19
getValidator/fractionMulExps precompiles should check input length exactly equal to expected


asaj:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

lucasege:
While this initially seemed like a simple change (modifying the precompiles to use `!=` rather than `<`) but this ended up failing tests on `fractionMulExp` because the tests were passing in inputs of length 196, not 192 (as expected by static analysis). 

Attempted to fix this by also addressing https://github.com/celo-org/celo-monorepo/issues/1666 , but changing the assembly in `fractionMulExp` to resemble `checkProofOfPossession` (using `.call` or `.staticcall`) violates the `view` property of `fractionMulExp`. 

@asaj and I tried to get around this by implementing a wrapper interface around the `fractionMulExp` fn to convince the compiler that we weren't modifying state, but this involves injecting 4 bytes into the input (the function signature). 

Injecting these bytes caused the existing precompiles on ganache to fail since they weren't expecting them, and we resolved that these issues were ballooning too much and were low priority, so we are tabling it for now. 

My understanding is that when things calm down, we would still like to address these by implementing this "hack" of a wrapper interface, but it is too involved for now.

nategraf:
@lucasege It sounds like you did some digging on this one. Could you clarify whether addressing this issue would be a hard fork? If so, we should get it done before mainnet if at all possible.

PR #2038, opened 2023-03-07
Stricter input legth checking in precompiles getValidator/fractionMulExps/transfer

karlb:
Fixes https://github.com/celo-org/celo-blockchain/issues/615

This this changes the behaviour of precompiles, it requires a hard-fork.


palango:
PR looks good, but do we need an CIP for this kind of change or can we just merge it?

karlb:
> PR looks good, but do we need an CIP for this kind of change or can we just merge it?

Technically a CIP would be necessary, but IMO the change is such a small detail that we should consider it to be below the "should have a CIP" threshold. Especially since the specific bounds checking has never been specified.

core/vm/contracts.go
```diff
@@ -698,7 +698,7 @@ func (c *transfer) Run(input []byte, caller common.Address, evm *EVM) ([]byte, e
 	//   to:    32 bytes representing the address of the recipient
 	//   value: 32 bytes, a 256 bit integer representing the amount of Celo Gold to transfer
 	// 3 arguments x 32 bytes each = 96 bytes total input
-	if len(input) < 96 {
+	if (evm.chainRules.IsGFork && len(input) != 96) || len(input) <= 96 {
```
eelanagaraj, line 701: Hey @karlb was it intended to make this `<=` instead of `<`? This is newly breaking a tracing test I wrote in a draft branch and I think this should be `<` but maybe I'm missing something

karlb: Yes, you are right! I messed this up when introducing the IsGFork check.

PR #2043, opened 2023-03-13
Overwrite developer genesis block period with flag value

### LINKED ISSUE #2007: `--dev.period` flag does not override DevChainConfig block period
## Description

The `--dev.period` should override the block time, but the block time remains 1s.

## Context

Improve dev net setup.

eelanagaraj:
### Description
- Overwrites the developer block period flag (`--dev.period`), which wasn't happening before.
- Ensures that the validator & txFeeRecipient vals are set in the config for dev mode, which should (mostly) fix `--dev` mode
  - Although this is a bit messy (in that we are re-setting this value after it has been set above), I think it strikes a balance of making dev mode work & not cluttering the rest of the code too much.
- Updates developer docs in `dev-mode.md` & parameter details to reflect the status quo

Note that the `geth --dev` command is still somewhat broken in terms of account configuration. For now, it runs as expected when using the hard-coded dev account (both in "ephemeral" mode & with a persistent data directory) but doesn't work properly for "general" accounts/data directories. (I didn't think it was worth investing time into making it work more generally under those circumstances, given that mycelo also exists & has more functionality anyways).

### Tested
- tested dev mode locally
- tested out the `dev-mode.md` instructions & deployment with remix

### Related issues

- Fixes #2007

### Backwards compatibility
- Backwards compatible
- Note that though the docs said that the default was 0 (on-demand), this wasn't actually the case before (default was hard-coded as 1).

### DIFF
--- a/docs/docs/_getting-started/dev-mode.md
+++ b/docs/docs/_getting-started/dev-mode.md
@@ -13,20 +13,19 @@ Starting geth in dev mode does the following:
 -   Sets the gas price to 0
 -   Uses the Clique PoA consensus engine with which allows blocks to be mined as-needed without excessive CPU and memory consumption
 -   By default produces a block per second
-    -   To instead produce blocks on-demand when transactions are waiting to be mined, set `--dev.period 0`
+    -   To instead produce blocks on-demand when transactions are waiting to be mined, set `--dev.period 0`)
 
 ## Start Geth in Dev Mode
 
-Note: this command currently does not work as intended in ephemeral mode and must be run with a data dictionary that contains a keystore file with the hard-coded developer account `0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95`. The easiest way of doing this is to:
+You can specify a data directory to maintain state between runs using the `--datadir` option, otherwise, databases are ephemeral and in-memory:
 
-1. Run the command with the desired `--datadir` (and `--dev.period` if a specific block period is desired): `geth --datadir test-chain-dir --dev.period 5`.
-    This will create the required genesis file in `test-chain-dir` and import the hard-coded developer account `0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95`
+```shell
+geth --dev
+```
 
-2. Then run: `geth --dev --miner.validator 0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95 --datadir test-chain-dir`
-    Note that this will use the block period configured in the first step.
+Optionally specify `--dev.period` if a block period other than 1 per second is desired.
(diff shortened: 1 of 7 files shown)


docs/docs/_getting-started/dev-mode.md
```diff
@@ -16,16 +17,24 @@ Starting geth in dev mode does the following:
 
 ## Start Geth in Dev Mode
 
-You can specify a data directory to maintain state between runs using the `--datadir` option, otherwise, databases are ephemeral and in-memory:
+Note: this command currently does not work as intended in ephemeral mode and must be run with a data dictionary that contains a keystore file with the hard-coded developer account `0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95`. The easiest way of doing this is to:
```
palango: s/dictionary/directory

docs/docs/_getting-started/dev-mode.md
```diff
@@ -13,28 +13,27 @@ Starting geth in dev mode does the following:
 -   Sets the gas price to 0
 -   Uses the Clique PoA consensus engine with which allows blocks to be mined as-needed without excessive CPU and memory consumption
 -   By default produces a block per second
-    -   To instead produce blocks on-demand when transactions are waiting to be mined, set `--dev.period 0`
+    -   To instead produce blocks on-demand when transactions are waiting to be mined, set `--dev.period 0`)
```
palango, line 16: superfluous closing brace?

palango (commented):
Looks good, some smaller comments inline.

- Why do we set the default to 1? Wouldn't 0, e.g. create a block when a tx is there, be the better option for dev mode?

docs/docs/_getting-started/dev-mode.md
eelanagaraj: Hmm this is from an older commit, resolving since this line should be gone anyways now I think!

palango: Ah yeah, I went through commit for commit. Sorry for the noise.

eelanagaraj:
@palango no strong preference on default value, though technically setting to 0 is breaking from previous behavior. (It previously said 0 was the default in the docs but was in practice setting the block period to 1). Lmk what you think!

palango:
> @palango no strong preference on default value, though technically setting to 0 is breaking from previous behavior. (It previously said 0 was the default in the docs but was in practice setting the block period to 1). Lmk what you think!

Then let's keep 1 for now until we get some feedback with strong opinions about that.

palango (approved):

carterqw2 (approved):

PR #2045, opened 2023-03-15
Visualize diff between geth and celo-blockchain

eelanagaraj:
### Description
Builds on top of #2024:
- Adds steps to the GitHub action to find/replace `celo-org/celo-blockchain` -> `ethereum/go-ethereum` in imports and apply `gofmt`/`goimports` to both codebases
  - Upstream commit is now specified in the workflow instead of the fork.yaml
- Uses an updated forkdiff image allowing us to diff against a local upstream repo (instead of a commit), merged upstream change to forkdiff but the image hasn't been updated yet, and it may be easiest to just use our own copy anyways
- Structures forkdiff mostly by package (unlike op-geth, which structures more generally by logical units/features)

### Implementation notes
- Necessary to separate the workflow into two jobs: `create-forkdiff` & `deploy`, since the latter needs to use the checkout action previously in the root workspace, while we need to do this in two separate subdirs in order to format & commit changes in both local copies of celo-blockchain and go-ethereum
- `forkdiff` doesn't allow for `**` globbing, and paths may only be listed once in the yaml file (otherwise it errors)
- The diff is so large (+ I don't have enough context on how everything works together in geth), so I wasn't able to truly split everything up by feature. I don't really know if that is even feasible in a reasonable amount of time with the current state of the repo, but I think it's best to use the repo structure as a starting point (as is) and continue to make adjustments.
  - Happy to make any bigger changes here though!

### Tested

- Changes are deployed here: https://celo-org.github.io/celo-blockchain/index.html/

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain-planning/issues/51

### TODO
- [x] update the workflow to trigger on push to master (once format/structure is settled in this PR)
- [x] reinstate CI tests (commented out most of them while developing in CI) (revert `d95678b3d`)


.github/workflows/config.yml
palango, line 1: This shoudn't be touched, right?

eelanagaraj: yes -- tucked away in the PR description but I will revert the commit that comments this out after addressing any PR comments! Commented it out to speed up dev time with CI stuff & to not unnecessarily run the full suite every time 😅

karlb:
Looks much better!

One remaining way to reduce the diff would be to ignore the whitespace changes introduced by gofmt column alignment, e.g.
```diff
-	stales int64
+	stales    int64
+	maxStales int64 // Maximum amount of stale price points allowed before a forced re-heap
```
would ideally only show the last line as diff. I don't think this can be done easily by changing the checkout, but maybe there is an "ignore whitespace" flag somewhere that we could flip?

eelanagaraj:
@karlb agreed, will see what i can find!

.github/workflows/config.yml
palango: Ups, should have read that more carefully!

eelanagaraj:
@karlb @carterqw2 from what I can tell, there are no "easy" changes to be made that would allow us to diff without whitespace :-/ The `go-git` package doesn't seem to implement any of the whitespace-ignore options (it diffs at git filetree level but does not check contents). While it makes changes available in chunks, which are parsed in `forkdiff`, we'd have to make more significant changes to `forkdiff` to manually process the file chunks & perform the ignore-whitespace diff ourselves, and from initial poking around this is not trivial 😅 (It would also be a bit janky since I think we would essentially need to reprocess/chunk each file, reimplementing part of what `go-git` does under the hood).

Happy to continue looking into this if we feel it's worth the effort, though honestly it may make sense (especially short-term or when actually using the diffs to apply changes) to manually revert the formatting for the old lines and add no-lint comments 🤷‍♀️) Let me know what you think!

Also if it visually looks alright to you too as is @palango, lmk and I'll reinstate the rest of the CI tests and update the workflow trigger 😄

palango:
To me this looks good, the whitespace changes would be nice, but I think the wins aren't worth the effort if we really have to change the tool.

Also, if this becomes a problem later on we can always fix it then.

carterqw2:
First of all, great work, the diff looks so much better now!
Regarding the whitespace issue, is there an easy way to run `gofmt` on both codebases to get identical formatting?

eelanagaraj:
@palango added a medium-fix of jankily replacing 2+ spaces with a single space for now, as this does seem to reduce the diff in most cases (even if it's not a great fix overally) and doesn't increase it too much/make it too unreadable I think.

@carterqw2 thanks! This workflow already applies `gofmt` & `goimports`, but the issue is that in cases where we've added new struct fields or parameters that are longer, gofmt aligns everything else in the struct, changing the whitespace (even if the content of those lines are the same). There doesn't seem to be a way to enforce a minimum spacing alignment with `gofmt` to reduce the changes with that either :-/

karlb:
> Regarding the whitespace issue, is there an easy way to run gofmt on both codebases to get identical formatting?

This actually happens *because* both of them are formatted with gofmt, which will align certain code blocks (e.g. struct field) in a tabular way to fit the longest entry. If one version has a different length for the longest element, there will be changes to all lines in the block.

I agree that we should not spend too much time on this now and can leave it in the current state if there is not easy way to solve.

One last suggestion what might help with the whitespace: Normalize whitespace in both checkouts by collapsing multiple spaces between words to a single space. Something like `s/(\w)\s+(\w)/\1 \2/g` (might need to add a few extra characters like `*()`).

.github/workflows/pages.yaml
```diff
@@ -0,0 +1,111 @@
...
+        # is not available. If this causes other issues down the line, delete those two lines.
+        run: |
+          grep -rl "//\s*nolint:\s*unused$" ${{ env.CELO_BLOCKCHAIN_PATH }}/. | xargs sed -i "s#//\s*nolint:\s*unused\$##"
+          grep -rl "github.com/celo-org/celo-blockchain" ${{ env.CELO_BLOCKCHAIN_PATH }}/. | xargs sed -i "s#github.com/celo-org/celo-blockchain#github.com/ethereum/go-ethereum#"
+          grep -rl "github.com/ethereum/go-ethereum/crypto/bls" ${{ env.CELO_BLOCKCHAIN_PATH }}/. | xargs sed -i "s#github.com/ethereum/go-ethereum/crypto/bls#github.com/celo-org/celo-blockchain/crypto/bls#"
+          grep -rl 'ethereum "github.com/ethereum/go-ethereum"' . | xargs sed -i 's#ethereum "github.com/ethereum/go-ethereum"#"github.com/ethereum/go-ethereum"#'
+          gofmt -w ./${{ env.CELO_BLOCKCHAIN_PATH }} ./${{ env.GO_ETHEREUM_PATH }}
+          goimports -w ./${{ env.CELO_BLOCKCHAIN_PATH }} ./${{ env.GO_ETHEREUM_PATH }}
+          grep -rl " \{2,\}" ${{ env.CELO_BLOCKCHAIN_PATH }}/. --include \*.go | xargs sed -i "s# \{2,\}# #"
+          grep -rl " \{2,\}" ${{ env.GO_ETHEREUM_PATH }}/. --include \*.go | xargs sed -i "s# \{2,\}# #"
```
karlb, line 54: sed needs a `g` flag to allow repeated matches.

```sh
$ echo 'datadirDefaultKeyStore   =     "keystore"           // Path within the datadir to the keystore' | sed "s# \{2,\}# #"
datadirDefaultKeyStore =     "keystore"           // Path within the datadir to the keystore
$ echo 'datadirDefaultKeyStore   =     "keystore"           // Path within the datadir to the keystore' | sed "s# \{2,\}# #g"
datadirDefaultKeyStore = "keystore" // Path within the datadir to the keystore
```

carterqw2:
Ah, I see now, thanks. If the `sed` workaround is not good enough or causes issues, let's not spend more time on it.

palango (approved):

carterqw2 (approved):

PR #2046, opened 2023-03-15
Split roundstate saving in two

hbandura:
Split the roundstate saving, allowing consensus message processing to be a bit faster.

### DIFF
--- a/consensus/istanbul/core/message_set.go
+++ b/consensus/istanbul/core/message_set.go
@@ -32,6 +32,7 @@ import (
 type MessageSet interface {
 	fmt.Stringer
 	Add(msg *istanbul.Message) error
+	AddAll(msgs []*istanbul.Message) int
 	GetAddressIndex(addr common.Address) (uint64, error)
 	Remove(address common.Address)
 	Values() (result []*istanbul.Message)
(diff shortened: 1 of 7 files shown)


consensus/istanbul/core/roundstate_db.go
```diff
@@ -40,6 +41,7 @@ const (
 	dbVersionKey = "version"  // Version of the database to flush if changes
 	lastViewKey  = "lastView" // Last View that we know of
 	rsKey        = "rs"       // Database Key Pefix for RoundState
+	rcvdKey      = "rcvd"     // Database Key Prefix for rcvd messages from the RoundState (split saving)
```
mcortesi, line 44: why not have differente keys for prepares, commits and parent commits?

mcortesi (commented):
This PR is missing transition logic from old db format to new. I mean if we run this on a old client it will fail.

Still for testing purposes it's ok

consensus/istanbul/core/roundstate_db.go
hbandura: I believe they shouldn't have an outstanding size, though I agree that further splitting is better. I'll check how the code looks with it.

hbandura:
> This PR is missing transition logic from old db format to new. I mean if we run this on a old client it will fail.
> 
> Still for testing purposes it's ok

Yes, it definitely lacks the migration logic. I'll add it.

consensus/istanbul/core/message_set.go
```diff
@@ -68,18 +69,40 @@ func deserializeMessageSet(binaryData []byte) (MessageSet, error) {
...
 
 	return nil
 }
 
+func (ms *messageSetImpl) AddAll(msgs []*istanbul.Message) int {
+	ms.messagesMu.Lock()
+	defer ms.messagesMu.Unlock()
+	var added = 0
+	for _, msg := range msgs {
+		err := ms.checkValidToAdd(msg)
```
gastonponti, line 96: Should we log the invalid messages or the possible overrides?

consensus/istanbul/core/roundstate_db_test.go
```diff
@@ -21,7 +48,12 @@ func TestRSDBRoundStateDB(t *testing.T) {
 			{Address: common.BytesToAddress([]byte(string(rune(2)))), BLSPublicKey: pubkey1},
 			{Address: common.BytesToAddress([]byte(string(rune(4)))), BLSPublicKey: pubkey2},
 		})
-		return newRoundState(newView(2, 1), valSet, valSet.GetByIndex(0))
+		view := newView(2, 1)
+		rs := newRoundState(view, valSet, valSet.GetByIndex(0))
+		rs.AddPrepare(mockViewMsg(view, istanbul.MsgPrepare, valSet.GetByIndex(0).Address()))
+		rs.AddCommit(mockViewMsg(view, istanbul.MsgPrepare, valSet.GetByIndex(0).Address()))
```
gastonponti, line 54: Shouldn't this be a istanbul.MsgCommit? (same with the next one)

gastonponti (approved):
Other than the log, and the nit in the test, LGTM

PR #2049, opened 2023-03-22
UltraGreen: Send base fee to FeeHandler contract

karlb:
### Description

Assuming the [UltraGreen Celo CIP](https://github.com/celo-org/celo-proposals/pull/321) passes, we want to transfer all base fees to the FeeHandler contract, which then takes care of distributing and burning the base fee according to the CIP.

Requires https://github.com/celo-org/celo-monorepo/pull/10227 in the monorepo.

### Other changes

Just some additional error messages to make failures more informative.

### Tested

I added a check to the TestTransferCELO test which verifies that the base fee is transferred to the FeeHandler. The FeeHandler behaviour is not tested here, it should already be sufficiently tested in the celo-monorepo.

### Backwards compatibility

This change needs to be part of a hard fork. Compatibility for application develops is maintained, since transactions are initiated in exactly the same way and only the fee receiver is changed.


palango (approved):
Looks good to me!

Do you want to wait until the contract changes have been merged in the monorepo or merge this right away?

karlb:
> Do you want to wait until the contract changes have been merged in the monorepo or merge this right away?

Let's merge it right away, before we get new merge conflicts within this repo. Bigger changes in the UltraGreen code are unlikely and we can easily update to a newer monorepo commit at any time.

PR #2052, opened 2023-03-23
testing out larger runners

pputman-clabs:
### Description

This is an attempt to use larger runners with 4 cores to see if it improves testing time

Will have to test in the PR since its testing a gh actions workflow

- Fixes #[issue number here]

### Backwards compatibility

Should cause no issues, just faster build/testing times.

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -20,7 +20,7 @@ env:
 jobs:
   prepare-system-contracts:
     name: Build celo-monorepo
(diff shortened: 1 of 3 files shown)


pputman-clabs:
added:
```cache-dependency-path: go.sum```
to see if it would speed some things up

pputman-clabs:
Switched the runners back down to 4 cores.

alfiomy (commented):
Good

PR #2060, opened 2023-04-04
Add baseFeePerGas to headers sent via EthSubscribe

karlb:
### Description

Adding the `baseFeePerGas` to blocks returned by `EthSubscribe` makes the results more consistent with the data returned on Ethereum and via other RPC-Endpoints on Celo. And it is useful!

Once we add the `baseFeePerGas` to the header, this implementation is not needed anymore. So this change is only temporary.

I did not check for `RPCEthCompatibility` since this is not done for `baseFeePerGas` in the normal JSON-RPC responses, either.

### Other changes

* Enabled websockets for mycelo

### Tested

Tested against mycelo with the following code

```json
{
  "name": "subs",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "web3": "^1.9.0",
    "web3-providers-http": "^1.9.0"
  }
}
```
```js
const Web3WsProvider = require('web3-providers-ws');
var provider = new Web3WsProvider('ws://localhost:9545');

const Web3 = require('web3');
const web3 = new Web3(provider)

async function wrapper(){
    const subs = web3.eth.subscribe('newBlockHeaders', function(error, result) {
      if (!error)
          console.log(result);
      else
          console.log(error);
    });
}

wrapper()
```

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain-planning/issues/59

### Backwards compatibility

Adding a field to the response should not break sanely implemented consumers. IMO it is acceptable to consider this backwards compatible, although it is not true in a strict sense.

### DIFF
--- a/accounts/abi/bind/backends/simulated.go
+++ b/accounts/abi/bind/backends/simulated.go
@@ -874,6 +874,10 @@ func (fb *filterBackend) ServiceFilter(ctx context.Context, ms *bloombits.Matche
 	panic("not supported")
 }
 
+func (fb *filterBackend) RealGasPriceMinimumForHeader(ctx context.Context, currencyAddress *common.Address, header *types.Header) (*big.Int, error) {
+	return nil, fmt.Errorf("filterBackend does not implement RealGasPriceMinimumForHeader")
+}
+
 func nullSubscription() event.Subscription {
 	return event.NewSubscription(func(quit <-chan struct{}) error {
 		<-quit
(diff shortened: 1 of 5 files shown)


palango:
Looks good to me, as a temporary measure until we add it to the header. But let's have @gastonponti have a look as well.

gastonponti (approved):
LGTM 👍

PR #1978, opened 2022-10-31
Go native tracer

jsvisa:
implements https://github.com/celo-org/celo-blockchain/issues/1970

DON'T MERGE, test in process.

PR #1695, opened 2021-09-16
Merge v1.10.7

trianglesphere:
### Description

This merges changes made after 1.9.25 up to 1.10.7 as well as some espresso specific HF changes.

Notable Changes from Upstream
* https://github.com/ethereum/go-ethereum/pull/22898. This had to be adapted to multiple fee currencies
* https://github.com/ethereum/go-ethereum/pull/21502. This modified the transaction types
* https://github.com/ethereum/go-ethereum/pull/22837. This implemented 1559. There is a celo specific implementation in https://github.com/celo-org/celo-blockchain/pull/1696
* Please note that reverting to Geth v1.9.19 or prior after upgrading to v1.9.20 is not possible without a resync because the blockchain database layout has changed.

Changes from Upstream
* Snap sync is completely disabled and hidden from end users
* The catalyst API is not included
* Lots of flags changes: See https://github.com/celo-org/celo-blockchain/issues/1699#issuecomment-938082239

### Notes for reviewers

* With the new transaction types, `core/types` got overhauled to supported multiple types.
* With 1559, the txpool was rewritten. We've placed a layer of modification over it to handle multiple fee currencies.
* There were some changes in the supported p2p protocol versions + some of the files were moved around
* With 1559, the blockcontext was moved & is also used in `state_transition`
* The eth protocols got moved around (file wise).

### Tested

Sync tests are being ran. It appears to sync without issue.

### Related issues

- Fixes #1699 

### Backwards compatibility

This updates the database layout so while it is possible to upgrade old nodes to this version, **downgrading is not supported**. See https://github.com/ethereum/go-ethereum/releases/tag/v1.9.20
> * Please note that reverting to Geth v1.9.19 or prior after upgrading to v1.9.20 is not possible without a resync because the blockchain database layout has changed.

PR #2061, opened 2023-04-04
Set interpreter config to debug as well in evm.SetDebug and update tracing tests

eelanagaraj:
### Description
Debugging the failure of the test case `TestBlockTracingConcurrentMapAccess` and then later in the `tracers_test` suite in https://github.com/celo-org/celo-blockchain/pull/1978 led to the two following observations:
- in the `evm.SetDebug` function, we currently only set `evm.Config.Debug` but not `evm.interpreter.cfg.Debug`, but both of these are used in EVM execution code. Fixing this is necessary for the go native tracing implementation to work
- the `tracers_test` suite has been commented out for the past few years but now mostly work with the original geth tests

The debug setting issue seems to have arisen with an upstream merge where we changed EVM/interpreter initialization from passing the config by reference to passing by value. This caused the divergence in debug config values. For now, I opted by setting both debug config values in the setter function instead of reverting to passing by reference, since upstream still passes the config by value and I imagine this could lead to the exact same issue in the future (and minimizes the diff).

This PR:
- fixes the config debug setting here (so that this subtle change doesn't get buried in the larger PR & that CI passes)
- takes some of the `tracer_test` fixes from the go-native tracing PR and implements them here so we have a clearer signal of what expected values are before the native tracing implementation

### Tested
- tracing unit tests
- ensured that all the tracing tests pass with/without the Debug config change locally
- [x] testing on blockscout rc1staging (TODO -- manually check that differences in outputs make sense, and make sure this doesn't break anything unexpected with older transaction tracing)
- [ ] ran rosetta CLI checks for limited time with new blockchain node to check that balance reconciliation looks good

### Related Issues
Prep to get https://github.com/celo-org/celo-blockchain/pull/1978 into mergeable state

### Backwards compatibility
This changes the traces outputted, as this now does not trace calls when distributing fees or completing tobin transfer logic. This breaks from current tracing behavior, but seems consistent with previous (pre #1695) functionality and intended functionality (from the comments).


palango (commented):
Looks good, thanks for digging so deep into this code!

Some comments:
- What about fixing this by passing the config as a reference?
- Does `TestTraceTransactionWithRegistryDeployed` fail without the prior commits fixing the flags?
- I'd merge some commits, so that the history gets a bit nicer.

eelanagaraj:
@palango 🙏 
> What about fixing this by passing the config as a reference?
geth passes the config by value in `NewEVM` and wanted to minimize the diff there. In more recent commits, they have entirely gotten rid of passing the config to the interpreter (and instead just use the one). I figured it'd be best for now to contain the changes within the `SetDebug` function for now and ultimately move towards the geth solution of refactoring the entire interpreter/evm. (Also I was nervous that within a future upstream merge, this change could again come back in, whereas the SetDebug code is entirely Celo-specific). Lmk if that makes sense or if you still think it's preferable to go back to passing by value!

> Does TestTraceTransactionWithRegistryDeployed fail without the prior commits fixing the flags?
yes!

> I'd merge some commits, so that the history gets a bit nicer.
def, planning to squash merge this. I initially kept stuff in separate undo/redo steps to save CI results while testing (obvi overwrote those when rebasing master hehe). Happy to split things up to separate commits if squash-merging isn't an option! 😄

palango:
> Lmk if that makes sense or if you still think it's preferable to go back to passing by value!

Makes a lot of sense, lets keep it like that for now.

Last question: How did you generate the updates in the tracer tests?

eelanagaraj:
> Makes a lot of sense, lets keep it like that for now.

Sounds good!

> Last question: How did you generate the updates in the tracer tests?

I took most of the changes from the go-native tracer PR  -- basically, that PR got these tests runnable (they have been commented out for 4 years 😅 ). The idea was to get those tests runnable with input pre-/post- the fix (no changes -- they pass either way; this was also one of the reasons for the undo/redo in the commit history since I wanted to confirm this in CI in addition to locally). The hope is that with these tests added on master, we will have a bit stronger of a signal about the go-native tracer implementation (i.e. that PR should require no changes to the input files to get those cases running).

 For the JSON files:
- I tried to make the least amount of changes to get them running (i.e. usually change gas values in outputs)
- In a few cases the inputs simply didn't work (I tried everything I could think of to decode these including using the transaction formats from the commit when they were added, but couldn't find a way 😭), I used the files from geth with slight changes to outputs to get them running (thank you for that tip btw @palango !).

Happy to add clarification anywhere or lmk if you can think of additional steps to take regarding the native tracer changes! 🙏

eelanagaraj:
@palango oop I also added one last TODO to the PR -- I won't merge this until after the blockscout results comparison for some txs & making sure that it doesn't break things. Then, blockscout outputs should be identical for the go-native tracer once this change is merged!

palango (approved):
Great, that all sounds good to me.
I really like the approach of fixing tests *before* the native tracer PR so that we can be confident that nothing changes.

PR #2062, opened 2023-04-12
Add GASLIMIT opcode

karlb:
### Description

* Add the blocks gas limit to the block header
* Add GASLIMIT opcode to get block gas limit from contracts

### Tested

I created a test contract calling `block.gaslimit` and executed it successfully against mycelo. Unit testing suggestions welcome!
CI still fails, but I can take care of fixing the tests once the CIP has been accepted.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain-planning/issues/15

### Backwards compatibility

Requires a hard fork, but maintains backwards compatibility from a user's point of view.


core/types/block.go
```diff
@@ -58,15 +58,18 @@ type Header struct {
 	extraLock  sync.Mutex
 	extraValue *IstanbulExtra
 	extraError error
+
+	GasLimit uint64 `json:"gasLimit" rlp:"optional"`
```
palango, line 62: Different position in struct compared to upstream.

internal/ethapi/api.go
```diff
@@ -745,17 +751,6 @@ func (s *PublicBlockChainAPI) GetBlockByHash(ctx context.Context, hash common.Ha
 // rpc response that ethers.js depends upon.
 // See https://github.com/celo-org/celo-blockchain/issues/1945
 func addEthCompatibilityFields(ctx context.Context, block map[string]interface{}, b Backend, header *types.Header) {
-	hash := header.Hash()
-	numhash := rpc.BlockNumberOrHash{
-		BlockHash: &hash,
-	}
-	gasLimit, err := b.GetRealBlockGasLimit(ctx, numhash)
```
palango, line 752: Can `GetRealBlockGasLimit` be removed as well?

palango:
One thing that might be worthwhile here is to remove the contract access to get the block gas limit during block execution (see `blockchain_parameters.GetBlockGasLimitOrDefault`).

Instead we can set the gas limit once during block creation and then use the header value in all subsequent steps.

internal/ethapi/api.go
```diff
@@ -709,13 +709,17 @@ func (s *PublicBlockChainAPI) GetBlockByNumber(ctx context.Context, number rpc.B
...
+			if s.b.RPCEthCompatibility() {
+				addEthCompatibilityFields(ctx, response, s.b, block.Header())
+				if number == rpc.PendingBlockNumber {
+					// Pending blocks need to nil out a few fields
+					for _, field := range []string{"hash", "nonce", "miner"} {
+						response[field] = nil
+					}
 				}
+			} else {
+				delete(response, "gasLimit")
```
palango, line 722: Wouldn't we want to keep the `gasLimit` field in the response irrespective of the `RPCEthCompatibility` flag?

karlb: Yes, you are right. I thought we should not add it to keep the same result layout for users expecting no RPCEthCompatibility, but actually the point is to return exactly values which are included in the block hash. I will also have to check that the order is the same, otherwise the RPCEthCompatibility=off feature won't make any sense.

In general, I would be happy to remove the feature unless there is some known usage.

core/types/block.go
palango, line 62: The plan here is to reorder the filed when we add the remaining ones?

internal/ethapi/api.go
```diff
@@ -744,27 +742,39 @@ func (s *PublicBlockChainAPI) GetBlockByHash(ctx context.Context, hash common.Ha
 // and ethers.js (and potentially other web3 clients) by adding fields to our
 // rpc response that ethers.js depends upon.
 // See https://github.com/celo-org/celo-blockchain/issues/1945
-func addEthCompatibilityFields(ctx context.Context, block map[string]interface{}, b Backend, header *types.Header) {
-	hash := header.Hash()
-	numhash := rpc.BlockNumberOrHash{
-		BlockHash: &hash,
+func addEthCompatibilityFields(ctx context.Context, response map[string]interface{}, b Backend, block *types.Block) {
+	isGFork := b.ChainConfig().IsGFork(block.Number())
```
palango, line 746: nit: move this into the first if-block

palango (approved):
Looks great. Thanks for adding tests, one remaining question, but if the plan is to move the field once we merge the remaining fields into the header is good to go from my side.

karlb:
> but if the plan is to move the field once we merge the remaining fields into the header is good to go from my side.

My plan was to defer the field reordering until https://github.com/celo-org/celo-blockchain-planning/issues/100. As long as all the changes go into the same hard fork it doesn't matter and makes for a cleaner split of PRs, IMO.

internal/ethapi/api.go
```diff
@@ -709,9 +709,9 @@ func (s *PublicBlockChainAPI) GetBlockByNumber(ctx context.Context, number rpc.B
 	if block != nil && err == nil {
 		response, err := s.rpcMarshalBlock(ctx, block, true, fullTx)
 
-		if err == nil && s.b.RPCEthCompatibility() {
-			addEthCompatibilityFields(ctx, response, s.b, block.Header())
-			if number == rpc.PendingBlockNumber {
+		if err == nil {
+			addEthCompatibilityFields(ctx, response, s.b, block)
+			if s.b.RPCEthCompatibility() && number == rpc.PendingBlockNumber {
```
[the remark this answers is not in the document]
karlb: You are right, here's the upstream code for comparison: https://github.com/ethereum/go-ethereum/blob/eae3b1946a276ac099e0018fc792d9e8c3bfda6d/internal/ethapi/api.go#L709-L714

I'll remove the `s.b.RPCEthCompatibility() &&` part.

carterqw2 (approved):

PR #2063, opened 2023-04-12
Add G-Fork infrastructure

palango:
### Description

Add the basic infrastructure for the G* hardfork.

### DIFF
--- a/cmd/mycelo/genesis_commands.go
+++ b/cmd/mycelo/genesis_commands.go
@@ -54,6 +54,10 @@ var templateFlags = []cli.Flag{
 		Name:  "forks.espresso",
 		Usage: "Optional flag to allow espresso fork overwritting (default: 0, disable: -1)",
 	},
+	cli.Int64Flag{
+		Name:  "forks.gfork",
+		Usage: "Optional flag to allow gfork fork overwritting (default: 0, disable: -1)",
+	},
 }
 
 var buildpathFlag = cli.StringFlag{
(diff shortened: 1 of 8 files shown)


palango:
Wouldn't it make more sense to to that in the PR that adds the new instruction?

karlb:
> Wouldn't it make more sense to to that in the PR that adds the new instruction?

If it was a single PR, yes. But since we touch `DIFFICULTY`, `RANDOM`, `PREVRANDAO`, `GASLIMIT` and `BASE_FEE`, I assume there will be multiple PRs.

palango:
@karlb Added the new instruction set.

cmd/mycelo/genesis_commands.go
```diff
@@ -167,6 +167,15 @@ func envFromTemplate(ctx *cli.Context, workdir string) (*env.Environment, *genes
 		}
```
eelanagaraj, line 167: Any reason not to add the CLI flag for `forks.gFork` too?

palango: Do you mean this below or something else?
https://github.com/celo-org/celo-blockchain/blob/f75b4a3b3f1068d4672b7c8791e4444cb4810331/cmd/mycelo/genesis_commands.go#L170-L178

palango: The flag definition was actually missing :)

karlb (approved):
I'm not used to handling forks, but it looks reasonable to me.

PR #2069, opened 2023-04-18
ci: Remove `go-modules` job

palango:
This job was just installing and caching go, something that will only be useful in the case of updating the go version. As this does not happen often, this commit optimises the common case by removing the job.

@carterqw2 Is there another use of this job?

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -79,23 +79,8 @@ jobs:
         name: system-contracts
         path: 'celo-monorepo/${{ env.CONTRACTS_BUILD_PATH }}'
 
-  go-modules:
-    name: Setup go
-    runs-on: ubuntu-latest
-
-    steps:
-    - name: Checkout repo
-      uses: actions/checkout@v3
-    - name: Setup go
-      uses: actions/setup-go@v3
-      with:
-        go-version: ${{ env.GO_VERSION }}
-        check-latest: false
-        cache: true
-
   check-imports:
     name: Check imports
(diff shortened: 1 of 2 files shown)


carterqw2:
Yeah, in the case when modules are updated, all the parallel jobs that use the `setup-go` action will run it and it will take significant amount of time, so it made sense to move it in front of them. You could upgrade/remove one of the modules to check out the effect of your change.

palango:
> in the case when modules are updated

What kind of modules do you mean? Go modules?

carterqw2:
> What kind of modules do you mean? Go modules?

Go modules, yes.

carterqw2:
> > > What kind of modules do you mean? Go modules?
> > 
> > 
> > Go modules, yes.
> 
> Mh, I added a change in one module, but I'm not seeing fundamental changes in CI runtimes.

I meant modules in dependencies, this job caches based on the checksum of `go.sum`, meaning that if you change anything there, it will result in a cache miss and would affect other jobs.

palango:
> > > > What kind of modules do you mean? Go modules?
> > > 
> > > 
> > > Go modules, yes.
> > 
> > 
> > Mh, I added a change in one module, but I'm not seeing fundamental changes in CI runtimes.
> 
> I meant modules in dependencies, this job caches based on the checksum of `go.sum`, meaning that if you change anything there, it will result in a cache miss and would affect other jobs.

That makes sense: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

Issue #950, opened 2020-04-01
Bad block error when syncing


kevjue:
### Expected Behavior

Bad block error not encountered.

### Current Behavior

Validators output this error intermittently:

```
########## BAD BLOCK #########
Chain config: {ChainID: 200312 Homestead: 0 DAO: <nil> DAOSupport: false EIP150: 0 EIP155: 0 EIP158: 0 Byzantium: 0 Constantinople: 0 Petersburg: 0 Istanbul: <nil> Engine: istanbul}

Number: 318442
Hash: 0xa99e67f828b587da45240b5ef4fa0f5d972909638c87add90f094f1e5257af1b


Error: unknown block
##############################
```

kevjue:
Fixed via https://github.com/celo-org/celo-blockchain/pull/952

PR #2070, opened 2023-04-19
Draft: Remove `eth` protocol changes.

palango:
### Description

For gathering feedback for now.

This removes a change to the ethereum protocol in our code, in which we send the block hashes with the bodies, for which I see no need.

If we decide to merge this, we need to create a new protocol version for it.


eth/fetcher/block_fetcher.go
```diff
@@ -596,16 +595,14 @@ func (f *BlockFetcher) loop() {
...
+				for i := 0; i < len(task.transactions) && i < len(task.randomness) && i < len(task.epochSnarkData); i++ {
 					// Match up a body to any possible completion request
 					var matched = false
 					for hash, announce := range f.completing {
 						if f.queued[hash] != nil || announce.origin != task.peer {
 							continue
 						}
-						if task.blockHashes[i] != announce.header.Hash() {
-							continue
-						}
```
palango, line 608: @mcortesi @gastonponti This seems to be the fundamental change that passing the hashes allows for. Do you see any reason for that?

carterqw2: Looks like this was a [part](https://github.com/celo-org/celo-blockchain/pull/952) of a [bug](https://github.com/celo-org/celo-blockchain/issues/950) fix. I'm worrying that authors haven't added a test that tests this situation, reverting it looks risky, no?

palango: Sure, but this is something we need to fix anyways when we want to separate protocols.

The only explanation that I have is that we could mistake an uncle block for  the correct one a take the wrong chain?
```go
if task.blockHashes[i] != announce.header.Hash() {
	continue
}
```

palango:
This cannot be removed right now.

PR #2072, opened 2023-04-20
Build GitHub action

alvarof2:
### Description

Add GH Actions to automatically build, sign and add tags (following [this](Dockerfile.alltools)) for `geth` and `geth-all`.

Signing is done with cosign, so signatures can be verified with (macOS):

```bash
brew install cosign
cosign verify us-west1-docker.pkg.dev/devopsre/clabs-public-images/test-wid-tag:<TAG>--certificate-identity-regexp 'https://github.com/celo-org/reusable-workflows/.github/workflows/container-cicd.yaml@refs/tags/.*' --certificate-oidc-issuer https://token.actions.githubusercontent.com | jq
```

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/.github/workflows/add-docker-tag.yaml
+++ b/.github/workflows/add-docker-tag.yaml
@@ -1,0 +1,50 @@
+name: Add Tags to Blockchain Images
+
+on:
+  workflow_dispatch:
+    inputs:
+      origin-tag:
+        description: 'Original tag'
+        required: true
+        type: string
+      destination-tag:
+        description: 'Tag to add'
+        required: true
+        type: choice
+        options:
+          - alfajores
+          - mainnet
+  workflow_call:
+    inputs:
+      origin-tag:
+        description: 'Original tag'
+        required: true
+        type: string
+      destination-tag:
+        description: 'Tag to add'
+        required: true
+        default: 'baklava'
+        type: string
+
+jobs:
+  Add-Tag:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+      - id: 'auth-gcp-master'
+        name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v1'
+        with:
+          workload_identity_provider: 'projects/1094498259535/locations/global/workloadIdentityPools/gh-celo-blockchain-add-tag/providers/github-by-repos'
+          service_account: 'celo-blockchain@devopsre.iam.gserviceaccount.com'
+          access_token_lifetime: '20m'
+
+      - name: 'Set up Cloud SDK'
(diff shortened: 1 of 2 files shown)


.github/workflows/add-docker-tag.yaml
```diff
@@ -0,0 +1,50 @@
...
+  workflow_call:
+    inputs:
+      origin-tag:
+        description: 'Original tag'
+        required: true
+        type: string
+      destination-tag:
+        description: 'Tag to add'
+        required: true
+        default: 'baklava'
```
carterqw2, line 26: Excuse me my ignorance but why `baklava` is a special case?

alvarof2: I was [following this guide](https://github.com/celo-org/celo-labs/blob/master/packages/operations/celo-blockchain-deploy-guide.md#signing-and-tagging-docker-images):

> Tag the docker images with the semantic versions (x.y.z, x.y, x) and with the network tag (baklava, alfajores, mainnet). This can be done on the GCP container registry or with the given script.
The network tags should be updated as the release is rolled out the that particular network. So the baklava tag is immediately updated, then 1 week later the alfajores tag is updated, then 1 week later the mainnet tag is updated.

So the `baklava` docker image is added when pushing a GH tag (along with the commit, the Major, the Major.Minor and the full semver). `alfajores` and `mainnet tags can be added by running the add-tag workflow manually later on (doc says 1 week between tags). For example:

Screenshot 2023-04-21 at 09 51 24

Happy to change if that doc is not up to date with the current release cycle.

.github/workflows/build-sign-images.yaml
```diff
@@ -0,0 +1,129 @@
...
+jobs:
+  Replace-Branch-Name:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      replaced-branch: ${{ steps.replace.outputs.value }}
+    if: startsWith(github.ref, 'refs/heads/release')
+    steps:
+      - id: replace
+        uses: mad9000/actions-find-and-replace-string@3
```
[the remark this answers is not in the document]
alvarof2: Updated!

.github/workflows/add-docker-tag.yaml
carterqw2: No need to update, I just wasn't aware of the process. Thanks for the clarification!

jcortejoso (approved):

jcortejoso:
@celo-org/blockchain does anyone have any opinion about making the build process for docker images (and binaries) at commit/hash level, dependent on other workflow/tests?
It may be interested only build the image some tests success, but that can also add some building time to the images (waiting until the tests success).

PR #2073, opened 2023-04-24
GitHub workflows in self-hosted runners

jcortejoso:
### Description

Migrate the GitHub actions job execution to cLabs self-hosted runners.
These runners are run in GKE cluster integration-tests in celo-testnet project.

### Other changes

### Tested

Tested. Also bumped the cache version to force rebuilding.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


carterqw2 (approved):
Looks good 🚀

PR #2074, opened 2023-04-25
eth/protocols: Explain reason for protocol change

palango:
### Description

Document the reason for returning block data and block hash in the eth protocol.


karlb (approved):
Looks good, apart from one typo.

PR #2075, opened 2023-04-26
Fix transfer prestate trace

### LINKED ISSUE #2066: method handler crashed error for prestateTracer
## Expected Behavior

I expect a non `-32000` response

## Actual Behavior

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "method handler crashed"
  }
}
```

## Steps to reproduce the behavior

```
{
  "method": "debug_traceCall",
  "params": [
    {
      "from": "0x61640A8D48Bca205BA91b16B0B7745e7aBc61084",
      "to": "0xe75b0bc5dda1cbbac7075de5b1a39e9755938648",
      "value": "0x5af3107a4000"
    },
    "latest",
    {
      "tracer": "prestateTracer"
    }
  ],
  "id": 1,
  "jsonrpc": "2.0"
}
```

When replacing the `prestateTracer` with `callTracer`, it works fine

## System Information

Client version:
`celo/v1.7.3-stable-6efea76c/linux-amd64/go1.20.2`

Chain/Network: `Mainnet`

palango:
### Description

Running the pretracer on a simple celo transfer failed with an exception in the js pretracer.
```
failed to retrieve trace result: TypeError: cannot read property '0xbeb299010df30f07d6e6ac00426409...' of null    in server-side tracer function 'result'
```

This happened, because in that case the result wasn't properly initialized.

This problem has been fixed upstream in https://github.com/ethereum/go-ethereum/pull/24268. As that PR also includes the addition of the native prestate tracer, I just applied the fix and left other changes for a later merge.

I also added a testcase for a simple transfer to prevent regressions.

## Issues

Fixes #2066


eelanagaraj (approved):
LGTM!

PR #2090, opened 2023-05-11
Remove/celo 1559 tx

gastonponti:
Remove TransactionType CeloDynamicFeeTx (0x7c)


karlb (approved):
Looks like a nice, clean change. Great that you could use the exact code from geth in so many places!

PR #2092, opened 2023-05-11
Cleanup Precompiles

mcortesi:
### Description

* Move Celo preocompiles to celo_contracts.go (do the same with tests)
* Modify calling behaviour so that we don't need to modify every eth precompile (just use a wrapper for them)
* Simplify diff with contracts and contracts_test


palango (approved):
LGTM, not sure if we should let the trace calls in there or not, but I have no strong feelings about either.

palango:
@mcortesi : this should have gone to `master` instead of `feature-split`, right?

PR #2093, opened 2023-05-11
Use NewTransaction for eth compatible tx

karlb:
The goal of this commit is to reduce the diff to the original code base. To achieve this, the naming of the new transaction functions is changed:

* Create an Ethereum compatible transaction:
  before: NewTransactionEthCompatible
  after: NewTransaction (like in geth)
* Create an transaction with Celo fields:
  before: NewTransaction
  after: NewCeloTransaction

Analogously, `NewContractCreation` defaults to Ethereum compatible transactions.

When going through the code base to fix the compilation errors introduced by these name changes, I made the decision by the following criteria:
* If Celo specific tx fields are used, use NewCeloTransaction
* Otherwise, try to use NewTransaction to minimize the diff compared to geth
* If this leads to a failing test, use NewCeloTransaction instead. This might not be the optimal choice in every case, but since this commit is already quite large, this helps to keep the changes very simple.

### DIFF
--- a/accounts/abi/bind/backends/simulated_test.go
+++ b/accounts/abi/bind/backends/simulated_test.go
@@ -64,7 +64,7 @@ func TestSimulatedBackend(t *testing.T) {
 
 	code := `6060604052600a8060106000396000f360606040526008565b00`
 	var gas uint64 = 3000000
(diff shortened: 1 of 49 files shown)


gastonponti (approved):
LGTM

PR #2094, opened 2023-05-11
remove-feat:full-node-incentives

palango:
### Description

Remove full nodes incentives from the LES protocol.

This keeps all fields in celo specific transactions and still deducts the incentive if set during tx execution.


gastonponti (approved):
LGTM

PR #2095, opened 2023-05-11
remove-feat: proxy

carterqw2:


cmd/utils/flags.go
```diff
@@ -1472,108 +1440,6 @@ func setIstanbul(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
...
-	if ctx.GlobalIsSet(ProxyFlag.Name) {
-		nodeCfg.Proxy = ctx.GlobalBool(ProxyFlag.Name)
-		ethCfg.Istanbul.Proxy = ctx.GlobalBool(ProxyFlag.Name)
-
-		// Mining must not be set for proxies
-		if ctx.GlobalIsSet(MiningEnabledFlag.Name) {
-			Fatalf("Option --%s must not be used if option --%s is used", MiningEnabledFlag.Name, ProxyFlag.Name)
-		}
-		// Replica must not be set for proxies
-		if ctx.GlobalIsSet(IstanbulReplicaFlag.Name) {
```
mcortesi, line 1499: should also remove ReplicaFlag i believe

consensus/consensus.go
```diff
@@ -168,11 +168,8 @@ type PoW interface {
...
-	// IsProxiedValidator returns true if this node is a proxied validator
-	IsProxiedValidator() bool
-
-	// IsProxy returns true if this node is a proxy
-	IsProxy() bool
+	// IsValidator returns true if this node is a validator
+	IsValidator() bool
 
 	// IsPrimary returns true if this node is the primary validator
 	IsPrimary() bool
```
mcortesi, line 175: i would remove IsPrimary and related since  it's related to HotSwap

cmd/geth/usage.go
```diff
@@ -191,9 +191,7 @@ var AppHelpFlagGroups = []flags.FlagGroup{
 	},
 	{
 		Name: "ISTANBUL",
```
palango, line 193: Can we remove this whole group?

palango (approved):
Looks good, some small nit below. And needs to be rebased.

cmd/utils/flags.go
```diff
@@ -1453,119 +1414,15 @@ func setIstanbul(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 	if ctx.GlobalIsSet(LegacyIstanbulProposerPolicyFlag.Name) {
 		log.Warn("Flag value is ignored, and obtained from genesis config", "flag", LegacyIstanbulProposerPolicyFlag.Name)
 	}
-	cfg.Istanbul.ReplicaStateDBPath = stack.ResolvePath(cfg.Istanbul.ReplicaStateDBPath)
```
mcortesi, line 1456: A good "clean up" would be to move all CELO specific flag code to a celo_flags.go file

consensus/istanbul/announce/valproxy.go
```diff
@@ -20,43 +19,6 @@ type ValProxyAssigmnentProvider interface {
 	GetValProxyAssignments(valAddresses []common.Address) (map[common.Address]*enode.Node, error)
 }
 
-type proxyVPAP struct {
```
mcortesi, line 23: shouldn't we remove the whole concept of "ValProxyAssigmnentProvider". I understand is to "assign one of my proxies to another validator", basically tell them which is my "enode address" but it would actually be one of my proxy's address.

Without proxies, you just send them your address

PR #2097, opened 2023-05-11
DRAFT: remove:feat:random beacon

eelanagaraj:
Still debugging a few test failures in:
- RLP decoding/encoding tests (obviously failing due to the changed block format with `Randomness` missing, but need to re-generate test chain data)
- State differences in tracing, etherJS, e2e tests on dynamic pricing (don't yet know why there are differences in gas prices, & expected behavior when state is missing)
  - tracing seems to be broken on `feature-split` too so that's ok ✔️ 

### Description

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


palango:
Closing this as we're using a different approach now. We';; keep the branch for future reference.

PR #2098, opened 2023-05-11
Update linter and format with go1.19

karlb:
Geth does not show any formatting difference when running `gofmt -l .` with go1.19. So formatting our code according to the same rules should reduce unnecessary formatting differences.

With the new linter, `make lint` also succeeds locally, which it didn't do for me before.


karlb:
Closed in favour of https://github.com/celo-org/celo-blockchain/pull/2107.

PR #2099, opened 2023-05-11
remove-feat: ethereum compatible tx/mandatory replay protection

gastonponti:
Removal of the ethCompatible for txs, and the replay protection mandatory for Donut


mcortesi (approved):

palango:
Closing this as we're using a different approach now. We';; keep the branch for future reference.

PR #2103, opened 2023-05-12
Cleanup Precompiles (#2092)

palango:
* Move Celo preocompiles to celo_contracts.go (do the same with tests)
* Modify calling behaviour so that we don't need to modify every eth precompile (just use a wrapper for them)
* Simplify diff with contracts and contracts_test


core/vm/contracts.go
```diff
@@ -294,16 +242,14 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
 	if suppliedGas < gasCost {
 		return nil, 0, ErrOutOfGas
 	}
-	gasLeft := suppliedGas - gasCost
-	output, err := p.Run(input, caller, evm)
-	log.Trace("Finished running precompiled contract", "input", input, "gas", suppliedGas, "gas_left", gasLeft)
-	return output, gasLeft, err
+	suppliedGas -= gasCost
+	output, err := p.Run(input, ctx)
+	return output, suppliedGas, err
```
karlb, line 252: The naming `suppliedGas` is not the best fit after the gasCost is deducted. I prefer the old version with the `gasLeft`. Or just return `output, suppliedGas - gasCost, err` if you don't like the additional local variable.

karlb (approved):
Looks good, just one variable naming nitpick.

The review would have been easier to do thoroughly if moving the code to celo_contracts and the other changes were separate commits.

core/vm/contracts.go
palango: Makes sense, I changed it back.

PR #2104, opened 2023-05-15
Hex-encode difficulty in RPC

karlb:
See https://github.com/celo-org/celo-blockchain/pull/2105 instead, due to CI-from-forked-repo problem.


karlb:
See https://github.com/celo-org/celo-blockchain/pull/2105 instead, due to CI-from-forked-repo problem.

PR #2107, opened 2023-05-16
Apply `go1.18.10 fmt ./...`

karlb:
Having all code formatted by the same tools helps with consistency and reduced diffs. Why go 1.18?
* Geth v1.10.9 shows no formatting differences when that version is used for formatting
* Later versions do
* Geth v1.10.9 uses go1.15 in its go mod. But that go version is so old that using it causes a large amount of additional errors during compilation and I want to avoid using different versions for compilation and formatting.
* Formatting with go 1.18 and 1.15 yields the same results (at least for geth and celo-blockchain)

Problems not solved which would be solved by moving to golint1.51.1 and go1.19 formatting (which have to go together):
* Linting fails locally
* Linting in CI did not catch that our formatting diverged from the gofmt output

Replaces https://github.com/celo-org/celo-blockchain/pull/2098 until we use a more recent geth version as a base.

### DIFF
--- a/crypto/bls12377/arithmetic_decl.go
+++ b/crypto/bls12377/arithmetic_decl.go
@@ -1,3 +1,4 @@
+//go:build amd64 && !generic
 // +build amd64,!generic
 
 package bls12377
--- a/cmd/evm/internal/t8ntool/transition.go
+++ b/cmd/evm/internal/t8ntool/transition.go
@@ -88,7 +88,7 @@ func Transition(ctx *cli.Context) error {
 	log.Root().SetHandler(glogger)
 
 	// TODO this command is broken: vmContext and vmRunner can not be created as required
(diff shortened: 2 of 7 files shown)


crypto/bls12377/arithmetic_decl.go
```diff
@@ -1,3 +1,4 @@
+//go:build amd64 && !generic
 // +build amd64,!generic
```
karlb, line 2: Go 1.18 prefers the new format of build instructions and will add them during formatting. Once this has been added by gofmt 1.18, this output stays stable across all gofmt versions between 1.15 and 1.18.

Eventually, we can use `go fix` to remove the old instructions. But I want to wait until I am sure at least go 1.18 is used everywhere.

carterqw2 (approved):

PR #2111, opened 2023-05-17
remove-feat: gas opcodes cost differences

gastonponti:
Re adds the eip1884 which changed the gas cost of some opcodes
Also revert the MaxCodeSize for contracts


core/vm/eips.go
carterqw2, line 1: What about `BASEFEE`? Later?

karlb (approved):
The Bls12381 gas prices still differ. We could choose to include them in this PR or do it separately.

core/vm/eips.go
gastonponti: Yes, BaseFee should be restore when we are removing our way to get the minimumGasPrice

PR #2117, opened 2023-05-30
Restore opcode gas costs from Ethereum

palango:
### Description

Draft implementation for https://github.com/celo-org/celo-proposals/discussions/367

### Tested

- Adapted existing tests to new behaviour
- Added tests that exercise old expectations with g fork disabled
- Updated monorepo e2e tests

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain-planning/issues/23

### Backwards compatibility

Changes introduced at hardfork.


core/vm/jump_table.go
```diff
@@ -67,6 +67,7 @@ type JumpTable [256]*operation
 // constantinople, istanbul, petersburg, espresso and g-fork instructions.
 func newGforkInstructionSet() JumpTable {
 	instructionSet := newEspressoInstructionSet()
+	enable3529GFork(&instructionSet) // EIP-3529: Reduction in refunds https://eips.ethereum.org/EIPS/eip-3529
```
carterqw2, line 70: Hmm, shouldn't it be `enable3529`?

palango: There's `enable3529` in `newEspressoInstructionSet`, but here the dynamic gas calculation function changes, so we need a different name. 

I could also rename other to `enable3529Celo` or similar if that makes it clearer. Let me know what you think.

carterqw2: Yeah, I think renaming makes sense but also I mean `enable3529GFork` uses `gForkGasSStoreEIP3529` which in turn uses `params.CeloSstoreClearsScheduleRefundEIP3529` which equals to `6100`, not `4800` or I'm missing something?

palango: Yes, good catch, I mixed those up.

core/vm/operations_acl.go
```diff
@@ -216,7 +243,8 @@ var (
 
 	// gasSStoreEIP2539 implements gas cost for SSTORE according to EPI-2539
 	// Replace `SSTORE_CLEARS_SCHEDULE` with `SSTORE_RESET_GAS + ACCESS_LIST_STORAGE_KEY_COST` (6,100)
-	gasSStoreEIP3529 = makeGasSStoreFunc(params.SstoreClearsScheduleRefundEIP3529)
+	gasSStoreEIP3529Celo = makeGasSStoreFunc(params.CeloSstoreClearsScheduleRefundEIP3529)
```
carterqw2, line 246: nit: maybe we could swap these two lines and restore the original comment.

carterqw2 (approved):

monorepo_commit
```diff
@@ -1 +1 @@
-d216e7599f499adec2e0200e599f601ea5424cde
+f2ed7dd9fa5cf11f9ffa773921cf4d1f99a5948d
```
palango, line 1: @gastonponti This commit results in the correct gas values: https://github.com/celo-org/celo-monorepo/commit/f2ed7dd9fa5cf11f9ffa773921cf4d1f99a5948d

However, it's still a bit unclear where those values come from. Our backwards calculation resulted in the commit linked above.

I also tried understanding it better by tracing those transaction. However this ended in even more questions 😆 
- When calling the normal tracer, it reports a gas usage of 102953.
- When using the call tracer a gas usage of 31321 is reported on the outermost frame
  - When adding the intrinsic tx costs to that value I get 102,321, which gives a difference of 632 that I cannot explain (yet)

palango: > This gives me 35300, not 34300. Is it a typo?

Yes, probably mixed something up.

carterqw2: Looks like you haven't added memory data costs to the intrinsic costs
`31,321 + 21,000 + 50,000 = 102,321` but then there is also
```
gas_cost += 4 * bytes_zero: gas added to base cost for every zero byte of memory data
gas_cost += 16 * bytes_nonzero: gas added to base cost for every nonzero byte of memory data
```
that need to be added.

core/vm/operations_acl.go
```diff
@@ -66,7 +69,11 @@ func makeGasSStoreFunc(clearingRefund uint64) gasFunc {
 			}
 			// EIP-2200 original clause:
 			//		return params.SstoreResetGasEIP2200, nil // write existing slot (2.1.2)
-			return cost + (params.SstoreResetGasEIP2200 - params.ColdSloadCostEIP2929), nil // write existing slot (2.1.2)
+			if evm.chainRules.IsGingerbread {
```
gastonponti, line 72: Every case it seems to be perfect, but I would add a little refactor.
Instead of checking for the gingerBread fork every time we use it, I would add a function
```
func getColdSloadCostEIP2929(evm *EVM) uint64 {
   if evm.chainRules.IsGingerbread {
      return params.ColdSloadCostEIP2929
   }
   return params.CeloColdSloadCostEIP2929
}
```
and then in every place where the `params.ColdSloadCostEIP2929` was called, just put `getColdSloadCostEIP2929(evm)`.
This way, everyone of this function will have less changes from upstream, and our logic of the gasCost will be in only one little function

gastonponti (commented):
I think that we have a possibility of a refactor there. But other than that, the PR looks good

core/vm/operations_acl.go
palango: That makes stuff a lot more readable!

PR #2118, opened 2023-05-31
mycelo: enable GFork by default

karlb:
We always want to run and test the latest hard fork by default.

### DIFF
--- a/mycelo/genesis/genesis.go
+++ b/mycelo/genesis/genesis.go
@@ -28,6 +28,7 @@ func CreateCommonGenesisConfig(chainID *big.Int, adminAccountAddress common.Addr
 		ChurritoBlock: common.Big0,
 		DonutBlock:    common.Big0,
 		EspressoBlock: common.Big0,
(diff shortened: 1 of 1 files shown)


palango:
Slightly related: do we want to enable gfork in the `TestChainConfig` by default as well?

karlb:
> TestChainConfig

Yes, IMO we should always test on the latest hardfork unless an older fork is explicitly required by the test.

karlb:
[End-to-end CIP35-eth compatibility test](https://github.com/celo-org/celo-blockchain/actions/runs/5130653671/jobs/9231302550#logs) fails with
```
spawnCmd failed for: /runner/_work/celo-blockchain/celo-blockchain/geth/build/bin/geth --datadir /tmp/e2e/validator/datadir init /tmp/e2e/genesis.json
Error: Process completed with exit code 1.
```
but I don't see the specific error reason, yet.

karlb:
Included in https://github.com/celo-org/celo-blockchain/pull/2062.

PR #2120, opened 2023-06-01
Deprecate gateway fee

palango:
### Description

Deprecating the gateway fee.

- Removes the `--light.gatewayfee` argument
- Adds checks to txpools to not accept transactions that include either `GatewayFeeRecipient` or `GatewayFee`
- Adds check to the state machine to not accept transactions that include either `GatewayFeeRecipient` or `GatewayFee`
- Update tests for new behaviour, keep old tests with a pre-g-fork config
- Corresponding monorepo changes: https://github.com/celo-org/celo-monorepo/commit/2e5d33f6b4f582b09ee344917f7245b5845fedce

This doesn't include changes to the LES protocol yet, as those can be done independently of a hardfork and are therefore not time sensitive.


core/state_processor.go
```diff
@@ -124,6 +124,16 @@ func applyTransaction(msg types.Message, config *params.ChainConfig, gp *GasPool
 		return nil, ErrUnprotectedTransaction
 	}
 
+	// TODO(paul): Add CIP number
+	// Since CIP XXX full node incentives are deprecated.
+	// Check that neither `GatewayFeeRecipient` nor `GatewayFee` are set, otherwise reject the transaction
+	if config.IsGFork(blockNumber) {
+		gatewayFeeSet := !(msg.GatewayFee() == nil || msg.GatewayFee().Cmp(common.Big0) == 0)
```
gastonponti, line 131: So, we have txs, that maybe have the gatewayFee in zero?

core/tx_pool_test.go
```diff
@@ -326,6 +326,7 @@ func TestInvalidTransactions(t *testing.T) {
 
 	pool, key := setupTxPool()
 	defer pool.Stop()
+	pool.gfork = false
```
gastonponti, line 329: Why shouldn't we add the case here of a gatewayFee bigger than zero?
If this test aims to check the whole validity of the transactions in the last fork, we should add the case too.
If we want to maintain this as testing just before the Gfork, I would change the name of the test, and maybe copy-paste it and change the copy to mimic a valid tx in the gFork.

My point here is that we should have a test that "checks every check" working in the last fork. Otherwise we can break one of the other checks in the gFork, without noticing (it's not the case, but I think that you get my point)

e2e_test/e2e_transfer_test.go
```diff
@@ -31,7 +31,189 @@ const (
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			watcher := test.NewBalanceWatcher(client, []common.Address{sender.Address, recipient.Address, gateWayFeeRecipient.Address, node.Address})
+			blockNum, err := client.BlockNumber(ctx)
+			require.NoError(t, err)
+			signer := types.MakeSigner(devAccounts[0].ChainConfig, new(big.Int).SetUint64(blockNum))
+			tx, err := prepareTransaction(*tc.txArgs, sender.Key, sender.Address, signer, client)
+			require.NoError(t, err)
+			err = client.SendTransaction(ctx, tx)
+			if tc.expectErr {
```
gastonponti, line 144: Can we check that is the error that we expect and not another?

eth/api_backend.go
```diff
@@ -418,10 +418,6 @@ func (b *EthAPIBackend) ServiceFilter(ctx context.Context, session *bloombits.Ma
 	}
 }
 
-func (b *EthAPIBackend) GatewayFeeRecipient() common.Address {
-	return b.eth.GatewayFeeRecipient()
-}
-
 func (b *EthAPIBackend) GatewayFee() *big.Int {
```
gastonponti, line 421: shouldn't this also be removed?

internal/ethapi/backend.go
```diff
@@ -100,7 +100,6 @@ type Backend interface {
 
 	ChainConfig() *params.ChainConfig
 
-	GatewayFeeRecipient() common.Address
 	GatewayFee() *big.Int
```
gastonponti, line 103: Same as the other

les/api_backend.go
```diff
@@ -390,7 +390,7 @@ func (b *LesApiBackend) ServiceFilter(ctx context.Context, session *bloombits.Ma
 }
 
 func (b *LesApiBackend) GatewayFeeRecipient() common.Address {
```
gastonponti, line 392: remove?

internal/web3ext/web3ext.go
```diff
@@ -821,26 +821,6 @@ web3._extend({
 			name: 'addBalance',
 			call: 'les_addBalance',
 			params: 2
-		}),
-		new web3._extend.Method({
```
gastonponti, line 825: Shouldn't we keep these and maybe return an error?
2 things here
1) it will be weird for the user, maybe the user won't understand if it is an error or not if an rpc call disapeared 
2) technically, this should be effective from the gFork, not before. So everyone that updated to this version, should be able to use this until the activation

Maybe it's better to keep this, with a if-fork between the an error from the gFork, and the actual behaviour and prune it after the checkpoint...

internal/ethapi/transaction_args.go
```diff
@@ -149,12 +149,6 @@ func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend) error {
 	if args.To == nil && len(args.data()) == 0 {
 		return errors.New(`contract creation without any data provided`)
 	}
-	if args.GatewayFeeRecipient == nil && !args.EthCompatible {
```
gastonponti, line 152: Same question from the web3 file... Do we want this version to keep it's behaviour until the fork?

eth/api_backend.go
gastonponti: or keep these and return an error after the fork?

core/state_processor.go
palango: Yes, that's what I found while adapting the tests.

internal/web3ext/web3ext.go
palango: You're right. I took the easy route assuming that we only had few transactions using that path, but I fixed it to do the proper hardfork check.

palango:
Changes after review:
- Don't remove `light.gatewayfee` flag. Instead move it to the deprecated section.
- Keep current transaction behaviour until hardfork. This means old code cannot be removed yet.
- Throw errors on JS API after hardfork instead of removing it outright.

internal/ethapi/transaction_args.go
palango: Done

e2e_test/e2e_transfer_test.go
palango: Done

monorepo_commit
```diff
@@ -1 +1 @@
-d4725a468fa20add9034efd109975b9fc7ec51d1
+b9cf8d7bbad5dcd74a8a91b6d6439f314b540db5
```
karlb, line 1: If this is merged after https://github.com/celo-org/celo-blockchain/pull/2130, this should be changed to `be301c4b17044e08bf5fb01395804816ee950957`.

core/tx_pool_test.go
```diff
@@ -341,7 +341,57 @@ func TestInvalidTransactions(t *testing.T) {
...
+	tx = lesTransaction(0, 100, big.NewInt(50), key)
+	if err := pool.AddRemote(tx); err != ErrGatewayFeeDeprecated {
+		t.Error("expected", ErrGatewayFeeDeprecated, "got", err)
+	}
+
+	// Should return to intrinsic gas error when gateway fee is covered.
+	pool.currentState.AddBalance(from, tx.GatewayFee())
+	if err := pool.AddRemote(tx); err != ErrGatewayFeeDeprecated {
+		t.Error("expected", ErrGatewayFeeDeprecated, "got", err)
+	}
```
karlb, line 354: The comments don't match the code.

karlb (commented):
This PR brings a lot of GFork names again, which should be gingerbread by now.

palango:
> This PR brings a lot of GFork names again, which should be gingerbread by now.

My idea was to merge this before the renaming PR, because the commit fixing the monorepo tests is already in the branch. So I'd fix those in #2127

karlb:
> > This PR brings a lot of GFork names again, which should be gingerbread by now.
> 
> My idea was to merge this before the renaming PR, because the commit fixing the monorepo tests is already in the branch. So I'd fix those in #2127

Ah, I somehow thought the rename PR was already merged. OK, then just fix the wrong comments and we're ready to go!

karlb (approved):

PR #2127, opened 2023-06-13
Rename gfork to gingerbread

palango:
### Description

- Give the hardfork it's proper name
- Adapt the `override.espresso` flag to gingerbread


karlb (approved):
I am not sure how the overrides are used, so I can't judge if the override should be switched from espresso to gingerbread. Apart from that, it looks good!

palango:
Failing tests will pass once #2130 is merged.

PR #2128, opened 2023-06-13
Restore baseFee opcode

gastonponti:
### Description

Implementation of the CIP-61: Restore BaseFee opcode/header field.

### Tested

Run test, e2e tests, and spin a testnet checking the hf switch in a heavy load

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

This is part of the Gingerbread hardfork


mycelo/genesis/genesis.go
```diff
@@ -20,15 +20,16 @@ var genesisMsgHash = common.HexToHash("ecc833a7747eaa8327335e8e0c6b6d8aa3a38d006
 
 // CreateCommonGenesisConfig generates a config starting point which templates can then customize further
 func CreateCommonGenesisConfig(chainID *big.Int, adminAccountAddress common.Address, istanbulConfig params.IstanbulConfig) *Config {
-	genesisConfig := BaseConfig()
+	gingerbreadBlock := newBigInt(2)
+	genesisConfig := BaseConfig(gingerbreadBlock)
```
karlb, line 24: Is this something we really want, or is it just included temporarily for testing the change?

accounts/abi/bind/backends/simulated.go
```diff
@@ -501,6 +502,38 @@ func (b *SimulatedBackend) EstimateGas(ctx context.Context, call ethereum.CallMs
...
+		if allowance.IsUint64() && hi > allowance.Uint64() {
+			transfer := call.Value
+			if transfer == nil {
+				transfer = new(big.Int)
+			}
+			log.Warn("Gas estimation capped by limited funds", "original", hi, "balance", balance,
+				"sent", transfer, "feecap", feeCap, "fundable", allowance)
+			hi = allowance.Uint64()
+		}
+	}
```
karlb, line 536: Why is this necessary to add the baseFee header field or opcode?

eth/gasprice/gasprice.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/celo-org/celo-blockchain/params"
+)
+
+var (
+	suggestionMultiplier *big.Int = big.NewInt(5) // The multiplier that we apply to the minimum when suggesting gas price
+)
+
+func GetBaseFeeForCurrency(vmRunner vm.EVMRunner, currencyAddress *common.Address, baseFee *big.Int) (*big.Int, error) {
+	// Gingerbread Fork
+	if baseFee != nil {
```
karlb, line 35: Not sure what the most relevant convention is, but I'm used reading it as either
```go
	if baseFee != nil {  // Gingerbread Fork
```
or
```go
	if baseFee != nil {
        // Gingerbread Fork
        ...
```

eth/gasprice/gasprice.go
```diff
@@ -0,0 +1,87 @@
...
+		if currencyAddress == nil {
+			return baseFee, nil
+		} else {
+			exchangeRate, err := cm.GetExchangeRate(vmRunner, currencyAddress)
+			if err != nil {
+				// Assign zero if the exchangeRate fails to mimic the same behaviour as Espresso
+				return big.NewInt(0), err
```
karlb, line 42: Is zero a good fallback here? Shouldn't we make transactions expensive rather than free if in doubt?

eth/gasprice/gasprice.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			return exchangeRate.FromBase(baseFee), nil
+		}
+	}
+
+	return gpm.GetGasPriceMinimum(vmRunner, currencyAddress)
```
karlb, line 48: We could reduce the nesting of most of the function if we started the function with
```go
if baseFee == nil {
	return gpm.GetGasPriceMinimum(vmRunner, currencyAddress)
}
```
I also prefer to have the code for old forks in the `if` and the current fork on lowest indentation.

karlb (commented):
This got bigger than I originally anticipated. Looks good so far. Feel free to ignore the nitpicks.

eth/gasprice/gasprice.go
gastonponti: I this case, as I'm going to make the other recommendation of the `if baseFee == nil {`, I'll keep it after the if

e2e_test/e2e_transfer_test.go
```diff
@@ -30,6 +30,8 @@ const (
 // - Governance account has base fee added.
 // - validator account has tip fee added.
 func TestTransferCELO(t *testing.T) {
+	// log.Root().SetHandler(log.LvlFilterHandler(log.LvlTrace, log.StreamHandler(os.Stderr, log.TerminalFormat(true))))
+	// fmt.Println(tt.dump())
```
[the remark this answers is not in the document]
gastonponti: No, it was used a way of debugging, but I'll remove it

mycelo/genesis/genesis.go
gastonponti: Temporary for testing

palango:
What's the relation of this PR and https://github.com/celo-org/celo-monorepo/pull/10274? Do we need to wait until the update is merged or do we just pull the changes in our monorepo branch? If so, will that cause problems with the v10 contracts release?

core/vm/jump_table.go
```diff
@@ -67,6 +67,7 @@ type JumpTable [256]*operation
 // constantinople, istanbul, petersburg, espresso and gingerbread instructions.
 func newGingerbreadInstructionSet() JumpTable {
 	instructionSet := newEspressoInstructionSet()
+	enable3198(&instructionSet) // Base fee opcode https://eips.ethereum.org/EIPS/eip-3198
```
palango, line 70: nit: From the name and ordering I'd expect that below `enable3529`.

accounts/abi/bind/backends/simulated.go
gastonponti: This is something I've bringed from upstream to be as close as possible, but is something that I can remove if we are not confident that we want it

karlb: It is fine with me, I was just irritated because I didn't see the reason for the change. That's clearer now.

eth/gasprice/gasprice.go
gastonponti: Yes, haha, agree on that. I've added it here as zero because it was the same behaviour we had on production. Technically speaking, if we don't have the gasPriceMinimum contract nor the sorted oracles of a stable, we are probably in a bigger problem, but yes, we could change the default on how we handle this

karlb: I agree that his is a rare edge case and the specific handling is not too important. We can leave it as zero to minimize the changes for now, if you think that is safer.

consensus/misc/eip1559.go
```diff
@@ -0,0 +1,103 @@
...
+			newBaseFee = calcBaseFeeEthereum(config, parent)
+		}
+		return newBaseFee
+	}
+
+	// // If an error occurs, the default block gas limit will be returned and a log statement will be produced by GetBlockGasLimitOrDefault
+	// gasLimit, err := blockchain_parameters.GetBlockGasLimit(vmRunner)
+	// if err != nil {
+	// 	return nil, err
+	// }
```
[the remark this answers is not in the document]
gastonponti: Yes, merged from master the gasLimit change in the header and keep the old one for testing. I'll remove it

core/state_processor_test.go
```diff
@@ -325,7 +325,13 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
...
 	}
 	header.Extra = CreateEmptyIstanbulExtra(header.Extra)
-
+	if config.IsGingerbread(header.Number) {
+		if config.FakeBaseFee != nil {
+			header.BaseFee = config.FakeBaseFee
+		} else {
+			header.BaseFee = common.Big0
+		}
+	}
```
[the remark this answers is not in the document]
gastonponti: This badBlock function stays that "the block is sufficiently valid to be considered for import".
Everything there is mocked. In our scenario, as we require a runner to calculate the gasPrice, would have being changing a lot for a field that is not important in the tests that are invoking it.
Also I've noticed that we have the creation of the Extra field, twice (I'll check if it works without the second set, which should, and remove that line)

karlb (approved):

core/types/block.go
```diff
@@ -353,6 +353,9 @@ func CopyHeader(h *Header) *Header {
 	if h.Number != nil {
 		cpy.Number.Set(h.Number)
 	}
+	if h.BaseFee != nil {
+		cpy.BaseFee = new(big.Int).Set(h.BaseFee)
```
mcortesi, line 357: what if h.BaseFee is nil?

cmd/mycelo/genesis_commands.go
```diff
@@ -99,7 +99,7 @@ var createGenesisFromConfigCommand = cli.Command{
 func readBuildPath(ctx *cli.Context) (string, error) {
 	buildpath := ctx.String(buildpathFlag.Name)
 	if buildpath == "" {
-		buildpath = path.Join(os.Getenv("CELO_MONOREPO"), "packages/protocol/build/contracts")
+		buildpath = path.Join(os.Getenv("CELO_MONOREPO"), "packages/protocol/build")
```
mcortesi, line 102: is this ok?

gastonponti: Yes, the monorepo changed their build folder to have `contracts`, and `contracts-mento`. As they have some contracts that collide with their names, I preferred to change our mycelo implementation to decide which folder to use

core/types/block.go
gastonponti: The cpy will still have a nil as baseFee. I don't know if I understand the question

PR #2130, opened 2023-06-14
Eth header hash compatibility

karlb:
### Description

Use same header layout as Ethereum, so that Eth tools can caculate the header hash and get the same results as we do.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain-planning/issues/100


core/types/block_compatibility_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEthereumHeaderHashCompatibility(t *testing.T) {
+	// Ethereum tools calculating bloch header hashes show work on Celo just as
```
palango, line 11: typo: block instead of bloch

core/types/block.go
carterqw2, line 1: I was checking the `NewBlock` function against geth, shouldn't we set the `EmptyUncleHash` there?

palango: Similar question: there's some checks in `SanityCheck` upstream that we could bring back.

karlb: I can easily use geth's `SanityCheck` to reduce the diff (checking the constants is not too useful itself).

For the `NewBlock` the situation is less clear:
* I can't use geth's code, because `Block.uncles` does not exist and I don't want to introduce it in this PR (if at all)
* Setting it to the constant only after gingerbread does not work because I can't check for forks with the available parameters
* Setting it to constant regardless of fork would mean a change in behaviour before gingerbread. I don't think it would actually break anything because we do checks in other places, but I don't like it.

palango: > I can easily use geth's `SanityCheck` to reduce the diff (checking the constants is not too useful itself).

I agree for `difficulty` we don't gain anything, but `basefee` will actually set and it might be nice to add the check for it.

core/types/block_compatibility_test.go
```diff
@@ -0,0 +1,44 @@
+package types
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEthereumHeaderHashCompatibility(t *testing.T) {
```
palango, line 10: Should we add a similar test with a recent celo block to make it obvious that the hashing is stable there as well?

karlb: Yes, that's easy to do and covers another case, so I'll do it!

palango (approved):
Looks good to me. And it turns out to require surprisingly small changes. 🎉

PR #2135, opened 2023-06-20
Pull later patch version of node-gyp to fix macOS

palango:
### Description

Version 6 of `node-gyp` has a known problem on macOS, therefore we now allow later versions of the 6.x series in the monorepo and pull this change here.


karlb (approved):
[The latest version 6 release is still from 2020](https://github.com/nodejs/node-gyp/blob/main/CHANGELOG.md#v610-2020-01-08), but better a small update than none.

For reference: the actual change is https://github.com/celo-org/celo-monorepo/commit/ec320d51889787078d66cdb498a12d629b2b7be3

carterqw2 (approved):

PR #2139, opened 2023-06-26
Update to contracts v10

karlb:
Since we want to wrap up the Gingerbread fork soo, we should start to depend on the branches in the monorepo which will eventually be used for the fork:
* https://github.com/celo-org/celo-monorepo/pull/10227
* and the v10 contracts branch it is based on: https://github.com/celo-org/celo-monorepo/pull/10227

I also added the monorepo commits we depend on which have not been merged into the v10 contracts branch yet. See https://github.com/celo-org/celo-monorepo/tree/karlb/core-contracts.v10-ultragreen

So main part of this PR is actually in the changed monorepo commit and the changes to this repo are quite simple.


palango (approved):
Looks good, we should probably also try to merge our changes back into `master` in the monorepo.

PR #2143, opened 2023-06-27
Remove LES v4 and v5 protocols

palango:
### Description

Remove the latest LES protocol versions. This will make `GetGatewayFeeMsg` and `GatewayFeeMsg` unused. They cannot be removed yet as they are used in the JS API, but removal will be possible after gingerbread.


palango:
There's a problem with how older LES protocols handle transaction pruning, so we cannot remove them like this.

PR #2145, opened 2023-06-28
Revert "Cleanup/tx processing (#2102)"

hbandura:
This reverts commit c8843c1c80910494dba8d7075213fa2933ccd91e. #2102 made a change in the state_transition that ended up forking mainnet blocks. We can re-create the cleanup in smaller chunks afterwards, to avoid this problem.


gastonponti (approved):
LGTM

PR #2147, opened 2023-07-02
fix: udp already listening issue

### LINKED ISSUE #1867: udp listen port is not always unbound when using lightest client.
## Description

Repeatedly running `./geth --datadir lightest --syncmode lightest --port 30303` and then closing using `ctrl-c` sometimes results in the oprating system thinking that udp port 30303 is still bound even after the server has shut down. This prevents re-running geth with the same port `Fatal: Error starting protocol stack: listen udp [::]:30303: bind: address already in use`. This state remains despite waiting several hours, the workaround is to change the port.

System: `uname -a
Darwin C02FV1SBMD6R 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:54 PST 2021; root:xnu-8019.61.5~1/RELEASE_X86_64 x86_64 i386 MacBookPro16,1 Darwin`

kehiy:
### Description
I added a missing defer in main function of node

### Other changes
also i used go fmt command for make code structure clean

### Related issues

- Fixes #1867


kehiy:
@shazow @fjl @Shadowfiend @jmcnevin

karlb:
The new defer makes sense and looks like it should resolve the linked issue (I didn't actually try it out yet).

The format change is nothing we want to include at the moment, because it increases the diff to the upstream geth version and will cause additional merge conflicts when updating to more recent upstream versions. The code is already formatted with gofmt, but with version 1.18 specifically to avoid this problem. You probably used a more recent gofmt version.

Thanks for the PR!

kehiy:
> The new defer makes sense and looks like it should resolve the linked issue (I didn't actually try it out yet).
> 
> The format change is nothing we want to include at the moment, because it increases the diff to the upstream geth version and will cause additional merge conflicts when updating to more recent upstream versions. The code is already formatted with gofmt, but with version 1.18 specifically to avoid this problem. You probably used a more recent gofmt version.
> 
> Thanks for the PR!

is that going to be merged or not?
only the connection part (if the formatting is not planned)

karlb (approved):
The tests succeed, the CI just seems to be unable to post a PR comment (I assume because  the PR is made from a different repo).

kehiy:
> The tests succeed, the CI just seems to be unable to post a PR comment (I assume because the PR is made from a different repo).

yeah, maybe(?)

PR #2148, opened 2023-07-03
fix: defer usage

kehiy:
### Description
there is more connections in celo that they don't use defer 
now I add them and we can make sure all connections, sockets and ports will closed when a node going shutdown.

### DIFF
--- a/p2p/discover/v4_udp_test.go
+++ b/p2p/discover/v4_udp_test.go
@@ -571,6 +571,8 @@ func startLocalhostV4(t *testing.T, cfg Config) *UDPv4 {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer socket.Close()
+	
 	realaddr := socket.LocalAddr().(*net.UDPAddr)
 	ln.SetStaticIP(realaddr.IP)
 	ln.SetFallbackUDP(realaddr.Port)
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -660,6 +660,8 @@ func (srv *Server) setupDiscovery() error {
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


kehiy:
@karlb can you review it?

p2p/discover/v4_udp_test.go
```diff
@@ -571,6 +571,8 @@ func startLocalhostV4(t *testing.T, cfg Config) *UDPv4 {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer socket.Close()
```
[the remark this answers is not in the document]
kehiy: about the socket connections i'm agree with you
but the connection in `server.go` is should be closed I think

p2p/server.go
```diff
@@ -660,6 +660,8 @@ func (srv *Server) setupDiscovery() error {
 	if err != nil {
 		return err
 	}
+	defer conn.Close()
```
karlb, line 663: I didn't find the place where the close is called (or should be called) yet, but calling it when `setupDiscovery` returns is too early, since real usage only starts afterwards.

kehiy: that't make sense, but we should find a good place for closing it

karlb: Maybe it is being closed properly. Is there a specific reason to believe otherwise?

kehiy: the main reason to make this change for me is because it can make another issue when nodes go shutdown
if we don't make sure all listening ports and other stuff is closed gracefully and correctly 
it can cause some issue like port already listening or make running node again hard
at all i'm not expert in blockchain and network and maybe it can be wrong idea

PR #2150, opened 2023-07-04
Faucet validator accounts

jcortejoso:
### Description

Quick change on mycelo `load-test` template to also faucet validator accounts in genesis.


palango (approved):
👌

Issue #2134, opened 2023-06-20
Seg fault / panic syncing mainnet on v1.7.4


aaronmboyd:
## Expected Behavior

Sync from genesis without panic

## Actual Behavior

Panic and geth terminated. Resumes syncing normally after restart.

## Steps to reproduce the behavior

Build celo-blockchain from source tag `v1.7.4`
Go: `go1.19.2.linux-arm64, dependency in custom Dockfile`
Docker: `24.0.2`
Running on: `Ubuntu 22.04.2 LTS ARM64`

Dockerfile:
```
FROM arm64v8/ubuntu:jammy as builder

RUN apt update && apt install gcc make musl-dev wget -y

RUN wget https://go.dev/dl/go1.19.2.linux-arm64.tar.gz
RUN rm -rf /usr/local/go && tar -C /usr/local -xzf go1.19.2.linux-arm64.tar.gz
ENV PATH $PATH:/usr/local/go/bin

ADD . /go-ethereum
WORKDIR /go-ethereum
RUN make geth-musl

FROM arm64v8/ubuntu:jammy
ARG COMMIT_SHA

COPY --from=builder /go-ethereum/build/bin/geth /usr/local/bin/
RUN echo $COMMIT_SHA > /version.txt
ADD scripts/run_geth_in_docker.sh /

EXPOSE 8545 8546 30303 30303/udp
ENTRYPOINT ["sh", "/run_geth_in_docker.sh"]

# Add some metadata labels to help programatic image consumption
ARG COMMIT=$COMMIT_SHA
ARG VERSION=""
ARG BUILDNUM=""

LABEL commit="$COMMIT" version="$VERSION" buildnum="$BUILDNUM"
```

## Backtrace
```
INFO [06-19\|13:17:04.423] Imported new chain segment               blocks=190  txs=1079  mgas=243.352 elapsed=8.336s    mgasps=29.192  number=18,830,021 hash=234037..f78e82 age=2mo2d1h  dirty=74.86MiB
--
INFO [06-19\|13:17:07.629] Sending val enode share msg to proxy     func=sendValEnodeShareMsgs       proxy peer="Peer 16331f68e399c8af 10.192.11.7:30503" valAddresses length=119
INFO [06-19\|13:17:07.630] Skipping sending ValEnodesShareMsg b/c not validating func=sendValEnodesShareMsg
[...]

PR #2157, opened 2023-07-11
Improve error logging in toCeloFn

karlb:
This might help us debug
https://github.com/celo-org/celo-blockchain/issues/2134 if it happens again.

### DIFF
--- a/miner/block.go
+++ b/miner/block.go
@@ -387,7 +387,10 @@ func createConversionFunctions(sysCtx *core.SysContractCallCtx, chain *core.Bloc
 		return sysCtx.GetGasPriceMinimum(feeCurrency)
 	}
 	toCeloFn := func(amount *big.Int, feeCurrency *common.Address) *big.Int {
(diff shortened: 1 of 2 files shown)


miner/block.go
```diff
@@ -387,7 +387,10 @@ func createConversionFunctions(sysCtx *core.SysContractCallCtx, chain *core.Bloc
 		return sysCtx.GetGasPriceMinimum(feeCurrency)
 	}
 	toCeloFn := func(amount *big.Int, feeCurrency *common.Address) *big.Int {
-		curr, _ := currencyManager.GetCurrency(feeCurrency)
+		curr, err := currencyManager.GetCurrency(feeCurrency)
+		if err != nil {
+			log.Error("toCeloFn: could not get currency", "err", err)
+		}
```
karlb, line 393: This should help a lot, since we just discarded all the useful error information before.

carterqw2 (approved):

PR #2163, opened 2023-07-25
Use UncleHash instead of Difficulty in EncodeRLP check

eelanagaraj:
### Description

Checking Difficulty breaks Rosetta since it seems that the Difficulty is 0x0 instead of nil.

### Tested
Tests passed locally and fixed the block hash issue seen in Rosetta.
(Will wait until CI is green but if we run into issues will add back the `|| Difficulty == nil` condition)


karlb (approved):
Looks good. Just a few more words for context:

`Difficulty` is non-nil in the Rosetta case because Rosetta gets the header via JSON and our JSON endpoints added a `difficulty` field during serialisation to increase Ethereum compatibility.

The UncleHash is not included in the JSON output before Gingerbread, leading to a default value (zero hash) after deserialization in our Go code or creating a pre-gingerbread header programmatically.

PR #2165, opened 2023-07-26
tests: Remove unused storage maps

palango:
Small cleanup.


karlb (approved):
`0x91646b8507bf2e54d7c3de9155442ba111546b81af1cbdd1f68eeb6926b98d58` is unused in our current Proxy implementation AFAICS (assuming it is compiled from the source in the monorepo). Considering https://github.com/celo-org/celo-blockchain/blob/4a1e3fc3b9437cc45497a5da42aa797c0a5bbacf/contracts/testutil/utils.go#L13 talks about registryMock, there might be mock contracts involved. I don't see any solidity source for such mock contracts, though.

Considering it is only used in the tests, removal should be fine as long as the tests pass.

PR #2169, opened 2023-08-04
`mycelo genesis` gingerbread activation fix

palango:
### Description

Fixes gingerbread activation in mycelo.

This results in this genesis config, instead of `gingerbreadBlock` not being set:
```json
[...]
   "churritoBlock": 0,
   "donutBlock": 0,
   "espressoBlock": 0,
   "gingerbreadBlock": 1,
[...]
```


cmd/mycelo/genesis_commands.go
```diff
@@ -135,6 +136,8 @@ func envFromTemplate(ctx *cli.Context, workdir string) (*env.Environment, *genes
 		} else {
 			gingerbreadBlock = big.NewInt(gingerbreadBlockNumber)
 		}
+	} else {
+		gingerbreadBlock = common.Big1
```
karlb, line 140: Why 1? I would have expected 0?

palango: Then I run into this: https://github.com/celo-org/celo-blockchain/blob/459662f782d27ece531e967eb1cd7abb1ad8457c/mycelo/genesis/genesis.go#L28-L34

Maybe there's a better solution though.

gastonponti:
We have 2 things there
 1) the zero as disabled in the core contract
 2) the way we have to set the base fee for the specific block of the hardfork

For 1), I didn't wanted to change the logic of what the other team was auditing nor add some hacky ways as infinite o something similar just because of a contract limitation

For 2), the way to calculate the block N depends on the block N-1, and the activation block, is the one that instead of using the block N-1, consumes the state at the beginning of the block
So, if we activate the block in the genesis, it means that the block 1 will rely on searching the baseFee on the genesis, and, I didn't wanted to avoid adding the baseFee in the genesis block
If starts with 1, it's going to search for the state already set in the genesis, and will work as it should

I thought that it was not going to generate any issue, mostly because the chain construction starts at block 1, so technically, any block is constructed as gingerbread, but maybe I should have explained it better

karlb (approved):

cmd/mycelo/genesis_commands.go
```diff
@@ -126,7 +127,7 @@ func envFromTemplate(ctx *cli.Context, workdir string) (*env.Environment, *genes
 	if ctx.IsSet("mnemonic") {
 		env.Accounts().Mnemonic = ctx.String("mnemonic")
 	}
-
+	fmt.Println("envFromTemplate: GB block:", ctx.Int64("forks.gingerbread"))
```
palango, line 130:
```suggestion
```

cmd/mycelo/genesis_commands.go
```diff
@@ -185,8 +188,6 @@ func envFromTemplate(ctx *cli.Context, workdir string) (*env.Environment, *genes
 		}
 	}
 
-	genesisConfig.Hardforks.GingerbreadBlock = gingerbreadBlock
```
gastonponti, line 188: If you remove this line you won't be setting the block.
The first if above is required because the `baseConfig` requires the gingerblock

gastonponti: And that baseConfig gives you the genesisConfig which you are setting. I'm changing it also with the default in 1 for the P2

palango: That's also where the forks get set: https://github.com/celo-org/celo-blockchain/blob/925fba88d7f84bc5e35c776ee73258eb9f8df8aa/mycelo/genesis/genesis.go#L43-L44

There's some additional logic that sets them to 1, when no other value is given. So in case of not passing the argument, we would overwrite with `nil` here, oven though it was set before. That's the whole problem this PR fixes.

gastonponti: ohhhh, missed that one

gastonponti:
Unit tests are passing but are flaky in github actions

PR #2170, opened 2023-08-05
Celo Dynamic Fee V2

gastonponti:
### Description

Celo dynamic fee v2 implements the CIP-XX

### Backwards compatibility

Part of the Gingerbread Hard Fork


e2e_test/e2e_transfer_test.go
```diff
@@ -153,12 +173,12 @@ func TestTransferCELO(t *testing.T) {
...
 			require.NoError(t, err)
 			signer := types.MakeSigner(devAccounts[0].ChainConfig, new(big.Int).SetUint64(blockNum))
-			tx, err := prepareTransaction(*tc.txArgs, sender.Key, sender.Address, signer, client)
+			tx, err := prepareTransaction(*tc.txArgs, sender.Key, sender.Address, signer, client, true)
 			require.NoError(t, err)
 			err = client.SendTransaction(ctx, tx)
 			if tc.expectedErr != nil {
 				// Once the error is checked, there's nothing more to do
-				if err.Error() != tc.expectedErr.Error() {
+				if err == nil || err.Error() != tc.expectedErr.Error() {
```
karlb, line 181: Since we are inside the `if tc.expectedErr != nil` branch, the new check is redundant. Or is it somehow possible for
`tc.expectedErr != nil` and`tc.expectedErr.Error() == nil`?

internal/ethapi/api_test.go
```diff
@@ -11,6 +11,58 @@ import (
...
+			FeeCurrency: &currency,
+			ChainID:     chainId,
+
+			GasFeeCap: bigFeeCap,
+			GasTipCap: gasTipCap,
+		}), blockHash, blockNumber, index, baseFeeFn, true)
+		assert.Equal(t, (*hexutil.Big)(big.NewInt(0).Add(gasTipCap, baseFee)), rpcTx.GasPrice)
+	})
+
+	t.Run("Unminned transaction. GasPrice == GasFeeCap", func(t *testing.T) {
```
karlb, line 54:
```suggestion
	t.Run("Unmined transaction. GasPrice == GasFeeCap", func(t *testing.T) {
```

light/txpool.go
```diff
@@ -392,6 +406,18 @@ func (pool *TxPool) validateTx(ctx context.Context, tx *types.Transaction) error
...
+		return core.ErrFeeCapVeryHigh
+	}
+	if tx.GasTipCap().BitLen() > 256 {
+		return core.ErrTipVeryHigh
+	}
+	// Ensure gasFeeCap is greater than or equal to gasTipCap.
+	if tx.GasFeeCapIntCmp(tx.GasTipCap()) < 0 {
+		return core.ErrTipAboveFeeCap
+	}
+
```
karlb, line 420: This looks fine, but is there any specific reason for adding it this PR?

gastonponti: I was checking the validations that we missed in the tx pool of the light node

e2e_test/e2e_transfer_test.go
gastonponti: If for some reason, the tx is not throwing an error, you'll get a panic, instead of "Expected <something>, actual <nil>"
So yes, this is not strictly important, but for future modifications if we have an issue and those txs don't fail, I prefer that nil, and not a panic

internal/ethapi/transaction_args.go
```diff
@@ -289,6 +289,19 @@ func (args *TransactionArgs) toTransaction() *types.Transaction {
 				Data:                args.data(),
 				AccessList:          al,
 			}
+		} else if isGingerbreadP2 && args.FeeCurrency != nil {
```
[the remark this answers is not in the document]
gastonponti: Strictly speaking, is not the same case because if it is gingerbread p2 and has a feeCurrency AND something as the gatewayFee, it will create a Tx v2 without the gatewayFee, which will make a valid and executable tx, instead of something that is suppose to fail.
From my point of view, I prefer that the user knows that is has an error but knowing that the transaction that is going to execute is exactly the one that it wants.

We can have or
```
if args.GatewayFeeRecipient != nil || args.GatewayFee != nil || (!isGingerbreadP2 && args.FeeCurrency != nil) {
    [...CeloDynamicV1...]
} else if --isGingerbreadP2 &&--  args.FeeCurrency != nil { // We can remove the isGingerbreadP2 Check otherwise it would have use the last one but I thought it was going to be clearer
    [...CeloDynamicV2...]
} else {
    [...Dynamic...]
}
```
or
```
if args.FeeCurrency == nil && args.GatewayFeeRecipient == nil && args.GatewayFee == nil {
   [...Dynamic...]
} else if isGingerbreadP2 && args.FeeCurrency == nil && args.GatewayFeeRecipient == nil {
   [...CeloDynamicV1...]
} else {
   [...CeloDynamicV2...]
}
```
But to be honest everything is ugly, haha

core/types/transaction_marshalling.go
```diff
@@ -122,6 +122,20 @@ func (t *Transaction) MarshalJSON() ([]byte, error) {
...
 		enc.R = (*hexutil.Big)(tx.R)
 		enc.S = (*hexutil.Big)(tx.S)
+	case *CeloDynamicFeeTxV2:
+		enc.ChainID = (*hexutil.Big)(tx.ChainID)
+		enc.AccessList = &tx.AccessList
+		enc.Nonce = (*hexutil.Uint64)(&tx.Nonce)
+		enc.Gas = (*hexutil.Uint64)(&tx.Gas)
+		enc.MaxFeePerGas = (*hexutil.Big)(tx.GasFeeCap)
+		enc.MaxPriorityFeePerGas = (*hexutil.Big)(tx.GasTipCap)
+		enc.FeeCurrency = tx.FeeCurrency // todo: check if needs deep copy
```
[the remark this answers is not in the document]
gastonponti: This didn't bring any issue that we are aware of, and is not something we are modifying or anything like it. I don't know the actual purpose of that comment, but I could copy that just in case.
The `To` field does a deep copy, so probably should be the correct scenario (I just copy pasted the other instances)

karlb (approved):

internal/ethapi/transaction_args.go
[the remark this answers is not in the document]
palango: I agree, let keep it.

internal/ethapi/transaction_args.go
```diff
@@ -266,15 +266,15 @@ func (args *TransactionArgs) ToMessage(globalGasCap uint64, baseFee *big.Int) (t
...
+func (args *TransactionArgs) toTransaction(isGingerbreadP2 bool) *types.Transaction {
 	var data types.TxData
 	switch {
 	case args.MaxFeePerGas != nil:
 		al := types.AccessList{}
 		if args.AccessList != nil {
 			al = *args.AccessList
 		}
-		if args.FeeCurrency != nil || args.GatewayFeeRecipient != nil || args.GatewayFee != nil {
+		if args.GatewayFeeRecipient != nil || args.GatewayFee != nil || (!isGingerbreadP2 && args.FeeCurrency != nil) {
```
palango, line 277: Mh, doesn't this mean that no `CeloDynamicFeeTx` will ever be created when `isGingerbreadP2` is true?
This deprecation and not what we want, right?

palango: Ok, get it now. Makes sense.

PR #2173, opened 2023-08-14
State transition gateway fee check

gastonponti:
### Description

Add a check in the state transition for the gateway fee

### Backwards compatibility

Part of Gingerbread HF


palango:
I remember us discussing this. Back then we decided to move that check to the state processor:
https://github.com/celo-org/celo-blockchain/blob/b61438b1cf13c7ab2d2bc4696655793c74b2ce3d/core/state_processor.go#L143-L150

palango (approved):
This is good to merge, but needs a lot of test fixing.

hbandura (approved):

PR #2174, opened 2023-08-15
Block size limit

gastonponti:
### Description

Block size limit
CIP-XX implementation

### Backwards compatibility

Part of the Gingerbread Part 2 Hard Fork

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -621,6 +621,17 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (*istanbulCore.StateProces
 		}
 	}
 
+	if sb.chain.Config().IsGingerbreadP2(block.Number()) {
+		bytesBlock := new(core.BytesBlock).AddBytes(params.MaxTxDataPerBlock)
+
+		for _, tx := range block.Transactions() {
+			if err := bytesBlock.SubBytes(uint64(tx.Size())); err != nil {
+				sb.logger.Error("verify - Error in validating txs block size", "err", err)
+				return nil, 0, err
+			}
+		}
+	}
+
 	// Apply this block's transactions to update the state
 	receipts, logs, usedGas, err := sb.processBlock(block, state)
 	if err != nil {
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -64,6 +64,15 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 		}
 		return consensus.ErrPrunedAncestor
 	}
+	if v.config.IsGingerbreadP2(block.Number()) {
+		bytesBlock := new(BytesBlock).AddBytes(params.MaxTxDataPerBlock)
+
+		for _, tx := range block.Transactions() {
+			if err := bytesBlock.SubBytes(uint64(tx.Size())); err != nil {
+				return err
+			}
+		}
+	}
 	return nil
 }
 
--- a/core/bytesblock.go
+++ b/core/bytesblock.go
@@ -1,0 +1,38 @@
+package core
+
+import (
+	"fmt"
+	"math"
+)
+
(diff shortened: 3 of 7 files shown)


core/block_validator.go
```diff
@@ -64,6 +64,15 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 		}
 		return consensus.ErrPrunedAncestor
 	}
+	if v.config.IsGingerbreadP2(block.Number()) {
+		bytesBlock := new(BytesBlock).AddBytes(params.MaxTxDataPerBlock)
+
+		for _, tx := range block.Transactions() {
+			if err := bytesBlock.SubBytes(uint64(tx.Size())); err != nil {
+				return err
```
palango, line 72: Above you log an error here, maybe add this here as well?

consensus/istanbul/backend/backend.go
```diff
@@ -621,6 +621,17 @@ func (sb *Backend) Verify(proposal istanbul.Proposal) (*istanbulCore.StateProces
 		}
 	}
 
+	if sb.chain.Config().IsGingerbreadP2(block.Number()) {
+		bytesBlock := new(core.BytesBlock).AddBytes(params.MaxTxDataPerBlock)
```
palango, line 625: nit: The code to check this exists twice, you could factor it out. But it's not necessary for this PR.

core/bytesblock.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"math"
+)
+
+// BytesBlock tracks the amount of bytes available during execution of the transactions
+// in a block. The zero value is a block with zero bytes available.
+type BytesBlock uint64
+
+// AddBytes makes bytes available to use.
+func (bp *BytesBlock) AddBytes(amount uint64) *BytesBlock {
```
palango, line 13: Given how this is used, i.e. only once per instance, maybe call it `SetLimit`!?

core/state_processor_test.go
```diff
@@ -348,8 +355,20 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
...
+
+func fillZero(slice []byte, length int) {
+	for i := 0; i < length && i < len(slice); i++ {
+		slice[i] = 0x00
+	}
+}
 
+func getBigData(length int) []byte {
+	f := make([]byte, length)
+	fillZero(f, length)
```
palango, line 372: Do we need the explicit fill here or is it zeroed by default?

miner/block.go
```diff
@@ -255,6 +259,12 @@ loop:
 			txs.Pop()
 			continue
 		}
+		// Same short-circuit of the gas above, but for bytes in the block (b.bytesBlock != nil => GingerbreadP2)
```
palango, line 262: Above there is a check if there's enough gas for another simple transaction:
```go
if b.gasPool.Gas() < params.TxGas {
	log.Trace("Not enough gas for further transactions", "have", b.gasPool, "want", params.TxGas)
	break
}
```

I think we should add something similar for the remaining bytes, so we short-circuit in case there's to little block space left to add another transaction.

palango (approved):
Looks good in general, left some comments for your consideration.

miner/block.go
gastonponti: The question here is what is the smaller rlp encode possible, and is that going to change in the future? for example removing the gatewayFee and currencyFee from the legacy tx. I preferred to avoid that now, and have it as a possible optimization later

core/block_validator.go
gastonponti: The block validator response is logged later.
In the proposal, we are not using this same instance, that's why we log it in every case

miner/block.go
palango: > The question here is what is the smaller rlp encode possible

It is the native transaction (with no calldata), right? That's the lower bound and should be quite simple to calculate.

PR #2175, opened 2023-08-15
Remove per-peer rlp.Encode calls of istanbul messages

hbandura:
Istanbul consensus is currently rlp encoding messages twice, and decoding them twice upon receiving one. The second encoding is done per peer, so if the same message is multicasted to 110 peers, it is encoded once, then 110 times more. This commit ensures the double encoding is done just once, reducing from 111 to 2 the calls to rlp.EncodeToBytes. A better way to handle this (but more disruptive) is completely drop the double encoding and double decoding, but would imply a protocol change.

* Also removed an expensive log line.

### DIFF
--- a/consensus/istanbul/backend/message_senders.go
+++ b/consensus/istanbul/backend/message_senders.go
@@ -110,7 +110,7 @@ func (sb *Backend) Gossip(payload []byte, ethMsgCode uint64) error {
 // sendMsg will asynchronously send the the Celo messages to all the peers in the destPeers param.
 func (sb *Backend) asyncMulticast(destPeers map[enode.ID]consensus.Peer, payload []byte, ethMsgCode uint64) error {
 	logger := sb.logger.New("func", "AsyncMulticastCeloMsg", "msgCode", ethMsgCode)
-	payload2, err := rlp.EncodeToBytes(payload)
+	reencodedPayload, err := rlp.EncodeToBytes(payload)
 	if err != nil {
 		return err
 	}
@@ -118,7 +118,7 @@ func (sb *Backend) asyncMulticast(destPeers map[enode.ID]consensus.Peer, payload
 		peer := peer // Create new instance of peer for the goroutine
 		go func() {
 			logger.Trace("Sending istanbul message(s) to peer", "peer", peer, "node", peer.Node())
-			if err := peer.SendDoubleEncoded(ethMsgCode, payload2); err != nil {
+			if err := peer.SendDoubleEncoded(ethMsgCode, reencodedPayload); err != nil {
 				logger.Warn("Error in sending message", "peer", peer, "ethMsgCode", ethMsgCode, "err", err)
 			}
 		}()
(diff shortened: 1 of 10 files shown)


consensus/istanbul/backend/message_senders.go
```diff
@@ -103,24 +104,26 @@ func (sb *Backend) Gossip(payload []byte, ethMsgCode uint64) error {
...
-	return nil
+	return sb.asyncMulticast(peersToSendMsg, payload, ethMsgCode)
 }
 
 // sendMsg will asynchronously send the the Celo messages to all the peers in the destPeers param.
-func (sb *Backend) asyncMulticast(destPeers map[enode.ID]consensus.Peer, payload []byte, ethMsgCode uint64) {
+func (sb *Backend) asyncMulticast(destPeers map[enode.ID]consensus.Peer, payload []byte, ethMsgCode uint64) error {
 	logger := sb.logger.New("func", "AsyncMulticastCeloMsg", "msgCode", ethMsgCode)
-
+	payload2, err := rlp.EncodeToBytes(payload)
```
palango, line 113: nit: `payloadEncoded` or `rlpPayload` seems more meaningful.

hbandura: I'll change it

eth/protocols/eth/celo_peer.go
```diff
@@ -0,0 +1,24 @@
...
+func (p *Peer) Send(msgcode uint64, data []byte) error {
+	raw, err := rlp.EncodeToBytes(data)
+	if err != nil {
+		return err
+	}
+	return p.SendDoubleEncoded(msgcode, raw)
+}
+
+// SendDoubleEncoded sends an rlp encoded (twice) message with the given code.
+func (p *Peer) SendDoubleEncoded(msgcode uint64, data []byte) error {
```
carterqw2, line 21: I understand the desire to emphasize the fact that the encoding is performed twice but the implementation of the function doesn't have anything with encoding at all it which is misleading. Logically, `SendDoubleEncoded` should become `Send` and `Send` should become `SendEncoded`. And in the [calling code](https://github.com/celo-org/celo-blockchain/pull/2175/files#diff-8e3ffc7b008ae6e501c7aa2cb2eed2db1fbef6586d767f59b1209fef26ad5ac1R121) I would leave a comment explaining the incidental (hopefully?) double encoding.
Otherwise, I like the change.

eth/protocols/eth/celo_peer.go
```diff
@@ -0,0 +1,33 @@
...
+		return err
+	}
+	return p.Send(msgcode, raw)
+}
+
+// Send sends the message to this peer. Since current istanbul protocol
+// has messages encoded twice, data should be twice encoded
+// with rlp encoding.
+func (p *Peer) Send(msgcode uint64, data []byte) error {
+	// Istanbul was encoding messages before sending it to the peer,
```
carterqw2, line 24: Let's maybe move this comment to the `asyncMulticast` function? This way it will be in the context where the double encoding that you mention in the comment actually happens and would make more sense.
Otherwise, looks good to me.

carterqw2 (approved):

PR #2177, opened 2023-08-16
Remove ed25519 precompile

gastonponti:
### Description

Remove the CIP-25 as part of the CIP: https://github.com/celo-org/celo-proposals/discussions/392

### Backwards compatibility

Part of the Gingerbread P2 Hard Fork


karlb (approved):
This is the minimal change to keep the diff as small as possible so directly before the release. The implementation can be removed in a PR after the hardfork.

PR #2178, opened 2023-08-25
Update public network bootnodes

jcortejoso:
### Description

Update the list of bootnode enodes for mainnet and baklava to include some fullnodes, removing the SPOF that was the bootnode.
Also included alfajores bootnode enode to the list of bootnodes in code so we follow the same patterns for the three public networks.

### Other changes

No

### Tested

Tested manually that new enode respond. This can be tested as follows (after compiling the repository):

```
build/bin/devp2p discv4 --networkid 42220 crawl -bootnodes "enode://6d1702b599db61bbdf23a6201b3d1595b526e536194c5140904e7b34d79162b9905a75a257e9a64bd2ee1ed3ef40e404b14e4f1eb985cc180d9912497fd81b0d@34.105.94.71:30303" -timeout 2m /tmp/all-nodes.json
```

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


palango:
Thanks for updating this @jcortejoso .
- Are those IPs stable, or do we need to update those lists for every release?
- There is a linting problem that needs to be fixed.

jcortejoso:
> Thanks for updating this @jcortejoso .
> 
> * Are those IPs stable, or do we need to update those lists for every release?
> * There is a linting problem that needs to be fixed.

Yes, ip addresses are static and enodes deterministic.

palango (approved):

PR #2184, opened 2023-09-07
RSDB: metrics and performance improvements

carterqw2:
### Description

Reuses the existing LevelDB implementation from the ethdb/leveldb module to collect LevelDB metrics for Roundstate DB as well and introduces additional metrics related to RLP-encoding.

### Other changes

- Changes default configuration for the Roundstate DB to reduce the amount of non-level0 compactions and improve performance, this change significantly reduces CPU usage and positively affects block construction times.
- Changes configuration settings of the state db to make compactions happen less often.

### Tested

Tested on the testnet.

### Backwards compatibility

No breaking changes.

### DIFF
--- a/consensus/istanbul/core/roundstate_db.go
+++ b/consensus/istanbul/core/roundstate_db.go
@@ -27,12 +27,13 @@ import (
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/common/task"
 	"github.com/celo-org/celo-blockchain/consensus/istanbul"
+	"github.com/celo-org/celo-blockchain/ethdb/leveldb"
 	"github.com/celo-org/celo-blockchain/log"
+	"github.com/celo-org/celo-blockchain/metrics"
 	"github.com/celo-org/celo-blockchain/rlp"
-	"github.com/syndtr/goleveldb/leveldb"
-	lvlerrors "github.com/syndtr/goleveldb/leveldb/errors"
+	goleveldb "github.com/syndtr/goleveldb/leveldb"
+
 	"github.com/syndtr/goleveldb/leveldb/opt"
-	"github.com/syndtr/goleveldb/leveldb/storage"
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
@@ -63,10 +64,19 @@ type RoundStateDBOptions struct {
 }
 
 type roundStateDBImpl struct {
-	db                   *leveldb.DB
+	db *leveldb.Database
+
 	stopGarbageCollector task.StopFn
 	opts                 RoundStateDBOptions
-	logger               log.Logger
+
+	rsRLPMeter      metrics.Meter // Meter for measuring the size of rs RLP-encoded data
+	rsRLPEncTimer   metrics.Timer // Timer measuring time required for rs RLP encoding
+	rsDbSaveTimer   metrics.Timer // Timer measuring rs DB write latency
+	rcvdRLPMeter    metrics.Meter // Meter for measuring the size of rcvd RLP-encoded data
+	rcvdRLPEncTimer metrics.Timer // Timer measuring time required for rcvd RLP encoding
+	rcvdDbSaveTimer metrics.Timer // Timer measuring rcvd DB write latency
+
+	logger log.Logger // Contextual logger tracking the database path
 }
 
 var defaultRoundStateDBOptions = RoundStateDBOptions{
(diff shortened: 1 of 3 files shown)


consensus/istanbul/core/roundstate_db.go
```diff
@@ -113,6 +124,13 @@ func newRoundStateDB(path string, opts *RoundStateDBOptions) (RoundStateDB, erro
 		logger: logger,
 	}
 
+	rsdb.rsRLPMeter = metrics.NewRegisteredMeter(namespace+"rs/rlp/encoding/size", nil)
+	rsdb.rsRLPEncTimer = metrics.NewRegisteredTimer(namespace+"rs/rlp/encoding/duration", nil)
+	rsdb.rsDbSaveTimer = metrics.NewRegisteredTimer(namespace+"rs/db/save/time", nil)
+	rsdb.rcvdRLPMeter = metrics.NewRegisteredMeter(namespace+"rcvd/rlp/encoding/size", nil)
```
hbandura, line 130: I remember we used rcvd in the past for tests. I wonder if there's name for this that makes it a bit clearer for a reader. As in , the entry for the collection of received consensus messages.

consensus/istanbul/core/roundstate_db.go
```diff
@@ -121,35 +139,34 @@ func newRoundStateDB(path string, opts *RoundStateDBOptions) (RoundStateDB, erro
...
 // also flushing its contents in case of a version mismatch.
-func newPersistentDB(path string) (*leveldb.DB, error) {
-	opts := &opt.Options{OpenFilesCacheCapacity: 5}
-	db, err := leveldb.OpenFile(path, opts)
-	if _, iscorrupted := err.(*lvlerrors.ErrCorrupted); iscorrupted {
-		db, err = leveldb.RecoverFile(path, nil)
-	}
+func newPersistentDB(path string, namespace string) (*leveldb.Database, error) {
+	db, err := leveldb.NewCustom(path, namespace, func(options *opt.Options) {
+		options.BlockSize = 128 * opt.KiB
```
hbandura, line 150: If it makes sense, maybe add a small justification for the specific values?

consensus/istanbul/core/roundstate_db.go
```diff
@@ -283,26 +309,33 @@ func (rsdb *roundStateDBImpl) UpdateLastRoundState(rs RoundState) error {
 	logger := rsdb.logger.New("func", "UpdateLastRoundState")
 	viewKey := view2Key(rs.View())
 
+	before := time.Now()
 	entryBytes, err := rlp.EncodeToBytes(rs)
+	rsdb.rsRLPEncTimer.UpdateSince(before)
```
hbandura, line 314: Down in the code the metric is updated after the error check, here it's the opposite. If there's any particular reason for that, let's document it here, if not, standardize on one

ethdb/leveldb/leveldb.go
```diff
@@ -149,12 +150,24 @@ func NewCustom(file string, namespace string, customize func(options *opt.Option
 	return ldb, nil
 }
 
+// NewCustom returns a wrapped LevelDB object with an in-memory storage.
```
hbandura, line 153: Do we want to move this additions to a new file, to prevent conflicts with upstream? We were doing that before, I'm not sure if we want to continue doing that or not.

hbandura: s/NewCustom/NewInMemory

ethdb/leveldb/leveldb.go
```diff
@@ -149,12 +150,24 @@ func NewCustom(file string, namespace string, customize func(options *opt.Option
...
+	}, nil
+}
+
 // configureOptions sets some default options, then runs the provided setter.
 func configureOptions(customizeFn func(*opt.Options)) *opt.Options {
 	// Set default options
 	options := &opt.Options{
 		Filter:                 filter.NewBloomFilter(10),
 		DisableSeeksCompaction: true,
+		CompactionTableSize:    10 * opt.MiB,
```
hbandura, line 170: This value is being set here, but also being set in the persistentDb of the roundstate db. Is this one the default? Is it necessary to set it in both?

hbandura (changes requested):

ethdb/leveldb/leveldb.go
carterqw2: Good eye, thanks! Yes, moved to a new file, thanks for pointing that out.

ethdb/leveldb/leveldb.go
carterqw2: Good point, updated.

consensus/istanbul/core/roundstate_db.go
carterqw2: Good point, updated.

consensus/istanbul/core/roundstate_db.go
carterqw2: Added a comment.

consensus/istanbul/core/roundstate_db.go
carterqw2: The type struct is called `rcvd`, so I thought I would just follow the convention. I also updated the comment next to the definition of the metric. Would it be enough or you want to change it here? If so, what exactly you would like to change, the name of the metric or the path or both?

hbandura: Yes, iirc I named that struct and I regret it highly xD. Maybe it's ok, I just felt that if you read the metric name, without the context of it being in this specific piece of code, it's really tough to understand what it is. That being said, it's probably the same with other metrics, so maybe it's ok.

hbandura (approved):
LGTM!

PR #2186, opened 2023-09-18
Fix tracing for transactions using on-chain randomness in latest block

eelanagaraj:
## Description

Fixes transaction tracing bug where the tracer did not include the block randomness for block whose transaction was being traced. For `(tx_i, block_n)`, it would fetch the state for `block_n-1` and then reexecute all txs up until `tx_i`, but would omit the random commitment at the very start of `block_n`.

This affects the following endpoints:
- `TraceTransaction`
- `TraceBlock*`
- `StandardTraceBlock*`
- `IntermediateRoots`
- `TraceChain`

Note:
- `TraceCall` performs correctly since this applies the tx _at the end_ of the requested block, so the block randomness of the next block is irrelevant.

### Implementation note
- Initially attempted to limit scope of the changes to the API (`eth/tracers/api.go`) but this results in even more messiness because the `eth/state_accessor/stateAtTransaction` function needs to have the solution duplicated
- Fixing this in the eth backend means that technically, the LES tracing API logic is wrong, but tracing with the light client fails or is incorrect as is anyways (confirmed for endpoints above)
  - Upside to this is that we can nicely separate out the celo-specific changes and make it explicit in the tracing API when randomness is included or not

*One alternative* here is just directly updating the `stateAtTransaction` function and then updating the API to use `stateAtTransaction` with the block & 0th index transaction to set up the state properly pre-tracing. I thought this was potentially a bit unclear + more likely to be accidentally overwritten in future merges but happy to reconsider!

Factors out `ApplyBlockRandomnessTx` from `core/state_processor/Process` so that it can be used by `state_accessor` as well.

## Tested
Unfortunately, testing this via an e2e/unit test is extremely involved, since tracing a simple interaction with `Random.sol` will not suffice. It appears necessary to trace a transaction where a contract interacts with `Random.sol` (tracing a non-view call), and deploying a contract manually in the test feels unnecessary.

- [x] Manual/scripted testing using foundry + mycelo -- https://github.com/celo-org/tracer-testing
 - future tracer testing can use the same repo
- [x] tested all endpoints listed above
- [  ] debugging an issue with `TraceChain` & not using the `live` database for multiple blocks 🤔  

## Backwards compatibility

This changes the behavior of the tracing API endpoints but is not expected to be a consensus-relevant breaking change.

### DIFF
--- a/eth/celo_state_accessor.go
+++ b/eth/celo_state_accessor.go
@@ -1,0 +1,35 @@
+package eth
+
+import (
+	"fmt"
+
+	"github.com/celo-org/celo-blockchain/core"
+	"github.com/celo-org/celo-blockchain/core/state"
+	"github.com/celo-org/celo-blockchain/core/types"
+)
+
+// Wraps `stateAtBlock` with the additional Celo-specific parameter afterNextRandomCommit.
+// This parameter executes the random commitment at the start of the next block,
+// since this is necessary for properly tracing transactions in the next block.
+func (eth *Ethereum) celoStateAtBlock(block *types.Block, reexec uint64, base *state.StateDB, checkLive bool, preferDisk bool, afterNextRandomCommit bool) (statedb *state.StateDB, err error) {
+	statedb, err = eth.stateAtBlock(block, reexec, base, checkLive, preferDisk)
+	if err != nil {
+		return nil, err
+	}
+
+	if !afterNextRandomCommit {
+		return statedb, nil
+	}
+	// Fetch next block's random commitment
+	nextBlockNum := block.NumberU64() + 1
+	nextBlock := eth.blockchain.GetBlockByNumber(nextBlockNum)
+	if nextBlock == nil {
+		return nil, fmt.Errorf("next block %d not found", nextBlockNum)
+	}
+	vmRunner := eth.blockchain.NewEVMRunner(nextBlock.Header(), statedb)
+	err = core.ApplyBlockRandomnessTx(nextBlock, &vmRunner, statedb, eth.blockchain)
+	if err != nil {
+		return nil, err
+	}
+	return statedb, nil
+}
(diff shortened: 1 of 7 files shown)


eth/celo_state_accessor.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/celo-org/celo-blockchain/core"
+	"github.com/celo-org/celo-blockchain/core/state"
+	"github.com/celo-org/celo-blockchain/core/types"
+)
+
+// Wraps `stateAtBlock` with the additional Celo-specific parameter afterNextRandomCommit.
+// This parameter executes the random commitment at the start of the next block,
+// since this is necessary for properly tracing transactions in the next block.
+func (eth *Ethereum) celoStateAtBlock(block *types.Block, reexec uint64, base *state.StateDB, checkLive bool, preferDisk bool, afterNextRandomCommit bool) (statedb *state.StateDB, err error) {
```
carterqw2, line 14: `afterNextRandomCommit` sounds a little confusing, maybe we could rename it to match the other bool arguments in the signature and sound more directive like `commitRandomness` or something similar?

palango, line 14: Are we ever interested in the state without the randomness applied? At least in this PR it is always enabled, so we could remove that parameter.

eelanagaraj: yes in `traceCall`, where the trace is applied to the end of the block (as opposed to the start of the following block), this one is used with "latest" often too, so would result in different behavior for the "latest" block vs. earlier blocks :-/ In this case I preferred keeping the behavior aligned with upstream but we can revisit!

palango: Ah, I looked at a partial diff, sorry.

eelanagaraj: no worries!

PR #2190, opened 2023-09-26
Refactor container CI. ARM Image

jcortejoso:
### Description

- Added a new `Dockerfile.arm64` supporting ARM builds. This image is different from the existing Dockerfile because:
  - celo-blockchain does not compile on alpine ARM. I tried installing some dependencies but I could not find the problem
  - For debian as base, there is one package (`linux-headers` and `linux-headers-arm64`) with different name for both architectures. It may be possible to add some logic to the Dockerfile to be able to share the same for linux/amd64 and linux/arm64 platforms, but really I don't see a lot of value on investing time there tbh.
- Refactored the building process to use new composite actions at celo-org/reusable-workflows. Depending on PR https://github.com/celo-org/reusable-workflows/pull/72

### Other changes

- Removed old `Dockerfile.arm64` expected for Android builds
- At `mycelo`, skip the transfer ownsership of governance for local and loadTest mycelo templates

### Tested

Images generated: https://github.com/celo-org/celo-blockchain/actions/runs/6315913589/workflow

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


carterqw2 (approved):
Nice!

PR #2191, opened 2023-10-02
Fix gingerbreadP2 check for light tx pool

eelanagaraj:
### Description

Currently the light tx pool is missing the `pool.gingerbreadP2` property definition, causing light clients to incorrectly throw an error on receiving `0x7b` (`CeloDynamicFeeTxV2Type`) transactions.

### Tested

Manually checked that this allows `CeloDynamicFeeTxV2Type` txs to be submitted afterwards.

### Backwards compatibility

Not a consensus-level change so shouldn't require a hardfork.

### DIFF
--- a/light/txpool.go
+++ b/light/txpool.go
@@ -331,6 +331,7 @@ func (pool *TxPool) setNewHead(head *types.Header) {
 	pool.donut = pool.config.IsDonut(next)
 	pool.espresso = pool.config.IsEspresso(next)
 	pool.gingerbread = pool.config.IsGingerbread(next)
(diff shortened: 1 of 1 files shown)


carterqw2 (approved):

karlb (approved):
It's a bit annoying that there is no more generic hardfork handling, but at least it is simple.

eelanagaraj:
> It's a bit annoying that there is no more generic hardfork handling, but at least it is simple.

@karlb yeah and that the light & normal txpools are different in this way, seems odd to me 🤷‍♀️

PR #2196, opened 2023-10-17
Update minimal required version on readme README.md

martinvol:
### Description

Tried to build with go1.16 but it didn't work


palango (approved):
Thanks!

PR #2198, opened 2023-10-19
Improve performance of answerGetBlockBodiesQuery

carterqw2:
### Description

Avoid decoding and re-encoding of the block body by using available RLP-encoded block body from the database, combine it with the RLP-encoded block hash to avoid unnecessary processing and extra memory allocations.

Screenshot 2023-06-26 at 17 36 19

Benchmarks:
```
Original Ethereum version (highly efficient as block body is taken either from the cache or database directly).

BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-1-8         	14819121	        92.65 ns/op	      24 B/op	       1 allocs/op
BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-10-8        	 1318404	       896.9 ns/op	     744 B/op	       5 allocs/op
BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-100-8       	  153598	      7907 ns/op	    6120 B/op	       8 allocs/op


Current Celo-version

BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-1-8         	 1000000	      1278 ns/op	     448 B/op	      12 allocs/op
BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-10-8        	   81495	     12543 ns/op	    4984 B/op	     115 allocs/op
BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-100-8       	   10000	    115520 ns/op	   48523 B/op	    1108 allocs/op


Suggested version

BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-1-8         	 3173032	       370.5 ns/op	     256 B/op	       6 allocs/op
BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-10-8        	  297216	      3805 ns/op	    3064 B/op	      55 allocs/op
BenchmarkAnswerGetBlockBodiesQuery/GetBlockBodies-100-8       	   30360	     37067 ns/op	   29322 B/op	     508 allocs/op
```

3x more CPU-efficient and 1.5x more memory-efficient.
A larger change would be storing RLP-encoded blockBodyWithBlockHash structures instead of just block bodies but it would significantly increase the diff.

### Other changes

No.

### Tested

Existing tests pass, added unit tests and the benchmark.

### Backwards compatibility

Yes.

### DIFF
--- a/rlp/celo_raw.go
+++ b/rlp/celo_raw.go
@@ -1,0 +1,58 @@
+package rlp
+
+import (
+	"encoding/binary"
+	"math"
+)
+
+// for payloads longer than 55 bytes for both strings and lists additional prefix signalizing
+// the total length of the payload is needed
+const payloadLengthThreshold = 0x37 // 55 in dec
+
+// indicates that the payload is a list
+const shortListEncodingByte = 0xC0
+
+// listEncodingByte (0xC0) + payloadLengthThreshold (0x37) = 0xF7
+const longListEncodingByte = 0xF7
+
+// Combine takes two RLP-encoded values one and two and produces a combined one
+// as if it was an RLP encoding of the [one, two] list
+// based on https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/
+func Combine(one []byte, two []byte) []byte {
+	payloadLen := len(one) + len(two)
+	result := make([]byte, 0, ListSize(uint64(payloadLen)))
+
+	if payloadLen <= payloadLengthThreshold {
+		// If the total payload of a list (i.e. the combined length of all its items being RLP
+		// encoded) is 0-55 bytes long, the RLP encoding consists of a single byte with value
+		// 0xc0 plus the length of the payload followed by the concatenation of the RLP encodings
+		// of the items.
+		result = append(result, byte(shortListEncodingByte+payloadLen))
+	} else {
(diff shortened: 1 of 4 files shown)


rlp/celo_raw.go
```diff
@@ -0,0 +1,58 @@
...
+// indicates that the payload is a list
+const shortListEncodingByte = 0xC0
+
+// listEncodingByte (0xC0) + payloadLengthThreshold (0x37) = 0xF7
+const longListEncodingByte = 0xF7
+
+// Combine takes two RLP-encoded values one and two and produces a combined one
+// as if it was an RLP encoding of the [one, two] list
+// based on https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/
+func Combine(one []byte, two []byte) []byte {
```
palango, line 21: Weird that there's no existing code for this. Could you add a simple test?

palango (commented):
Looks good, maybe it's worthwhile to add the benchmarking code as well?

rlp/celo_raw.go
carterqw2: Yeah, I was also surprised there is nothing like that. I'll add some.

carterqw2: Added.

palango (approved):
Looks good, thanks for adding the test and benchmark!

PR #2199, opened 2023-10-24
Fee currencies without `credit`/`debit`

palango:
### Description

Prototype for gas currency payments without `credit`/`debit` functions in the token contract.

Inside the EVM state transition code it is possible to send transactions from arbitrary accounts. @karlb came up with the idea to use this to use the existing `approve`/`transferFrom` functions to remove the need of `credit`/`debit` functions in fee currency token contracts.

During the implementation it became clear that it is possible to just call `transfer`, which reduces the overhead further.

There's two remaining differences from the current implementation:
- This will emit transfer events. It should be possible to discard them though.
- It doesn't temporarily reduce the total supply like we do now. It still keeps the invariant "sum(balances) == total supply" by moving the tokens to `0xce106a5` instead of letting the debited fees vanish.

⚠️ The code is a prototype, the gas limit handling is just wrong and error handling needs to be improved.

### DIFF
--- a/contracts/erc20gas/erc20gas.go
+++ b/contracts/erc20gas/erc20gas.go
@@ -13,6 +13,12 @@ import (
 const (
 	maxGasForDebitGasFeesTransactions  uint64 = 1 * n.Million
 	maxGasForCreditGasFeesTransactions uint64 = 1 * n.Million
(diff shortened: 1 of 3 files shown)


palango:
Closing, as this will be Cel2 only.

PR #2203, opened 2023-10-30
Implemented the limit per fee currency

carterqw2:
### Description

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


core/celo_multi_gaspool.go
```diff
@@ -0,0 +1,57 @@
...
+	celoPool := new(GasPool).AddGas(block_gas_limit)
+
+	return MultiGasPool{
+		pools:       pools,
+		defaultPool: celoPool,
+	}
+}
+
+func (mgp MultiGasPool) PoolFor(feeCurrency *FeeCurrency) *GasPool {
+	if feeCurrency == nil || mgp.pools[*feeCurrency] == nil {
```
palango, line 52: I don't think we need the `defaultPool` here, as this is checked in `blockState.gasPool` and there should never be an unregistered fee currency here.

But I guess this is mostly to make the code return a `*GasPool`, which is fine.

palango (commented):
The approach looks right to me, left one additional comment inline.

core/celo_multi_gaspool.go
carterqw2: Yes, that's mostly to avoid errors or additional checks in case of `nil` as `feeCurrency`, as I didn't want that code to be dependent on where it's used and rely on checks made elsewhere.

palango: Makes sense!

miner/celo_defaults.go
```diff
@@ -0,0 +1,20 @@
...
+
+// cStables addresses on mainnet
+var cUSD_TOKEN = common.HexToAddress("0x765DE816845861e75A25fCA122bb6898B8B1282a")
+var cEUR_TOKEN = common.HexToAddress("0xD8763CBa276a3738E6DE85b4b3bF5FDed6D6cA73")
+var cREAL_TOKEN = common.HexToAddress("0xe8537a3d056DA446677B9E9d6c5dB704EaAb4787")
+
+// default limits configuration
+var DefaultFeeCurrencyLimits = map[uint64]map[common.Address]float64{
+	params.MainnetNetworkId: {
+		cUSD_TOKEN:  0.9,
```
gastonponti, line 16: I don't really get this.
I understand that we have a lot less cReal txs for example in our history, but this is going to limit peaks. I completely agree that a validator could choose this, but I'm not totally getting this as a default.
From my point of view, if this is an optional tool to limit traffic, this should maintain the same functionality as default, not changing the actual behaviour

gastonponti (commented):
I've added a few comments, one is something to change and the others are more opinionated about the defaults.
On the other hand, I've also checked what could happen in the case of having a validator accepting a proposal that whitelists a new currency, and in the same block sending a tx with that field, but we are not allowing that because we are checking against the whitelist that we have in the block context, which won't change until the next one, so we won't have that scenario.

cmd/utils/flags.go
```diff
@@ -421,6 +421,16 @@ var (
 		Name:  "miner.extradata",
 		Usage: "Block extra data set by the miner (default = client version)",
 	}
+	CeloFeeCurrencyDefault = cli.Float64Flag{
+		Name:  "celo.feecurrency.default",
+		Usage: "Default fraction of block gas limit available for TXs paid with a whitelisted currency",
+		Value: 0.9,
```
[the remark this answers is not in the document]
carterqw2: This is the default value for any new gas token, the idea was to have at least some block space allocated to CELO. If we set it to one and validators don't bother to change the value, we might end up in a situation we tried to avoid.

cmd/utils/flags.go
```diff
@@ -421,6 +421,16 @@ var (
 		Name:  "miner.extradata",
 		Usage: "Block extra data set by the miner (default = client version)",
 	}
+	CeloFeeCurrencyDefault = cli.Float64Flag{
+		Name:  "celo.feecurrency.default",
+		Usage: "Default fraction of block gas limit available for TXs paid with a whitelisted currency",
```
gastonponti, line 426: Just a nit, but I would change it to "whitelisted alternative currency"
I think that feeCurrency and whitelisted means the same for us, but from someone new reading the help, maybe is not clear if Celo is or not part of that batch

gastonponti (approved):
LGTM, optional nit

PR #2206, opened 2023-11-24
Fix underpriced check for multi currency

hbandura:
Previous underprice implementation would provoke different currency comparisons, erroneously calculate the underprice flag, and ignore basefee in underprice comparisons.

### DIFF
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -696,27 +696,29 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 }
 
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
-	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
-	if !l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper) {
-		// Check Len after calling underpricedFor since it might prune entries
-		if h.nativeCurrencyHeap.Len() > 0 {
+	// wellpriced returns if, after pruning, tx is above minimum price
+	// compared to the top of the heap (minimum priced tx in heap)
+	wellpriced := func(heap *priceHeap) bool {
+		l.prune(heap)
+		return heap.Len() > 0 && !h.IsCheaper(tx, heap.list[0])
+	}
+
+	// If tx is wellpriced for at least one heap, then it is not underpriced
+	if wellpriced(h.nativeCurrencyHeap) {
+		return false
+	}
+	for _, sh := range h.currencyHeaps {
+		if wellpriced(sh) {
 			return false
 		}
 	}
-	for _, sh := range h.currencyHeaps {
-		if !l.underpricedFor(sh, tx, h.IsCheaper) {
-			// Check Len after calling underpricedFor since it might prune entries
-			if sh.Len() > 0 {
-				return false
-			}
-		}
-	}
+	// While this impl. is returning true when all heaps are empty (h.Len() == 0), that border
+	// case is managed on invokation of this fn.
 	return true
 }
 
-// underpricedFor checks whether a transaction is cheaper than (or as cheap as) the
-// lowest priced (remote) transaction in the given heap.
-func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction, isCheaper func(tx1, tx2 *types.Transaction) bool) bool {
+// prune discards from the top of the heap txs that are no longer in the pool
+func (l *txPricedList) prune(h *priceHeap) {
 	// Discard stale price points if found at the heap start
 	for len(h.list) > 0 {
 		head := h.list[0]
@@ -727,13 +729,6 @@ func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction, isChe
 		}
 		break
 	}
-	// Check if the transaction is underpriced or not
-	if len(h.list) == 0 {
-		return false // There is no remote transaction at all.
-	}
-	// If the remote transaction is even cheaper than the
-	// cheapest one tracked locally, reject it.
-	return isCheaper(tx, h.list[0])
 }
 
 // Discard finds a number of most underpriced transactions, removes them from the
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -696,10 +696,11 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 }
 
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
-	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx)
+	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
+	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper)
 	for _, sh := range h.currencyHeaps {
-		if l.underpricedFor(sh, tx) {
-			underpriced = true
+		if !l.underpricedFor(sh, tx, h.IsCheaper) {
+			underpriced = false
 		}
 	}
 	return underpriced
@@ -707,7 +708,7 @@ func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.
 
 // underpricedFor checks whether a transaction is cheaper than (or as cheap as) the
 // lowest priced (remote) transaction in the given heap.
-func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction) bool {
+func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction, isCheaper func(tx1, tx2 *types.Transaction) bool) bool {
 	// Discard stale price points if found at the heap start
 	for len(h.list) > 0 {
 		head := h.list[0]
@@ -724,7 +725,7 @@ func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction) bool 
 	}
 	// If the remote transaction is even cheaper than the
 	// cheapest one tracked locally, reject it.
-	return h.cmp(h.list[0], tx) >= 0
+	return isCheaper(tx, h.list[0])
 }
 
 // Discard finds a number of most underpriced transactions, removes them from the
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -698,11 +698,17 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
 	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
 	if !l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper) {
-		return false
+		// Check Len after calling underpricedFor since it might prune entries
+		if h.nativeCurrencyHeap.Len() > 0 {
+			return false
+		}
 	}
 	for _, sh := range h.currencyHeaps {
 		if !l.underpricedFor(sh, tx, h.IsCheaper) {
-			return false
+			// Check Len after calling underpricedFor since it might prune entries
+			if sh.Len() > 0 {
+				return false
+			}
 		}
 	}
 	return true
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -697,13 +697,15 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
 	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
-	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper)
+	if !l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper) {
+		return false
+	}
 	for _, sh := range h.currencyHeaps {
 		if !l.underpricedFor(sh, tx, h.IsCheaper) {
-			underpriced = false
+			return false
 		}
 	}
-	return underpriced
+	return true
 }
 
 // underpricedFor checks whether a transaction is cheaper than (or as cheap as) the
--- a/core/tx_multicurrency_priceheap.go
+++ b/core/tx_multicurrency_priceheap.go
@@ -19,6 +19,9 @@ func (cc CurrencyCmpFn) GasTipCapCmp(tx, other *types.Transaction) int {
 	return cc(tx.GasTipCap(), tx.FeeCurrency(), other.GasTipCap(), other.FeeCurrency())
 }
 
+// EffectiveGasTipCmp returns the same comparison result as Transaction.EffectiveGasTipCmp
+// but taking into account the exchange rate comparison of the CurrencyCmpFn.
+// Each baseFee is expressed in each tx's currency.
 func (cc CurrencyCmpFn) EffectiveGasTipCmp(tx, other *types.Transaction, baseFeeA, baseFeeB *big.Int) int {
 	if baseFeeA == nil && baseFeeB == nil {
 		return cc.GasTipCapCmp(tx, other)
@@ -30,6 +33,9 @@ func (cc CurrencyCmpFn) GasFeeCapCmp(a, b *types.Transaction) int {
 	return cc(a.GasFeeCap(), a.FeeCurrency(), b.GasFeeCap(), b.FeeCurrency())
 }
 
+// Cmp returns the same comparison as the priceHeap comparison but taking into account
+// the exchange rate comparison of the CurrencyCmpFn.
+// Each baseFee is expressed in each tx's currency.
 func (cc CurrencyCmpFn) Cmp(a, b *types.Transaction, baseFeeA, baseFeeB *big.Int) int {
 	if baseFeeA != nil || baseFeeB != nil {
 		// Compare effective tips if baseFee is specified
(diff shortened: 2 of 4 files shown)


carterqw2:
Could you add a test that would reproduce the issue?

core/tx_list.go
```diff
@@ -696,18 +696,19 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 }
 
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
```
carterqw2, line 698: Why do we check against all price heaps and not only the heap for the fee currency of the TX? I think I'm missing the logic here.

hbandura: the underprice check tries to find out if this tx is too cheap for this pool (in its entirety), so as to drop it if no slots are available. It might not be the cheapest of its currency, but there might be another tx from a different currency that's cheaper, so it makes sense to eventually keep this one and remove a another one

palango: > It might not be the cheapest of its currency, but there might be another tx from a different currency that's cheaper, so it makes sense to eventually keep this one and remove a another one

Why don't we remove all underpriced transactions?

core/tx_list.go
```diff
@@ -696,18 +696,19 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
...
-	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx)
+	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
+	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper)
 	for _, sh := range h.currencyHeaps {
-		if l.underpricedFor(sh, tx) {
-			underpriced = true
+		if !l.underpricedFor(sh, tx, h.IsCheaper) {
+			underpriced = false
```
palango, line 703: With the above we could short-circuit here, right?

core/tx_multicurrency_priceheap.go
```diff
@@ -34,8 +64,10 @@ func newMultiCurrencyPriceHeap(currencyCmp CurrencyCmpFn, gpm GasPriceMinimums)
 
 		// inner state
 
-		nativeCurrencyHeap: &priceHeap{},
-		currencyHeaps:      make(map[common.Address]*priceHeap),
+		nativeCurrencyHeap: &priceHeap{
+			baseFee: gpm.GetNativeGPM(),
```
palango, line 68: Is this updates on every block?

core/tx_multicurrency_priceheap.go
```diff
@@ -10,9 +10,39 @@ import (
...
+		return cc.GasTipCapCmp(tx, other)
+	}
+	return cc(tx.EffectiveGasTipValue(baseFeeA), tx.FeeCurrency(), other.EffectiveGasTipValue(baseFeeB), other.FeeCurrency())
+}
+
+func (cc CurrencyCmpFn) GasFeeCapCmp(a, b *types.Transaction) int {
+	return cc(a.GasFeeCap(), a.FeeCurrency(), b.GasFeeCap(), b.FeeCurrency())
+}
+
+func (cc CurrencyCmpFn) Cmp(a, b *types.Transaction, baseFeeA, baseFeeB *big.Int) int {
```
palango, line 33: When would the base fees ever be different?

core/tx_list.go
hbandura: yeap, changing it

core/tx_multicurrency_priceheap.go
hbandura: Yes, in `UpdateFeesAndCurrencies`

core/tx_multicurrency_priceheap.go
hbandura: they are expressed in each tx's currency. I'll add a comment for that

core/tx_list.go
hbandura: Underpriced is a term for a new tx in the (already full) pool that has a gas price lower than all of the txs in the pool.

core/tx_multicurrency_priceheap_test.go
```diff
@@ -321,6 +321,6 @@ func TestIsCheaper(t *testing.T) {
 		return 1
 	}
 	assert.True(t, cheaper.IsCheaper(tx1, tx2))
-	assert.False(t, equal.IsCheaper(tx1, tx2))
+	assert.True(t, equal.IsCheaper(tx1, tx2))
```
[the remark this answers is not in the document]
hbandura: One of the fixes of this PR is that the underprice check actually verifies that price is lower or equal than the cheapest in the pool. Before we were just checking for strict lower, and this test of our creation was testing that border case. Due to the behavior fix, I had to update the test.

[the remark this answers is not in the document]
carterqw2: Yeah, I just asked why do we need to check for equality? What's the point of this specific change?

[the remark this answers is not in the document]
hbandura: Ah, I see. Just to be consistent with what upstream does: https://github.com/ethereum/go-ethereum/blob/ea9e62ca3db5c33aa7438ebf39c189afd53c6bf8/core/txpool/list.go#L582

core/tx_list.go
palango: Ok, so in theory we could find a transaction in a different pool that is cheaper than the one we're processing and remove that instead.

But does that make sense given that we can only add the current on to the pool matching the fee currency?

And if we want to take this path, I would expect that we need the exchange rates to convert prices between the different heaps. How can you work around that?

core/tx_multicurrency_priceheap.go
```diff
@@ -10,9 +10,40 @@ import (
 
 type CurrencyCmpFn func(*big.Int, *common.Address, *big.Int, *common.Address) int
 
-// IsCheaper returns true if tx1 is cheaper than tx2 (GasPrice with currency comparison)
-func (cc CurrencyCmpFn) IsCheaper(tx1, tx2 *types.Transaction) bool {
-	return cc(tx1.GasPrice(), tx1.FeeCurrency(), tx2.GasPrice(), tx2.FeeCurrency()) < 0
+func (cc CurrencyCmpFn) GasTipCapCmp(tx, other *types.Transaction) int {
```
palango, line 13: Any particular reason for making those functions methods on `CurrencyCmpFn`? In my opinion it would be clearer to make those free functions, but that's just a nitpick.

core/tx_list.go
hbandura: Discussed offline: There's only one pool,but many heaps. Exchange rate conversion is already being done on getting the cheapest tx and also in underprice check (in this PR)

core/tx_multicurrency_priceheap.go
```diff
@@ -10,9 +10,39 @@ import (
...
-	return cc(tx1.GasPrice(), tx1.FeeCurrency(), tx2.GasPrice(), tx2.FeeCurrency()) < 0
+	return cc.Cmp(tx1, tx2, nil, nil) <= 0
+}
+
+func (cc CurrencyCmpFn) GasTipCapCmp(tx, other *types.Transaction) int {
+	return cc(tx.GasTipCap(), tx.FeeCurrency(), other.GasTipCap(), other.FeeCurrency())
+}
+
+func (cc CurrencyCmpFn) EffectiveGasTipCmp(tx, other *types.Transaction, baseFeeA, baseFeeB *big.Int) int {
+	if baseFeeA == nil && baseFeeB == nil {
```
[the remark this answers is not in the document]
hbandura: Since it's a public function, I feel like it still makes sense to leave it

core/tx_multicurrency_priceheap.go
hbandura: No reason, at the moment of writing it felt like it made sense. But I have no strong preference over it.

core/tx_multicurrency_priceheap_test.go
```diff
@@ -305,7 +313,7 @@ func TestIsCheaper_FwdFields(t *testing.T) {
 		assert.Equal(t, curr2, *c2)
 		return -1
 	}
-	assert.True(t, cmp.IsCheaper(tx1, tx2))
+	assert.True(t, cmp.Cmp(tx1, tx2, nil, nil) <= 0)
 }
 
 func TestIsCheaper(t *testing.T) {
```
carterqw2, line 319: This test doesn't test anything. Assuming that `CmpValues` is covered here, we need to test the `(cc CurrencyCmpFn) Cmp` method, making sure that TXs are compared properly in different situations.

core/tx_multicurrency_priceheap_test.go
```diff
@@ -320,7 +328,47 @@ func TestIsCheaper(t *testing.T) {
...
+}
+
+// TestMulticurrencyUnderpriced tests that the underpriced method from pricedList functions
+// properly when handling many different currencies.
+func TestMulticurrencyUnderpriced(t *testing.T) {
+	all := newTxLookup()
+	curr1 := common.HexToAddress("curr1")
+	rate1, _ := currency.NewExchangeRate(common.Big1, common.Big1)
+	curr2 := common.HexToAddress("curr2")
+	rate2, _ := currency.NewExchangeRate(common.Big1, common.Big1)
```
carterqw2, line 343: It's actually results in a single currency as `common.HexToAddress("curr1")` and `commont.HexToAddress("curr2")` both would return `0x000000000000000000000000000000000000000C` as neither `curr1` nor `curr2` is a valid hex string.
Also, it would make sense to use something less trivial than exchange rate 1.

carterqw2 (changes requested):

core/tx_multicurrency_priceheap_test.go
```diff
@@ -121,57 +126,58 @@ func TestMultiPushPop(t *testing.T) {
...
-	m.Push(txC(4, c2))  // 400
-	m.Push(txC(21, c2)) // 2100
-	m.Push(txC(7, c2))  // 700
-
-	m.Push(tx(1100)) // 1100
-	m.Push(tx(2700)) // 2700
-	m.Push(tx(560))  // 560
-	m.Push(tx(2150)) // 2150
-	m.Push(tx(750))  // 750
+	m.UpdateFeesAndCurrencies(cmp, gpm)
```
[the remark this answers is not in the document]
hbandura: iirc the tx pool expects the heap has no basefee at start for it's own battery of tests (upstream tests)

core/tx_multicurrency_priceheap.go
```diff
@@ -10,9 +10,40 @@ import (
...
+
+func (cc CurrencyCmpFn) GasFeeCapCmp(a, b *types.Transaction) int {
+	return cc(a.GasFeeCap(), a.FeeCurrency(), b.GasFeeCap(), b.FeeCurrency())
+}
+
+// Cmp returns the same comparison as the priceHeap comparison but taking into account
+// the exchange rate comparison of the CurrencyCmpFn.
+// Each baseFee is expressed in each tx's currency.
+func (cc CurrencyCmpFn) Cmp(a, b *types.Transaction, baseFeeA, baseFeeB *big.Int) int {
+	if baseFeeA != nil && baseFeeB != nil {
```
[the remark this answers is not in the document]
hbandura: I don't think it's possible, since the basefees all come from the same basefee in celo with the exchange currency conversion made. Still, if it happened, it would return a meaningless tip comparison.

core/tx_list.go
```diff
@@ -696,18 +696,27 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
...
 		}
 	}
-	return underpriced
+	return true
 }
 
 // underpricedFor checks whether a transaction is cheaper than (or as cheap as) the
 // lowest priced (remote) transaction in the given heap.
-func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction) bool {
+func (l *txPricedList) underpricedFor(h *priceHeap, tx *types.Transaction, isCheaper func(tx1, tx2 *types.Transaction) bool) bool {
```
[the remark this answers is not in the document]
hbandura: Fair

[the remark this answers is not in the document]
hbandura: Actually, I do not know inside of underpricedFor if I'm using the urgent or floating multicurrency_heap. And, while I could bypass it calling the comparison fn, I still need the basefees. So I think I still need to send an extra parameter here.

core/tx_multicurrency_priceheap_test.go
hbandura: This tests only checks that the Cmp fn forwards the values in the proper order. I'll add a new test for comparing txs (independently from the multicurrency tests)

core/tx_list.go
```diff
@@ -696,18 +696,27 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 }
 
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
-	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx)
+	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
+	if !l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper) {
+		// Check Len after calling underpricedFor since it might prune entries
+		if h.nativeCurrencyHeap.Len() > 0 {
+			return false
+		}
```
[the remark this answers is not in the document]
hbandura: This is a complicated logic, and even more complicated with our multi currency queues:

The underpricedForMulti fn has to return true if there is one heap (of any currency) non-empty for which the tx is underpriced. 
So, if we find that the tx is underpriced for a heap, we can already return false. But only if the heap is non empty. If it is empty, that false means nothing, so we continue looking.
The first link you give is broken for me, but it's important to understand that the logic in the Underprice fn, which tests for one of the floating or urgent has to have items AND be underprice for the return, is not the same as one of the subheaps to be empty (which doesn't imply that another heap in the same multicurrency_heap can't have items).

The last commit you linked is correct, but in a CR (which I agreed) it was noted that I could probably prune and exit early in some of these cases.

core/tx_multicurrency_priceheap.go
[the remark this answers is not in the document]
hbandura: changed it

core/tx_list.go
```diff
@@ -696,18 +696,27 @@ func (l *txPricedList) Underpriced(tx *types.Transaction) bool {
 }
 
 func (l *txPricedList) underpricedForMulti(h *multiCurrencyPriceHeap, tx *types.Transaction) bool {
-	underpriced := l.underpricedFor(h.nativeCurrencyHeap, tx)
+	// Has to be underpriced for ALL heaps to be underpriced for the whole pool.
+	if !l.underpricedFor(h.nativeCurrencyHeap, tx, h.IsCheaper) {
```
carterqw2, line 700: I suggest we don't use `underpricedFor` at all, since we have to cancel out the check for the empty heap anyway AND also pass the external `IsCheaper` comparison function. It makes sense to do the opposite and extract the "pruning" part into a separate function and just call it.

core/tx_multicurrency_priceheap_test.go
carterqw2: > This tests only checks that the Cmp fn forwards the values in the proper order.

What do you mean exactly?

core/tx_multicurrency_priceheap_test.go
```diff
@@ -320,7 +370,52 @@ func TestIsCheaper(t *testing.T) {
...
-	assert.False(t, equal.IsCheaper(tx1, tx2))
-	assert.False(t, notCheaper.IsCheaper(tx1, tx2))
+	assert.True(t, cheaper.Cmp(tx1, tx2, nil, nil) <= 0)
+	assert.True(t, equal.Cmp(tx1, tx2, nil, nil) <= 0)
+	assert.False(t, notCheaper.Cmp(tx1, tx2, nil, nil) <= 0)
+}
+
+// TestMulticurrencyUnderpriced tests that the underpriced method from pricedList functions
+// properly when handling many different currencies.
+func TestMulticurrencyUnderpriced(t *testing.T) {
```
carterqw2, line 380: Could you reorganize this test to be a table test with different initial state and different cases for `Underpriced` calls? Now the state of all heaps is mixed up so it's very hard to follow the logic of the final assertions. Then we would have more meaningful test cases like:
- Empty pool, any TX is not underpriced.
- Single native TX in the pool, a cheaper incoming TX is underpriced.
- Two native TXs in the pool, an incoming TX must not be cheaper than the cheapest one.
- 1 TX in native pool, 1 TX in another pool, an incoming TX must not be cheaper than the cheapest of the two.

carterqw2 (changes requested):

core/tx_multicurrency_priceheap_test.go
palango: @hbandura I added a more structured approach in https://github.com/celo-org/celo-blockchain/commit/3e7ae1c835261e969f4059e40d59cccced8b97fc , feel free cherry-pick that commit if you like the apporach.

hbandura: @palango thanks, I'll use that

core/tx_multicurrency_priceheap_test.go
hbandura: That the fields to Cmp arrived in the proper order, but I deleted it since it doesn't add much.

core/tx_multicurrency_priceheap_test.go
palango: I added the remaining cases Valentin mentioned.

core/tx_multicurrency_priceheap_test.go
carterqw2: Ah, that must be a misunderstanding them, I meant the `TestIsCheaper` test. Since it's not testing `IsCheaper` and the comparison functions are mocked, it just tested that -1 <= 0, 0 <= = and 1 >= 0, so I thought we could remove it.

hbandura: That one is testing that the return value from the Cmp function is being returned successfully to the IsCheaper function.

carterqw2: How does it do it if there is no `IsCheaper` call?

hbandura: Sorry, meant returning from the currency comparison to the transaction comparison

carterqw2 (approved):

PR #2216, opened 2024-01-19
Hardcode pre gingerbread block gas limits for existing chains

piersy:
### Description

Ensures that for mainnet, alfajores and baklava, nodes will return the correct block gas limit up to the gingerbread fork even if they are not archive nodes when running in eth compatibility mode (I.E. `--disablerpcethcompatibility` is not set).

### Other changes

A small re-factor of GetBlockByNumber and GetBlockByHash to reduce the depth of nesting.

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

This change is not backwards compatible with the previous implementations on mainnet, alfajores or baklava which would have returned a zero gas limit for a block if there was no state for that block.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -752,10 +752,11 @@ func addEthCompatibilityFields(ctx context.Context, response map[string]interfac
 		var gasLimit uint64
 		var err error
 
-		// For mainnet, alfajores and baklava we have a set of hardcoded values derived from historical state that we can use.
-		v := params.PreGingerbreadNetworkGasLimits[b.ChainConfig().ChainID.Uint64()]
-		if v != nil {
-			gasLimit = v.Limit(header.Number)
+		// For mainnet, alfajores and baklava we have a set of hardcoded values derived from historical state that we can
+		// use, note ChainID might be unset, so we need to account for that.
+		chainId := b.ChainConfig().ChainID
+		if chainId != nil && params.PreGingerbreadNetworkGasLimits[chainId.Uint64()] != nil {
+			gasLimit = params.PreGingerbreadNetworkGasLimits[chainId.Uint64()].Limit(header.Number)
 		} else {
 			// If no hardcoded limits are available for this network then we will try to look up the gas limit in the state.
 			hash := header.Hash()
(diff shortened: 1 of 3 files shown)


internal/ethapi/api.go
```diff
@@ -706,64 +706,68 @@ func (s *PublicBlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.H
...
 // See https://github.com/celo-org/celo-blockchain/issues/1945
 func addEthCompatibilityFields(ctx context.Context, response map[string]interface{}, b Backend, block *types.Block) {
-	isGingerbread := b.ChainConfig().IsGingerbread(block.Number())
-	if !b.RPCEthCompatibility() {
-		if !isGingerbread {
-			delete(response, "gasLimit")
-		}
-		return
-	}
-
```
karlb, line 753: Why this change? Isn't it better to check inside `addEthCompatibilityFields` rather than at every call site?

internal/ethapi/api.go
```diff
@@ -752,10 +752,11 @@ func addEthCompatibilityFields(ctx context.Context, response map[string]interfac
 		var gasLimit uint64
 		var err error
 
-		// For mainnet, alfajores and baklava we have a set of hardcoded values derived from historical state that we can use.
-		v := params.PreGingerbreadNetworkGasLimits[b.ChainConfig().ChainID.Uint64()]
-		if v != nil {
-			gasLimit = v.Limit(header.Number)
+		// For mainnet, alfajores and baklava we have a set of hardcoded values derived from historical state that we can
+		// use, note ChainID might be unset, so we need to account for that.
```
karlb, line 756: In which case is chainId unset?

piersy: `cmd/geth.TestCustomGenesis` explicitly checks for unset chainID's, see [here](https://github.com/celo-org/celo-blockchain/blob/3a9fda2314604e7533cfc1086c3b1512c78071ae/cmd/geth/genesis_test.go#L40)

internal/ethapi/api.go
piersy: If you are referring to the check for rpc compatibility, I prefer to have the check outside rather than inside, so that from the top level code reads if X then do Y rather than just do Y (but then internally, oh actually don't do anything at all if !X).

If you are referring to the removal of the deletion of the gasLimit I updated RPCMarshalHeader to only add the gas limit if it the rlp decoded block had a gas limit set and RPCMarshalHeader is always called regardless of the eth compatibility setting.

karlb: > If you are referring to the check for rpc compatibility, I prefer to have the check outside rather than inside, so that from the top level code reads if X then do Y rather than just do Y (but then internally, oh actually don't do anything at all if !X).

I prefer not to repeat the checking code and avoiding the possibility to forget the check. But I can live with either approach.

karlb (approved):

PR #2218, opened 2024-01-19
Update container CI

jcortejoso:
### Description

This PR include two changes to container CI workflows:

1. Remove the `build-sign-binaries.yaml` workflow: This workflow/release relied on xgo compilation, that relied on removed dependencies (https://github.com/karalabe/xgo). This release would be needed for Android/iOS releases, but as longer as it is not supported by upstream go-ethereum right now, we'd need to figure out and replace dependencies to build it again. Removing the workflow until we work on this refactor.

2. Update the container CI workflow version, and enable `trivy` scan so CVE reports are available under the security tab.


palango (approved):
Thanks for fixing that so quickly! 🏅

PR #2230, opened 2024-01-24
Fix underpriced check for multi currency

hbandura:
Fix the underpriced check by checking against all currency heaps

### DIFF
--- a/contracts/currency/currency.go
+++ b/contracts/currency/currency.go
@@ -153,6 +153,11 @@ func NewManager(vmRunner vm.EVMRunner) *CurrencyManager {
 	return newManager(GetExchangeRate, vmRunner)
 }
 
+// NewCacheOnlyManager creates a cache-full currency manager. TESTING PURPOSES ONLY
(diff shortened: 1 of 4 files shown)


palango (approved):

piersy:
@hbandura this failure `TestEthersJSCompatibility: e2e_test` is fixed by my pr #2229 , not sure about the other two failures though

piersy:
> @hbandura this failure `TestEthersJSCompatibility: e2e_test` is fixed by my pr #2229 , not sure about the other two failures though

Hey @hbandura I started looking into those other failing tests, because they were failing for me and it seems that they are also failing for similar reasons to `TestEthersJSCompatibility` basically a lack of gingerbread fields on the genesis block. I'm working on a PR to fix all these gingerbread genesis issues and once that is done it should solve your problems here.

piersy:
@hbandura #2245 has been merged and I think will fix all the flaky tests you are encountering on this PR!

PR #2236, opened 2024-01-24
chore: Update e2e_test dependencies

palango:
### Description

Update dependencies in `e2e_test`.
@piersy Any idea how I can avoid the lock file version upgrade? Or is this fine?


piersy (approved):
Looks fine assuming the tests pass.

The lock file is simply tracking all the versions of the the packages used (which includes all transitively included packages) so if you change your dependencies then you expect the lock file to change accordingly.

PR #2239, opened 2024-01-27
eth/filter, ethclient/gethclient: added fullTx-flag to NewPendingTransactions

diwu1989:
Porting https://github.com/ethereum/go-ethereum/pull/25186/files to Celo.
There were some trivial merge conflicts in `eth/filters/api.go` that needed to be resolved before this could apply.

Would like this to be merged in and deployed to Forno so that websocket streams from the pub RPC TX pool includes the body, otherwise, there's just extra load doing round trip to fetch via pending tx header.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -39,8 +39,9 @@ import (
 // and associated subscription in the event system.
 type filter struct {
 	typ      Type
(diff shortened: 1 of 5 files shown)


palango (approved):
Great, thanks for creating this PR  ⭐
Code looks good and the remaining CI failures are due to missing permissions.

PR #2245, opened 2024-02-01
Fix genesis block gingerbread fields

piersy:
### Description

This PR fixes the way the the gasLimit and gasPriceMinimum are retrieved from state such that for all non genesis blocks the value returned is the value present in the state of the parent block, and for the genesis block the value returned is the value present in the state of the genesis block (because the genesis has no parent). Prior to this fix requesting the gasLimit or gasPrice minimum for the genesis block would result in an error because we would try to get the state for the genesis block's parent which of course did not exist.

It also removes a restriction in the mycelo genesis package (used by e2e tests) that prevented us from setting the gingerbread activation block to zero. This restriction was in place because originally it was not possible to set the baseFeeOpCodeActivationBlock (AKA the gingerbread block) to zero in the GasPriceMinimum contract but the contract was later updated to allow this.

It also updates all e2e tests that were setting the gingerbread block to 1 to now use 0.

These fixes fix a number of flakily failing eth compatibility and gingerbread tests, that would fail if run on the genesis block.
```
 TestSendCelo
 TestTraceSendCeloViaGoldToken
 TestCallTraceTransactionNativeTransfer
 TestPrestateTransactionNativeTransfer
 TestEpochBlockMarshaling
 TestStartStopValidators
 TestBlockTracingConcurrentMapAccess
 TestBlockTracingSequentialAccess
 TestRPCDynamicTxGasPriceWithBigFeeCap
 TestRPCDynamicTxGasPriceWithState
 TestRPCDynamicTxGasPriceWithoutState
 TestEthersJSCompatibility
 TestEthersJSCompatibilityDisableAfterGingerbread
 TestTransferCELO
 TestTransferERC20
```

### Tested

Two end to end tests have been added to verify the new behaviour

### Backwards compatibility

This is non backwards compatible, since it is changing the gasLimit and baseFee values that people will see on the genesis block when running in eth compatibility mode. They will now see the values defined in the genesis state rather than zero.

### DIFF
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -44,7 +44,7 @@ func init() {
 // network to process the transaction.
 func TestSendCelo(t *testing.T) {
 	ac := test.AccountConfig(3, 2)
-	gingerbreadBlock := common.Big1
+	gingerbreadBlock := common.Big0
 	gc, ec, err := test.BuildConfig(ac, gingerbreadBlock)
 	require.NoError(t, err)
 	network, shutdown, err := test.NewNetwork(ac, gc, ec)
@@ -70,7 +70,7 @@ func TestSendCelo(t *testing.T) {
 // and can be useful for debugging these traces.
 func TestTraceSendCeloViaGoldToken(t *testing.T) {
 	ac := test.AccountConfig(3, 2)
-	gingerbreadBlock := common.Big1
+	gingerbreadBlock := common.Big0
 	gc, ec, err := test.BuildConfig(ac, gingerbreadBlock)
 	require.NoError(t, err)
 	network, shutdown, err := test.NewNetwork(ac, gc, ec)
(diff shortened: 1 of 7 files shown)


e2e_test/e2e_test.go
```diff
@@ -733,15 +733,16 @@ func TestEthersJSCompatibilityDisableAfterGingerbread(t *testing.T) {
 	defer cancel()
 
 	result = make(map[string]interface{})
-	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &result, "eth_getBlockByNumber", "latest", true)
+	err = network[0].WsClient.GetRPCClient().CallContext(ctx, &result, "eth_getBlockByNumber", "0x0", true)
 	require.NoError(t, err)
 
+	fmt.Printf("baseFee %v\n", result["baseFeePerGas"])
```
carterqw2, line 739: Debugging leftover?

piersy: Thank you!

e2e_test/e2e_test.go
carterqw2, line 1: Looks like we could also fix `e2e_transfer_test` [here](https://github.com/celo-org/celo-blockchain/blob/master/e2e_test/e2e_transfer_test.go#L39) and [here](https://github.com/celo-org/celo-blockchain/blob/master/e2e_test/e2e_transfer_test.go#L454).

PR #2248, opened 2024-02-01
renovate: ignore non-security npm updates

palango:
Those updates take a lot of time of time to merge and provide little actual benefit. So lets disable non-security updates.


piersy (approved):
Looks good to me

PR #2258, opened 2024-02-20
Palango/estimation fix

palango:
### Description

This PR adds a new config flag: `SkipDebitCredit`. If set it skips calls to `debit` and `credit`.

This is useful in the case of gas estimation or calls, where some assumptions about the values don't hold and would otherwise cause reverts in the fee currency contracts.


core/state_transition.go
```diff
@@ -525,8 +528,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	}, nil
 }
 
-// distributeTxFees calculates the amounts and recipients of transaction fees and credits the accounts.
-func (st *StateTransition) distributeTxFees() error {
+// creditTxFees calculates the amounts and recipients of transaction fees and credits the accounts.
+func (st *StateTransition) creditTxFees() error {
```
carterqw2, line 532: Worth updating the function name in the log line [here](https://github.com/celo-org/celo-blockchain/pull/2258/files#diff-8c398ef0a79f97ba6d497a99247a815b3a51918de9e16954215b6073f907c92eR578) as well then.

piersy:
Hey @palango, could you expand upon the use cases for this, is there an associated ticket? It just seems like a bit of a random change, and I'm wondering why do it, who is it helping?

core/state_transition.go
palango: Done

palango:
> Hey @palango, could you expand upon the use cases for this, is there an associated ticket? It just seems like a bit of a random change, and I'm wondering why do it, who is it helping?

I expanded the PR's description, let me know if that makes sense now.

piersy:
Hi @palango thanks for expanding the description, makes a lot more sense now. However I still don't see why we need the flag, it sounds like the first commit solves the problem at hand so why add a flag to skip credit/debit?

palango:
> Hi @palango thanks for expanding the description, makes a lot more sense now. However I still don't see why we need the flag, it sounds like the first commit solves the problem at hand so why add a flag to skip credit/debit?

It fixes the problem for the tokens we're aware of, with the second it fixed the problem for all possible implementations.

Imagine a token that reverts on credit when a receiver is the zero address, which might happen for example when the coinbase is not set for the node doing the gas estimation.
If the token implementation early returns (like [here](https://github.com/celo-org/celo-monorepo/blob/6604de72dc026a3bc8b765e4eb3cec5c7b84ad85/packages/protocol/contracts-0.8/stability/FeeCurrencyAdapter.sol#L95-L98)), then the fix commit is sufficient. But if that's not the case the revert would still happen.
So in general, without further knowledge of the token, ww just skip calling debit/credit.

carterqw2 (approved):

PR #2265, opened 2024-02-21
HFork flags and configs

hbandura:
Adds the override flags, configs, and necessary boilerplate for the H fork activation, while removing the old Gingerbread overrides.


piersy:
Hi @hbandura what's the plan for the H hardfork, I'm not aware of any changes agreed yet? Also do we want to call it just 'H'? It seems historically we went for a more memorable name such as donut, espresso ...

hbandura:
> Hi @hbandura what's the plan for the H hardfork, I'm not aware of any changes agreed yet? Also do we want to call it just 'H'? It seems historically we went for a more memorable name such as donut, espresso ...

In the past we have gone with Xfork when the name hasn't been defined yet. It can be renamed afterwards, it usually takes some time for the name to be discussed and decided.

While it hasn't been yet approved, there's a change to be written and tested for a possible future fork. I'll continue this in private.

PR #2271, opened 2024-02-27
Improve error messages for credit/debit reverts

karlb:
### Description

The revert messages now show up in the RPC responses (for calls). For most cases, this should provide enough debugging information without looking at the server logs.

### Tested

Tested both cases manually by creating a fee currency that always fails in the respective method and sending a CIP-64 calls and txs via viem.

### Backwards compatibility

Fully backwards compatible


palango (approved):
Awesome, this will make debugging much easier!

PR #2274, opened 2024-02-28
Remove special handling for gas estimation

### LINKED ISSUE #2267: Asymmetry in mechanisms to skip calling debitGasFees & creditGasFees
The blockchain client [skips calling debitGasFees if the fee is zero](https://github.com/celo-org/celo-blockchain/blob/8175f3a1ae4bc52ea963d3ef1e7cf808636db144/contracts/erc20gas/erc20gas.go#L18-L21), this is to support estimating gas (gas is not charged for the credit or debit gas fees calls since they are covered by an increased intrinsic cost). However the blockchain client does not skip calling credit gas fees if the fee is zero, in fact this check is currently performed [in the contract ](https://github.com/celo-org/celo-monorepo/blame/6604de72dc026a3bc8b765e4eb3cec5c7b84ad85/packages/protocol/contracts-0.8/stability/FeeCurrencyAdapter.sol#L95-L98). 

We should be consistent about where this functionality is skipped, or whether it is skipped at all.

Given that we just [added a flag to allow explicit skipping of credit and debit calls](https://github.com/celo-org/celo-blockchain/pull/2258). It seems sensible to remove all other logic to skip.

palango:
This is now done by the `SkipDebitCredit` flag.

Resolves #2267

### DIFF
--- a/contracts/erc20gas/erc20gas.go
+++ b/contracts/erc20gas/erc20gas.go
@@ -18,9 +18,6 @@ const (
 )
 
 func DebitFees(evm *vm.EVM, address common.Address, amount *big.Int, feeCurrency *common.Address) error {
-	if amount.Cmp(big.NewInt(0)) == 0 {
-		return nil
-	}
 	// Function is "debitGasFees(address from, uint256 value)"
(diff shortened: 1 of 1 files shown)


piersy (approved):
Looking good!

PR #2275, opened 2024-02-28
Test celo transaction replacement

palango:
This PR adds tests for replacing celo transactions in the txpool.

Adding this required extending the mocked contracts. It also found two issues:
1. A comparison of fee currencies only checked the pointers, not the underlying addresses.
2. There is the problem of rounding errors happening when calculating the required replacement fee values. For example if a 10% increase is required and a fee of 2 wei is offered by the old transaction, the calculation also rounded the new required fee down to 2 (2 * 110 / 100). I solved this by removing the division by 100. This might cause overflow, but I believe that this won't cause problems in reality as the fee amounts will never be this high. But please have a special look at this during the review.

### DIFF
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -316,7 +316,7 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
 		var newGasFeeCap, newGasTipCap *big.Int
 
 		// Short circuit conversion if both are the same currency
-		if old.FeeCurrency() == tx.FeeCurrency() {
+		if common.AreSameAddress(old.FeeCurrency(), tx.FeeCurrency()) {
 			if old.GasFeeCapCmp(tx) >= 0 || old.GasTipCapCmp(tx) >= 0 {
 				return false, nil
 			}
--- a/core/tx_pool_test.go
+++ b/core/tx_pool_test.go
@@ -176,9 +176,6 @@ func setupTxPool() (*TxPool, *ecdsa.PrivateKey) {
 }
 
 func setupTxPoolWithConfig(config *params.ChainConfig) (*TxPool, *ecdsa.PrivateKey) {
-	// statedb, _ := state.New(common.Hash{}, state.NewDatabase(rawdb.NewMemoryDatabase()), nil)
-	// blockchain := &testBlockChain{10000000, statedb, new(event.Feed)}
-
 	blockchain := newTestBlockchain()
 	key, _ := crypto.GenerateKey()
 	pool := NewTxPool(testTxPoolConfig, config, blockchain)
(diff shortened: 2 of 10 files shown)


core/tx_list.go
```diff
@@ -344,18 +344,18 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
...
-		aTip := a.Mul(a, oldGasTipCap)
+		thresholdFeeCap := new(big.Int).Mul(a, oldGasFeeCap)
+		thresholdTip := a.Mul(a, oldGasTipCap)
 
 		// thresholdTip    = oldTip * (100 + priceBump) / 100
 		b := big.NewInt(100)
-		thresholdFeeCap := aFeeCap.Div(aFeeCap, b)
-		thresholdTip := aTip.Div(aTip, b)
+		newFeeCap := new(big.Int).Mul(b, newGasFeeCap)
+		newTip := a.Mul(b, newGasTipCap)
```
karlb, line 353: The comments don't match the code anymore.

palango: Nice catch, let's agree on the solution before I fix those.

core/tx_list.go
```diff
@@ -344,18 +344,18 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
...
 
 		// Have to ensure that either the new fee cap or tip is higher than the
 		// old ones as well as checking the percentage threshold to ensure that
 		// this is accurate for low (Wei-level) gas price replacements
-		if newGasFeeCap.Cmp(thresholdFeeCap) < 0 || newGasTipCap.Cmp(thresholdTip) < 0 {
+		if newFeeCap.Cmp(thresholdFeeCap) < 0 || newTip.Cmp(thresholdTip) < 0 {
```
karlb, line 358: This comment is hard for me to read. Does it say that *either* has to be higher by the required percentage? Because the code looks like *both* have to be increased by `priceBump` percent.

core/tx_list.go
karlb: Is the old code really problematic? The threshold is rounded down, so replacing is somewhat easier than with the fully accurate calculation, right?

core/tx_pool_test.go
```diff
@@ -2459,6 +2457,228 @@ func TestTransactionReplacementDynamicFee(t *testing.T) {
...
+	}
+
+	if err := validateTxPoolInternals(pool); err != nil {
+		t.Fatalf("pool internal state corrupted: %v", err)
+	}
+}
+
+// Tests that the pool rejects replacement celo dynamic fee transactions that don't
+// meet the minimum price bump required.
+func TestTransactionReplacementCeloDynamicFeeDifferentCurrencies(t *testing.T) {
```
carterqw2, line 2572: That assumes 1:1 exchange rate, right?

core/tx_list.go
carterqw2: Looks like neither of the two issues (rounding errors on wei-level fee values and overflow) could realistically happen in real life, so not dividing by 100 looks okay but the old code doesn't looks outrageously wrong either.

karlb: Agreed, both look fine. To avoid unnecessary change, I would stay with the old version unless some need for the change can be shown.

core/tx_pool_test.go
palango: Yes

core/tx_list.go
palango: > Is the old code really problematic? The threshold is rounded down, so replacing is somewhat easier than with the fully accurate calculation, right?

Exactly, without this change it is possible to replace transactions without increasing the gas price. In Ethereum this case is handled by [this early return](https://github.com/celo-org/celo-blockchain/blob/e91fdf12b79379259e5bee0d56e687af86ad98a5/core/tx_list.go#L320). It also breaks [this assumption in the ported test](https://github.com/celo-org/celo-blockchain/blob/e91fdf12b79379259e5bee0d56e687af86ad98a5/core/tx_pool_test.go#L2620-L2623), that's why I started thinking about it in the first place.

karlb: We could increase the numbers in the test to more realistic levels to make it pass or use your suggested solution. No strong opinions on this.

palango: If we decide that we don't want to change the behavior I'd just remove that assertion from the test.

palango: I removed the changes and adapted the test to more realistic values.

karlb (approved):

PR #2277, opened 2024-02-29
Use sync calls to add txs to pool in tests

piersy:
Without the use of sync calls there is no guarantee that txpoool reorg will run between adding txs and chekcking some condition, this causes tests to fail randomly.

This PR ensures the use of sync calls.

### Tested

To verify my suspicion that the lack of reorg running was the cause of the error I was seeing in `TestTransactionPoolUnderpricing` i used the following patch to reliably recreate the error.

```diff
diff --git a/core/tx_pool.go b/core/tx_pool.go
index 9a3003d6f..0a35c8a76 100644
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -263,6 +263,8 @@ type txPoolContext struct {
 // current state) and future transactions. Transactions move between those
 // two states over time as they are received and processed.
 type TxPool struct {
+	reorgSkip   int32
+	skipLock    sync.Mutex
 	config      TxPoolConfig
 	chainconfig *params.ChainConfig
 	chain       blockChain
@@ -1248,6 +1250,17 @@ func (pool *TxPool) scheduleReorgLoop() {

 // runReorg runs reset and promoteExecutables on behalf of scheduleReorgLoop.
 func (pool *TxPool) runReorg(done chan struct{}, reset *txpoolResetRequest, dirtyAccounts *accountSet, events map[common.Address]*txSortedMap) {
+	pool.skipLock.Lock()
+	pool.reorgSkip++
+	if pool.reorgSkip >= 3 {
+		pool.skipLock.Unlock()
+		close(done)
+		return
+	}
+	pool.skipLock.Unlock()
+
+	println("reorging")
+
 	defer func(t0 time.Time) {
 		reorgDurationTimer.Update(time.Since(t0))
 	}(time.Now())
diff --git a/core/tx_pool_test.go b/core/tx_pool_test.go
index 2a2ae348f..e48733f05 100644
--- a/core/tx_pool_test.go
+++ b/core/tx_pool_test.go
@@ -1899,7 +1899,9 @@ func TestTransactionPoolUnderpricing(t *testing.T) {
 	ltx := pricedTransaction(0, 100000, big.NewInt(1), keys[2])

 	// Import the batch and that both pending and queued transactions match up
+	println("adding remote")
 	pool.AddRemotes(txs)
+	println("adding local")
 	pool.AddLocal(ltx)

 	pending, queued := pool.Stats()
@@ -1915,17 +1917,21 @@ func TestTransactionPoolUnderpricing(t *testing.T) {
 	if err := validateTxPoolInternals(pool); err != nil {
 		t.Fatalf("pool internal state corrupted: %v", err)
 	}
+	println("adding remote")
 	// Ensure that adding an underpriced transaction on block limit fails
 	if err := pool.AddRemote(pricedTransaction(0, 100000, big.NewInt(1), keys[1])); err != ErrUnderpriced {
 		t.Fatalf("adding underpriced pending transaction error mismatch: have %v, want %v", err, ErrUnderpriced)
 	}
+	println("adding remote")
 	// Ensure that adding high priced transactions drops cheap ones, but not own
 	if err := pool.AddRemote(pricedTransaction(0, 100000, big.NewInt(3), keys[1])); err != nil { // +K1:0 => -K1:1 => Pend K0:0, K0:1, K1:0, K2:0; Que -
 		t.Fatalf("failed to add well priced transaction: %v", err)
 	}
+	println("adding remote")
 	if err := pool.AddRemote(pricedTransaction(2, 100000, big.NewInt(4), keys[1])); err != nil { // +K1:2 => -K0:0 => Pend K1:0, K2:0; Que K0:1 K1:2
 		t.Fatalf("failed to add well priced transaction: %v", err)
 	}
+	println("adding remote")
 	if err := pool.AddRemote(pricedTransaction(3, 100000, big.NewInt(5), keys[1])); err != nil { // +K1:3 => -K0:1 => Pend K1:0, K2:0; Que K1:2 K1:3
 		t.Fatalf("failed to add well priced transaction: %v", err)
 	}
```

### Related issues

- Partial fix of #2276

### DIFF
--- a/core/tx_pool_test.go
+++ b/core/tx_pool_test.go
@@ -348,25 +348,25 @@ func TestInvalidTransactions(t *testing.T) {
 	from, _ := deriveSender(tx)
 
 	testAddBalance(pool, from, big.NewInt(1))
-	if err := pool.AddRemote(tx); !errors.Is(err, ErrInsufficientFunds) {
+	if err := pool.addRemoteSync(tx); !errors.Is(err, ErrInsufficientFunds) {
 		t.Error("expected", ErrInsufficientFunds)
 	}
 
 	balance := new(big.Int).Add(tx.Value(), new(big.Int).Mul(new(big.Int).SetUint64(tx.Gas()), tx.GasPrice()))
 	testAddBalance(pool, from, balance)
-	if err := pool.AddRemote(tx); !errors.Is(err, ErrIntrinsicGas) {
+	if err := pool.addRemoteSync(tx); !errors.Is(err, ErrIntrinsicGas) {
 		t.Error("expected", ErrIntrinsicGas, "got", err)
 	}
 
 	// Adding a gateway fee should result in deprecation error.
 	tx = lesTransaction(0, 100, big.NewInt(50), key)
-	if err := pool.AddRemote(tx); err != ErrGatewayFeeDeprecated {
+	if err := pool.addRemoteSync(tx); err != ErrGatewayFeeDeprecated {
 		t.Error("expected", ErrGatewayFeeDeprecated, "got", err)
 	}
 
 	// Should still return a deprecation error.
 	pool.currentState.AddBalance(from, tx.GatewayFee())
-	if err := pool.AddRemote(tx); err != ErrGatewayFeeDeprecated {
+	if err := pool.addRemoteSync(tx); err != ErrGatewayFeeDeprecated {
 		t.Error("expected", ErrGatewayFeeDeprecated, "got", err)
 	}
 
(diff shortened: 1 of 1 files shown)


carterqw2 (approved):

piersy:
> This sounds reasonable for at least some of the cases. I'm not sure we need the sync everywhere but I also don't see any harm in it (it can only slow things down, but not to a degree that matters, AFAICS).
> 
> As so often, upstream already solved the issue in at least some cases, see [ethereum/go-ethereum@c866dfd#diff-6083272455b210c86aaa70e7a348d038d2867870cedb4d402b4909abef8fa23b](https://github.com/ethereum/go-ethereum/commit/c866dfdc78fda6a9cd940970aa8c787dd4762205#diff-6083272455b210c86aaa70e7a348d038d2867870cedb4d402b4909abef8fa23b) .

Yep, I did check upstream but it seems unlikely we will do any more upstream merges, so this shouldn't cause us any trouble.

PR #2279, opened 2024-03-04
Validate debitGasFees execution in tx pool

karlb:
### Description

Txs must not fail during `debitGasFees` execution during the state transition, so we must remove problematic txs during the tx pool validation. We do this by executing `debitGasFees` inside the tx pool validation and discarding the resulting state changes. This is not overly efficient, but still better than the alternatives that have been considered so far.

### Other changes

* Added `ExecuteAndDiscardChanges` to `vm.EVMRunner`. I would have liked to avoid this change, but `StaticCall` does not allow (even temporary) changes and I didn't see a better way.
* extracted `debitGasFeesSelector` and `creditGasFeesSelector` and updated comment to show how to generate the selector with `cast` instead of python.
* removed old check for sufficient ERC20 balance, since that is implicitly checked by executing `debitGasFees` and I want to avoid duplicate work
* no espresso hardfork check for fee currencies in `ValidateTransactorBalanceCoversTx`, since we don't need to support older forks in the tx pool

### Tested

Manually by sending txs via Viem to a local mycelo. Adding a unit test would be desirable.

I'm haven't tested that the state revert works as expected and am also not sure about the performance impact.

### Backwards compatibility

At the time of writing, we don't have any fee currencies on mainnet that can fail in `debitGasFees`. The error message for fee currencies without sufficient balance will change with this PR. The new error will contain the revert message from the token's `debitGasFees`.

### DIFF
--- a/contracts/erc20gas/erc20gas.go
+++ b/contracts/erc20gas/erc20gas.go
@@ -8,6 +8,7 @@ import (
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/common/hexutil"
 	"github.com/celo-org/celo-blockchain/contracts/internal/n"
+	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/core/vm"
 	"github.com/celo-org/celo-blockchain/log"
 )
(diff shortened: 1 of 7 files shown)


core/tx_pool.go
```diff
@@ -1721,8 +1722,7 @@ func (pool *TxPool) demoteUnexecutables() {
 //   - Post-Espresso: it ensures balance >= GasFeeCap * gas + value + gatewayFee (2)
 //
 // For non-native tokens(cUSD, cEUR, ...) as feeCurrency:
-//   - Pre-Espresso: it ensures balance > GasPrice * gas + gatewayFee (3)
-//   - Post-Espresso: it ensures balance >= GasFeeCap * gas + gatewayFee (4)
+//   - It executes a static call on debitGasFees, implicitly ensuring balance >= GasFeeCap * gas and that `from` is not on the token's block list
 func ValidateTransactorBalanceCoversTx(tx *types.Transaction, from common.Address, currentState *state.StateDB, currentVMRunner vm.EVMRunner, espresso bool) error {
 	if tx.FeeCurrency() == nil {
 		balance := currentState.GetBalance(from)
```
palango, line 1728: When cleaning up maybe also remove the `if espresso` condition below

contracts/erc20gas/erc20gas.go
```diff
@@ -17,17 +18,35 @@ const (
 	maxGasForCreditGasFeesTransactions uint64 = 1 * n.Million
 )
 
+var (
+	// Recalculate with `cast sig 'debitGasFees(address from, uint256 value)'`
+	debitGasFeesSelector = hexutil.MustDecode("0x58cf9672")
```
palango, line 23: Would be good to add the `creditGasFeesSelector` here as well.

karlb: done

core/tx_pool.go
karlb: done

core/tx_pool.go
```diff
@@ -1717,28 +1718,20 @@ func (pool *TxPool) demoteUnexecutables() {
...
-			}
-		} else {
-			// cost = GasPrice * gas + value + gatewayFee, as in (1)
-			cost = tx.Cost()
+
+		// cost = GasFeeCap * gas + value + gatewayFee, as in (2)
+		cost := new(big.Int).SetUint64(tx.Gas())
+		cost.Mul(cost, tx.GasFeeCap())
+		cost.Add(cost, tx.Value())
+		if tx.GatewayFeeRecipient() != nil {
```
piersy, line 1733: Wasn't support for gateway fees removed in gingerbread, could this be cleaned up as well?

karlb: Yes, and I could also remove the "(2)" comment, since only once case is left.

core/tx_pool.go
```diff
@@ -1749,31 +1742,7 @@ func ValidateTransactorBalanceCoversTx(tx *types.Transaction, from common.Addres
...
-			}
-		} else {
-			// cost = GasPrice * gas + gatewayFee, as in (3)
-			cost := tx.Fee()
-			if balance.Cmp(cost) <= 0 {
-				log.Debug("ValidateTransactorBalanceCoversTx: insufficient funds", "feeCurrency", tx.FeeCurrency(), "balance", balance)
-				return ErrInsufficientFunds
-			}
-		}
+		return erc20gas.TryDebitFees(tx, from, currentVMRunner)
```
piersy, line 1745: Rather than having an else statement we could simplify this function like this:

```go
if tx.FeeCurrency() == nil {
  return erc20gas.TryDebitFees(tx, from, currentVMRunner)
}
...
```

palango (approved):
Looks good.

PR #2282, opened 2024-03-12
HFork flags and configs

hbandura:
Add all H fork flags and configs.  Fork is not activated.


karlb (approved):
It would have been nice to have the real hard fork name, but since we don't know that yet 🤷.

karlb (changes requested):
I'm just blocking this temporarily until the next release branch has been branched off of master (should not be long.)

karlb (approved):

PR #2283, opened 2024-03-12
Send fee tip to fee handler if coinbase not set

palango:
### Description

Send fee tip to fee handler if coinbase not set

### DIFF
--- a/contracts/erc20gas/erc20gas.go
+++ b/contracts/erc20gas/erc20gas.go
@@ -78,7 +78,26 @@ func CreditFees(
 	gatewayFee *big.Int,
 	baseTxFee *big.Int,
 	feeCurrency *common.Address) error {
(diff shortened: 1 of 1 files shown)


contracts/erc20gas/erc20gas.go
```diff
@@ -78,7 +78,26 @@ func CreditFees(
 	gatewayFee *big.Int,
 	baseTxFee *big.Int,
 	feeCurrency *common.Address) error {
-	transactionData := common.GetEncodedAbi(creditGasFeesSelector, [][]byte{common.AddressToAbi(from), common.AddressToAbi(feeRecipient), common.AddressToAbi(*gatewayFeeRecipient), common.AddressToAbi(feeHandler), common.AmountToAbi(refund), common.AmountToAbi(tipTxFee), common.AmountToAbi(gatewayFee), common.AmountToAbi(baseTxFee)})
+
+	// The fee recipient is the validator building the block.
+	// If the coinbase is not set for any reason, send the amount to the fee handler instead.
+	if feeRecipient == common.ZeroAddress {
+		feeRecipient = feeHandler
+	}
```
karlb, line 86: I like the change, but can we do this outside of a hardfork?

palango: Good point, I don't think we can.

Issue #2187, opened 2023-09-20
Feature request (ethapi): eth_getBlockReceipts


kamikazechaser:
This feature was released last week on geth (Therum).

- https://github.com/ethereum/go-ethereum/pull/27702

Benefits:

- No longer need to batch individual get receipt calls.

PR #2284, opened 2024-03-19
feat (internal/ethapi): implement eth_getBlockReceipts

kamikazechaser:
### Description

Addition of a new ETH RPC method `eth_getBlockReceipts`.

https://github.com/ethereum/execution-apis/pull/438

Standardized and implemented in several eth clients e.g. geth, nethermind, nimbus, besu e.t.c.

### Other changes

None.

### Tested

[`internal/ethapi/api.go`](https://app.codecov.io/gh/celo-org/celo-blockchain/blob/master/internal%2Fethapi%2Fapi.go) is not covered by any tests.

To be tested on a devnet/testnet with existing chaindata.

### Related issues

 #2187

### Backwards compatibility

Backwards compatible.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -331,6 +331,15 @@ func (ec *Client) TransactionReceipt(ctx context.Context, txHash common.Hash) (*
 	return r, err
 }
 
+func (ec *Client) BlockReceipts(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]*types.Receipt, error) {
+	var r []*types.Receipt
+	err := ec.c.CallContext(ctx, &r, "eth_getBlockReceipts", blockNrOrHash)
+	if err == nil && r == nil {
+		return nil, ethereum.NotFound
+	}
+	return r, err
+}
+
 type rpcProgress struct {
 	StartingBlock hexutil.Uint64
 	CurrentBlock  hexutil.Uint64
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -824,6 +824,33 @@ func (s *PublicBlockChainAPI) GetStorageAt(ctx context.Context, address common.A
 	return res[:], state.Error()
 }
 
+// GetBlockReceipts returns the block receipts for the given block hash or number or tag.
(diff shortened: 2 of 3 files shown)


ethclient/ethclient.go
```diff
@@ -331,6 +331,15 @@ func (ec *Client) TransactionReceipt(ctx context.Context, txHash common.Hash) (*
 	return r, err
 }
 
+func (ec *Client) BlockReceipts(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]*types.Receipt, error) {
```
palango, line 334: Please add a docstring here as well

palango (commented):
Looks good so far, please add the missing docstring.

palango:
For future reference, upstream PR: https://github.com/ethereum/go-ethereum/pull/27702

ethclient/ethclient.go
kamikazechaser: Done

palango:
@kamikazechaser I had another look and I think we need to add the effective gas price to the receipts as it is done in `GetTransactionReceipt`, otherwise the receipts will differ.

https://github.com/grassrootseconomics/celo-blockchain/blob/d77ad8e6f617d33f16ab521210f2ed8d703ce044/internal/ethapi/api.go#L1693-L1710

kamikazechaser:
@palango I have refactored the `generateReceiptResponse` helper to also accept a context and a backend interface. Tested on devnet.

I have also noticed there is an `GetBlockReceipt` method that is not used anywhere but also calls the generateReceiptResponse fn.

palango:
> @palango I have refactored the `generateReceiptResponse` helper to also accept a context and a backend interface. Tested on devnet.

Thanks, the changes look good. Needs a merge from master, where a new transaction type was added in the meantime, but the conflict is easy to resolve.

> I have also noticed there is an `GetBlockReceipt` method that is **not** a standardized API (added in #1628) that calls into `generateReceiptResponse`. because the `tx` passed by this method is nil, no effectiveGasPrice is assigned to that response. I would personally propose to deprecate that method and strictly follow the standardized spec. Anyways let me know what should be done for this case.

Yes, this is the RPC API to get receipts for celo system transactions. Not having the field added for them is fine.

kamikazechaser:
>  Not having the field added for them is fine.

Sounds good. I have also added the DenominatedFeeCurrency changes.

palango (approved):

PR #2290, opened 2024-03-25
fix(eth/tracers): wrong balance on prestate tracer

chuwt:
### Description

Fix the wrong balance on prestate_tracer.js

go-ethereum fix: [prestate_tracer](https://github.com/ethereum/go-ethereum/blob/14cc967d1964d3366252193cadd4bfcb4c927ac1/eth/tracers/js/internal/tracers/prestate_tracer_legacy.js#L65)


palango:
Hi @chuwt ,

thank you for opening this PR. The upstream fix that solves this problem seems to be https://github.com/ethereum/go-ethereum/commit/0f4942214d76385c52c7917212331c9bad8b3566, which includes more changes. Are those required for our codebase as well?

It would also be nice if you could add a test to see that the behaviour now matches the expectation.

chuwt:
Hi @palango, there are too many discrepancies in the trace section. I will close this PR and then prepare a new one specifically for the trace part.

PR #2292, opened 2024-04-08
Add CIP-66 receipt field feeInFeeCurrency

hbandura:
Add the receipt field for CELO denominated txs
Add e2etest for CELO denominated txs
Fix minor bugs in tx JSON marshalling


e2e_test/e2e_transfer_test.go
```diff
@@ -298,6 +301,20 @@ func TestTransferERC20(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
+		{
+			name: "Celo denominated transfer with fee currency",
```
carterqw2, line 305: Worth adding a test case where `MaxFeeInFeeCurrency` gets exceeded.

core/types/receipt.go
```diff
@@ -88,13 +92,29 @@ type receiptRLP struct {
 	CumulativeGasUsed uint64
 	Bloom             Bloom
 	Logs              []*Log
+	FeeInFeeCurrency  *big.Int
+}
+
+type preCIP66receiptRLP struct {
```
carterqw2, line 98: Maybe we should keep the old structures untouched for easier maintenance and use a specifically-typed structure for the new TX type.

e2e_test/e2e_transfer_test.go
hbandura: Agreed. I've now added one that's just -1 below what it should be for acceptance

core/types/receipt.go
hbandura: The receipt types are a bit confusing, technically here the old ones are untouched, and this one is only used for CELO denominated txs, as seen in the Receipts encode index (and decode as well).

But it's true that it feels like a weird change from this perspective

PR #2294, opened 2024-04-09
Palango/min go version

palango:
### Description

Update the go version in the modfile to match the minimal version required.


piersy:
Hey @palango, where is this minimal version defined?

piersy (approved):
All good

PR #2295, opened 2024-04-09
Fix eth_getBlockReceipts

### LINKED ISSUE #2293: eth_getBlockReceipts is broken, because it does not take into account the celo block receipt
## Expected Behavior

Getting the block receipts with cast should work

## Actual Behavior

```
cast rpc eth_getBlockReceipts `cast to-hex 53906`
Error:
(code: -32000, message: receipts length mismatch: 0 vs 1, data: None)
```

Solution:
 Ensure we handle the following edge case and any others in this function taking into account the celo block receipt:
https://github.com/celo-org/celo-blockchain/blob/6351612d67d4cf5e18708a0536f0cb43afa19e43/internal/ethapi/api.go#L840-L842

## System Information

Commit: 6351612d67d4cf5e18708a0536f0cb43afa19e43

piersy:
Updated the api method to take into account the celo block receipt.

There was previous confusion around whether this was working or not, as it had worked for some people but not others.

This confusion was caused because block receipts are only added when logs have been generated by system calls. Before gingerbread, we would update the gas price minimum on every block and hence create a log for the update, but post gingerbread the only updates that are happening are for epoch blocks.

### Tested

Manually tested using cast
```
cast rpc eth_getBlockReceipts `cast to-hex 53906`
```

### Related issues

- Fixes #2293
- Relates to #2284 #2187


palango (approved):
Logic makes sense, and I tested it locally on both epoch and non-epoch blocks.

PR #2298, opened 2024-04-12
Allow use of "finalized" tag in rpc requests

piersy:
This PR allows the use of the "finalized" tag to retrieve blocks or
headers via the RPC. Using the "finalized" tag is equivalent to using
the "latest" tag because all blocks are final in celo.

This is to supprt integration with chainlink since their system depends
on retrieving finalized blocks.

### Tested

There is a new e2e test that checks the "finalized" tag retrieves the same header as the "latest" tag.


rpc/types_test.go
```diff
@@ -87,18 +88,20 @@ func TestBlockNumberOrHash_UnmarshalJSON(t *testing.T) {
...
-		19: {`{"blockHash":"0x0000000000000000000000000000000000000000000000000000000000000000","requireCanonical":false}`, false, BlockNumberOrHashWithHash(common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000000"), false)},
-		20: {`{"blockHash":"0x0000000000000000000000000000000000000000000000000000000000000000","requireCanonical":true}`, false, BlockNumberOrHashWithHash(common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000000"), true)},
-		21: {`{"blockNumber":"0x1"}`, false, BlockNumberOrHashWithNumber(1)},
-		22: {`{"blockNumber":"pending"}`, false, BlockNumberOrHashWithNumber(PendingBlockNumber)},
-		23: {`{"blockNumber":"latest"}`, false, BlockNumberOrHashWithNumber(LatestBlockNumber)},
-		24: {`{"blockNumber":"earliest"}`, false, BlockNumberOrHashWithNumber(EarliestBlockNumber)},
-		25: {`{"blockNumber":"0x1", "blockHash":"0x0000000000000000000000000000000000000000000000000000000000000000"}`, true, BlockNumberOrHash{}},
+		14: {`"finalized"`, false, BlockNumberOrHashWithNumber(EarliestBlockNumber)},
```
carterqw2, line 91: nit: why `EasliestBlockNumber` here?

piersy: Not a nit, that's a copy paste error. Thanks!

rpc/types.go
```diff
@@ -106,6 +108,9 @@ func (bn *BlockNumber) UnmarshalJSON(data []byte) error {
 	case "pending":
 		*bn = PendingBlockNumber
 		return nil
+	case "finalized":
```
carterqw2, line 111: Looks like you forgot to add it to `func (bnh *BlockNumberOrHash) UnmarshalJSON(data []byte) error {`

piersy: Yes, its added now. Thanks!

carterqw2 (approved):

PR #2299, opened 2024-04-16
Remove exchange rate calculation for non CELO denominated txs in state_transition

hbandura:
Fix state_transition producing invalid gas used values for pre-H fork transactions by preventing from taking the exchange rate of fee currencies meant for only CELO denominated txs.

Rename some variables to be more specific about them applying only to this tx type.


core/state_transition.go
```diff
@@ -217,12 +217,12 @@ func IntrinsicGas(data []byte, accessList types.AccessList, isContractCreation b
...
 	if evm.ChainConfig().IsEspresso(evm.Context.BlockNumber) {
-		var feeCurrency *common.Address = msg.FeeCurrency()
 		if msg.MaxFeeInFeeCurrency() != nil {
 			// Celo denominated tx
-			feeCurrency = nil
+			gasPriceMinimum = sysCtx.GetGasPriceMinimum(nil)
+		} else {
+			gasPriceMinimum = sysCtx.GetGasPriceMinimum(msg.FeeCurrency())
 		}
-		gasPriceMinimum = sysCtx.GetGasPriceMinimum(feeCurrency)
```
piersy, line 225: Why is this necessary? This seems to be functionally unchanged by the change. I.E `sysCtx.GetGasPriceMinimum` is called once with either nil or a fee currency.

core/state_transition.go
```diff
@@ -378,16 +378,12 @@ func (st *StateTransition) debitFee(from common.Address, feeCurrency *common.Add
...
-			currencyFee = effectiveFee
+		if st.msg.MaxFeeInFeeCurrency() != nil {
+			st.erc20FeeDebited = denominatedCurrencyRate.FromBase(effectiveFee)
 		} else {
-			// Celo denominated tx
-			currencyFee = feeCurrencyRate.FromBase(effectiveFee)
+			st.erc20FeeDebited = effectiveFee
 		}
-		st.erc20FeeDebited = currencyFee
-		return erc20gas.DebitFees(st.evm, from, currencyFee, feeCurrency)
```
piersy, line 390: This block also seems to be functionally unchanged

core/state_transition.go
```diff
@@ -482,9 +478,14 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
-		return nil, err
+	var denominatedCurrencyRate *currency.ExchangeRate
+	// Get rate only for CELO denominated txs (H fork)
+	if st.msg.MaxFeeInFeeCurrency() != nil {
+		var err error
+		denominatedCurrencyRate, err = currency.GetExchangeRate(st.vmRunner, st.msg.FeeCurrency())
+		if err != nil {
+			return nil, err
+		}
 	}
```
piersy, line 489: So this is the actual fix right? Only setting denominatedCurrencyRate for celo denominated transactions.

core/state_transition.go
hbandura: It is uncchanged. But since the bad block I wanted to make sure everything was like before, so while this is a diff here, it makes it closer than what we had before the cip66 change in state_transition. The line 224 is exactly what we had, I wanted to preserve that line, even if, functionally, this block is the same

core/state_transition.go
hbandura: Just removing an unnecessary local variable.

core/state_transition.go
hbandura: Correct. The rest of the changes are: variable renames for (in my opinion) more explicit names, and changing a few ifblocks to be exactly like they were before the previous change (a mini revert if you will).

piersy (approved):
Hey @hbandura thanks for answering my questions. It's all looking good!

hbandura:
A full sync is currently running, may take a few days. When it finishes we can merge this

PR #2302, opened 2024-04-23
Omit ethCompatible on RPC responses for non legacy txs

piersy:
Our RPC responses for non legacy transactions (txType != 0) were returning the ethCompatible field set to false, which is confusing because some of those transactions are eth compatible. The ethCompatible flag was added only to allow distinguishing LegacyTransactions that were compatible with ethereum vs those that were not. This was needed because the LegacyTransaction struct was overloaded to handle celo legacy transactions (with feeCurrency, gatewayFee & gatewayFeeRecipient optionally set) and eth legacy transactions that wouldn't set those fields. The only thing that distinguished a legacy eth transaction from a legacy celo transaction that didn't set any of feeCurrency, gatewayFee & gatewayFeeRecipient would be the signature. So rather than requiring multiple signature verification attempts in order to determine the type of the transaction an extra field was added that would indicate if it was eth compatible.


karlb (commented):
The code looks fine and the change makes sense. I'm just not sure how and when we can release this, since it changes the external API.

arthurgousset (commented):
Thanks for the short pairing session earlier @piersy :) I learnt about "struct tags" in Go today ✅ 

One thing I mentioned on the call was that it might make sense to also omit `feeCurrency`, `gatewayFeeRecipient`, and `gatewayFee` from the response when they are empty (effectively on Ethereum-compatible transactions).

```diff
- FeeCurrency         *common.Address   `json:"feeCurrency"`
+ FeeCurrency         *common.Address   `json:"feeCurrency,omitempty"`
- GatewayFeeRecipient *common.Address   `json:"gatewayFeeRecipient"`
+ GatewayFeeRecipient *common.Address   `json:"gatewayFeeRecipient,omitempty"` 
- GatewayFee          *hexutil.Big      `json:"gatewayFee"`
+ GatewayFee          *hexutil.Big      `json:"gatewayFee,omitempty"`
```

Can I make a commit for that in this PR @piersy ? If not, I'll open an issue and raise a separate PR for that.

### Additional context

When I call `eth_gasTransactionByHash` on an EIP-1559 transaction, the response object includes `feeCurrency`, `gatewayFeeRecipient`, and `gatewayFee` although they aren't relevant.

For example:
1. Transaction hash: `0x04d8b58785bdd7199469489607e9f1e61be88ff115cb9c8f6b48b286effcda59`
2. [explorer.celo.org/alfajores/tx/0x04d8b](http://explorer.celo.org/alfajores/tx/0x04d8b)
4. [alfajores.celoscan.io/tx/0x04d](http://alfajores.celoscan.io/tx/0x04d)

```sh
$ curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0x04d8b58785bdd7199469489607e9f1e61be88ff115cb9c8f6b48b286effcda59"],"id":1}' "https://alfajores-forno.celo-testnet.org/" | jq

{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "blockHash": "0x260a8b8d0ab8fa4b71d9f2ef303b020d009f3f8ba167923f3cc3b3122a6ad6c4",
    "blockNumber": "0x169612a",
    "from": "0xae4ca7989e35e505948ea5faeb054f055567e50c",
    "gas": "0x8aa2",
    "gasPrice": "0x37e11d600",
    "maxFeePerGas": "0x6fc23ac00",
    "maxPriorityFeePerGas": "0x2540be400",
    "feeCurrency": null,
    "gatewayFeeRecipient": null,
    "gatewayFee": "0x0",
    "hash": "0x04d8b58785bdd7199469489607e9f1e61be88ff115cb9c8f6b48b286effcda59",
    "input": "0x",
    "nonce": "0xdb",
    "to": "0xae4ca7989e35e505948ea5faeb054f055567e50c",
    "transactionIndex": "0x0",
    "value": "0x1",
    "type": "0x2",
    "accessList": [],
    "chainId": "0xaef3",
    "v": "0x0",
    "r": "0xa3925714aabcf85921eb9a380de3e1243d112f0ccd0155941eb83c26ccc16eed",
    "s": "0x5111d0ffb7464b6ae1bcc192b86928b0f21ec1bc8cf6ed5bc7f2b4a017127e42",
    "ethCompatible": false
  }
}
```

piersy:
Hi @karlb I'm not sure what you mean, generally our changes go into master and at some point we make a release. What's different here?

piersy:
Hi @arthurgousset, that seems like a sensible change and in line with the change here, so feel free to add it to this PR. Thanks!

karlb:
> Hi @karlb I'm not sure what you mean, generally our changes go into master and at some point we make a release. What's different here?

I was wondering if we can just update forno after we release this or if we need some longer period for people to notice such a change, if we have any way to check if this breaks anyones code, etc. I guess it's fine and we expect people to notice on the testnet before going to mainnet or read the changelog carefully on release.

karlb (approved):

arthurgousset:
Looks like the "[End-to-end transfers test](https://github.com/celo-org/celo-blockchain/actions/runs/8815328035/job/24197076173?pr=2302)" and "[Unit tests](https://github.com/celo-org/celo-blockchain/actions/runs/8815328035/job/24197077031?pr=2302)" are failing :(

karlb:
The tests are broken by https://github.com/celo-org/celo-monorepo/blob/f08e269d88c6680641427952ce4743a7d26336a6/packages/celotool/src/e2e-tests/transfer_tests.ts#L494 and probably other similar checks. I'll take care of it.

karlb (approved):

PR #2322, opened 2024-07-31
Stop Celo L1 at a specific block

alecps:
Closes https://github.com/celo-org/celo-blockchain-planning/issues/419

Changes:

- '--l2migrationblock' flag added to geth cmd. This specifies what will be the first block of the l2 network, and the block after the last block of the l2 network. This does not affect the forkid or hardfork config.

- 3 checks are added to stop block production, insertion and communication
1. The `mainLoop` in `miner/worker.go` is updated to check if the next block is the l2-migration-block every time it starts generating a new block. When the l2-migration-block is reached, it the worker stops itself. This halts block production. 
2. `writeHeadBlock` in `core/blockchain.go` is updated to check if the block it is writing is >= the l2-migation-block before it does anything else, and also whether the block it just wrote is >= l2-migration-block - 1 at the end of the function. If either of these checks pass `StopInsert()` is called. This prevents any more blocks from being inserted and stops all processes related to block insertion. Normally the check at the end of the function will pass first, but if a node is restarted with the same l2-migration-block configured after already reaching and stopping on l2-migration block, the check at the beginning of `writeHeadBlock` will pass first and log an error.`InsertChain` will return an error if block insertion is stopped during its execution. 
3. `Commit` in `consensus/istanbulbackend.go` is updated to check if the block that was just committed is the block before the l2-migration-block right before it returns. If the check passes then`StopAnnounce()` and 'Close()` are called. This stops blocks from being shared between nodes via the istanbul announce protocol and frees up associated resources. 

Other changes:
- Adds e2e test support for networks where some nodes are full nodes (not validating)
- Updates logging to display latest blockNumber when blockchain is stopped

Tested with e2e tests
- `go test -v ./e2e_test -run TestStopNetworkAtL2BlockSimple -count 1000 -timeout 1h` 
- `go test -v ./e2e_test -run TestStopNetworkAtL2Block` 

Tested on Alfajores: 
- Use the following launch.json config
`{
            "name": "Launch geth afaljores",
            "type": "go",
            "request": "launch",
            "mode": "auto",
            "program": "./cmd/geth",
            "args": [
                "--alfajores",
                "--datadir", "<DATADIR>",
                "--http",
                "--http.api", "eth,net,web3,debug,txpool", 
                "--syncmode", "full",
                "--verbosity", "3",
                "--light.serve", "90",
                "--light.maxpeers", "1000",
                "--maxpeers", "1100",
                "--l2migrationblock", "<MIGRATION_BLOCK>",
            ]
        }`
        
- Or run with cli command
`./build/bin/geth --alfajores --datadir <DATADIR> --http --http.api eth,net,web3,debug,txpool --syncmode full --verbosity 3 --light.serve 90 --light.maxpeers 1000 --maxpeers 1100 --port 30304 --l2migrationblock <MIGRATION_BLOCK>`

Check on syncing with 

`./build/bin/geth attach <DATADIR>/geth.ipc --exec "eth.syncing"`

and 

`./build/bin/geth attach <DATADIR>/geth.ipc --exec "eth.blockNumber"`

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -320,7 +320,7 @@ func prepare(ctx *cli.Context) {
 // geth is the main entry point into the system if no special subcommand is ran.
 // It creates a default node based on the command line arguments and runs it in
 // blocking mode, waiting for it to be shut down.
-func geth(ctx *cli.Context) error { // TODO(Alec) code pointer
+func geth(ctx *cli.Context) error {
 	if args := ctx.Args(); len(args) > 0 {
 		return fmt.Errorf("invalid command: %q", args[0])
 	}
@@ -415,7 +415,7 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend) {
 				if timestamp := time.Unix(int64(latest.Time), 0); time.Since(timestamp) < 10*time.Minute {
 					log.Info("Synchronisation completed", "latestnum", latest.Number, "latesthash", latest.Hash(),
 						"age", common.PrettyAge(timestamp))
-					stack.Close() // TODO(Alec) code pointer
+					stack.Close()
 				}
 			}
 		}()
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/celo-org/celo-blockchain/cmd/utils"
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/console/prompt"
+	ethCore "github.com/celo-org/celo-blockchain/core"
 	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/eth"
 	"github.com/celo-org/celo-blockchain/eth/downloader"
@@ -327,6 +328,11 @@ func geth(ctx *cli.Context) error {
 
 	prepare(ctx)
 	stack, backend := makeFullNode(ctx)
+
+	if backend.ChainConfig().IsL2Migration(backend.CurrentBlock().Number()) {
+		return fmt.Errorf("Attempted to start node with an l2 migration block that has already been reached. latestBlock: %d, hash: %s", backend.CurrentBlock().NumberU64(), backend.CurrentBlock().Hash())
+	}
+
 	defer stack.Close()
 
 	startNode(ctx, stack, backend)
@@ -416,6 +422,30 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend) {
 					log.Info("Synchronisation completed", "latestnum", latest.Number, "latesthash", latest.Hash(),
 						"age", common.PrettyAge(timestamp))
 					stack.Close()
+				}
+			}
+		}()
+	}
+
+	// Spawn a standalone goroutine to monitor the chain head for the l2 migration block.
+	// When the l2 migration block is reached, close the entire stack.
+	if ctx.GlobalIsSet(utils.L2MigrationBlockFlag.Name) && ctx.GlobalUint64(utils.L2MigrationBlockFlag.Name) > 0 {
+		go func() {
+			chainHeadCh := make(chan ethCore.ChainHeadEvent, 10)
+			chainHeadSub := backend.SubscribeChainHeadEvent(chainHeadCh)
+			defer chainHeadSub.Unsubscribe()
+			for {
+				select {
+				case chainHeadEvent := <-chainHeadCh:
+					block := chainHeadEvent.Block
+					if backend.ChainConfig().IsL2Migration(block.Number()) {
+						log.Info("L2 Migration Block Reached, closing entire stack from geth cmd", "block", block.NumberU64(), "hash", block.Hash())
+						stack.Close()
+						return
+					}
+				case err := <-chainHeadSub.Err():
+					log.Error("Error in outer subscription to blockchain's chainhead event listening for l2 migration block", "err", err)
+					return
 				}
 			}
 		}()
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -390,6 +390,28 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend) {
 		}
 	}()
 
+	// if ctx.GlobalIsSet(utils.L2ForkFlag.Name) {
+	// 	go func() {
+	// 		sub := stack.EventMux().Subscribe(ethCore.ChainHeadEvent{})
+	// 		defer sub.Unsubscribe()
+	// 		for {
+	// 			event := <-sub.Chan()
+	// 			if event == nil {
+	// 				continue
+	// 			}
+	// 			var latest *types.Header
+	// 			if done, ok := event.Data.(ethCore.ChainHeadEvent); ok {
+	// 				latest = done.Block.Header()
+	// 			}
+
+	// 			if ctx.GlobalUint64(utils.L2ForkFlag.Name) >= latest.Number.Uint64() {
+	// 				log.Info("L2 Migration Block Reached", "latest", latest.Number.Uint64(), "latesthash", latest.Hash())
+	// 				stack.Close()
+	// 			}
+	// 		}
+	// 	}()
+	// }
+
 	// Spawn a standalone goroutine for status synchronization monitoring,
 	// close the node when synchronization is complete if user required.
 	if ctx.GlobalBool(utils.ExitWhenSyncedFlag.Name) {
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -30,7 +30,6 @@ import (
 	"github.com/celo-org/celo-blockchain/cmd/utils"
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/console/prompt"
-	ethCore "github.com/celo-org/celo-blockchain/core"
 	"github.com/celo-org/celo-blockchain/core/types"
 	"github.com/celo-org/celo-blockchain/eth"
 	"github.com/celo-org/celo-blockchain/eth/downloader"
@@ -329,10 +328,6 @@ func geth(ctx *cli.Context) error {
 	prepare(ctx)
 	stack, backend := makeFullNode(ctx)
 
-	if backend.ChainConfig().IsL2Migration(backend.CurrentBlock().Number()) {
-		return fmt.Errorf("Attempted to start node with an l2 migration block that has already been reached. latestBlock: %d, hash: %s", backend.CurrentBlock().NumberU64(), backend.CurrentBlock().Hash())
-	}
-
 	defer stack.Close()
 
 	startNode(ctx, stack, backend)
@@ -422,30 +417,6 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend) {
 					log.Info("Synchronisation completed", "latestnum", latest.Number, "latesthash", latest.Hash(),
 						"age", common.PrettyAge(timestamp))
 					stack.Close()
-				}
-			}
-		}()
-	}
-
-	// Spawn a standalone goroutine to monitor the chain head for the l2 migration block.
-	// When the l2 migration block is reached, close the entire stack.
-	if ctx.GlobalIsSet(utils.L2MigrationBlockFlag.Name) && ctx.GlobalUint64(utils.L2MigrationBlockFlag.Name) > 0 {
-		go func() {
-			chainHeadCh := make(chan ethCore.ChainHeadEvent, 10)
-			chainHeadSub := backend.SubscribeChainHeadEvent(chainHeadCh)
-			defer chainHeadSub.Unsubscribe()
-			for {
-				select {
-				case chainHeadEvent := <-chainHeadCh:
-					block := chainHeadEvent.Block
-					if backend.ChainConfig().IsL2Migration(block.Number()) {
-						log.Info("L2 Migration Block Reached, closing entire stack from geth cmd", "block", block.NumberU64(), "hash", block.Hash())
-						stack.Close()
-						return
-					}
-				case err := <-chainHeadSub.Err():
-					log.Error("Error in outer subscription to blockchain's chainhead event listening for l2 migration block", "err", err)
-					return
 				}
 			}
 		}()
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -319,7 +319,7 @@ func prepare(ctx *cli.Context) {
 // geth is the main entry point into the system if no special subcommand is ran.
 // It creates a default node based on the command line arguments and runs it in
 // blocking mode, waiting for it to be shut down.
-func geth(ctx *cli.Context) error {
+func geth(ctx *cli.Context) error { // TODO(Alec) code pointer
 	if args := ctx.Args(); len(args) > 0 {
 		return fmt.Errorf("invalid command: %q", args[0])
 	}
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -74,7 +74,7 @@ var (
 		utils.USBFlag,
 		// utils.SmartCardDaemonPathFlag,
 		utils.OverrideHForkFlag,
-		utils.L2ForkFlag,
+		utils.L2MigrationBlockFlag,
 		utils.TxPoolLocalsFlag,
 		utils.TxPoolNoLocalsFlag,
 		utils.TxPoolJournalFlag,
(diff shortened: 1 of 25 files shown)


eth/handler.go
```diff
@@ -593,6 +608,25 @@ func (h *handler) txBroadcastLoop() {
...
+			block := event.Block
+			if h.chain.Config().IsL2Migration(block.Number()) {
+				log.Info("L2 Migration Block Reached, stopping handler and p2p server", "block", block.NumberU64(), "hash", block.Hash())
+				h.wg.Done()
+				h.Stop()
+				h.server.Stop()
+				return
+			}
+		case <-h.newChainHeadSub.Err():
+			h.wg.Done()
```
ezdac, line 624: Nitpick I randomly noticed: you could put it in a `defer` before the loop instead of all the return cases so that you don't forget to call this when a new return case gets added.

alecps: Yea good catch and thanks for taking a look! This was actually done on purpose after I discovered that if we use defer for `h.wg.Done()` like in the other handler loops then we get a deadlock [here](https://github.com/celo-org/celo-blockchain/blob/e0fc697ca4f2cfe04618d9b337477c31bfe31f37/eth/handler.go#L494) inside `h.Stop()` because the defer statement hasn't been called yet. I'll add a comment to explain this

cmd/geth/main.go
```diff
@@ -420,6 +427,30 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend) {
 		}()
 	}
 
+	// Spawn a standalone goroutine to monitor the chain head for the l2 migration block.
+	// When the l2 migration block is reached, close the entire stack.
```
karlb, line 431: Should we use a goroutine? Wouldn't the result be more predictable if we executed this without spawning s new goroutine?

cmd/geth/main.go
```diff
@@ -326,6 +328,11 @@ func geth(ctx *cli.Context) error {
 
 	prepare(ctx)
 	stack, backend := makeFullNode(ctx)
+
+	if backend.ChainConfig().IsL2Migration(backend.CurrentBlock().Number()) {
```
mcortesi, line 332: not sure if this would actually work...

when you start a node are you're not synced or out of sync... you don't know the block number

mcortesi (commented):
All in all, i would change a few things here.

First it is important to differentiate between two asks:
 1. Feature is node should not produce (or accept) block after L2 Block
 2. Feature is node should STOP after L2 block
 
 To me, initially the feature would be (1).
 
 As impl plant than means only 2 places to change:
 
  1. Not Produce => modify worker create block block to STOP once it reaches a certain number
  2. Not accept => Modify "Blockchain.Insert()" and other insert methods on blockchain object to FAIL after L2 block is reached.
  
 What it doesn't mean is triggering the whole shtudown mechanism

params/config.go
```diff
@@ -464,6 +465,11 @@ func (c *ChainConfig) IsGingerbreadP2(num *big.Int) bool {
 	return isForked(c.GingerbreadP2Block, num)
 }
 
+// IsL2 returns whether num represents a block number greater than or equal to the L2 migration block (last block before L2)
+func (c *ChainConfig) IsL2Migration(num *big.Int) bool {
```
mcortesi, line 469: would be important to specify for users where L2MigrationBlock is:

 * The last block of the L1, 
 or
 * the first block of the L2
 
Here the comment says "Last block of L1 => L2Block -1", but the check says otherwise.

core/blockchain.go
```diff
@@ -811,6 +811,10 @@ func (bc *BlockChain) ExportN(w io.Writer, first uint64, last uint64) error {
 //
 // Note, this function assumes that the `mu` mutex is held!
 func (bc *BlockChain) writeHeadBlock(block *types.Block) {
+	if bc.Config().IsL2Migration(new(big.Int).Sub(block.Number(), big.NewInt(1))) {
```
mcortesi, line 814: i feel there are better places to do this check, and return an error (vs. just log Crit which i'm not sure if it does not have the side effect of killing the process)

I remeber there are 2 entry point to add block here.

 1. As part of sync process. Blockchain.Insert()
 2. As part of consensus validators use another method that bypass the first pass of Blockchain, but it is a public method.

cmd/geth/main.go
alecps: removed this

params/config.go
alecps: Updated the PR so that l2MigrationBlock refers to first block of L2 (this is different than before) as it seems cleaner. Updated comments and help strings to clarify this

cmd/geth/main.go
alecps: Removed this change

core/blockchain.go
alecps: I removed the log.Crit here, still looking into whether there's a better place to put the check

alecps: There seem to be many code paths that can call [writeHeadBlock](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L813) including
[InsertPreprocessedBlock](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L1479)
[reorg](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L2206) (shouldn’t be relevant for Celo as I understand it)
[InsertChain](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L1714)
[InsertChainWithoutSealVerification](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L1746)
[insertSideChain](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L2063)
My reasoning behind calling bc.StopInsert() in writeHeadBlock is just that it seems to be called every time we add a new head block regardless of the code path and so feels safer than adding the check everywhere else. From your comment it seemed you may have been thinking [insertChain](https://github.com/celo-org/celo-blockchain/blob/02d8916d201dde648c4ea1650b7c3c9cfe5b8c89/core/blockchain.go#L1765) would be a good spot to add the check?

alecps: I've added a test to check that InsertChain behaves as expected and returns an error when inserting large chains containing the migration block. 

Given that StopInsert() can be called multiple times I'm not worried about calling it in a private method, and this approach avoids changes in all the functions that call `writeHeadBlock`. Given I'm a bit new to the code, it feels safer to put the checks here so that it's easy to reason about when StopInsert() gets called, and so it gets called as early as possible after the last l1 block is added. 

There may be something I'm missing so please let me know if you still prefer a different spot for the checks!

e2e_test/e2e_test.go
```diff
@@ -369,6 +370,125 @@ func TestStartStopValidators(t *testing.T) {
...
+		go func(n *test.Node) {
+			defer wg.Done()
+			err = n.Tracker.AwaitBlock(shortCtx, l2Block.Uint64())
+			require.EqualError(t, err, context.DeadlineExceeded.Error())
+		}(n)
+	}
+	wg.Wait()
+}
+
+func TestStopNetworkAtl2BlockSimple(t *testing.T) {
```
piersy, line 393:
```suggestion
func TestStopNetworkAtL2BlockSimple(t *testing.T) {
```

e2e_test/e2e_test.go
```diff
@@ -369,6 +370,125 @@ func TestStartStopValidators(t *testing.T) {
...
+	shortCtx, cancel := context.WithTimeout(ctx, time.Second*2)
+	defer cancel()
+
+	// fail if any node adds a new block >= the migration block
+	var wg sync.WaitGroup
+	for _, n := range network {
+		wg.Add(1)
+		go func(n *test.Node) {
+			defer wg.Done()
+			err = n.Tracker.AwaitBlock(shortCtx, l2Block.Uint64())
```
piersy, line 386:
```suggestion
			err := n.Tracker.AwaitBlock(shortCtx, l2Block.Uint64())
```
There was a data race on this error variable, since it was being accessed concurrently without any synchronisation, redeclaring it here solves the issue.

test/node.go
```diff
@@ -440,6 +500,47 @@ func NewNetwork(accounts *env.AccountsConfig, gc *genesis.Config, ec *eth.Config
...
+		for _, err := range network.Shutdown() {
+			fmt.Println(err.Error())
+		}
+	}
+
+	return network, shutdown, nil
+}
+
+// AddNonValidatorNodes Adds non-validator nodes to the network with the specified sync mode.
+func AddNonValidatorNodes(network Network, ec *eth.Config, numNodes uint64, syncmode downloader.SyncMode) (Network, func(), error) {
```
piersy, line 513: Loving the approach to adding non validator nodes, very neat (apart from the internal duplication but I'm sure that could be easily solved later if we want to) 👌

core/blockchain.go
```diff
@@ -837,6 +845,13 @@ func (bc *BlockChain) writeHeadBlock(block *types.Block) {
 	}
 	bc.currentBlock.Store(block)
 	headBlockGauge.Update(int64(block.NumberU64()))
+
+	nextBlockNum := new(big.Int).Add(block.Number(), big.NewInt(1))
+	if bc.Config().IsL2Migration(nextBlockNum) {
+		log.Info("The next block is the L2 migration block, stopping block insertion", "currentBlock", block.NumberU64(), "hash", block.Hash(), "nextBlock", nextBlockNum.Uint64())
+		bc.StopInsert()
+		return
+	}
```
piersy, line 854: I don't think we need this, the check up front should be sufficient. Sopping insertion isn't really important unless there is an attempt to insert the next block.

core/blockchain.go
```diff
@@ -811,6 +811,14 @@ func (bc *BlockChain) ExportN(w io.Writer, first uint64, last uint64) error {
 //
 // Note, this function assumes that the `mu` mutex is held!
 func (bc *BlockChain) writeHeadBlock(block *types.Block) {
```
piersy, line 813: There's also `core.HeaderChain.writeHeaders` where `rawdb.WriteCanonicalHash` is called. I'm just wondering if this could cause any issues. For example we may sync a header later than the stop block.

core/blockchain.go
```diff
@@ -811,6 +811,14 @@ func (bc *BlockChain) ExportN(w io.Writer, first uint64, last uint64) error {
...
+	// Normally the check at the end of the function will pass first, but if a node is restarted
+	// with the same l2-migration-block configured after already reaching and stopping on l2-migration-block,
+	// this check will pass first and log an error.
+	if bc.Config().IsL2Migration(block.Number()) {
+		log.Error("Attempt to insert block number >= l2MigrationBlock, stopping block insertion", "block", block.NumberU64(), "hash", block.Hash())
+		bc.StopInsert()
```
piersy, line 819: I don't think we should need to call StopInsert here since this check will effectively prevent any additional blocks being added.

consensus/istanbul/backend/backend.go
```diff
@@ -554,6 +554,13 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, aggregatedSeal types.Istan
 		}
 	}
 	sb.onNewConsensusBlock(block, result.Receipts, result.Logs, result.State)
+
+	nextBlockNum := new(big.Int).Add(block.Number(), big.NewInt(1))
+	if sb.chain.Config().IsL2Migration(nextBlockNum) {
+		sb.logger.Info("The next block is the L2 migration block, stopping announce protocol and closing istanbul backend", "currentBlock", block.NumberU64(), "hash", block.Hash(), "nextBlock", nextBlockNum)
+		sb.StopAnnouncing()
+		sb.Close()
```
piersy, line 562: I don't think this is safe to call from here since Close is also called from other threads, it might be enough to add a lock inside Close so that only one thread can be executing in there at any time. Needs a bit more thought though.

miner/worker.go
```diff
@@ -239,10 +241,12 @@ func (w *worker) start() {
 func (w *worker) stop() {
 	atomic.StoreInt32(&w.running, 0)
 
-	if istanbul, ok := w.engine.(consensus.Istanbul); ok {
-		err := istanbul.StopValidating()
-		if err != nil {
-			log.Error("Error while calling engine.StopValidating", "err", err)
+	if istanbul, ok := w.engine.(*istanbulBackend.Backend); ok {
+		if istanbul.IsValidating() {
```
piersy, line 245: Why the added condition here?

consensus/istanbul/backend/backend.go
piersy: Do we even need to call Close though?

e2e_test/e2e_test.go
```diff
@@ -369,6 +370,125 @@ func TestStartStopValidators(t *testing.T) {
...
+			require.EqualError(t, err, context.DeadlineExceeded.Error())
```
piersy, line 387: `T.FailNow` which is ultimately called by all `require.XXXX` only functions correctly when called from the main test goroutine. Essentially any assertions you do in tests need to be made in the main goroutine, so this is not going to work correctly. Instead you can use the `errgroup` package in place of the waitgroup and do your error assertion at the wait point.

alecps: Thanks! I was not aware of that. I looked into using errgroup for this but then remembered that we actually want to wait for all the nodes to time out here, and errgroup will only wait until the first error is returned. I ended up just using a channel to pass the error back to the main go routine https://github.com/celo-org/celo-blockchain/pull/2322/commits/c7bea3217e78c76eb96958ccfd27592f99afaba4

miner/worker.go
alecps: So the [miner.Stop()](https://github.com/celo-org/celo-blockchain/blob/c7bea3217e78c76eb96958ccfd27592f99afaba4/eth/backend.go#L618) function gets called when the user (or the e2e test) actually shuts down the node, in which case istanbul.StopValidating() throws an error here since it's already been called (in the code below)

core/blockchain.go
alecps: When I remove this the long test actually starts failing. I was under the impression bc.StopInsert() was essential to call based on what I saw during testing. The function flips an atomic value that is checked at the beginning of numerous functions related to block insertion.

core/blockchain.go
alecps: See comment below on StopInsert()

core/blockchain.go
alecps: writeHeaders checks something called procInterrupt(), which wraps a call to insertStopped(), which is injected when NewHeaderChain() is called. When we call StopInsert(), it makes procInterrupt() return true in writeHeaders before it updates the head header references etc. 
My understanding is that by calling bc.StopInsert() we prevent the HeaderChain code from continuing to update the head header.

consensus/istanbul/backend/backend.go
alecps: It looks like calling Close() may be a bit overkill actually. StopAnnouncing() stops the announce protocol and it looks like Close() basically prevents it from being easily restarted with StartAnnouncing(). I removed it and the e2e test still passed so I'll remove it from the PR

core/blockchain.go
karlb: I wasn't sure why we need changes to `writeHeadBlock` at all. So I put a panic in for the `IsL2Migration` migration case to see when the check is relevant. The panic was triggered by the block fetcher.

It might be more elegant to discard the blocks when the block fetcher receives than rather then preventing the write here. Suggested commit: https://github.com/celo-org/celo-blockchain/pull/2330/commits/90dbc6cbd73b311f7472e190e1d025632f0e8f84

I still don't have the full picture, so let me know if I'm missing something.

miner/worker.go
piersy: Makes sense, but this is introducing a race condition, since you could still have 2 goroutines calling stop. But  it might be good enough for our purposes if the time between the 2 calls to stop is sufficient.

core/blockchain.go
piersy: @karlb that looks pretty neat to me, and the tests seem to work, so I'd be in favour of this approach.

piersy: Ok I've pushed a couple of updates to karlb/stop and removed basically everything except the extra flag, the rejection in the validate function and the e2e test. Tests seem to be working.

piersy:
Closed in favour of https://github.com/celo-org/celo-blockchain/pull/2330

PR #2327, opened 2024-08-21
chore: remove repetitive words

shandongzhejiang:
### Description

remove repetitive words

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


palango (approved):
Thanks @shandongzhejiang

PR #2330, opened 2024-09-17
Karlb/stop

karlb:
Try alternative approaches to https://github.com/celo-org/celo-blockchain/pull/2322

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/celo-org/celo-blockchain/common"
 	"github.com/celo-org/celo-blockchain/consensus"
+	istanbulBackend "github.com/celo-org/celo-blockchain/consensus/istanbul/backend"
 	"github.com/celo-org/celo-blockchain/core"
 	"github.com/celo-org/celo-blockchain/core/state"
 	"github.com/celo-org/celo-blockchain/core/types"
@@ -239,10 +240,12 @@ func (w *worker) start() {
 func (w *worker) stop() {
 	atomic.StoreInt32(&w.running, 0)
 
-	if istanbul, ok := w.engine.(consensus.Istanbul); ok {
-		err := istanbul.StopValidating()
-		if err != nil {
-			log.Error("Error while calling engine.StopValidating", "err", err)
+	if istanbul, ok := w.engine.(*istanbulBackend.Backend); ok {
+		if istanbul.IsValidating() {
+			err := istanbul.StopValidating()
+			if err != nil {
+				log.Error("Error while calling engine.StopValidating", "err", err)
+			}
 		}
 	}
 }
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -18,6 +18,7 @@ package miner
 
 import (
 	"context"
+	"math/big"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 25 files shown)


miner/worker.go
```diff
@@ -396,19 +392,10 @@ func (w *worker) mainLoop() {
...
 			cancel()
 		}
-
-		if w.chainConfig.IsL2Migration(nextBlockNum) {
-			if w.isRunning() {
-				log.Info("The next block is the L2 migration block, stopping block construction", "currentBlock", w.chain.CurrentBlock().NumberU64(), "hash", w.chain.CurrentBlock().Hash(), "nextBlock", nextBlockNum.Uint64())
-				w.stop()
-			}
-			return
-		}
```
karlb: With this removed, could the local node produce blocks after the migration? Or are those also checked by the block fetcher header validity check?

We also wouldn't stop the worker anymore. Is that a problem (resource usage, irritating logs or rpc responses)?

piersy: I don't think the local node can produce blocks any more, because if you look at consensus we can see that `sendPreprepareV2` is the point at which the proposer sends the initial preprepare message.

```
func (c *core) sendPreprepareV2(request *istanbul.Request, roundChangeCertificateV2 istanbul.RoundChangeCertificateV2) {
	logger := c.newLogger("func", "sendPreprepareV2")

	// If I'm the proposer and I have the same sequence with the proposal
	if c.current.Sequence().Cmp(request.Proposal.Number()) == 0 && c.isProposer() {
		m := istanbul.NewPreprepareV2Message(&istanbul.PreprepareV2{
			View:                     c.current.View(),
			Proposal:                 request.Proposal,
			RoundChangeCertificateV2: roundChangeCertificateV2,
		}, c.address)
		logger.Debug("Sending preprepareV2", "m", m)
		c.broadcast(m)
	}
}

```

If you dig down into the `broadcast`  function you can see that the preprepare message eventually gets sent back to the proposer, for it to handle in the same way as all the other messages eventually arriving at `handlePreprepareV2` which calls `verifyProposal` on the block.


So to summarise consensus nodes, including proposers would not be able to progress past the preprepare phase of consensus if the block does not pass validation, and therefore not be able to produce new blocks.


>  We also wouldn't stop the worker anymore. Is that a problem (resource usage, irritating logs or rpc responses)?

I think that's a benefit, because we don't end up messing about with the internal stopping mechanisms.

karlb: Sounds good, thanks for the explanation!

piersy (approved):
Well, I approve, but these are partly my changes so maybe @palango or @karlb could take a look

karlb (commented):
Looks good to me. The test seems to cover the important cases (and has proven useful when trying different approaches) and the actual changes (ignoring the flag passing) are really small.

mcortesi:
So @piersy @karlb the approach here is just to change `Engine.verifyHeader()` to fail if header is for a L2 block?

Does this affect both flows:
 1. Syncing flow: No full node will accept a block of L2
 2.  Consensus Flow: There won't be consensus on a block of L2

And for (2). What happens instead? Proposer keep trying to produce blocks and failing? Then comes a new proposer, and so on... is this some kind of "busy wait"?

piersy:
Hey @mcortesi 

> here is just to change `Engine.verifyHeader()` to fail if header is for a L2 block?

Yes.

>Does this affect both flows:

>Syncing flow: No full node will accept a block of L2
>Consensus Flow: There won't be consensus on a block of L2

As far as I can see this would prevent any fetching of L2 blocks and and from my previous investigation it seems that it also prevents creation of L2 blocks.

> And for (2). What happens instead? Proposer keep trying to produce blocks and failing? Then comes a new proposer, and so on... is this some kind of "busy wait"?

Yep, but probably not that "busy" because the istanbul timeouts will increase exponentially.

alecps:
This looks great! Much cleaner than https://github.com/celo-org/celo-blockchain/pull/2322. Thanks @karlb @piersy

Issue #2318, opened 2024-07-17
Validator experiencing a panic


piersy:
Validator raised this issue on discord - https://discord.com/channels/600834479145353243/601046166620078100/1262163359428775967

## Expected Behavior

No panics

## Actual Behavior

A panic

## Steps to reproduce the behavior
Using one of the following images:
us.gcr.io/celo-org/geth:1.8.2/1.8.4

Running a backup validator with the following args:
```
      --verbosity 3 
      --syncmode full 
      --etherbase $CELO_VALIDATOR_SIGNER_ADDRESS 
      --nodiscover 
      --nousb
      --mine 
      --proxy.proxied 
      --proxy.proxyenodeurlpairs=enode://$PROXY_ENODE02@$PROXY_EXTERNAL_IP02:30503\;enode://$PROXY_ENODE02@$PROXY_EXTERNAL_IP02:30303 
      --unlock=$CELO_VALIDATOR_SIGNER_ADDRESS 
      --password /root/.celo/.password 
      --celostats=stakely.io@stats-server.celo.org
      --light.serve 0
      --datadir /root/.celo
```

Syncing on top of  a downloaded datadir, the latest at the 14th July.

## Backtrace

```

 WARN [07-14|20:52:41.199] Error in retrieving exchange rate.  Will do comparison of two values without exchange rate conversion. currency1=0x2F25deB3848C207fc8E0c34035B3Ba7fC157602B err1="numerator zero: Exchange rate returned from the network is zero" currency2=nil err2=nil
WARN [07-14|20:52:41.199] Error in retrieving exchange rate.  Will do comparison of two values without exchange rate conversion. currency1=nil err1=nil currency2=0x2F25deB3848C207fc8E0c34035B3Ba7fC157602B err2="numerator zero: Exchange rate returned from the network is zero"
 ERROR[07-14|20:52:41.200] Error invoking evm function: can't unpack result to=0x2F25deB3848C207fc8E0c34035B3Ba7fC157602B method=balanceOf err="abi: attempting to unmarshall an empty string while arguments are expected" maxgas=100,000
 ERROR[07-14|20:52:41.200] GetBalanceOf evm invocation error        err="abi: attempting to unmarshall an empty string while arguments are expected"
 panic: runtime error: invalid memory address or nil pointer dereference
 [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5ce2a5]
 
 goroutine 159592 [running]:
[...]

PR #2333, opened 2024-11-01
Test concurrent feecurrency tx handling in tx pool

piersy:
Adds a test to verify that the tx pool is executing concurrent fee currency checks safely.

I created this to see if I could reproduce https://github.com/celo-org/celo-blockchain/issues/2318 but I was unable to reproduce that issue. It did however uncover a different race condition, which was that `core.current` was being set in one goroutine and accessed from others without synchronisation. This was easy to fix by just locking around the point where it is set.

### DIFF
--- a/consensus/istanbul/core/handler.go
+++ b/consensus/istanbul/core/handler.go
@@ -36,7 +36,9 @@ func (c *core) Start() error {
 		return err
 	}
 
+	c.currentMu.Lock()
 	c.current = roundState
+	c.currentMu.Unlock()
 	c.roundChangeSetV2 = newRoundChangeSetV2(c.current.ValidatorSet())
 
 	// Reset the Round Change timer for the current round to timeout.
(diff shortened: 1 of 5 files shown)


gastonponti (approved):
LGTM
The only doubt I have, is that you are sending the txs from the same N accounts, and I thought that every txpool had a maximum amount of txs per account in a pending state, so I would have guess that it was going to fail due to that limitation, but it seems that works, so maybe was configured to not have a cap before?

piersy:
@gastonponti Good point, that is strange there is a limit (see below), and the default which is used in the test is 64, so I'd really expect this test to at least be flaky if not fail all the time, but even running with a block time of 20 seconds, which means all txs will have been sent and in the tx pool before a block is constructed (300 txs per account) it still passes!

So I guess there is some understanding we are missing.

https://github.com/celo-org/celo-blockchain/blob/ff274ca79238969ec94d23a0a5e61170f783f867/core/tx_pool.go#L193

palango:
@piersy Do we still want to merge this?

PR #2339, opened 2024-12-04
config: Add L2 migration config for testnets

palango:
Adds the L2 migration config for Alfajores and Baklava testnets.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -97,7 +97,7 @@ var (
 		GingerbreadBlock:    big.NewInt(18785000),
 		GingerbreadP2Block:  big.NewInt(19157000),
 		HForkBlock:          nil, // TBD
-		L2MigrationBlock:    big.NewInt(27109999),
+		L2MigrationBlock:    big.NewInt(27110000),
 
 		Istanbul: &IstanbulConfig{
 			Epoch:          17280,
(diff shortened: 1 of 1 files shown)


params/config.go
```diff
@@ -97,6 +97,7 @@ var (
 		GingerbreadBlock:    big.NewInt(18785000),
 		GingerbreadP2Block:  big.NewInt(19157000),
 		HForkBlock:          nil, // TBD
+		L2MigrationBlock:    big.NewInt(27109999),
```
gastonponti, line 100: The L2MigrationBlock is the one that is not going to generate. In this case, if we set the 9999, we will have a gap of 1 block between the L1 and L2
```suggestion
		L2MigrationBlock:    big.NewInt(27110000),
```

params/config.go
```diff
@@ -126,6 +127,7 @@ var (
 		GingerbreadBlock:    big.NewInt(19814000),
 		GingerbreadP2Block:  big.NewInt(19814000),
 		HForkBlock:          nil, // TBD
+		L2MigrationBlock:    big.NewInt(26383999),
```
gastonponti, line 130: Same as above
```suggestion
		L2MigrationBlock:    big.NewInt(26384000),
```

gastonponti (approved):
LGTM

PR #2345, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


palango:
@RiceChuan Thanks for your PR. This is upstream code and we don't want to further increase our diff. Please have a look if you can apply this to upstream geth as well. 🙏

PR #2347, opened 2024-12-13
refactor: unify the error handling methods in the crypto package that are different from the project style

ChengenH:


palango:
Thanks for the PR. Those changes are in upstream code and should be fixed there. Please raise the patch with go-ethereum!

PR #2357, opened 2025-03-04
Add batch request limit and response size limit in a RPC batch request

Kourin1996:
### Description

This PR adds limitations of number of batch items and response size in JSON-RPC request for avoiding an excessive memory usage.

After this change, a batch request can contain up to a maximum of 1000 requests, and the total size of all result data is limited to 25 MB.

The hardcoded boundary values are based on the default values in the current implementation of go-ethereum.

### Other changes

### Tested

I added unit tests for each change

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -28,6 +28,11 @@ import (
 	"github.com/celo-org/celo-blockchain/log"
 )
 
+const (
+	BatchRequestLimit    = 1000
+	BatchResponseMaxSize = 25 * 1000 * 1000
+)
+
 // handler handles JSON-RPC messages. There is one handler per connection. Note that
 // handler is not safe for concurrent use. Message handling never blocks indefinitely
 // because RPCs are processed on background goroutines launched by handler.
@@ -100,6 +105,13 @@ func (h *handler) handleBatch(msgs []*jsonrpcMessage) {
 		})
 		return
 	}
+	// Returns an error if number of requests exceeds an allowed maximum
+	if len(msgs) > BatchRequestLimit {
+		h.startCallProc(func(cp *callProc) {
+			h.conn.writeJSON(cp.ctx, errorMessage(&invalidRequestError{"batch too large"}))
+		})
+		return
+	}
 
 	// Handle non-call messages first:
 	calls := make([]*jsonrpcMessage, 0, len(msgs))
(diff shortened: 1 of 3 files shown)


rpc/handler.go
```diff
@@ -112,11 +124,21 @@ func (h *handler) handleBatch(msgs []*jsonrpcMessage) {
...
 	h.startCallProc(func(cp *callProc) {
 		answers := make([]*jsonrpcMessage, 0, len(msgs))
-		for _, msg := range calls {
+		for idx, msg := range calls {
 			if answer := h.handleCallMsg(cp, msg); answer != nil {
 				answers = append(answers, answer)
+				// Once total size of responses exceeds an allowed maximum,
+				// generate error messages for all remaining calls and stop further processing
+				if responseBytes += len(answer.Result); responseBytes > BatchResponseMaxSize {
+					for i := idx; i < len(calls); i++ {
```
karlb, line 136: Since we added one `answer` above, don't we have to either
* start at `idx +1` to reject all the following requests (and live with the fact that we slightly exceeded `responseBytes`) or
* do this check before appending the current `answer` to `answers`?

rpc/handler.go
```diff
@@ -112,11 +124,21 @@ func (h *handler) handleBatch(msgs []*jsonrpcMessage) {
...
+	responseBytes := 0
 	h.startCallProc(func(cp *callProc) {
 		answers := make([]*jsonrpcMessage, 0, len(msgs))
-		for _, msg := range calls {
+		for idx, msg := range calls {
 			if answer := h.handleCallMsg(cp, msg); answer != nil {
 				answers = append(answers, answer)
+				// Once total size of responses exceeds an allowed maximum,
+				// generate error messages for all remaining calls and stop further processing
+				if responseBytes += len(answer.Result); responseBytes > BatchResponseMaxSize {
```
gastonponti, line 135: Technically `+=` is not thread safe.
I would recommend to use https://pkg.go.dev/sync/atomic#Int32.Add

rpc/handler.go
gastonponti: +1 to check before

rpc/handler.go
gastonponti: Nevermind, it's sequential on a thread not multi-thread

karlb: `append` isn't thread-safe either. Is this an upstream bug or am I missing something that avoids problems here?

karlb (approved):

PR #2358, opened 2025-03-04
Add Mainnet L2 migration block

palango:
### Description

See https://forum.celo.org/t/returning-home-to-ethereum-the-launch-of-celo-l2-mainnet/10466

Part of https://github.com/celo-org/celo-blockchain-planning/issues/903


gastonponti (approved):
LGTM

PR #2366, opened 2025-03-28
refactor: use the built-in max/min to simplify the code

pkucode:
### Description

Use the built-in max/min from the standard library in Go 1.21 to simplify the code.  https://pkg.go.dev/builtin@go1.21.0#max

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


palango:
Thanks for submitting this PR!

Our development now concentrates on [op-geth](https://github.com/celo-org/op-geth/) and we'll only fix critical issues in this repository. Please check and submit your PR there.

PR #2368, opened 2025-04-16
chore: remove duplicate package imports

fivecut:
### Description

remove duplicate package imports

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


palango:
Thanks for your PR, but as stated in the readme: "celo-blockchain is in maintainance mode and will only be updated for critical issues"

PR #2372, opened 2025-04-27
Fix broken links

PixelPil0t1:
Removed non-working links from the Clef Tutorial documentation and replaced them with available working alternatives where possible. If better or more specific links exist, please suggest replacements. Otherwise, it is better to keep the corrected working links.


palango:
Thanks for submitting this PR!

Our development now concentrates on [op-geth](https://github.com/celo-org/op-geth/) and we'll only fix critical issues in this repository. Please check and submit your PR there.

